PR #2, opened 2018-09-15
Staging 20180914

gctucker:
Summary:

* test-configs: fix file system for v4l2 and igt test suites
* test-configs: fix default filters
* add simple test plan variant for QEMU and split cases into sets and enable in lab-collabora
* Debian file systems: use the Debian arch names
  (when new file system binaries are available, the test configs will need to be updated accordingly)


gctucker:
PS Thanks @khilman for the idea of adding sets to the simple test plan, that's useful to test the groups feature.

PR #4, opened 2018-10-09
Staging 20181009

mattface:
Summary: 
* Fix a typo in the test-configs.yaml
* Remove the generic_defconfig from MIPS builds as it is not a config that is expected to be buildable, and some builds are hanging until timing out due to an unset CONFIG_HZ

Note: The patch to build with the all_abi config fragment from Anders' work-in-progress branch is being kept on staging until this has been accepted in upstream Linux kernel.

### DIFF
--- a/jenkins/build-trigger.jpl
+++ b/jenkins/build-trigger.jpl
@@ -67,6 +67,9 @@ def addDefconfigs(configs, kdir, arch) {
             for (String config: found.tokenize(' \n'))
                 configs.add(config)
         }
+        if (arch == "mips") {
+            configs.remove("generic_defconfig")
+        }
     } else {
         echo("WARNING: No configs directory: ${configs_dir}")
     }
--- a/test-configs.yaml
+++ b/test-configs.yaml
@@ -685,7 +685,7 @@ device_types:
 
   sun8i_h3_libretech_all_h3_cc:
     name: 'sun8i-h3-libretech-all-h3-cc'
-    mach: alliwinnder
+    mach: allwinner
     class: arm-dtb
     boot_method: uboot
 


gctucker:
I would be in favour of waiting until the `all_abi` config fragment has been merged in mainline kernel (or at least in linux-next) before trying to use it in KernelCI production.  We may have to rename this to something else or remove it if it doesn't get merged as-is, causing some unnecessary noise in production.  Maybe keeping the patch on staging would be enough to test this work-in-progress on Anders' tree?

mattface:
This patch only adds the fragment if the file exists, if it doesn't ever get into any other trees, it will have no effect.
We already look for a "debug".config, which doesn't exist in next or mainline either.
I see staging as for testing our changes, not kernel changes, so production is the right place.

khilman:
I agree with @gctucker.  Let's have a discussion about all_abi_defconfig on the list, what its purpose is, etc.

mattface:
Can we have this discussion alongside merging this PR? Is this really a blocker for a change that only adds *possibly* 1 build on 1 tree?
Wouldn't it be handy to already have build and boot information available when whatever this config frag becomes is agreed upon?

gctucker:
Once an agreement has been reached and the patch has landed in mainline or next then we can start testing it in production.

mattface:
@broonie they were all separate patches to staging, but when we do a staging->prod update it's all the patches since the last staging->prod merge, so by default they always have a pile of changes in there.
I merged it into staging because I really didn't see it being anything to discuss.

I have removed that patch, but I don't see at all how it's even an issue and my mind is boggled by how much air has been expelled discussing it.

PR #6, opened 2018-10-19
Staging 20181019

gctucker:
Summary:
* test-configs: enable missing `vexpress_defconfig` on qemu arm
* v4l2: fix results parsing to remove spurious "v4p2-compliance" test case result
* bisect.jpl: use kernelci-core repo and use Kernel library class
* stretchtests: disable mips64el build since v4l2-compliance fails to build with this arch

### DIFF
--- a/jenkins/bisect.jpl
+++ b/jenkins/bisect.jpl
@@ -75,6 +75,9 @@ TREES_WHITELIST
  */
 
 
+@Library('kernelci') _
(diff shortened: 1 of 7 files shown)


broonie:
LGTM

gctucker:
Closing manually as there seems to have been a bug in Github.  This has been merged now.

PR #8, opened 2018-11-12
Staging 20181112

mattface:
- Add dmaengine and soundwire urls to the trigger job
- revert all_abi config patch
- add vinod trees
- add qdf2400 device
- consolidate docker files into build-base
- DOCKERBASE -> DOCKER_BASE
- bisect.jpl use docker
- add xz-utils and g++ to build-base docker image
- build.py: use file_server_resource to describe artifacts upload location
- add ARC arch and defconfigs
- ensure output directory exists for debos


mattface:
because of the DOCKER_BASE changes, we will need to turn off the trigger and update jobs on production before merging this

gctucker:
You can add the new DOCKER_BASE parameter while leaving the current DOCKERBASE one, and remove it after the merge.  But yes, generally speaking I think it's best to stop the trigger and wait for all builds to complete before doing a merge anyway.

gctucker:
Could you please remove the "revert" commits and the corresponding ones that were merged by mistake as they're just noise in the history?

PR #10, opened 2018-11-22
Staging 20181122

gctucker:
Summary:
* v4l2: group test cases to avoid duplicates and make it possible to start tracking regressions
* Debian: split stretchtests root file system as stretch-igt and stretch-v4l2
* Docker images: add riscv64 and fix mips-6.3.0
* bisection: fix error propagation and automatically add email report recipients
* add retry logic to Docker pull to work around timeout issues with hub.docker.com
* add support to build riscv64
* add linux-media tree URL to start building and testing the master branch with v4l2


gctucker:
Building all the Debian rootfs images on production now, will then update the URLs and run a test with my tree then enable the trigger again.

PR #21, opened 2019-02-20
Init requirements for python scripts

aliceinwire:
sorted in alphabetical order

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,4 @@
+jinja2
+keyring
+pyyaml
+requests


gctucker:
With our current workflow, PRs should be sent to https://github.com/kernelci/kernelci-core-staging instead.  Then they get merged into `kernelci-core` and used in production once they have been tested on staging.kernelci.org.  As this PR is pretty low risk it can be merged directly during the next production update (probably end of next week).

I know it's confusing, we should do something about that to not have 2 projects...

PR #24, opened 2019-02-27
Add android-4.19 branch

metti:


gctucker:
Please submit the PR to https://github.com/kernelci/kernelci-core-staging instead.  Then it'll get merged into production after it has been tested on staging.kernelci.org.

I know it's confusing, we should do something about that to not have 2 projects...

metti:
See https://github.com/kernelci/kernelci-core-staging/pull/102

PR #29, opened 2019-03-19
Meson g12

touilkhouloud:


touilkhouloud:
Hello @gctucker, device types merged in LAVA.

test-configs.yaml
```diff
@@ -461,6 +461,24 @@ device_types:
     filters:
       - blacklist: {defconfig: ['allnoconfig', 'allmodconfig']}
 
+  meson_g12a_sei510:
+    name: 'meson-g12a-sei510'
+    mach: amlogic
+    class: arm64-dtb
+    boot_method: uboot
+    flags: ['little_endian']
```
gctucker: Like I wrote before, we shouldn't be introducing a `little_endian` flag now.  Little-endian is the default unless big-endian is specified.

test-configs.yaml
```diff
@@ -461,6 +461,24 @@ device_types:
...
+    filters:
+      - blacklist: {defconfig: ['allnoconfig', 'allmodconfig']}
+
+  meson-g12a-x96-max:
+    name: 'meson-g12a-x96-max'
```
gctucker: The `name` property is not necessary when it's the same as the config name, so that can be dropped as well.

test-configs.yaml
```diff
@@ -461,6 +461,24 @@ device_types:
...
+    mach: amlogic
+    class: arm64-dtb
+    boot_method: uboot
+    flags: ['little_endian']
```
gctucker: Same comment about `little_endian` flag.

test-configs.yaml
```diff
@@ -461,6 +461,24 @@ device_types:
     filters:
       - blacklist: {defconfig: ['allnoconfig', 'allmodconfig']}
 
+  meson_g12a_sei510:
```
gctucker: Initial configurations we using underscores instead of dashes, but it's OK in YAML to have dashes.  So to simplify things, you can have the config name set to `meson-g12a-sei510` and drop the `name` property as it's the same.

touilkhouloud:
Changes are done and the boards are online now in the lab.

test-configs.yaml
```diff
@@ -941,6 +948,9 @@ test_configs:
   - device_type: kirkwood_db_88f6282
     test_plans: [boot_nfs]
 
+  - device_type: meson_g12a_sei510
+    test_plans: [boot, kselftest, simple]
```
khilman: Please drop kselftest for now.  This test plan is not currently useful on amlogic boards.

test-configs.yaml
```diff
@@ -941,6 +955,12 @@ test_configs:
   - device_type: kirkwood_db_88f6282
     test_plans: [boot_nfs]
 
+  - device_type: meson_g12a_sei510
```
khilman: Shouldn't this have dashes, not underscores?

khilman:
@gctucker this is ready for staging too.

gctucker:
Great, this will be tested today.

gctucker:
The kernel branch used on staging was from `linux-5.0.y` which doesn't have the dtb for these boards, trying again today with mainline.  Parsing the test configs worked, so this seems good so far.

touilkhouloud:
Is it ready to be merged ?

gctucker:
The staging results look good:
https://staging.kernelci.org/boot/all/job/kernelci/branch/staging.kernelci.org/kernel/v5.1-rc3-23-gfb648afee9ff/

Only the `meson-g12a-x96-max` appears to be there, I guess the `meson-g12a-sei510` was offline.

Interestingly we get a same-lab conflict with the `meson-g12a-x96-max`: it passed with gcc but failed to boot with the clang kernel build.

So overall, yes it's fine to merge this noe.

PR #35, opened 2019-03-26
Ltp testrootfs

roxell:
Introduce a new debos rootfs with Linux Text Project (LTP).


jenkins/stretch-ltp.jpl
```diff
@@ -0,0 +1,13 @@
...
+import org.kernelci.debian.RootFS
+
+def r = new RootFS()
+
+def config = ['name':"stretch-ltp",
+              'vm_memory':"3G",
+              'arch_list':["armhf", "arm64", "i386", "amd64"],
+              'debian_release':"stretch",
+              'extra_packages':"acl attr cpio cron curl expect file gawk gzip iproute2 libaio1 libc-bin logrotate net-tools perl procps python-all quota systemd tar unzip util-linux ",
+              'script':"scripts/stretch-ltp.sh"]
```
gctucker: Cosmetics: it's a bit better to add a trailing `,` and a new line, to reduce diff noise when adding things at the end.

roxell: fixed

roxell:
target x15, arm: https://lkft.validation.linaro.org/scheduler/job/656811/definition
target hikey, arm64: https://lkft.validation.linaro.org/scheduler/job/656319/definition

jenkins/stretch-igt.jpl
```diff
@@ -18,6 +18,7 @@ def config = ['name':"stretch-igt",
               'extra_packages':"libpciaccess0 libkmod2 libprocps6 libcairo2 \
                                 libssl1.1 libunwind8 libudev1 libglib2.0-0 \
                                 libdw1 liblzma5",
+              'extra_packages_remove':"bash e2fsprogs e2fslibs"
```
gctucker: Missing trailing `,` - this breaks the Jenkins job.

jenkins/stretch-v4l2.jpl
```diff
@@ -15,6 +15,7 @@ def config = ['name':"stretch-v4l2",
               'arch_list':["armhf", "arm64", "amd64"],
               'debian_release':"stretch",
               'extra_packages':"",
+              'extra_packages_remove':"bash e2fsprogs e2fslibs"
```
gctucker: Missing trailing `,` - this breaks the Jenkins job.

jenkins/stretch.jpl
```diff
@@ -16,6 +16,7 @@ def config = ['name':"stretch",
                            "mips", "mipsel", "mips64el"],
               'debian_release':"stretch",
               'extra_packages':"",
+              'extra_packages_remove':"bash e2fsprogs e2fslibs"
```
gctucker: Missing trailing `,` - this breaks the Jenkins job.

gctucker:
So the problem with x86 flavours is that the total size of the binaries appears to exceed a limit set on the server.  One way to solve this is to upload each file separately, another one is to increase the limit.  Then I guess it may also be possible to optimize the size of the rootfs files.  We'll see if we have the time to investigate, maybe next week (cc @mgalka).

gctucker:
It should already possible to test the `arm` and `arm64` file systems by hand, will give that a spin tomorrow.

roxell:
Where can I find the arm64 rootfs?

gctucker:
> Where can I find the arm64 rootfs?

http://staging-storage.kernelci.org/images/rootfs/debian/stretch-ltp/20190327.1/

gctucker:
Hrm, well it timed out after 4h:
https://lava.collabora.co.uk/scheduler/job/1581286

So yes the arm64 rootfs seems to work but we probably have to adjust quite a few things before we can run this kind of long tests in production (split it into small jobs...).

gctucker:
So as discussed, we should try to run a smaller set of tests from LTP so the jobs don't last more than 10 min to avoid causing disruptions in production.  Waiting for @roxell to confirm the arm64 or arm rootfs produced on staging works as expected with such as smaller test set using nfsroot or ramdisk in the LKFT lab, compared with an LKFT job running the same tests.

gctucker:
Thanks @roxell for the sample test job definition, I've verified that this worked as expected with the debos-based rootfs from Jenkins using nfs on an arm64 device:
https://lava.collabora.co.uk/scheduler/job/1587298

gctucker:
Next steps are to resolve the upload size issue which we should be able to do this week, or drop the x86 variants for now.  Then we can start adding some test templates for ltp-ipc in another PR.

gctucker:
The upload size issue has been resolved and the arm64 flavour of this filesystem has been tested as explained above, so this can now be merged.

Latest builds from staging, for all 4 architectures:
http://staging-storage.kernelci.org/images/rootfs/debian/stretch-ltp/20190425.0/

Next step is to add a template for the ipc tests to start running that in KernelCI labs.

PR #36, opened 2019-03-26
Add basic configs tests using pytest

danrue:
- Test parsing ./build-configs.yaml
- Add some additional test yaml files to verify edge cases
- Fix issues in configs.py related to empty fragments, build_config_defaults,
and when architecture is None

Makefile provided for convenience, with recommended invocation of
pytest. Due to the way this project is structured (specifically, that it
lacks a setup.poy), pytest must be run using python -m.

Example usage:

drue@xps:~/src/kernelci/kernelci-core$ make
python -m pytest
========================================== test session starts ==========================================
platform linux2 -- Python 2.7.15, pytest-4.3.1, py-1.8.0, pluggy-0.9.0
rootdir: /home/drue/src/kernelci/kernelci-core, inifile:
collected 3 items

tests/test_configs.py ...                                                                         [100%]

======================================= 3 passed in 0.13 seconds ========================================

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,4 @@
+test:
+	# Note that just running 'pytest' will not work here because there is no setup.py. See
+	# https://docs.pytest.org/en/latest/pythonpath.html#invoking-pytest-versus-python-m-pytest
+	python -m pytest
--- a/kernelci/configs.py
+++ b/kernelci/configs.py
@@ -373,7 +373,7 @@ class BuildVariant(YAMLObject):
             config, ['name', 'build_environment', 'fragments']))
         kw['build_environment'] = build_environments[kw['build_environment']]
         kw['architectures'] = list(
-            Architecture.from_yaml(data, name, fragments)
+            Architecture.from_yaml(data or {}, name, fragments)
             for name, data in config['architectures'].iteritems()
         )
         cf = kw.get('fragments')
@@ -818,7 +818,7 @@ def builds_from_yaml(yaml_path):
 
     fragments = {
         name: Fragment.from_yaml(config, name)
-        for name, config in data['fragments'].iteritems()
+        for name, config in data.get('fragments', {}).iteritems()
     }
 
     build_environments = {
@@ -826,7 +826,7 @@ def builds_from_yaml(yaml_path):
         for name, config in data['build_environments'].iteritems()
     }
 
-    defaults = data.get('build_configs_defaults')
+    defaults = data.get('build_configs_defaults', {})
 
     build_configs = {
         name: BuildConfig.from_yaml(config, name, trees, fragments,
--- a/tests/test_configs.py
+++ b/tests/test_configs.py
@@ -1,0 +1,28 @@
+import kernelci.configs
+
+def test_build_configs_parsing():
+    """ Verify build-configs.yaml """
+    configs = kernelci.configs.builds_from_yaml("build-configs.yaml")
+    assert len(configs) == 4
+    for key in ['build_configs', 'build_environments', 'fragments', 'trees']:
+        assert key in configs
+        assert len(configs[key]) > 0
+
+def test_build_configs_parsing_minimal():
+    configs = kernelci.configs.builds_from_yaml("tests/configs/minimal.yaml")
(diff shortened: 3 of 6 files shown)


kernelci/configs.py
```diff
@@ -262,9 +262,10 @@ def from_yaml(cls, data, name, fragments):
         kw.update(cls._kw_from_yaml(data, [
             'name', 'base_defconfig', 'extra_configs',
         ]))
-        cf = data.get('fragments')
-        kw['fragments'] = [fragments[name] for name in cf] if cf else None
-        kw['filters'] = FilterFactory.from_data(data)
+        if data is not None:
```
gctucker: It would be better to not call this method with data being `None` rather than checking its value.  I think it's important to draw the line with what the values for the methods should be, I guess that would require some docstrings in the `from_yaml()` methods.

So how about this patch instead, to fix it on the calling side:
```diff
diff --git a/kernelci/configs.py b/kernelci/configs.py
index 42b4c2de9da9..87ecfb863d9b 100644
--- a/kernelci/configs.py
+++ b/kernelci/configs.py
@@ -373,7 +373,7 @@ class BuildVariant(YAMLObject):
             config, ['name', 'build_environment', 'fragments']))
         kw['build_environment'] = build_environments[kw['build_environment']]
         kw['architectures'] = list(
-            Architecture.from_yaml(data, name, fragments)
+            Architecture.from_yaml(data or {}, name, fragments)
             for name, data in config['architectures'].iteritems()
         )
         cf = kw.get('fragments')
```

Makefile
```diff
@@ -0,0 +1,4 @@
+test:
```
gctucker: Is there a good reason for wrapping this with make?  Running `pytest` directly would seem fine, or maybe using `setuptools` as a first step towards turning this into a proper Python package?

tests/test_configs.py
```diff
@@ -0,0 +1,25 @@
+import kernelci.configs
+
+def test_build_configs_parsing():
+    """ Verify build-configs.yaml """
+    configs = kernelci.configs.builds_from_yaml("build-configs.yaml")
+    assert len(configs) == 4
```
gctucker: Might be worth checking that the required things are in the returned dict (`trees`, `fragments`...).

kernelci/configs.py
danrue: Thanks, that is better.

Makefile
danrue: For convenience. It's not really a python project, so I don't think it's a great idea to just drop in a setup.py file to make pytest happy.

tests/test_configs.py
danrue: Good suggestion. Added

Makefile
gctucker: I actually think we should eventually have a kernelci python package, but that wouldn't include the jenkins stuff and we're not quite there yet.  So I agree that using a `setup.py` wouldn't be very appropriate at the moment.  Still I'm not convinced either by the `Makefile` approach just to run `pytest`...

How about adding something about `pytest` in the README instead?

danrue: That's the point though - you can't just run 'pytest'. You must run `python -m pytest`.

I think it's a larger question than just pytest. What if we want to enforce for example, flake8, or black, along with pytest? There should be a simple, obvious one-liner that any developer can run to guarantee compliance and match whatever automated CI will also run against pull requests. Make provides a nice interface for that automatically. We can also write some shell script, or rely on documentation, but I don't think it's a good developer experience to ask them to copy paste multiple commands out of documentation.

Once we decide on the interface though, I agree it needs to be added to README and I can do that in this PR or a subsequent PR.

gctucker: OK, fair enough.  If it happens to make more sense to have these test commands elsewhere than in the `Makefile` in the future, it won't be hard to make the change then.

gctucker:
Tested OK on my laptop, ready to merge now.

danrue:
@gctucker thanks. how do we go from 'ready to merge' to merged?

gctucker:
I can merge it once the tests on staging have completed today.

PR #37, opened 2019-03-26
Add am57xx-beagle-x15

danrue:
BeagleBoard-X15 is the next generation Open Source Hardware BeagleBoard based
on TI's AM5728 SoC featuring dual core 1.5GHz A15 processor. The platform
features 2GB DDR3L (w/dual 32bit busses), eSATA, 3 USB3.0 ports, integrated
HDMI (1920x1080@60), separate LCD port, video In port, 4GB eMMC, uSD, Analog
audio in/out, dual 1G Ethernet.

For more information, refer to:
BeagleBoard-X15 Wiki:
http://www.elinux.org/Beagleboard:BeagleBoard-X15

Support for x15 was first added in linux v3.19. The LAVA implementation for
this board uses 'setenv console ttyS2,115200n8', which requires
CONFIG_SERIAL_8250_OMAP to be set. However, CONFIG_SERIAL_8250_OMAP was not
added to multi_v7_defconfig until v5.0.

This initial x15 implementation in kernelci will therefore restrict x15 to
v5.0+, so that it can use multi_v7_defconfig without an additional fragment.
Future kernelci work could add a fragment for CONFIG_SERIAL_8250_OMAP to boot
this board on v4.* kernel versions.

It is also worth noting that the default LAVA implementation uses the 'x15'
device-type name, while kernelci uses the dtb-based name 'am57xx-beagle-x15'.
Labs wishing to host the x15 for kernelci will therefore need to implement them
as device type name 'am57xx-beagle-x15'.

### DIFF
--- a/test-configs.yaml
+++ b/test-configs.yaml
@@ -151,6 +151,16 @@ device_types:
     class: arm-dtb
     boot_method: uboot
 
+  am57xx-beagle-x15:
+    mach: omap2
+    class: arm-dtb
+    boot_method: uboot
+    filters:
+      - blacklist: *allmodconfig_filter
+      # This board could boot v4.* if CONFIG_SERIAL_8250_OMAP is set. It was
+      # added to multi_v7_defconfig as of 5.0.
+      - blacklist: {kernel: ['v3.', 'v4.']}
+
   apm_mustang:
     name: 'mustang'
     mach: apm
@@ -839,6 +849,9 @@ test_configs:
 
   - device_type: alpine_db
     test_plans: [boot]
+
+  - device_type: am57xx-beagle-x15
+    test_plans: [boot, boot_nfs, simple, sleep]
 
   - device_type: apm_mustang
     test_plans: [boot, boot_nfs, kselftest]
--- a/jenkins/lava-boot-v2.sh
+++ b/jenkins/lava-boot-v2.sh
@@ -51,6 +51,6 @@ elif [ ${LAB} = "lab-theobroma-systems" ] || [ ${LAB} = "lab-theobroma-systems-d
   python lava-v2-jobs-from-api.py --defconfigs ${DEFCONFIG_COUNT} --callback ${CALLBACK} --api ${API} --storage ${STORAGE} --lab ${LAB} --describe ${GIT_DESCRIBE} --tree ${TREE} --branch ${BRANCH} --arch ${ARCH} --plans boot kselftest --token ${API_TOKEN}
   python lava-v2-submit-jobs.py --username kernel-ci --jobs ${LAB} --token ${LAVA_THEOBROMA_SYSTEMS_TOKEN} --lab ${LAB}
 elif [ ${LAB} = "lab-drue" ] || [ ${LAB} = "lab-drue-dev" ]; then
-  python lava-v2-jobs-from-api.py --defconfigs ${DEFCONFIG_COUNT} --callback ${CALLBACK} --api ${API} --storage ${STORAGE} --lab ${LAB} --describe ${GIT_DESCRIBE} --tree ${TREE} --branch ${BRANCH} --arch ${ARCH} --plans boot boot-kvm boot-kvm-uefi boot-nfs boot-nfs-mp simple --token ${API_TOKEN} --priority medium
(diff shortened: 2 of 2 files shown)


danrue:
I haven't been able to finish my staging environment to do a full end-to-end test of this. Perhaps it can get tested along with other changes.

test-configs.yaml
```diff
@@ -839,6 +846,9 @@ test_configs:
   - device_type: alpine_db
     test_plans: [boot]
 
+  - device_type: am57xx-beagle-x15
+    test_plans: [boot, boot_nfs, kselftest, simple, sleep]
```
gctucker: I think @khilman suggested to not run `kselftest` on that board in the previous version of this PR.

gctucker: Also please note that only these test plans are listed in `jenkins/lava-boot-v2.sh` (something else that should be put in YAML):
```
boot boot-kvm boot-kvm-uefi boot-nfs boot-nfs-mp simple
```
So any other test plan will not be scheduled in `lab-drue` unless they get added.

I guess `kselftest` should be removed from `test-configs.yaml` and `sleep` should be added to `lava-boot-v2.sh`?

danrue: Fixed.

gctucker:
> > I've run it on staging.kernelci.org, it looks like the jobs were submitted but it's not booting (tested on v4.20.17):
> > https://lava.therub.org/scheduler/job/321
> > I see your healthchecks are passing with v5.0.0-rc7, is this board not supported in v4.20?
> > I'll upgrade the staging kernel test branch to use `linux-5.0.y` in any case.
> 
> We use the board since 4.4 in LKFT. I did some digging, and as it turns out we're adding CONFIG_SERIAL_8250_OMAP, which was added to multi_v7_defconfig as of 5.0. I was able to reproduce the issue you found, and adding that config fixed it. I'm not sure the best way to handle this in kernelci. Any suggestions?

One way is to blacklist kernels earlier than 5.0, another way is to define a config fragment but it would only apply on top of `multi_v7_defconfig`.  It's better than nothing if testing stable branches before 5.0 is needed.

I think we should favour what is supported in upstream, so I would opt for blacklisting kernels prior to 5.0.  You can add a filter for that easily, there are other boards in a similar situation.  For example:

```yaml
    filters:
      - blacklist: *allmodconfig_filter
      - blacklist: {kernel: ['v3.', 'v4.']}
```

danrue:
> > > I've run it on staging.kernelci.org, it looks like the jobs were submitted but it's not booting (tested on v4.20.17):
> > > https://lava.therub.org/scheduler/job/321
> > > I see your healthchecks are passing with v5.0.0-rc7, is this board not supported in v4.20?
> > > I'll upgrade the staging kernel test branch to use `linux-5.0.y` in any case.
> > 
> > 
> > We use the board since 4.4 in LKFT. I did some digging, and as it turns out we're adding CONFIG_SERIAL_8250_OMAP, which was added to multi_v7_defconfig as of 5.0. I was able to reproduce the issue you found, and adding that config fixed it. I'm not sure the best way to handle this in kernelci. Any suggestions?
> 
> One way is to blacklist kernels earlier than 5.0, another way is to define a config fragment but it would only apply on top of `multi_v7_defconfig`. It's better than nothing if testing stable branches before 5.0 is needed.
> 
> I think we should favour what is supported in upstream, so I would opt for blacklisting kernels prior to 5.0. You can add a filter for that easily, there are other boards in a similar situation. For example:
> 
> ```yaml
>     filters:
>       - blacklist: *allmodconfig_filter
>       - blacklist: {kernel: ['v3.', 'v4.']}
> ```

I implemented it as suggested. It would be nice to add a fragment for this board for v4.x kernels, but I think I'd like to just get this moved forward with v5.x for now. It was also not clear to me how to allow the x15 to test all configs in v5.x, but only one fragment on v4.x.

gctucker:
> > I think we should favour what is supported in upstream, so I would opt for blacklisting kernels prior to 5.0. You can add a filter for that easily, there are other boards in a similar situation. For example:
> > ```yaml
> >     filters:
> >       - blacklist: *allmodconfig_filter
> >       - blacklist: {kernel: ['v3.', 'v4.']}
> > ```
> 
> I implemented it as suggested. It would be nice to add a fragment for this board for v4.x kernels, but I think I'd like to just get this moved forward with v5.x for now. It was also not clear to me how to allow the x15 to test all configs in v5.x, but only one fragment on v4.x.

Thanks, will get it tested again on staging today.  In principle it should be possible to run with a v4.x kernel and a fragment using a `combination` filter with `['kernel', 'defconfig']`.  That would mean an extra build for some v4.x kernels though.

gctucker:
Tested OK: https://staging.kernelci.org/boot/id/5cbe16cdcf3a0f75ea164c66/

PR #42, opened 2019-04-03
T13960 stretch-igt: disable MIPS builds as currently broken in i-g-t

gctucker:
The current master branch of i-g-t doesn't build on MIPS due to the
use of some hardware atomics that are not supported on MIPS.  An
upstream fix is being worked on, but disable MIPS builds in the
meantime so we can get the latest version on other architectures.


gctucker:
Tested OK on staging:
http://staging-storage.kernelci.org/images/rootfs/debian/stretch-igt/20190403.0/

PR #44, opened 2019-04-12
Add android-q branches to build-configs

mattface:
These branches are expected to only be updated to LTS once a week.

I have tested that they parse with kci_build and produce the expected results


gctucker:
Tested OK on staging up to the point where it created a tarball.

PR #45, opened 2019-04-18
More lab-pengutronix barebox platforms now available in lava

mgrzeschik:


test-configs.yaml
```diff
@@ -794,6 +794,16 @@ device_types:
     class: arm-dtb
     boot_method: uboot
 
+  socfpga_cyclone5_socrates:
+    name: 'socfpga-cyclone5-socrates'
+    mach: socfpga
+    class: arm-dtb
+    boot_method: barebox
+    filters:
```
gctucker: Typo - there's one `filters` level too many, this should be:
```yaml
      - blacklist:
          - 'allmodconfig'
```

mgrzeschik:
Fixed it and also changed all modconfig nodes to macro.

gctucker:
Ah yes, thanks.  In YAML they're called aliases and anchors, they're more like pointers than macros btw.

Giving that another spin today.

mgrzeschik:
Some of the boards are still in maintenance in our lab. Do we have to put them online before merging?

gctucker:
It would be more useful, but the job generation part can still be verified even if jobs aren't actually submitted due to no online devices for some types.  The advantage of fully testing them on staging is that it's easier to fix anything than if we discover some issues later on when it's already in production.

kernelci/configs.py
```diff
@@ -534,6 +534,12 @@ def __init__(self, name, mach, arch='arm', *args, **kw):
         kw.setdefault('dtb', '{}.dtb'.format(name))
         super(DeviceType_arm, self).__init__(name, mach, arch, *args, **kw)
 
+class DeviceType_mips(DeviceType):
```
gctucker: Please keep 2 blank lines before the beginning of each class, as reported by `pep8`:
```
kernelci/configs.py:537:1: E302 expected 2 blank lines, found 1
kernelci/configs.py:544:1: E302 expected 2 blank lines, found 1
```

gctucker:
So this was tested OK today, here's some boot results for `imx6ul-pico-hobbit`:
https://staging.kernelci.org/boot/id/5cbf21becf3a0feb49164c1c/

@mgrzeschik Do you think the other boards can be online tomorrow?  We could get another round of testing before merging this PR.  Also it's not super useful in practice to add these boards if they're not actually online.

PR #46, opened 2019-04-19
T14297 kci_build to build kernels

gctucker:
Add build commands to the `kci_build` tool and the `kernelci.build` library and use that in the build job in place of the legacy `build.py` script.  It makes the code more maintainable and focused on the KernelCI use-case with still some flexibility to build kernels manually with arbitrary configs.  It also relies on the `build-configs.yaml` data to simplify job parameters.

Following this change, the current `build.py` script will not be used any more in this repo and could therefore be removed.

Sample commands to use this locally to build a kernel from linux-next:

1. Optional: set up a local mirror
If you already have a linux check out:
```
git clone \
  --mirror git://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git \
  --reference=~/src/linux linux-mirror.git
```
Then to update the mirror, or create it if you skipped the command above:

```
./kci_build update_mirror --config=next --mirror=linux-mirror.git
```

2. Create a local check-out (or update an existing one when running this again)
```
./kci_build update_repo --config=next --kdir=linux --mirror=linux-mirror.git
```
Optionally, to generate additional config fragments (e.g. to then build `defconfig+kselftest`):
```
./kci_build generate_fragments --config=next --kdir=linux
```

3. Build the kernel with defconfig
```
./kci_build build_kernel --defconfig=defconfig --arch=x86 --build-env=gcc-7 --kdir=linux
```

To see the compiler output, add `--verbose`.
The output can be found in `linux/build`.

To build again without regenerating the kernel config file, just omit the `--defconfig` argument:
```
./kci_build build_kernel --arch=x86 --build-env=gcc-7 --kdir=linux
```

Note: the build-env option is only used to populate the meta-data for the KernelCI database, it is not downloading a build environment.  A future improvement would be to enable using a Docker image as with Jenkins builds (`jenkins/build.jpl`).

4. Install the kernel binaries in a local directory
```
./kci_build install_kernel --config=next --kdir=linux 
```

See the output in `linux/_install_`.

5. If you have a kernelci-backend instance running, you can send the kernel and also the meta-data with these commands:
```
./kci_build push_kernel --kdir=linux --api=https://localhost:12345 --token=1234-5678
./kci_build publish_kernel --kdir=linux --api=https://localhost:12345 --token=1234-5678
```
Alternatively, to store the meta-data locally in a JSON file:
```
./kci_build publish_kernel --kdir=linux --json-path=build-meta.json
```

PR #47, opened 2019-04-23
build.py: fix riscv kernel image install

khilman:
The install phase is not installing any kernel images for RISC-V.  Fix
that.

Signed-off-by: Kevin Hilman <khilman@kernel.org>

### DIFF
--- a/build.py
+++ b/build.py
@@ -415,7 +415,7 @@ if install:
     # Patterns for matching kernel images by architecture
     if arch == 'arm':
         patterns = ['zImage', 'xipImage']
-    elif arch == 'arm64':
+    elif arch == 'arm64' or arch == "riscv":
         patterns = ['Image']
     elif arch == "arc" or arch == "mips":
         patterns = ['uImage.gz']


build.py
```diff
@@ -415,7 +415,7 @@ def extra_configs(dot_config, kbuild_output):
     # Patterns for matching kernel images by architecture
     if arch == 'arm':
         patterns = ['zImage', 'xipImage']
-    elif arch == 'arm64':
+    elif arch == 'arm64' or arch == "riscv":
```
gctucker: There's also `Image.gz` which is actually defined as the default one in `arch/riscv/Makefile`:
```
KBUILD_IMAGE    := $(boot)/Image.gz
```
I'm not sure why that is the case though, and maybe `Image` is more suitable to boot in practice.

gctucker: @khilman In the new build.py there's a similar thing using a dictionary and I initially put `Image.gz` for `riscv`:
https://github.com/kernelci/kernelci-core/pull/46/files#diff-d9df6c8ff856deb49f5362dc3c59e305R53

so the kernel image uploaded ends up being `Image.gz`:
https://staging.kernelci.org/build/id/5cc73390cf3a0f0df35f958d/

If this is no good then I can change it to `Image`, or have both as the first one in the list will take priority.

khilman: @gctucker please add Image as the first choice, as this works better with the current upstream u-boot, until they add booti and compressed image support.

I'll update/rebase this patch to list both as well.

gctucker: OK I've changed `kernelci/build.py` to have this:
```python
# Hard-coded binary kernel image names for each CPU architecture
KERNEL_IMAGE_NAMES = {
    'arm': ['zImage', 'xipImage'],
    'arm64': ['Image'],
    'arc': ['uImage.gz'],
    'i386': ['bzImage'],
    'mips': ['uImage.gz', 'vmlinux.gz.itb'],
    'riscv': ['Image', 'Image.gz'],
    'x86_64': ['bzImage'],
    'x86': ['bzImage'],
}
```

PR #48, opened 2019-04-25
Add a script to build + push all the docker images

mattface:
A (hopefully) helpful script to rebuild and push all the dockerimages used by kernelci builds.
It can also be used to build for a developer tag.


gctucker:
I used it today to update the docker image for production roll-out.  Now waiting for the test run to see if the builds and things work as expected.

gctucker:
Worked fine, that's great.

One suggestion, maybe having an option to print a help message or using `-h` for that would be nicer than having to open the script to read the description of all the options.

PR #49, opened 2019-04-26
Enable clang-8 builds on linux-next for arm64

mattface:
Once the build and boots email fixes are tested and merged, we should be good to enable clang-8 builds for arm64 on linux-next.

Tested locally that the right outputs come out of kci_build

### DIFF
--- a/build-configs.yaml
+++ b/build-configs.yaml
@@ -581,6 +581,11 @@ build_configs:
               - 'multi_v7_defconfig+CONFIG_EFI=y+CONFIG_ARM_LPAE=y'
               - 'allnoconfig'
               - 'allmodconfig'
+      clang-8:
+        build_environment: clang-8
+        fragments: *default_fragments
+        architectures:
+          arm64: *arm64_arch
 
   next_pending-fixes:
     tree: next


build-configs.yaml
```diff
@@ -581,6 +581,11 @@ build_configs:
               - 'multi_v7_defconfig+CONFIG_EFI=y+CONFIG_ARM_LPAE=y'
               - 'allnoconfig'
               - 'allmodconfig'
+      clang-8:
+        build_environment: clang-8
+        fragments: *default_fragments
+        architectures:
+          arm64: *arm64_arch
```
gctucker: Are we sure we want to enable `allmodconfig` already, while it seems to be far from building based on the results from staging?
https://staging.kernelci.org/build/id/5cc6e080cf3a0f9d66257f6d/
http://staging-storage.kernelci.org/next-clang/master/next-20190429/arm64/allmodconfig/clang-8/build.log

mattface: It doesn't pollute the email results (just shows as a fail) so I thought it would be fine.
It is actively being worked on but would be nice to increase the audience to see if the community could help speed up resolving some of the build issues

nickdesaulniers: It's not "far" relative to past issues.  This was a recent regression that kernel maintainers are taking their sweet time picking up the fix for.
https://lkml.org/lkml/2019/4/29/743

gctucker:
Running one more spin on staging now, will merge if we agree that the reports look good.

build-configs.yaml
gctucker: OK fair enough.  We can always turn it off if linux-next recipients find it too noisy or not useful for any reason.

gctucker:
All the bugs in the backend have been fixed so the reports are now looking fine based on the latest test runs on staging.kernelci.org.  So this is ready to be deployed now.

Issue #51, opened 2019-04-29
Fix gcc-7 docker images


gctucker:
Debian Buster does not appear to have gcc-7 cross-compiler toolchain packages any more.

gctucker:
In fact it looks like Stretch has only gcc-6 and Buster only gcc-8.  So the gcc-7 packages can't be found anywhere any more:
https://packages.debian.org/search?suite=all&section=all&arch=any&searchon=names&keywords=gcc-7-arm-linux-gnueabihf

gctucker:
See PR #52.

gctucker:
#52 has now been merged, `gcc-8` is the new default compiler.

PR #52, opened 2019-04-29
Use gcc-8 by default and drop gcc-7

gctucker:
Debian Buster now only has gcc-8 cross-compiler toolchains, so use it instead of gcc-7 as the default and drop the gcc-7 Docker images.  See also #51.


mattface:
I just came across this when trying my build script. Annoying of Debian to retrospectively remove gcc-7 from older releases.
Anyway LGTM. I'm fine with upgrading but would be nice to do a full build run on gcc-8 before we drop the gcc-7 Dockerfiles

gctucker:
@mattface The gcc-7 Dockerfiles don't build any more.  We can keep the existing tagged images in the Docker Hub for a little while.  And sure, this will be tested on staging tomorrow with a subset of builds and will go through a test build based on mainline when going into production, so if some builds start failing it can always be reverted at that point.

mattface:
Yeah I meant the images on dockerhub, no point removing them in any hurry

PR #53, opened 2019-04-29
test-configs.yaml: fix x86 default defconfig filters

gctucker:
Fix the x86 default filters with the standard defconfig names as these
are now specific to i386 and x86_64 since the x86 arch was split and
the base defconfigs are named in build-configs.yaml.

### DIFF
--- a/test-configs.yaml
+++ b/test-configs.yaml
@@ -54,8 +54,8 @@ test_plan_default_filters:
       values:
         - ['arm', 'multi_v7_defconfig']
         - ['arm64', 'defconfig']
-        - ['i386', 'defconfig']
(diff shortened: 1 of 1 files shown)


mattface:
I'm not sure why we need to do this?

Doesnt the ARCH= section of a make command give the right config using 'defconfig'?

For i386 I just do
`make ARCH=i386 defconfig`
and i get the right result

gctucker:
This is how it was originally in `build-configs.yaml`, then I changed it to explicitly name which defconfigs to build as a way to make it clearer what is actually being built.  I believe it was actually like that before the move to the YAML-based configuration.  So this change in `test-configs.yaml` is now needed to match the meta-data.

PR #54, opened 2019-05-01
test-configs.yaml: drop lpae flag on arm64 devices

gctucker:
The 'lpae' flag only makes sense on ARMv7, so drop them on arm64
(ARMv8) devices as they were most likely added by mistake.


gctucker:
Tested OK on staging.

PR #58, opened 2019-05-03
Add IGT test plan for gru-kevin

mgalka:
Enable IGT test plan for kevin, limit kernel versions to 4.19+

### DIFF
--- a/test-configs.yaml
+++ b/test-configs.yaml
@@ -1159,6 +1159,11 @@ test_configs:
   - device_type: rk3399_gru_kevin
     test_plans: [boot, v4l2_compliance_uvc]
 
+  - device_type: rk3399_gru_kevin
+    test_plans: [igt]
+    filters:
+      - blacklist: {kernel: ['v3.', 'v4.4', 'v4.9', 'v4.14']}
+
   - device_type: rk3399_puma_haikou
     test_plans: [boot, kselftest]
 


test-configs.yaml
```diff
@@ -744,6 +744,8 @@ device_types:
     mach: rockchip
     class: arm64-dtb
     boot_method: depthcharge
+    filters:
+       - blacklist: {kernel: ['v4.20-']}
```
gctucker: One issue with putting the filter here is that it will blacklist kernel revisions for all tests, whereas we only need to blacklist `igt` on kernels earlier than 4.20.

mgalka: Is it possible to use the filter feature for an individual test plan or is it something that needs to be implemented?

gctucker: I believe this is something that needs implemented, to add support for filters in test config objects.

gctucker:
This branch conflicts with #55 so it was skipped in today's tests on staging.  Maybe you could rebase it on top of #55 to fix this issue?

Also I think there should be another test config entry in the file, to run igt on rk3399-gru-kevin but with a filter using the kernel version.  That way, the filter only applies to igt and other tests can be run with all available builds.

test-configs.yaml
gctucker: Actually it was already there :) (in TestConfig)

gctucker:
Tested OK on staging with v4.14 to check the filters work as expected:
```
15:23:11 ('igt', False, False, 'arm64', True, {'kernel': 'v4.14-11011-ga39fa8b3e663', 'arch': 'arm64', 'lab': 'lab-collabora-dev', 'defconfig': u'defconfig'}, True, True)
15:23:11 test config did not match: rk3399-gru-kevin
15:23:11 ('igt', True, True, 'arm64', True, {'kernel': 'v4.14-11011-ga39fa8b3e663', 'arch': 'arm64', 'lab': 'lab-collabora-dev', 'defconfig': u'defconfig'}, True, False)
15:23:11 test config did not match: rk3399-gru-kevin
```
The debug information is:
```python
            print(
                plan,
                plan in self._test_plans,
                plan in self._test_plans and self._test_plans[plan].match(config),
                arch,
                self.device_type.arch == arch,
                config,
                self.device_type.match(flags, config),
                all(f.match(**config) for f in self._filters)
            )
```
It shows that the first case is the test config without igt, and the 2nd one has igt but the filter didn't match the kernel revision.

With mainline (v5.2-rc1), the igt test suite was run and the results can be found in the email report:
```
kernelci/staging.kernelci.org igt: 642 tests, 1 regressions (v5.2-rc1-173-g63c5b90e0c02)

Test results summary
--------------------

  Test:    igt
  Tree:    kernelci
  Branch:  staging.kernelci.org
  Kernel:  v5.2-rc1-173-g63c5b90e0c02
  URL:     https://github.com/kernelci/linux.git
  Commit:  63c5b90e0c0228416394938f4c7c63bc718ef3ea


1  | exynos5422-odroidxu3   | clang-8  | arm   | 215 total:  50 PASS  15 FAIL 150 SKIP
2  | rk3399-gru-kevin       | gcc-8    | arm64 | 106 total:  71 PASS   7 FAIL  28 SKIP
3  | exynos5422-odroidxu3   | gcc-8    | arm   | 216 total:  50 PASS  16 FAIL 150 SKIP
4  | rk3399-gru-kevin       | clang-8  | arm64 | 105 total:  71 PASS   7 FAIL  27 SKIP  
```

gctucker:
@mgalka As the `baseline` changes are taking a while to resolve, could you please rebase this again on top of master with just the igt-related changes?  It should just be adding a new test config entry, so this should not conflict with the baseline changes any more.

gctucker:
Tested OK on staging again with mainline, from the email report:
```
kernelci/staging.kernelci.org igt: 291 tests, 0 regressions (v5.2-rc2-1-ge2ecfac1544f)

Test results summary
--------------------

  Test:    igt
  Tree:    kernelci
  Branch:  staging.kernelci.org
  Kernel:  v5.2-rc2-1-ge2ecfac1544f
  URL:     https://github.com/kernelci/linux.git
  Commit:  e2ecfac1544f3308b920cb2f0a44d346d02b4044


1  | rk3288-veyron-jaq      | gcc-8    | arm   | 189 total:  76 PASS   6 FAIL 107 SKIP
2  | rk3399-gru-kevin       | clang-8  | arm64 | 102 total:  68 PASS   7 FAIL  27 SKIP  
```

PR #59, opened 2019-05-03
Add first ARC device: hsdk

montjoie:
This patchs adds support for the ARC architecture.
It adds the first boards using this arch, the HSDK

### DIFF
--- a/test-configs.yaml
+++ b/test-configs.yaml
@@ -440,6 +440,11 @@ device_types:
     class: arm64-dtb
     boot_method: uboot
 
+  hsdk:
+    mach: arc
+    class: arc-dtb
+    boot_method: uboot
+
   imx23_olinuxino:
     name: 'imx23-olinuxino'
     mach: imx
@@ -1087,6 +1092,9 @@ test_configs:
 
   - device_type: hikey_uboot
     test_plans: [boot, boot_nfs, kselftest, simple]
+
+  - device_type: hsdk
+    test_plans: [boot]
 
   - device_type: imx23_olinuxino
     test_plans: [boot, boot_nfs, kselftest, sleep]
--- a/kernelci/configs.py
+++ b/kernelci/configs.py
@@ -538,6 +538,14 @@ class DeviceType(YAMLObject):
         )
 
 
+class DeviceType_arc(DeviceType):
+
+    def __init__(self, name, mach, arch='arc', *args, **kw):
+        """arc device type with a device tree."""
+        kw.setdefault('dtb', '{}.dtb'.format(name))
+        super(DeviceType_arc, self).__init__(name, mach, arch, *args, **kw)
+
+
 class DeviceType_arm(DeviceType):
 
     def __init__(self, name, mach, arch='arm', *args, **kw):
@@ -566,6 +574,7 @@ class DeviceTypeFactory(YAMLObject):
     """Factory to create device types from YAML data."""
 
     _classes = {
+        'arc-dtb': DeviceType_arc,
         'mips-dtb': DeviceType_mips,
         'arm-dtb': DeviceType_arm,
         'arm64-dtb': DeviceType_arm64,


test-configs.yaml
```diff
@@ -7,6 +7,7 @@ file_system_types:
   buildroot:
     url: 'https://storage.kernelci.org/images/rootfs/buildroot/kci-2018.11'
     arch_map:
+      arc:     [{arch: arc}]
```
gctucker: While this should work, it's not actually necessary to add an entry if the arch name is the same for both the kernel and the distro conventions.

gctucker:
@montjoie Looks like this needs to be rebased again, then it'll get tested on staging if there's a device of that type available to run test jobs.

montjoie:
I have rebased, but tryong to use kernelci-core with it lead to
  File "./lava-v2-jobs-from-api.py", line 342, in <module>
    main(args)
  File "./lava-v2-jobs-from-api.py", line 282, in main
    config_data = kernelci.configs.tests_from_yaml(config.get('test_configs'))
  File "/home/montjoie/kernelci-core/kernelci/configs.py", line 901, in tests_from_yaml
    for name, device_type in data['device_types'].iteritems()
  File "/home/montjoie/kernelci-core/kernelci/configs.py", line 901, in <dictcomp>
    for name, device_type in data['device_types'].iteritems()
  File "/home/montjoie/kernelci-core/kernelci/configs.py", line 593, in from_yaml
    return device_cls(**kw)
  File "/home/montjoie/kernelci-core/kernelci/configs.py", line 570, in __init__
    super(DeviceType_arm64, self).__init__(name, mach, arch, *args, **kw)
TypeError: __init__() takes at least 5 arguments (6 given)

I dont understand why it fail on arm64

test-configs.yaml
```diff
@@ -442,6 +442,11 @@ device_types:
     name: 'hi6220-hikey'
     mach: hisilicon
     class: arm64-dtb
+
+  hsdk:
+    name: 'hsdk'
```
gctucker: The `name` attribute is not necessary when it's the same as the name of the YAML entry.

test-configs.yaml
```diff
@@ -442,6 +442,11 @@ device_types:
...
+    mach: arc
+    class: arc-dtb
     boot_method: uboot
```
gctucker: This `boot_method: uboot` is for the entry above `hikey_uboot`, the `hsdk` entry was added in the middle of it.

test-configs.yaml
gctucker: Still same comment.  It's a benign issue, but it's not necessary so it makes maintenance a bit easier to drop this `name` attribute.

gctucker:
@montjoie @khilman Do we need to build `hsdk_defconfig` for this board?

The `hsdk.dtb` is not built at the moment, we only seem to be getting only this one:
http://staging-storage.kernelci.org/kernelci/staging.kernelci.org/v5.2-rc4-46-g946d99caa45c/arc/nsim_hs_defconfig/gcc-8/dtbs/?C=M&O=D

montjoie:
I have rebased on top of master. with this PR I have sucessfully generated "stable" jobs against 5.0 and 5.1 with lava-v2-jobs-from-api.py.
Note that generated jobs still didnt boot with success but this is another problem.

montjoie:
ARC are still compiled with the old toolchain so that explains the "dont boot"

gctucker:
> @montjoie @khilman Do we need to build `hsdk_defconfig` for this board?
> 
> The `hsdk.dtb` is not built at the moment, we only seem to be getting only this one:
> http://staging-storage.kernelci.org/kernelci/staging.kernelci.org/v5.2-rc4-46-g946d99caa45c/arc/nsim_hs_defconfig/gcc-8/dtbs/?C=M&O=D

@khilman @montjoie Ah I see `hsdk_defconfig` is built normally, just not on staging due to the reduced number of builds there.  So I'll make a temporary patch on staging to enable that defconfig, and then we should start seeing some boot tests (which may fail, but that's a different issue).

gctucker:
Tested OK on staging: https://staging.kernelci.org/boot/id/5d14e4b9cf3a0f9a7a992bb7/

gctucker:
Except it failed to boot.  Is that expected, due to the toolchain maybe? @khilman @montjoie

montjoie:
Yes it is due to the toolchain, I need to investigate

gctucker:
OK, it's not breaking anything but it would of course be better to have at least one successful boot on staging before merging this.  That's to be sure it's all configured correctly but also to make it more useful.

montjoie:
I just tryed a kernel compiled by myself:
Linux version 5.2.0-rc5-next-20190620-00092-gd6d8f05c2d3c-dirty (clabbe@build2-bionic-1804) (gcc version 8.3.1 20190225 (ARCompact/ARCv2 ISA elf32 toolchain 2019.03-rc1)) #52 SMP PREEMPT Mon Jul 1 12:35:55 UTC 2019

So this confirm that the toolchain is the good one

montjoie:
i found a difference, in my job I see:
"Image Name:   Linux-5.2.0-rc5-next-20190620-00"
and  in one generated by kernelci:
"Image Name:"
So the problem is the kernelci handling of the kernel image.

gctucker:
Can you copy the image URL from your job here?

montjoie:
The problem is that kernelci set the image as "uimage.gz", so LAVA "convert" it and this destroy everything.

gctucker:
OK that worked, with the help of #109: https://staging.kernelci.org/boot/id/5d1dc0edcf3a0fbbf0f3dc50/

PR #60, opened 2019-05-04
build.py: mips: use uImage.gz build_target

khilman:
The 'make all' target doesn't actually build uImage.gz, which is the
image file needed to boot the current MIPS boards in
kernelCI (pistachio_marduk)

### DIFF
--- a/build.py
+++ b/build.py
@@ -365,10 +365,8 @@ extra_configs(dot_config, kbuild_output)
 #
 if len(args) >= 1:
     build_target = args[0]
-elif arch == "arc":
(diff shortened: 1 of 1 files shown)


gctucker:
@khilman I made a corresponding change in `kernelci/build.py` and got this build which seems OK:
https://staging.kernelci.org/build/id/5ccfe99bcf3a0f705b6576d8/

gctucker:
See #61 for the change in `kernelci.build`.

khilman:
@gctucker nice, thanks.

PR #63, opened 2019-05-07
kernelci.build: make all for mips

gctucker:
Revert to running "make all" for mips as not all defconfigs enable
uImage.gz to be built.


gctucker:
Tested OK on staging: https://staging.kernelci.org/build/id/5cd1c368cf3a0f4c9284acfa/
This defconfig doesn't enable `uImage.gz` and as a result no kernel image was uploaded.  The build was successful however.

PR #66, opened 2019-05-09
Add binutils-arm-linux-gnueabihf to clang-8 docker

mattface:
Install armv7 binutils to enable building armv7 kernels with the clang-8 container

Currently fails with:
/usr/bin/as: unrecognized option '-EL'
clang: error: assembler command failed with exit code 1 (use -v to see invocation)


mattface:
Looks like theres another PR for this https://github.com/kernelci/kernelci-core/pull/67 so lets use that

PR #69, opened 2019-05-10
dockerfiles/build-and-push.sh: replace -h option with -t for tag

gctucker:
Replace the -h option which was to provide the Docker Hub account name
with -t to provide a Docker tag prefix.  This essentially does the
same when pushing the tagged images as the beginning of the tag name
gives the account name, but it also fixes the case where a a longer
string prefix is required in the tag i.e. "-t kernelci/staging-".


gctucker:
I used it today to create the images on staging and it worked fine, here are all the builds using the docker images it produced:
https://staging.kernelci.org/build/kernelci/branch/staging.kernelci.org/kernel/v5.1-9366-g63f731a788f8/

PR #71, opened 2019-05-10
Fix a missing quotation mark in the linux-next recipients list

mattface:


gctucker:
Found that this typo was initially introduced by this commit:
```
commit 2fe6c8e527bb058083515bd4383c63888b6e2177
Author: Milo Casagrande <milo.casagrande@linaro.org>
Date:   Wed Jun 22 09:58:01 2016 +0200

    Send txt only emails to vger mailing lists
```
So it's been there for a while :)

But yeah, LGTM.

PR #73, opened 2019-05-10
Revert "build.py: mips: use uImage.gz build_target"

khilman:
This was a mistake as not all MIPS defconfigs actually have a
uImage.gz target.  Switch back to 'make all'

This reverts commit c781f6d72d983334e375371fb31429c6a03919dc.

### DIFF
--- a/build.py
+++ b/build.py
@@ -365,8 +365,10 @@ extra_configs(dot_config, kbuild_output)
 #
 if len(args) >= 1:
     build_target = args[0]
-elif arch == "arc" or arch == "mips":
+elif arch == "arc":
     build_target = "uImage.gz"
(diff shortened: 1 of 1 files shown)


khilman:
Yes, I know that this build.py isn't used anymore here, but this at least keeps it in an equivalent state as tne new kci_build tooling.

gctucker:
@khilman fine - can we remove `build.py` after this PR?

I meant to create a PR to remove `build.py` but thought we should first wait a bit to see how things were going in production with `kci_build`.

khilman:
Yes, fine with me to remove it after this PR.

gctucker:
ack

PR #75, opened 2019-05-15
dockerfiles: arc: update toolchain to 2019.03-rc1

montjoie:
Current toolchain produces unbootable kernel.
When using the new 2019.03-rc1, I produces working kernel.

### DIFF
--- a/build-configs.yaml
+++ b/build-configs.yaml
@@ -153,7 +153,7 @@ build_environments:
       x86_64: 'x86'
       riscv: 'riscv64'
     cross_compile: &default_cross_compile
-      arc: 'arc-linux-'
+      arc: 'arc-elf32-'
       arm: 'arm-linux-gnueabihf-'
       arm64: 'aarch64-linux-gnu-'
       mips: 'mips-linux-gnu-'
--- a/jenkins/dockerfiles/gcc-8_arc/Dockerfile
+++ b/jenkins/dockerfiles/gcc-8_arc/Dockerfile
@@ -4,6 +4,6 @@ FROM kernelci/build-base
 # pre-built toolchains for ARC available here:
 #    https://github.com/foss-for-synopsys-dwc-arc-processors/toolchain/releases/
 #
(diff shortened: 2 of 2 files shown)


mattface:
I rebuilt this container and tried an ARC build but it failed with:
```
root@3a20dc6028a8:/linux# make KBUILD_BUILD_USER=KernelCI -C/linux -s ARCH=arc CROSS_COMPILE=arc-linux- HOSTCC=gcc CC="ccache arc-linux-gcc" O=build vdk_hs38_defconfig
ccache: error: Could not find compiler "arc-linux-gcc" in PATH
ccache: error: Could not find compiler "arc-linux-gcc" in PATH
ccache: error: Could not find compiler "arc-linux-gcc" in PATH
ccache: error: Could not find compiler "arc-linux-gcc" in PATH
ccache: error: Could not find compiler "arc-linux-gcc" in PATH
ccache: error: Could not find compiler "arc-linux-gcc" in PATH
ccache: error: Could not find compiler "arc-linux-gcc" in PATH
ccache: error: Could not find compiler "arc-linux-gcc" in PATH
ccache: error: Could not find compiler "arc-linux-gcc" in PATH
ccache: error: Could not find compiler "arc-linux-gcc" in PATH
ccache: error: Could not find compiler "arc-linux-gcc" in PATH
```

It looks like the cross-compile option needs changing in the build scripts, or some expected binaries are now missing

montjoie:
Yes I forgot CROSS_COMPILE change

build.py
```diff
@@ -56,7 +56,7 @@
     "i386": None,
     "x86": None,
     "x86_64": None,
-    "arc": "arc-linux-",
+    "arc": "arc-elf32-",
```
gctucker: `build.py` is not used any more.  See `kernelci/build.py` instead.

gctucker: Ah the other change in `build-configs.yaml` should be used in `kernelci/build.py` so that should be fine.

gctucker:
This worked on staging: https://staging.kernelci.org/build/id/5ced5834cf3a0f1147ce3728/

PR #76, opened 2019-05-16
Add hi6220-hikey with uboot to test-configs

mattface:
lab-mhart has a hikey which is driven by LAVA as a standard uboot device. Add this to test-configs so that it can start receiving jobs.

Tested jobs are created as expected.


mattface:
Tested on staging and booted the kci linux tree, which was recently rebased to mainline.

PR #77, opened 2019-05-16
Add extra android trees

mattface:
Add android-4.19-gki-dev and android-mainline branches to build-configs.

Extra GCE builders have been added to production jenkins to help with the load these may add.


mattface:
Tested on staging branch and booted kci linux branch which was recently rebased to mainline.

PR #85, opened 2019-06-05
test-configs.yaml: Add support for the meson-gxm-khadas-vim2

montjoie:
We have a meson-gxm-khadas-vim2 in lab-baylibre, ready to accept jobs.

### DIFF
--- a/test-configs.yaml
+++ b/test-configs.yaml
@@ -591,6 +591,14 @@ device_types:
     filters:
       - blacklist: {defconfig: ['allnoconfig', 'allmodconfig']}
 
+  meson-gxm-khadas-vim2:
+    mach: amlogic
+    class: arm64-dtb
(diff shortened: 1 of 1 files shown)


test-configs.yaml
```diff
@@ -591,6 +591,15 @@ device_types:
     filters:
       - blacklist: {defconfig: ['allnoconfig', 'allmodconfig']}
 
+  meson_gxm_khadas_vim2:
+    name: 'meson-gxm-khadas-vim2'
```
gctucker: You can just call it `meson-gxm-khadas-vim2` and drop the `name` attribute.

gctucker:
Tested OK on staging: https://staging.kernelci.org/boot/id/5cf9958ccf3a0f9a35fff6f9/

PR #88, opened 2019-06-05
test-configs: fix r8a7795-salvator-x boot method

montjoie:
While doing some statistics about our lab, I found that the salavtor-x
didnt worked since a long time.
The boot-method of r8a7795-salvator-x is wrongly set as qemu.
Since this board is not a qemu device, this is wrong.


gctucker:
LGTM, will get tested on staging to confirm.

gctucker:
Tested OK on staging: https://staging.kernelci.org/boot/id/5cf99594cf3a0f9a50fff6c2/

PR #90, opened 2019-06-05
test-configs: Add cubietruck

montjoie:
The cubietruck is lab-seattle since a long time and dont get any job.
This patchs (re?)add it.

### DIFF
--- a/test-configs.yaml
+++ b/test-configs.yaml
@@ -381,6 +381,12 @@ device_types:
       - blacklist: *allmodconfig_filter
       - blacklist: {defconfig: ['multi_v7_defconfig+CONFIG_LKDTM=n']}
 
+  cubietruck:
+    mach: allwinner
+    class: arm-dtb
+    dtb: "sun7i-a20-cubietruck.dtb"
+    boot_method: uboot
+
   d03:
     name: 'd03'
     mach: hisilicon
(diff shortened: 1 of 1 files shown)


test-configs.yaml
```diff
@@ -381,6 +381,12 @@ device_types:
       - blacklist: *allmodconfig_filter
       - blacklist: {defconfig: ['multi_v7_defconfig+CONFIG_LKDTM=n']}
 
+  cubietruck:
+    name: 'sun7i-a20-cubietruck'
```
gctucker: I think the upstream device type name is `cubietruck`:
https://git.lavasoftware.org/lava/lava/blob/master/lava_scheduler_app/tests/device-types/cubietruck.jinja2#L1

It's true that using the dt name is better, but that would need to be done using the new LAVA device type alias feature.  Otherwise, I think it has to be `cubietruck`.

montjoie:
I confirm that the devicetype is cubietruck
./lava_scheduler_app/tests/device-types/cubietruck.jinja2

test-configs.yaml
montjoie: With your change, kcicore does not generate a job, where I need to set the DT name for let kcicore find it?

montjoie: The linux DT name is sun7i-a20-cubietruck

montjoie: oh i just saw dtb:

gctucker:
Tested OK on staging: https://staging.kernelci.org/boot/id/5cf9958fcf3a0f9a35fff6fa/

PR #91, opened 2019-06-05
chrome-platform build filters

gctucker:
Add definitions to build only the main defconfig of each architecture, and use it in chrome-platform as requested by @eballetbo.


gctucker:
Tested locally:
```
$ ./kci_build list_kernel_configs --config=chrome-platform --kdir=linux
x86_64 x86_64_defconfig gcc-8
arm multi_v7_defconfig gcc-8
arm64 defconfig gcc-8
```

eballetbo:
LGTM, thanks to support this

PR #93, opened 2019-06-13
Increase deploy timeout

montjoie:
On lab-baylibre, we have many slaves running on potato.
While they are still powerful enough for running lot of jobs in
parallel, they lack a bit on IO.
So while running many deploy in parallel, they hit sometimes the deploy timeout.

This patch increases the deploy timeout, 2 minutes is too short.


gctucker:
LGTM, will get that tested on staging.

gctucker:
Tested OK on staging, all the jobs seem to have been generated correctly.

PR #96, opened 2019-06-18
kernelci.build: fix initial git clone with no reference

gctucker:
Fix the case where a new git repo needs to be initialised and no
reference repo/mirror is provided.


gctucker:

PR #97, opened 2019-06-18
Only send build reports for next/master to clang-built-linux

mattface:
Currently next/pending-fixes don't have clang results, so don't send the build reports email to the clang-built-linux mailing list.


nickdesaulniers:
thanks @mattface !

gctucker:
LGTM, testing on staging now to check it doesn't break the Jenkins job.

PR #98, opened 2019-06-19
build-configs.yaml: Update agross repo and branch

andersson:
Andy's tree has moved, so update the build config in line with [1]. Also
update the branch which is monitored to allow for running the CI step
before the tree lands in linux-next (the way we though it was setup
already).

[1] https://lore.kernel.org/linux-arm-msm/1559936691-15759-1-git-send-email-agross@kernel.org/

Signed-off-by: Bjorn Andersson <bjorn.andersson@linaro.org>

### DIFF
--- a/build-configs.yaml
+++ b/build-configs.yaml
@@ -5,7 +5,7 @@
 trees:
 
   agross:
-    url: "https://git.kernel.org/pub/scm/linux/kernel/git/agross/linux.git"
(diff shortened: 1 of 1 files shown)


gctucker:
Tested OK on staging up to the point where the source tarball was created and kernel builds were scheduled (than I cancelled them).

PR #99, opened 2019-06-19
Find LAVA devices by aliases, when available

danrue:
In addition to looking up available device_types in each lava instance,
also look for aliases.

Create a new lookup function, get_device_type_by_name(), to simplify the
lookup. Provide 6 unit tests for the new function.

Example run:

    ./lava-v2-submit-jobs.py --username dan.rue --token='XXX' --lab lab-linaro-lkft --jobs foo
    Loading jobs from foo
    LAVA API: https://lkft.validation.linaro.org/RPC2/
    Connecting to Server...
    Connection Successful!
    connect-to-server : pass
    Fetching all devices from LAVA
    Fetching all device-types from LAVA
    Fetching all device-type aliases from LAVA
    Submitting Jobs to Server...
    Submitting job kernelci-staging.kernelci.org-v5.2-rc5-178-gdac6ebf0e111-arm-multi_v7_defconfig-gcc-8-no-dtb-qemu-boot to device-type qemu
    Submitting job arm-soc-for-next-v5.2-rc5-110-ge57f4f2c4506-arm-multi_v7_defconfig-gcc-8-am57xx-beagle-x15.dtb-am57xx-beagle-x15-simple to device-type am57xx-beagle-x15
    Submitting job lkft-ltp-quickhit-master-1894 to device-type x15

### DIFF
--- a/kernelci/lava.py
+++ b/kernelci/lava.py
@@ -1,0 +1,48 @@
+#!/usr/bin/python
+#
+# Copyright (C) 2019 Linaro Limited
+# Author: Dan Rue <dan.rue@linaro.org>
+#
+# This module is free software; you can redistribute it and/or modify it under
+# the terms of the GNU Lesser General Public License as published by the Free
+# Software Foundation; either version 2.1 of the License, or (at your option)
+# any later version.
+#
+# This library is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public License for more
+# details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this library; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+
+
+def get_device_type_by_name(name, device_types, aliases=[]):
+    """
+        Return a device type named name. In the case of an alias, resolve the
+        alias to a device_type
+
+        Example:
+        IN:
+            name = "x15"
+            device_types = [
+              {'busy': 1, 'idle': 0, 'name': 'x15', 'offline': 0},
+              {'busy': 1, 'idle': 0, 'name': 'beaglebone-black', 'offline': 0},
+            ]
+            aliases = [
+              {'name': 'am57xx-beagle-x15', 'device_type': 'x15'}
+            ]
+        OUT:
+            {'busy': 1, 'idle': 0, 'name': 'x15', 'offline': 0}
+
+    """
+    for device_type in device_types:
+        if device_type["name"] == name:
+            return device_type
+    for alias in aliases:
+        if alias["name"] == name:
+            for device_type in device_types:
+                if alias["device_type"] == device_type["name"]:
+                    return device_type
+    return None
--- a/tests/test_lava.py
+++ b/tests/test_lava.py
@@ -1,0 +1,55 @@
+import kernelci.lava
+import pytest
+
+
+def test_get_device_type_by_name_1():
+    assert kernelci.lava.get_device_type_by_name("some_board", []) is None
+
+
+def test_get_device_type_by_name_2():
+    assert kernelci.lava.get_device_type_by_name("some_board", [], []) is None
+
+
+@pytest.fixture
+def device_types():
+    device_types = [
+        {"busy": 1, "idle": 0, "name": "x15", "offline": 0},
+        {"busy": 1, "idle": 0, "name": "beaglebone-black", "offline": 0},
+    ]
+    return device_types
+
+
+@pytest.fixture
+def aliases():
+    aliases = [{"name": "am57xx-beagle-x15", "device_type": "x15"}]
+    return aliases
+
+
+def test_get_device_type_by_name_3(device_types, aliases):
+    assert (
+        kernelci.lava.get_device_type_by_name("some_board", device_types,
+                                              aliases) is None
+    )
+
+
+def test_get_device_type_by_name_4(device_types, aliases):
+    assert (
+        kernelci.lava.get_device_type_by_name("x15", device_types, aliases)
+        == device_types[0]
+    )
+
+
+def test_get_device_type_by_name_5(device_types, aliases):
+    assert (
+        kernelci.lava.get_device_type_by_name(
+            "am57xx-beagle-x15", device_types, aliases
+        )
+        == device_types[0]
+    )
+
+
+def test_get_device_type_by_name_6(device_types, aliases):
+    assert (
+        kernelci.lava.get_device_type_by_name("beaglebone-black", device_types)
+        == device_types[1]
+    )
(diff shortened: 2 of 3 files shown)


kernelci/lava.py
```diff
@@ -0,0 +1,50 @@
...
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public License for more
+# details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this library; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+
+
+def get_device_type_by_name(name, device_types, aliases=[]):
```
danrue: I wasn't sure where to stick this function. I had to put it in lib/ or kernelci/ to be able to write tests against it.

kernelci/lava.py
```diff
@@ -0,0 +1,50 @@
...
+            ]
+        OUT:
+            {'busy': 1, 'idle': 0, 'name': 'am57xx-beagle-x15', 'offline': 0}
+
+    """
+    for device_type in device_types:
+        if device_type["name"] == name:
+            return device_type
+    for alias in aliases:
+        if alias["name"] == name:
```
gctucker: Do we actually have to care about how the alias is called in the lab?  Is it not enough to check that there is one alias with a `device_type` that matches the `name`?  Here we're comparing `am57xx-beagle-x15` with `x15` so it looks like something isn't quite right.  (see my previous comment as I'm not sure which way round this should be)

kernelci/lava.py
gctucker: That's fine, the scripts to generate and submit jobs need to be consolidated in the same way as `kci_build`.  So yes `kernelci.lava` sounds like a logic place to store all things LAVA-related.

kernelci/lava.py
```diff
@@ -0,0 +1,50 @@
...
+        IN:
+            name = "x15"
+            device_types = [
+              {'busy': 1, 'idle': 0, 'name': 'am57xx-beagle-x15', 'offline': 0},
+              {'busy': 1, 'idle': 0, 'name': 'beaglebone-black', 'offline': 0},
+              {'busy': 0, 'idle': 1, 'name': 'docker', 'offline': 0},
+              {'busy': 0, 'idle': 1, 'name': 'qemu', 'offline': 0}
+            ]
+            aliases = [
+              {'name': 'x15', 'device_type': 'am57xx-beagle-x15'}
```
gctucker: Are you sure it's not the other way round i.e. `{'name': 'am57xx-beagle-x15', 'device_type': 'x15'}` to match the code in the function?

My point is that if you call the device `x15` in your lab, you need an alias so `am57xx-beagle-x15` also works.  I see the LKFT lab are doing this, using `x15` as the device type with an `am57xx-beagle-x15` alias:
https://lkft.validation.linaro.org/scheduler/device_type/x15

tests/test_lava.py
```diff
@@ -0,0 +1,41 @@
+import kernelci.lava
+import pytest
+
+
+def test_get_device_type_by_name_1():
```
gctucker: Please add some blank lines between each function, ideally to pass all the `pep8` checks.

danrue: Are the checks you expect developers to run documented or automated? What check specifically should I run to ensure compliance? Incidentally, I vote for 'black' when it comes to python formatting. It does create a lot of whitespace in tests though, as it puts two empty lines between each function.

kernelci/lava.py
danrue: My example is backwards, I agree. I can change that so this is less confusing. But I think the code is correct - test case test_get_device_type_by_name_4 hits the code path that resolves an alias to a device_type.

danrue:
In the latest push, I have changed around the examples and tests to make more sense from a kernelci perspective, but the implementation has not changed. I also ran 'black' against tests/test_lava.py and kernelci/lava.py which is close to pep8 compliant.

kernelci/lava.py
gctucker: OK thanks for making the example in the docstring more like a typical KernelCI case.  That makes sense now, it's not just about the device type name as a string but the rest of the dictionary is needed to submit jobs to online devices etc...

tests/test_lava.py
gctucker: I was referring to the usual coding standard for Python [1] as I thought that's what had been adopted by default by the Python community.  If you want to run "black" instead then why not, I'm not familiar with it.

As far as this PR is concerned, `pep8` only reports these things with the newly added files:
```
$ pep8 tests/test_lava.py kernelci/lava.py 
tests/test_lava.py:6:68: E711 comparison to None should be 'if cond is None:'
tests/test_lava.py:10:72: E711 comparison to None should be 'if cond is None:'
tests/test_lava.py:30:80: E501 line too long (82 > 79 characters)
tests/test_lava.py:31:9: E711 comparison to None should be 'if cond is None:'
```

The comparison with `None` seems fair since there's only one `None` object in Python.  Regarding line lengths, there's no absolute solution but I personally prefer to follow the 79 characters rule.

So for this change maybe fix the `None` items, and the 79 characters if you agree with them, and then create a PR or Github issue to discuss the topic to get an agreement going forward?

[1] https://www.python.org/dev/peps/pep-0008/

kernelci/lava.py
gctucker: OK thanks that makes sense now.

gctucker:
@danrue Are there currently any devices in your LAVA lab, or in any lab that you know of, that would make use of the alias feature to verify this works as expected on staging?

danrue:
@gctucker the test is the lkft lab. If alias is working as expected, kernelci should submit jobs to the x15 devices in lkft.validation.linaro.org.

tests/test_lava.py
danrue: Thanks - fixed all those things and updated the PR. I'll follow up in a separate PR with either pycodestyle (pep8) or black integration.

gctucker: Thanks!

gctucker:
OK, so we should start seeing some jobs here soon:
https://lkft.validation.linaro.org/scheduler/device_type/x15?dt_length=10&dt_search=staging.kernelci.org#dt_

There aren't any at the moment but that's because the `kernelci` tree used on staging is currently filtered out for `lab-linaro-lkft`.  So I'll make a PR to enable that and try again.

gctucker:
See #102 about enabling jobs to run with kernels from the `kernelci` tree.

gctucker:
Seems to have worked:
https://lkft.validation.linaro.org/scheduler/job/797680
https://lkft.validation.linaro.org/scheduler/job/797677

gctucker:
I also checked that the unit tests are passing:
```
$ make test
# Note that just running 'pytest' will not work here because there is no setup.py. See
# https://docs.pytest.org/en/latest/pythonpath.html#invoking-pytest-versus-python-m-pytest
python -m pytest
===================================== test session starts ======================================
platform linux2 -- Python 2.7.13, pytest-3.0.6, py-1.4.32, pluggy-0.4.0
rootdir: /home/gtucker/project/kernelci/kernelci-staging-tools/checkout/kernelci-core, inifile: 
collected 10 items 

tests/test_configs.py ....
tests/test_lava.py ......

================================== 10 passed in 0.18 seconds ===================================
```

PR #100, opened 2019-06-19
add UEFI test plan

roxell:
Initital test plan for UEFI boot testing.
Run the test in qemu in these combinations:

- host arm64 guest arm64 enable gic-version=2 and 3.
- host arm64 guest arm enable gic-version=2 and 3.
- host x86_64 guest x86_64.
- host x86_i386 guest x86_i386.
- host x86_64 guest x86_i386.

### DIFF
--- a/templates/uefi/generic-qemu-uefi-boot-template.jinja2
+++ b/templates/uefi/generic-qemu-uefi-boot-template.jinja2
@@ -1,0 +1,62 @@
+{% extends 'base/kernel-ci-base.jinja2' %}
+{% set ctx_arch = arch %}
+{% if arch == "arm" %}
+{% set ctx_arch = 'arm64' %}
+{% set ctx_cpu = 'cortex-a15' %}
+{% set console_dev = 'ttyAMA0' %}
+{% endif %}
+{% if arch == "arm64" %}
+{% set ctx_arch = 'arm64' %}
+{% set console_dev = 'ttyAMA0' %}
+{% endif %}
+{% if arch == 'x86_64' %}
+{% set ctx_arch = 'amd64' %}
+{% set console_dev = 'ttyS0' %}
+{% endif %}
+{% if arch == 'i386' %}
+{% set ctx_arch = 'i386' %}
+{% set ctx_cpu = 'host' %}
+{% set console_dev = 'ttyS0' %}
+{% endif %}
+{% block metadata %}
+{{ super() }}
+{% endblock %}
+{% block main %}
+{{ super() }}
+{% endblock %}
+{% block actions %}
+context:
+  arch: {{ ctx_arch }}
+  cpu: {{ ctx_cpu }}
+  guestfs_interface: virtio
+{% if arch == 'arm' or arch == 'arm64' %}
+  machine: virt,{{ ctx_gicversion }}
+{% endif %}
+
+actions:
+- deploy:
+    timeout:
+      minutes: 3
+    to: tmpfs
+    os: oe
+    images:
+      bios:
+        image_arg: '-bios {bios}'
+        url: {{ image_url_uefi }}
+      kernel:
+{%- block image_arg %}
+        image_arg: '-kernel {kernel} -append "console={{ console_dev }},115200 root=/dev/ram0 debug verbose {{ extra_kernel_args }}"'
+{%- endblock %}
+        url: {{ kernel_url }}
+      ramdisk:
+        image_arg: '-initrd {ramdisk}'
+        url: {{ initrd_url }}
+
+- boot:
+    timeout:
+      minutes: 5
+    method: qemu
+    media: tmpfs
+    prompts:
+      - '{{ rootfs_prompt }}'
+{% endblock %}
--- a/test-configs.yaml
+++ b/test-configs.yaml
@@ -160,6 +160,72 @@ test_plans:
             - 'defconfig+virtualvideo'
             - 'multi_v7_defconfig+virtualvideo'
 
+  uefi-arm64-gicv2:
+    rootfs: buildroot_ramdisk
+    params:
+        ctx_gicversion: "gic-version=2"
+        ctx_cpu: "cortex-a57"
+        image_url_uefi: 'https://staging-storage.kernelci.org/images/uefi/111bbcf87621/QEMU_EFI.fd-AARCH64-RELEASE-111bbcf87621'
+    pattern: 'uefi/generic-qemu-uefi-boot-template.jinja2'
+    filters:
+      - whitelist: {defconfig: ['defconfig']}
+
+  uefi-arm64-gicv3:
+    rootfs: buildroot_ramdisk
+    params:
+        ctx_gicversion: "gic-version=3"
+        ctx_cpu: "cortex-a57"
+        image_url_uefi: 'https://staging-storage.kernelci.org/images/uefi/111bbcf87621/QEMU_EFI.fd-AARCH64-RELEASE-111bbcf87621'
+    pattern: 'uefi/generic-qemu-uefi-boot-template.jinja2'
+    filters:
+      - whitelist: {defconfig: ['defconfig']}
+
+  uefi-arm-gicv2:
+    rootfs: buildroot_ramdisk
+    params:
+        ctx_gicversion: "gic-version=2"
+        ctx_cpu: "cortex-a15"
+        image_url_uefi: 'https://staging-storage.kernelci.org/images/uefi/111bbcf87621/QEMU_EFI.fd-ARM-RELEASE-111bbcf87621'
+    pattern: 'uefi/generic-qemu-uefi-boot-template.jinja2'
+    filters:
+      - whitelist: {defconfig: ['multi_v7_defconfig']}
+
+  uefi-arm-gicv3:
+    rootfs: buildroot_ramdisk
+    params:
+        ctx_gicversion: "gic-version=3"
+        ctx_cpu: "cortex-a15"
+        image_url_uefi: 'https://staging-storage.kernelci.org/images/uefi/111bbcf87621/QEMU_EFI.fd-ARM-RELEASE-111bbcf87621'
+    pattern: 'uefi/generic-qemu-uefi-boot-template.jinja2'
+    filters:
+      - whitelist: {defconfig: ['multi_v7_defconfig']}
+
+  uefi-x86_64:
+    rootfs: buildroot_ramdisk
+    params:
+        ctx_cpu: "host"
+        image_url_uefi: 'https://staging-storage.kernelci.org/images/uefi/111bbcf87621/OVMF.fd-X64-RELEASE-111bbcf87621'
+    pattern: 'uefi/generic-qemu-uefi-boot-template.jinja2'
+    filters:
+      - whitelist: {defconfig: ['defconfig']}
+
+  uefi-x86_i386:
+    rootfs: buildroot_ramdisk
+    params:
+        ctx_cpu: "host"
+        image_url_uefi: 'https://staging-storage.kernelci.org/images/uefi/111bbcf87621/OVMF.fd-IA32-RELEASE-111bbcf87621'
+    pattern: 'uefi/generic-qemu-uefi-boot-template.jinja2'
+    filters:
+      - whitelist: {defconfig: ['i386_defconfig']}
+
+  uefi-x86-mixed:
+    rootfs: buildroot_ramdisk
+    params:
+        ctx_cpu: "i386"
+        image_url_uefi: 'https://staging-storage.kernelci.org/images/uefi/111bbcf87621/OVMF.fd-IA32-RELEASE-111bbcf87621'
+    pattern: 'uefi/generic-qemu-uefi-boot-template.jinja2'
+    filters:
+      - whitelist: {defconfig: ['i386_defconfig']}
 
 device_default_filters:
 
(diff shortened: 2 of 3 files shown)


templates/uefi/generic-qemu-uefi-boot-template.jinja2
```diff
@@ -0,0 +1,72 @@
+{% extends 'base/kernel-ci-base.jinja2' %}
+{% set ctx_arch = arch %}
+{% if arch == "arm" %}
+{% set ctx_arch = 'arm64' %}
+{% set ctx_cpu = 'cortex-a15' %}
+{% set console_dev = 'ttyAMA0' %}
+{% set image_url_uefi = 'http://people.linaro.org/~ard.biesheuvel/kernelci/QEMU_EFI.fd-ARM-RELEASE-325ad6226099' %}
```
danrue: We probably need to host these files somewhere on storage.kernelci.org, yea?

roxell: I'm not convinced, there's both pros and cons...
But please convince me =)

gctucker: I think this should be a parameter defined in `test-configs.yaml`, which is a site-specific config file unlike test templates.

roxell: I think we should start using Ard's firmware binaries and put them onto storage.kernelci.org/uefi/...
Today Linaro does releases for arm and aarch64 not for x64 or ia32 [1]. However, I got promised that they will add those to the next release that will happen 19.08. On the release page there is documented how stuff is built.
If  we just stick a file next to the storage.kernelci.org/uefi/19.08/build-info.txt and pint to the release page or something that should be enough for reproducing the build if needed.

[1] http://releases.linaro.org/reference-platform/enterprise/firmware/open-source/19.03/

test-configs.yaml
```diff
@@ -146,6 +146,83 @@ test_plans:
             - 'defconfig+virtualvideo'
             - 'multi_v7_defconfig+virtualvideo'
 
+  uefi-arm64-gicv2:
+    rootfs: buildroot_ramdisk
+    params: {ctx_gicversion: "gic-version=2", ctx_cpu: "cortex-a57",
```
gctucker: There's no need to use curly brackets for a large dictionary, it's easier to read with the indented syntax:
```yaml
    params:
      ctx_gicversion: "gic-version=2"
      ctx_cpu: "cortex-a57",
      image_url_uefi: 'http://people.linaro.org/~ard.biesheuvel/kernelci/QEMU_EFI.fd-AARCH64-RELEASE-325ad6226099'
```

test-configs.yaml
```diff
@@ -146,6 +146,83 @@ test_plans:
...
+             image_url_uefi: 'http://people.linaro.org/~ard.biesheuvel/kernelci/QEMU_EFI.fd-AARCH64-RELEASE-325ad6226099'
```
gctucker: I've copied these files on the staging storage server to test this on staging initially:
https://staging-storage.kernelci.org/images/uefi/325ad6226099/

test-configs.yaml
```diff
@@ -146,6 +146,83 @@ test_plans:
...
+            }
+    pattern: 'uefi/generic-qemu-uefi-boot-template.jinja2'
+    filters:
+      - whitelist:
```
gctucker: Well, for this kind of single-value dictionary, the curly bracket is more appropriate :)

```yaml
    filters:
```

templates/uefi/generic-qemu-uefi-boot-template.jinja2
```diff
@@ -0,0 +1,62 @@
...
+{% set console_dev = 'ttyS0' %}
+{% endif %}
+{% block metadata %}
+{{ super() }}
+{% endblock %}
+{% block main %}
+{{ super() }}
+{% endblock %}
+{% block actions %}
+context:
```
gctucker: This will conflict with the other changes to make the context more flexible, see #108.  Some rebase will be needed, either this PR or the other one depending on which one gets merged first.  It's OK to leave it as-is for now to test it on staging.

gctucker:
@roxell You need to also add the new UEFI test plans in `jenkins/lava-boot-v2.sh` for the LKFT lab.  Otherwise, no jobs are being generated.

test-configs.yaml
```diff
@@ -787,6 +853,14 @@ device_types:
     filters:
       - whitelist: {defconfig: ['defconfig']}
 
+  qemu_i386:
```
gctucker: This one should be before `qemu_x86_64` to keep the list in alphabetical order.

test-configs.yaml
```diff
@@ -1210,13 +1284,16 @@ test_configs:
     test_plans: [boot]
 
   - device_type: qemu_arm
-    test_plans: [boot_qemu, simple_qemu, v4l2_compliance_qemu_vivid]
+    test_plans: [boot_qemu, simple_qemu, v4l2_compliance_qemu_vivid, uefi-arm-gicv2, uefi-arm-gicv3]
```
gctucker: Might be worth using the expanded list syntax as this is starting to grow (easier to diff and maintain imho):
```yaml
    test_plans:
      - boot_qemu
      - simple_qemu
      - v4l2_compliance_qemu_vivid
      - uefi-arm-gicv2
      - uefi-arm-gicv3
```

test-configs.yaml
```diff
@@ -1210,13 +1284,16 @@ test_configs:
...
 
   - device_type: qemu_arm64
-    test_plans: [boot_qemu, simple_qemu, v4l2_compliance_qemu_vivid]
+    test_plans: [boot_qemu, simple_qemu, v4l2_compliance_qemu_vivid, uefi-arm64-gicv2, uefi-arm64-gicv3]
```
gctucker: Ditto

templates/uefi/generic-qemu-uefi-boot-template.jinja2
```diff
@@ -0,0 +1,62 @@
+{% extends 'base/kernel-ci-base.jinja2' %}
```
gctucker: Note: This template is effectively a boot test, because it doesn't have any test cases.  It's OK for now but the results will probably get lost in the noise until boot tests get tidied up.  What we need to have is the `baseline` test plan running on all the platforms, then we can run variants such as these UEFI cases and all the other kinds of boots (ramdisk, nfs...).

ardbiesheuvel:
> Then the next thing to resolve is to have a documented way to rebuild these UEFI binaries, ideally to have a Jenkins Pipeline job to create them directly on the kernelci.org servers or at the very least some script and where to get the source code from. If in practice this really can't be built by anyone else but @ardbiesheuvel then the bare minimal would be a manifest file with the URL and revision of the source code repository, and maybe any config choices that were made or build environment properties (toolchain...).

I shared some instructions with @roxell on how to rebuild the UEFI binaries. However, it should be made very clear that testing different upstream versions of EDK2 is outside of the scope of this testing effort. Therefore, we should choose a baseline now (I suggest the latest EDK2 stable tag), and only ever upgrade the firmware binaries if there is a pressing need to do so.

gctucker:
Thanks, we'll follow up with @roxell.

Sure, this is about testing that upstream kernels can be booted with a known good version of UEFI rather than doing some kind of UEFI CI.

Being able to reproduce these binaries is important for KernelCI to be able to remove a dependency on an external supplier, and also to be able to trace all the way back to which source revisions were used and how they were built.

roxell:
I'm confused, why do we bring this up again!
I thought we agreed on to do what I wrote here [1], since we resolve that.

And I really think that is what we should do. Use whats already there. also if people want to rebuild they can look there.

[1] https://github.com/kernelci/kernelci-core/pull/100#discussion_r299985936

leiflindholm:
With regards to reproducibility, the Linaro (Date Centre & Cloud Group) firmware releases are built using
https://git.linaro.org/uefi/rpf.git/tree/build.sh
It does sometimes include additional patches on top of the upstream trees (never carried across multiple releases), but they are all held in this repo.

For the 2019.08 release (which may spill over into September, depending on how late in August upstream edk2 stable tag is made), we will be switching to the Debian packaged GCC 8.3 for all architectures.

At this point, the only reproducibility issue I can think of is the version of iasl, from acpica-tools. We occasionally run into changes requiring to install a newer version of that package. We do list the specific version in the release notes: (e.g. end of http://releases.linaro.org/reference-platform/enterprise/firmware/open-source/19.03/), but I could put a proper tool version matrix closer to the start of the next release notes.

ardbiesheuvel:
@leiflindholm is making a good point: if you build your own firmware binaries, you will need to do your own validation as well. Without any validation of the firmware binaries themselves, the relevance of your findings in performing CI on the Linux kernel using these binaries will be significantly reduced.

broonie:
The concern with source originates from being able to tell people who wanted to trace issues into the firmware how to do that - it's been an issue at times in the past with binary compiler releases that we've used where end up scratching our heads about what was actually in the binary. Using our own builds is a convenient way to do that, but if validation is a concern I guess we should just make sure we have something clear recorded on storage.kernelci.org (and get that into place before merging this).

gctucker:
For example, having these things alongside the firmware binaries would be good:
* the build script used to produce the binaries
* the build log
* a file mentioning the git url and revision of the sources used
* the docker image or any information about the version of the build tools involved
* maybe a source tarball with patch files, esp if the git branch is going to be rebased or removed
* and maybe some validation report

templates/uefi/generic-qemu-uefi-boot-template.jinja2
gctucker: @roxell Please rebase this PR now that #108 has been merged.

gctucker:
@roxell Please rebase this PR now that #108 has been merged.

gctucker:
@roxell I've copied the new binaries here:
https://staging-storage.kernelci.org/images/uefi/111bbcf87621/

Please update the PR accordingly.

test-configs.yaml
```diff
@@ -160,6 +160,72 @@ test_plans:
             - 'defconfig+virtualvideo'
             - 'multi_v7_defconfig+virtualvideo'
 
+  uefi-arm64-gicv2:
+    rootfs: buildroot_ramdisk
+    params:
+        ctx_gicversion: "gic-version=2"
+        ctx_cpu: "cortex-a57"
```
gctucker: The `ctx_cpu` is already defined in the template, it's derived from the CPU architecture.  And it doesn't seem like it needs to be set to anything different than what is already in `generic-qemu-boot-template.jinja2`, so I think it should be dropped here.

templates/uefi/generic-qemu-uefi-boot-template.jinja2
```diff
@@ -0,0 +1,62 @@
...
+{% set ctx_arch = arch %}
```
gctucker: This is very similar to `generic-qemu-boot-template.jinja2`, so I think it should be inheriting from it to avoid duplicating all this.  Then you can add the i386 case in the base template.

templates/uefi/generic-qemu-uefi-boot-template.jinja2
```diff
@@ -0,0 +1,62 @@
...
+{% endblock %}
+{% block main %}
+{{ super() }}
+{% endblock %}
+{% block actions %}
+context:
+  arch: {{ ctx_arch }}
+  cpu: {{ ctx_cpu }}
+  guestfs_interface: virtio
+{% if arch == 'arm' or arch == 'arm64' %}
```
gctucker: Probably better to do:
```
{% if qemu-machine -%}
  machine: {{ qemu-machine }}
{% endif %}
```
and also include that in the base template.

templates/uefi/generic-qemu-uefi-boot-template.jinja2
```diff
@@ -0,0 +1,62 @@
...
+{% endif %}
+
+actions:
+- deploy:
+    timeout:
+      minutes: 3
+    to: tmpfs
+    os: oe
+    images:
+      bios:
```
gctucker: Same thing here, this could be solved in a generic way in the base template:
```
{% if bios_url %}
    bios:
      image_arg: '-bios {bios}'
      url: {{ bios_url }}
{% endif }
```

templates/uefi/generic-qemu-uefi-boot-template.jinja2
```diff
@@ -0,0 +1,62 @@
+{% extends 'base/kernel-ci-base.jinja2' %}
+{% set ctx_arch = arch %}
+{% if arch == "arm" %}
+{% set ctx_arch = 'arm64' %}
```
gctucker: `arm64` - is this intentional?  It appears to be working fine with `arm`.

gctucker:
@roxell I've made a fixup PR #151, testing this on staging now.

gctucker:
@roxell There appears to be a problem with the "mixed" x86 variant:
https://lkft.validation.linaro.org/scheduler/job/882544

Can you see what's wrong with it?

It was not run before, so I don't think it's a regression introduced by #151.  All the other variants work, including the i386 which is also run now.

gctucker:
I've resubmitted it in the Collabora lab and got some IA32 exception:
https://lava.collabora.co.uk/scheduler/job/1775563
```
IA32 Exception Type - 0D(#GP - General Protection)  CPU Apic ID - 00000000 !!!!
```

gctucker:
The issue above has now been fixed with a new UEFI firmware for x86-mixed mode.

gctucker:
This PR is now being replaced with #151.

PR #101, opened 2019-06-20
labs: Add lab-clabbe

montjoie:
My lab is now ready to handle jobs.


gctucker:
Tested OK on staging, here's an example boot: https://staging.kernelci.org/boot/id/5d1f578acf3a0feeef1783e8/

PR #102, opened 2019-06-25
Add kernelci tree

gctucker:
Add the definition for the kernelci tree to be able to use it in production with a `kernelci.org` branch.  The `staging.kernelci.org` branch should not be in this file on the master branch to avoid triggering production every time a new revision is pushed to test on staging.

Also enable kernels build from this tree to run in the LKFT lab.


danrue:
:+1: thanks @gctucker

PR #107, opened 2019-06-26
kernel-arch-complete.sh: trigger email when first arch is complete

gctucker:
The way build-trigger.jpl works is that it waits for all the kernel
builds to have completed before calling kernel-arch-complete.sh once
for each architecture.  So all the builds are already there the first
time the script is called, and we can drop the magic number of "4".
This fixes cases where some trees build fewer architectures.

This is still not optimal but it will become a lot better when
replaced with a pipeline step to generate the test jobs on a per-build
basis.


gctucker:
@eballetbo FYI this should enable email reports for the chrome-platform tree in particular, which only builds 3 architectures.

gctucker:
Tested OK on staging, all the reports were sent with the correct information in them.

PR #108, opened 2019-06-26
templates/boot: add console_msg_format to kernel cmdline

khilman:
syslog-style printk message from the kernel are available when booting
with "console_msg_format=syslog" on the kernel command-line.  This
makes it much easier for log parsers to get the log-level information
from the kernel boot logs.

Automatically add this command-line argument to all u-boot and qemu
jobs.

### DIFF
--- a/templates/base/kernel-ci-base.jinja2
+++ b/templates/base/kernel-ci-base.jinja2
@@ -1,3 +1,5 @@
+{% set extra_kernel_args = "console_msg_format=syslog" %}
+
 {%- block metadata %}
 metadata:
   image.type: 'kernel-ci'
--- a/templates/boot/generic-qemu-boot-template.jinja2
+++ b/templates/boot/generic-qemu-boot-template.jinja2
@@ -34,7 +34,7 @@ actions:
     images:
       kernel:
 {%- block image_arg %}
-        image_arg: '-kernel {kernel} -append "console={{ console_dev }},115200 root=/dev/ram0 debug verbose"'
+        image_arg: '-kernel {kernel} -append "console={{ console_dev }},115200 root=/dev/ram0 debug verbose {{ extra_kernel_args }}"'
 {%- endblock %}
         url: {{ kernel_url }}
       ramdisk:
--- a/templates/boot/generic-uboot-tftp-ramdisk-boot-template.jinja2
+++ b/templates/boot/generic-uboot-tftp-ramdisk-boot-template.jinja2
@@ -3,6 +3,9 @@
 {{ super() }}
 {% endblock %}
 {% block main %}
+{%- if extra_kernel_args %}
+{% do context.update({"extra_kernel_args": extra_kernel_args}) %}
+{% endif %}
 {{ super() }}
 {% endblock %}
 {% block actions %}
--- a/lava-v2-jobs-from-api.py
+++ b/lava-v2-jobs-from-api.py
@@ -245,7 +245,7 @@ def write_jobs(config, jobs):
     for job in jobs:
         job_file = os.path.join(job_dir, '.'.join([job['name'], 'yaml']))
         with open(job_file, 'w') as f:
-            env = Environment(loader=FileSystemLoader('templates'))
+            env = Environment(loader=FileSystemLoader('templates'), extensions=["jinja2.ext.do"])
             template = env.get_template(job['short_template_file'])
             data = template.render(job)
             f.write(data)
--- a/templates/boot/generic-depthcharge-tftp-ramdisk-boot-template.jinja2
+++ b/templates/boot/generic-depthcharge-tftp-ramdisk-boot-template.jinja2
@@ -3,6 +3,9 @@
 {{ super() }}
 {% endblock %}
 {% block main %}
+{%- if extra_kernel_args %}
+{% do context.update({"extra_kernel_args": extra_kernel_args}) %}
+{% endif %}
 {{ super() }}
 {% endblock %}
 {% block actions %}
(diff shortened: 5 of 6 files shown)


khilman:
This should work for QEMU and u-boot jobs (at least ones who don't modify/override this in their device-dict).

I don't know the other methods  (PXE, depthcharge, grub) well enough to add those.

@mgalka could you have look at how to extend the kernel command-line for barebox jobs?

templates/boot/generic-qemu-boot-template.jinja2
```diff
@@ -34,7 +34,7 @@ actions:
     images:
       kernel:
 {%- block image_arg %}
-        image_arg: '-kernel {kernel} -append "console={{ console_dev }},115200 root=/dev/ram0 debug verbose"'
+        image_arg: '-kernel {kernel} -append "console={{ console_dev }},115200 root=/dev/ram0 debug verbose console_msg_format=syslog"'
```
gctucker: Something similar should be done in this template:
templates/v4l2-compliance/generic-qemu-v4l2-compliance-template.jinja2

gctucker:
I can look into the depthcharge part.

khilman:
@gctucker updated to include the `v4l2-compliance` QEMU templated, and also made it into a variable so it can be re-used across all the templates.

khilman:
@mattface can you look at the grub templates for how to extend the kernel cmdline?

gctucker:
@khilman Yes, we have some x86 Minnowboards booting with GRUB so I'll take a look.

templates/boot/generic-uboot-tftp-ramdisk-boot-template.jinja2
```diff
@@ -6,6 +6,10 @@
 {{ super() }}
 {% endblock %}
 {% block actions %}
+{%- if extra_kernel_args %}
+context:
```
mattface: I think this will overwrite the context section thats written in kernel-ci-base.jinja2
Context can currently be added in the test-configs yaml per device-type, so we should probably add a "default" context that is written to all jobs.

gctucker:
> @khilman Yes, we have some x86 Minnowboards booting with GRUB so I'll take a look.

Sorry saw the notification and thought that was for me :p  I'll just look into Depthcharge for now but can help with GRUB if needed.

mattface:
I've had a quick look at GRUB, and extra_kernel_args context should work in any device that extends base-grub.

gctucker:
I'm going to make a PR against this PR branch, so we can merge incoming changes for other bootloaders here and then merge the main PR.  How does that sound?

templates/boot/generic-uboot-tftp-ramdisk-boot-template.jinja2
gctucker: OK it's possible to use the `context` jinja dictionary variable instead then, and add `extra_kernel_args` to it.

gctucker:
Sent a PR, see status update above.  It generated a job file locally with the right context.

For Depthcharge devices, I do need to send some patches upstream in LAVA though as some device types define `extra_kernel_args`.  If the job also defines it, the device type value gets overridden.  So we need to be able to add kernel args on top of the device one, which I'll fix in LAVA.  Only lab-collabora has Depthcharge devices so far, so I can put an update here when this has been put in production (can happen before the next LAVA release).

gctucker:
LAVA changes to allow device dictionaries and job definitions to append kernel args with `extra_kernel_args` using Depthcharge:
https://git.lavasoftware.org/lava/lava/merge_requests/633

templates/boot/generic-uboot-tftp-ramdisk-boot-template.jinja2
khilman: Thanks for catching this, this is fixed in the latest version by just updating the `context` jinja dictionary.

templates/base/kernel-ci-base.jinja2
```diff
@@ -1,3 +1,5 @@
+{% set extra_kernel_args = "console_msg_format=syslog" %}
```
gctucker: OK I think that should work for Depthcharge too, I'll make a patch in `generic-depthcharge-tftp-ramdisk-boot-template.jinja2`.

khilman: Thanks, merged into this PR now.

gctucker:
@mattface I think we're just waiting for the GRUB patch now, can you please send a PR against `khilman:dev/console_msg_format` with this?

If you won't have the time this week then let me know and I can take a look as we have some GRUB boards in lab-collabora too (Minnowboard).

khilman:
@gctucker any reason we need to wait for GRUB before merging this?  IMO, we should merge it now to enable broader testing and see if we break any existing log parsing code in the backend.

gctucker:
@khilman No real need to wait before merging what we currently have, but it sounded like @mattface was going to do this.

gctucker:
@khilman This has been running on staging for days btw, and the error parsing seems fine.  Here's an example:
https://staging-storage.kernelci.org/kernelci/staging.kernelci.org/staging-20190724.0/arm64/defconfig/clang-8/lab-collabora-dev/boot-rk3399-gru-kevin.html

gctucker:
I also checked that GRUB boards still booted fine, they just don't use the extra kernel command line parameter for syslog-type formatting:
https://staging-storage.kernelci.org/kernelci/staging.kernelci.org/staging-20190723.0/x86_64/x86_64_defconfig/gcc-8/lab-collabora-dev/boot-minnowboard-turbot-E3826.html

PR #109, opened 2019-07-02
Fix the arc imagetype

montjoie:
All jobs generated for ARC devices uses "type: uImage.gz".
Since this type is unknow, LAVA try to convert the kernel file to an
uImage and corrupt it.
Using "type: uImage" fix this problem.

Since the file is not strictly speaking gzip compressed, that the build
create a symlink uImage->uImage.gz, the easiest fix is to tell kernelci
to use uImage.

### DIFF
--- a/kernelci/build.py
+++ b/kernelci/build.py
@@ -39,14 +39,14 @@ TORVALDS_GIT_URL = \
 MAKE_TARGETS = {
     'arm': 'zImage',
     'arm64': 'Image',
-    'arc': 'uImage.gz',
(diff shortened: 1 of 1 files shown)


gctucker:
LGTM, will get this tested on staging.

gctucker:
Got a build on staging, looks fine: https://staging.kernelci.org/build/id/5d1db5b3cf3a0fb5d4f3dc41/

See #59 with details about the boot test on `hsdk`.

PR #110, opened 2019-07-02
Add more qemus

montjoie:
This PR add more qemus machines to boot

### DIFF
--- a/templates/base/kernel-ci-base.jinja2
+++ b/templates/base/kernel-ci-base.jinja2
@@ -38,7 +38,7 @@ device_type: {{ base_device_type }}
 
 {% if context %}
 context:
-{% for key, value in context.items() %}  {{ key }}: "{{ value }}"
+{% for key, value in context.items() %}  {{ key }}: {{ value }}
 {% endfor %}
 {% endif %}
 
--- a/templates/boot/generic-qemu-boot-template.jinja2
+++ b/templates/boot/generic-qemu-boot-template.jinja2
@@ -39,6 +39,11 @@ actions:
       ramdisk:
         image_arg: '-initrd {ramdisk}'
         url: {{ initrd_url }}
+{%- if dtb_url %}
+      dtb:
+        image_arg: '-dtb {dtb}'
+        url: {{ dtb_url }}
+{% endif %}
 
 - boot:
     timeout:
--- a/test-configs.yaml
+++ b/test-configs.yaml
@@ -808,6 +808,49 @@ device_types:
       - whitelist: {defconfig: ['defconfig']}
       - blacklist: {kernel: ['v3.']}
 
+  qemu_arm-vexpress-a9:
+    base_name: qemu
+    mach: qemu
+    arch: arm
+    context:
+      arch: 'arm'
+      machine: 'vexpress-a9'
+      cpu: 'cortex-a9'
+      model: 'model=lan9118'
+    dtb: 'vexpress-v2p-ca9.dtb'
+    boot_method: qemu
+    filters:
+      - whitelist: {defconfig: ['vexpress_defconfig']}
+
+  qemu_arm-vexpress-a15:
+    base_name: qemu
+    mach: qemu
+    arch: arm
+    context:
+      arch: 'arm'
+      machine: 'vexpress-a15'
+      cpu: 'cortex-a15'
+      model: 'model=lan9118'
+    dtb: 'vexpress-v2p-ca15-tc1.dtb'
+    boot_method: qemu
+    filters:
+      - whitelist: {defconfig: ['vexpress_defconfig']}
+
+  qemu_arm-versatilepb:
+    base_name: qemu
+    mach: qemu
+    arch: arm
+    context:
+      arch: 'arm'
+      machine: 'versatilepb'
+      cpu: 'arm926'
+      memory: 256
+      model: 'model=smc91c111'
+    dtb: 'versatile-pb.dtb'
+    boot_method: qemu
+    filters:
+      - whitelist: {defconfig: ['versatile_defconfig']}
+
   qemu_arm-virt-gicv2:
     base_name: qemu
     mach: qemu
@@ -1596,6 +1639,18 @@ test_configs:
     test_plans:
       - baseline
       - boot
+
+  - device_type: qemu_arm-vexpress-a9
+    test_plans:
+      - boot_qemu
+
+  - device_type: qemu_arm-vexpress-a15
+    test_plans:
+      - boot_qemu
+
+  - device_type: qemu_arm-versatilepb
+    test_plans:
+      - boot_qemu
 
   - device_type: qemu_arm-virt-gicv2
     test_plans:


gctucker:
@montjoie Please rebase now that #108 has been merged.

gctucker:
@montjoie Sorry, please rebase one more time now that #151 has been merged.  It caused some conflicts with your PR but I believe this should be easy to resolve.

templates/boot/generic-qemu-boot-template.jinja2
```diff
@@ -51,6 +51,11 @@ actions:
       ramdisk:
         image_arg: '-initrd {ramdisk}'
         url: {{ initrd_url }}
+{%- if dtb_url is defined and dtb_url != none %}
```
gctucker: Maybe `{%- if dtb_url %}` is enough?

templates/boot/generic-qemu-boot-template.jinja2
```diff
@@ -23,13 +20,6 @@
 {{ super() }}
 {% endblock %}
 {% block actions %}
-context:
```
gctucker: Please make a separate PR to remove this template block and fix all the existing QEMU entries that rely on these parameters (UEFI).

test-configs.yaml
```diff
@@ -1302,6 +1304,7 @@ test_configs:
   - device_type: qcom_qdf2400
     test_plans: [boot]
 
+<<<<<<< HEAD
```
gctucker: How much time did you actually spend rebasing this PR?

templates/boot/generic-qemu-boot-template.jinja2
montjoie: No because dtb_url is defined if equal to None

test-configs.yaml
```diff
@@ -756,6 +758,46 @@ device_types:
       - whitelist: {defconfig: ['defconfig']}
       - blacklist: {kernel: ['v3.']}
 
+  qemu_arm-vexpress-a9:
+    mach: qemu
```
gctucker: Missing this attribute, for all the new QEMU entries:
```yaml
base_name: qemu
```

test-configs.yaml
```diff
@@ -58,6 +58,8 @@ test_plan_default_filters:
       keys: ['arch', 'defconfig']
       values:
         - ['arm', 'multi_v7_defconfig']
+        - ['arm', 'versatile_defconfig']
```
gctucker: I don't think we should be enabling these defconfigs by default for all test plans.  If you want `boot_qemu` to be run with them, then add this filter to `boot_qemu`.  If it's for all QEMU devices, then it may be better to update the filters in the device types.

test-configs.yaml
```diff
@@ -756,6 +756,49 @@ device_types:
...
+  qemu_arm-vexpress-a15:
+    base_name: qemu
+    mach: qemu
+    arch: arm
+    context:
+      arch: 'arm'
+      machine: 'vexpress-a15'
+      cpu: 'cortex-a15'
+      model: 'model=lan9118'
+    dtb: 'vexpress-v2p-ca15-tc1.dtb'
```
gctucker: Only problem here is that when building `vexpress_defconfig`, it didn't create `vexpress-v2p-ca15-tc1.dtb`.  That dtb is there with the `multi_v7_defconfig` build though, but then it's filtered out.

So the staging run only included `versatile-pb` which has its dtb created with `vexpress_defconfig`:
https://staging.kernelci.org/boot/id/5d8be5fecf3a0fe289322dc6/

gctucker:
@montjoie Any update on this?

test-configs.yaml
montjoie: arch/arm/configs/vexpress_defconfig contains CONFIG_ARCH_VEXPRESS=y
And in arch/arm/boot/dts/Makefile, I see that vexpress-v2p-ca15-tc1 is built with ARCH_VEXPRESS.
So it is strange that vexpress_defconfig didnt create it.

montjoie: I have just compiled a vexpress_defconfig and express-v2p-ca15-tc1.dtb is built

gctucker:
OK I'll take another closer look at this today to find out what's going on.

templates/base/kernel-ci-base.jinja2
```diff
@@ -37,7 +37,7 @@ device_type: {{ base_device_type }}
 
 {% if context %}
 context:
-{% for key, value in context.items() %}  {{ key }}: "{{ value }}"
+{% for key, value in context.items() %}  {{ key }}: {{ value }}
```
mattface: I can't remember why exactly, but I'm sure these quotes around the {{ value }} were for some context that was being set needing to be forced into being a string.
Was there a reason to remove them?

montjoie: because some values need to be without quotes, let me refind which ones

montjoie: In fact it was in the commit description, you need to be able to pass numeric values, and adding quotes breaks the numeric status

gctucker: FYI this has been tested on staging for days and there doesn't seem to be any issue without the quotes.

test-configs.yaml
gctucker: Fair enough, tried this with mainline v5.4-rc8 and it seems to do the job:
```
$ make dtbs
scripts/kconfig/conf  --syncconfig Kconfig
  UPD     include/config/kernel.release
  HOSTCC  scripts/dtc/dtc.o
[...]
  HOSTLD  scripts/dtc/dtc
  DTC     arch/arm/boot/dts/vexpress-v2p-ca5s.dtb
  DTC     arch/arm/boot/dts/vexpress-v2p-ca9.dtb
  DTC     arch/arm/boot/dts/vexpress-v2p-ca15-tc1.dtb
  DTC     arch/arm/boot/dts/vexpress-v2p-ca15_a7.dtb
```

gctucker: I'll enable `vexpress_defconfig` in today's staging run to check it works all the way and gets tested on QEMU as expected.

gctucker: OK got a vexpress_defconfig build on staging: https://staging-storage.kernelci.org/kernelci/staging.kernelci.org/staging-20191125.0/arm/vexpress_defconfig/gcc-8/

It was tested in several labs so that seems fine:
https://staging-ci.kernelci.org/job/test-runner/751/console

However there's still no boot results here:
https://staging.kernelci.org/boot/all/job/kernelci/branch/staging.kernelci.org/kernel/staging-20191125.0/

I'm investigating this now.  We should also run the baseline test plan on these QEMU variants.

test-configs.yaml
```diff
@@ -1597,6 +1640,18 @@ test_configs:
       - baseline
       - boot
 
+  - device_type: qemu_arm-vexpress-a9
+    test_plans:
+      - boot_qemu
```
gctucker: Please add `baseline_qemu` here and to the other 2 variants below.

test-configs.yaml
gctucker: I think the problem was because of this: https://github.com/kernelci/kernelci-backend/pull/180

Testing again on staging now with the backend fix for Unicode characters.

gctucker: Finally got a boot result now: https://staging.kernelci.org/boot/id/5ddd3eff092e747463ed498e/

So this is fine - except the baseline test plan should also be run, but I can add that.

test-configs.yaml
gctucker: I'll add that with another patch...

PR #112, opened 2019-07-03
Remove the android-4.4 branch from the android tree

mattface:
As per Todds request, the android-4.4 branch is being deprecated so remove it from build-configs.yaml


gctucker:
Tested OK on staging.

PR #114, opened 2019-07-04
kernel-arch-complete.sh: send email reports for kernelci tree

gctucker:
Send email reports to a fixed list of recipients for the kernelci
tree.  This is mostly used on staging and when testing production
updates.


gctucker:
Tested OK on staging.

gctucker:
BTW this was already being applied as a flying patch on staging branches, so it's been tested for quite a while already.  It's now useful to have it merged since the `kernelci` tree is merged in `build-configs.yaml` for production test runs.

PR #116, opened 2019-07-05
Fix lab-clabbe config

gctucker:
A few fixes that should go on top of #101.


gctucker:
@montjoie Please pick the fixes from this PR and add them to #101, then we can close this PR.  I initially thought it could be merged separately but it's a fixup so should be used to rework your patches instead.

At least this enabled testing your lab on staging, I can see some boot results:
https://staging.kernelci.org/boot/id/5d1f1a45cf3a0fe6d41783da/

gctucker:
Fixes have now been applied on #101 so this PR can be closed.

PR #118, opened 2019-07-09
linusw build config

gctucker:
Reduce the number of builds for the `linusw` build configs and add support for regex filters to be able to do that.  Also send email reports to the linux-gpio mailing list.

See discussion: https://groups.io/g/kernelci/message/428


gctucker:
Sample run on staging: https://staging.kernelci.org/job/linusw/branch/devel/

PR #120, opened 2019-07-11
Test the new 'test_template' object in the POST/send API

touilkhouloud:
This patch send email reports using the test_template option which gives
the possibility to choose the test template to be used in the email
report.


gctucker:
@touilkhouloud Please don't send PRs for `staging.kernelci.org`, they need to be sent for the `master` branch.

gctucker:
Also instead of duplicating the entry, we should be replacing it using the new API template parameter.

touilkhouloud:
So we should remove the hard-coded part in `app/utils/report/test.py` ?

gctucker:
Yes, that's the idea.

PR #128, opened 2019-07-30
Set the overall job timeout as a variable

touilkhouloud:
As the job timeout is different from one job to another so we change this
to be a configurable variable with a 10 min per default value.

### DIFF
--- a/templates/base/kernel-ci-base.jinja2
+++ b/templates/base/kernel-ci-base.jinja2
@@ -60,7 +60,7 @@ notify:
 job_name: {{ name }}
 timeouts:
   job:
-    minutes: 10
+    minutes: {{ job_timeout|default("10") }}
   action:
    minutes: 10
   actions:


lava-v2-jobs-from-api.py
```diff
@@ -174,6 +174,7 @@ def get_job_params(config, test_config, defconfig, opts, build, plan):
         'context': device_type.context,
         'rootfs_prompt': rootfs.prompt,
         'plan_name': test_plan.name,
+        'job_timeout': config.get('job_timeout')
```
gctucker: Missing trailing comma, no jobs were run on staging because of this.

Can you please, please at least try your changes once locally before submitting a PR?

lava-v2-jobs-from-api.py
```diff
@@ -317,6 +318,8 @@ def main(args):
                         help="section in the KernelCI config file")
     parser.add_argument("--plans", nargs='+', required=True,
                         help="test plan to create jobs for")
+    parser.add_argument("--job_timeout", default="10", dest='job_timeout',
```
gctucker: Why add this as a command line argument in this script, which applies to all the jobs generated, rather than a YAML parameter in each test plan like I suggested on IRC?

lava-v2-jobs-from-api.py
touilkhouloud: I am sorry I do try my changes in a diff repo then the one to use for commits to keep things clean that's why I have these problems often, I am sorry again.

lava-v2-jobs-from-api.py
touilkhouloud: I didn't get what you exactly meant I should've asked, but is this a wrong one to do so or not preferable to do it this way ?

touilkhouloud:
@gctucker oky I got what you mean now, I will re-push changes as you requested without errors this time you have my word and sorry again ...

gctucker:
Tested OK on staging that it doesn't break anything.

Also tested locally that it does override the job timeout:
```yaml
  v4l2_compliance_qemu_vivid:
    name: v4l2-compliance-vivid
    rootfs: debian_stretch-v4l2_ramdisk
    pattern: 'v4l2-compliance/generic-qemu-v4l2-compliance-template.jinja2'
    params: {v4l2_driver: "vivid", job_timeout: 123}
```
produced:
```yaml
job_name: gtucker-kernelci-local-kernelci-local-snapshot-048-arm64-defconfig+virtualvideo-gcc-8-no-dtb-qemu-v4l2-compliance-vivid
timeouts:
  job:
    minutes: 123
```

gctucker:
Please note that the "action" timeout is still hard-coded at 10min, so maybe another `action_timeout` parameter would be needed in some cases.

Also the variable name doesn't mention the unit so it's assumed to be in minutes.

PR #129, opened 2019-07-31
T14893 kernelci.build: keep .ccache directory inside the build directory

gctucker:
In order to make ccache more likely to be useful, and to be able to
discard it when necessary, keep the .ccache directory inside the
kernel build directory.  That way, it should always cache build data
relevant to the build done in that build directory, and when removing
unused build directories the cache will be removed at the same time.

Also include both the CPU arch and the compiler name in the ccache
directory name, in case a build directory is being reused multiple
times (typically, for manual builds rather than on kernelci.org).

### DIFF
--- a/kernelci/build.py
+++ b/kernelci/build.py
@@ -424,8 +424,8 @@ def _run_make(kdir, arch, target=None, jopt=None, silent=True, cc='gcc',
     if use_ccache:
         px = cross_compile if cc == 'gcc' and cross_compile else ''
         args.append('CC="ccache {}{}"'.format(px, cc))
-        os.environ.setdefault('CCACHE_DIR',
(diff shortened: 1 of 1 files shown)


gctucker:
@khilman If the ccache directory is part of the build directory, then the build directory can be arch-specific.  So the ccache directory follows the build directory, if there's binaries from another arch there then it's as "wrong" as having a ccache for another arch.

gctucker:
@khilman OK I've updated the PR to include both the CPU arch and compiler names in the ccache directory.  While this won't make any difference with kernelci.org builds, I guess it can help when building locally and reusing the same build directory multiple times.

Testing this on staging today.

khilman:
@gctucker ok, thanks.

gctucker:
Tested OK on staging.

PR #130, opened 2019-07-31
T15937 bisect.jpl: only keep one kernel build directory

gctucker:
Bisection jobs keep their data in the workspace in order to avoid
having to do a fresh git checkout.  However, build diretories can keep
accumulating if they are not removed as there are many possible
combinations of architectures and compilers.

To keep the disk usage under control, remove existing build
directories before starting a bisection except if there is one that
matches exactly the required arch/compiler combination as it may
contain some useful cached data when running similar bisections in a
row.

### DIFF
--- a/jenkins/bisect.jpl
+++ b/jenkins/bisect.jpl
@@ -212,6 +212,12 @@ def buildKernel(kdir, kci_core) {
     dir(kci_core) {
         sh(script: "rm -f ${env._BUILD_JSON}")
 
+        sh(script: """\
+for d in \$(find ${kdir} -name "build-*" -type d); do
+  [ "\$d" = "${output}" ] || rm -rf "\$d"
+done
+""")
+
         sh(script:"""\
 ./kci_build \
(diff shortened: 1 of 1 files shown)


gctucker:
Tested OK on staging, here's an example with an `arm` job deleting an old `arm64` build directory:
```
+ find /home/buildslave/staging/workspace/gtucker/lava-bisect/linux -name build-* -type d
+ [ /home/buildslave/staging/workspace/gtucker/lava-bisect/linux/build-arm64-gcc-8 = /home/buildslave/staging/workspace/gtucker/lava-bisect/linux/build-arm-gcc-8 ]
+ rm -rf /home/buildslave/staging/workspace/gtucker/lava-bisect/linux/build-arm64-gcc-8
+ [ /home/buildslave/staging/workspace/gtucker/lava-bisect/linux/build-arm-gcc-8 = /home/buildslave/staging/workspace/gtucker/lava-bisect/linux/build-arm-gcc-8 ]
```

PR #132, opened 2019-07-31
test-config.yaml: Add support for meson-g12-odroid-n2

touilkhouloud:
We have a meson-g12-odroid-n2 in lab-baylibre, ready to accept jobs.

### DIFF
--- a/test-configs.yaml
+++ b/test-configs.yaml
@@ -590,6 +590,13 @@ device_types:
     filters:
       - blacklist: {defconfig: ['allnoconfig', 'allmodconfig']}
 
+  meson-g12b-odroid-n2:
+    mach: amlogic
+    class: arm64-dtb
+    boot_method: uboot
+    filters:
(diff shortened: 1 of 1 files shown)


khilman:
device-type in upstream LAVA already: https://git.lavasoftware.org/lava/lava/blob/master/lava_scheduler_app/tests/device-types/meson-g12b-odroid-n2.jinja2

test-configs.yaml
```diff
@@ -590,6 +590,14 @@ device_types:
     filters:
       - blacklist: {defconfig: ['allnoconfig', 'allmodconfig']}
 
+  meson_g12b_odroid_n2:
```
gctucker: Change the name to use hyphens and drop the `name` attribute.

gctucker:
Tested OK on staging: https://staging.kernelci.org/boot/id/5d4c6be4cf3a0f037f8003dd/

Issue #136, opened 2019-08-07
clang builds don't enable additional config options


broonie:
It appears that clang builds for configurations which enable additional configuration options don't actually enable those options. For example:

https://storage.kernelci.org/next/master/next-20190807/arm64/defconfig+CONFIG_CPU_BIG_ENDIAN=y/clang-8/build.log

at no point enables CONFIG_CPU_BIG_ENDIAN resulting in a config:

https://storage.kernelci.org/next/master/next-20190807/arm64/defconfig+CONFIG_CPU_BIG_ENDIAN=y/clang-8/kernel.config

where CPU_BIG_ENDIAN is not set. In contrast the equivalent GCC build does have it set in the generated config. I'm not seeing any visible difference in the start of the build logs.

@mattface

broonie:
@gctucker Potentially this is something in kci_build?

gctucker:
I'll take a look but it would be surprising that only CPU_BIG_ENDIAN doesn't work while there hasn't been any issues with the other extra config options in the past.

gctucker:
I've reproduced this problem with a build on staging, this defconfig doesn't have CPU_BIG_ENDIAN enabled:
https://staging-storage.kernelci.org/gtucker/kernelci-local/kernelci-local-snapshot-048/arm64/defconfig+CONFIG_CPU_BIG_ENDIAN=y/clang-8/kernel.config
```
# CONFIG_CPU_BIG_ENDIAN is not set
```

even though it's in the frag.config:
https://staging-storage.kernelci.org/gtucker/kernelci-local/kernelci-local-snapshot-048/arm64/defconfig+CONFIG_CPU_BIG_ENDIAN=y/clang-8/frag.config

When I run it locally it does enable it in the `.config` so either this gets changed later during the build process with `clang` for some obscure reason or there's a bug in `kci_build` that only shows on the Jenkins jobs.  I'll add some more debug to narrow the issue down.

broonie:
I suspect it's not only CPU_BIG_ENDIAN, it's just very visible on CPU_BIG_ENDIAN due to the fact that failing to set that option causes us to fail to boot userspace whereas many other options could reasonably be off and won't stop things booting.

gctucker:
OK I've found several different issues in various places:

[...]

PR #141, opened 2019-08-08
T16118 clang merge config

gctucker:
Export compiler names when calling `merge_config.sh` as otherwise it keeps using `gcc` even when building with `clang`.

Also do some related minor code clean-up.


gctucker:
Tested OK on staging.  This also relies on a patch in the kernel top-level Makefile to be able to override the compiler, see #136 for the details.

PR #144, opened 2019-08-13
arm-soc renamed to soc

khilman:

### DIFF
--- a/build-configs.yaml
+++ b/build-configs.yaml
@@ -19,8 +19,8 @@ trees:
   ardb:
     url: "git://git.kernel.org/pub/scm/linux/kernel/git/ardb/linux.git"
 
-  arm-soc:
-    url: "https://git.kernel.org/pub/scm/linux/kernel/git/arm/arm-soc.git"
+  soc:
+    url: "https://git.kernel.org/pub/scm/linux/kernel/git/soc/soc.git"
 
   arnd:
     url: "https://git.kernel.org/pub/scm/linux/kernel/git/arnd/playground.git"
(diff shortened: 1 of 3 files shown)


build-configs.yaml
```diff
@@ -19,8 +19,8 @@ trees:
   ardb:
     url: "git://git.kernel.org/pub/scm/linux/kernel/git/ardb/linux.git"
 
-  arm-soc:
-    url: "https://git.kernel.org/pub/scm/linux/kernel/git/arm/arm-soc.git"
+  soc:
```
gctucker: Please move this in the list to keep the entries in alphabetical order.

build-configs.yaml
```diff
@@ -450,12 +450,12 @@ build_configs:
...
 #
-#  arm-soc_fixes:
-#    tree: arm-soc
+#  soc_fixes:
+#    tree: soc
 #    branch: 'arm/fixes'
 
-  arm-soc_for-next:
-    tree: arm-soc
+  soc_for-next:
```
gctucker: Ditto.

kernelci/build.py
```diff
@@ -229,7 +229,7 @@ def git_describe(tree_name, path):
     The returned value is a string with the "git describe" for the commit
     currently checked out in the local git repository.
     """
-    describe_args = "--match=v\*" if tree_name == "arm-soc" else ""
+    describe_args = "--match=v\*" if tree_name == "soc" else ""
```
gctucker: By the way, is this special treatment still required to get a useful git describe?

khilman: Probably not, but I didn't want to remove that as part of this patch.  That should be investigated and removed separately.

khilman:
@gctucker updated entries are now sorted alphabetically.

kernelci/build.py
gctucker: OK

gctucker:
Did a quick test locally:
```
$ ./kci_build list_configs | grep soc
soc_for-next
$ ./kci_build tree_branch --config=soc_for-next
soc
https://git.kernel.org/pub/scm/linux/kernel/git/soc/soc.git
for-next
```

PR #146, opened 2019-08-14
Stop booting odroid-x2 with LPAE configs

mattface:
This board has failed to boot this config for a long time, so blacklist it

### DIFF
--- a/test-configs.yaml
+++ b/test-configs.yaml
@@ -691,7 +691,6 @@ device_types:
     class: arm-dtb
     boot_method: uboot
     dtb: 'exynos4412-odroidx2.dtb'
-    flags: ['lpae']
 
   odroid_xu3:
     name: 'odroid-xu3'


test-configs.yaml
```diff
@@ -692,6 +692,8 @@ device_types:
     boot_method: uboot
     dtb: 'exynos4412-odroidx2.dtb'
     flags: ['lpae']
```
gctucker: Remove this line instead of the blacklist.  The `lpae` flag means it will run boot tests with kernel builds that have LPAE in their defconfig.

I guess this flag may be replaced with pure YAML rules, but it was ported from the old device map implementation.

mattface:
Tested the suggested change has the same effect, the EFI+LPAE job is no longer created.
Updated PR.

gctucker:
Thanks - I think the commit summary should be updated as it's not really blacklisting that particular config anymore but any build with LPAE enabled.

gctucker:
@mattface Err, I meant the actual commit, not just the PR title :p

https://github.com/kernelci/kernelci-core/pull/146/commits/14da334589187e911abe28a25e9872c29384031f

mattface:
Done!

gctucker:
Thanks!

PR #150, opened 2019-08-21
build-configs.yaml: disable fragments with clang builds

gctucker:
Disable special fragments with clang builds on linux-next/master until
the issues related to merging fragments with defconfigs has been
solved.  This is to avoid noise in the build reports when we know
these are going to fail but not because of actual kernel issues.

### DIFF
--- a/build-configs.yaml
+++ b/build-configs.yaml
@@ -656,9 +656,11 @@ build_configs:
               - 'allmodconfig'
       clang-8:
         build_environment: clang-8
-        fragments: *default_fragments
         architectures:
-          arm64: *arm64_arch
+          arm64:
+            extra_configs:
+              - 'allmodconfig'
+              - 'allnoconfig'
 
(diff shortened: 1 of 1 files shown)


gctucker:
Please note that arm64 CPU_BIG_ENDIAN builds with clang are enabled on staging.kernelci.org so we can test a proper fix while this is disabled in production with linux-next.

gctucker:
Quick local check:
```
$ ./kci_build list_kernel_configs --config=next --kdir=linux | grep clang
arm64 allnoconfig clang-8
arm64 allmodconfig clang-8
arm64 defconfig clang-8
```
These 2 are no longer listed as a result of this change:
```
arm64 defconfig+CONFIG_RANDOMIZE_BASE=y clang-8
arm64 defconfig+CONFIG_CPU_BIG_ENDIAN=y clang-8
```

PR #153, opened 2019-08-22
Update android trees as per Todd request

mattface:
delete: android-4.14
add: android-4.14-r

delete: android-4.19
add: android-4.19-r

delete: android-4.9

### DIFF
--- a/build-configs.yaml
+++ b/build-configs.yaml
@@ -356,10 +356,6 @@ build_configs:
     tree: android
     branch: 'android-4.4-p-release'
 
-  android_4.9:
-    tree: android
-    branch: 'android-4.9'
-
   android_4.9-o:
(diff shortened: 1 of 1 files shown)


gctucker:
What about clang?

> 4) Please add limited clang build testing of two branches
> 
> a) android-4.19-r arm64/x86_64 'allmodconfig'
> b) android-mainline arm64/x86_64 'allmodconfig'

mattface:
We should be good to merge this as-is, because I'll pickup all the clang changes on the next PR

gctucker:
OK, merging this now to make it part of the production update as the missing branches are causing some errors in the kernel-tree-monitor job (not preventing it from running, but potentially hiding other unexpected errors in the noise).

PR #154, opened 2019-08-22
test-configs: disable hsdk for v4.14

khilman:
This board has never worked on v4.14, and fixes for more recent
kernels have not been backported this far and are unlikely to ever be.

Signed-off-by: Kevin Hilman <khilman@kernel.org>

### DIFF
--- a/test-configs.yaml
+++ b/test-configs.yaml
@@ -504,6 +504,9 @@ device_types:
     mach: arc
     class: arc-dtb
     boot_method: uboot
+    filters:
+      - blacklist:
+          kernel: ['v4.14']
 
   imx23_olinuxino:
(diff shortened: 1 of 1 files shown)


test-configs.yaml
```diff
@@ -458,6 +458,9 @@ device_types:
     mach: arc
     class: arc-dtb
     boot_method: uboot
+    filters:
+      blacklist:
```
gctucker: Missing `-` as this is a list of filters.

khilman: OK, updated.

gctucker:
Tested on staging that it doesn't have any side-effects (staging doesn't run with v4.14.y).

PR #157, opened 2019-08-23
kernelci.build: enable verbose merge_config.sh

gctucker:
When building with verbose turned on, show the full output of
merge_config.sh as well as the regular make commands.

### DIFF
--- a/kernelci/build.py
+++ b/kernelci/build.py
@@ -398,6 +398,13 @@ def list_kernel_configs(config, kdir, single_variant=None, single_arch=None):
     return kernel_configs
 
 
+def _output_to_file(cmd, log_file):
+    open(log_file, 'a').write("#\n# {}\n#\n".format(cmd))
+    cmd = "/bin/bash -c '(set -o pipefail; {} 2>&1 | tee -a {})'".format(
+        cmd, log_file)
+    return cmd
+
+
 def _run_make(kdir, arch, target=None, jopt=None, silent=True, cc='gcc',
               cross_compile=None, use_ccache=None, output=None, log_file=None,
               opts=None):
@@ -438,13 +445,11 @@ def _run_make(kdir, arch, target=None, jopt=None, silent=True, cc='gcc',
     cmd = ' '.join(args)
     print(cmd)
     if log_file:
-        open(log_file, 'a').write("#\n# {}\n#\n".format(cmd))
-        cmd = "/bin/bash -c '(set -o pipefail; {} 2>&1 | tee -a {})'".format(
-            cmd, log_file)
+        cmd = _output_to_file(cmd, log_file)
     return shell_cmd(cmd, True)
 
 
-def _make_defconfig(defconfig, kwargs, fragments, verbose):
+def _make_defconfig(defconfig, kwargs, fragments, verbose, log_file):
     kdir, output_path = (kwargs.get(k) for k in ('kdir', 'output'))
     result = True
 
@@ -488,6 +493,8 @@ scripts/kconfig/merge_config.sh -O {output} '{base}' '{frag}' {redir}
            frag=os.path.join(rel_path, kconfig_frag_name),
            redir='> /dev/null' if not verbose else '')
         print(cmd.strip())
+        if log_file:
+            cmd = _output_to_file(cmd, log_file)
         result = shell_cmd(cmd, True)
 
     tmpfile.close()
@@ -549,7 +556,8 @@ def build_kernel(build_env, kdir, arch, defconfig=None, jopt=None,
     start_time = time.time()
     fragments = []
     if defconfig:
-        result = _make_defconfig(defconfig, kwargs, fragments, verbose)
+        result = _make_defconfig(
+            defconfig, kwargs, fragments, verbose, log_file)
     elif os.path.exists(dot_config):
         print("Re-using {}".format(dot_config))
         result = True
--- a/kernelci/build.py
+++ b/kernelci/build.py
@@ -444,7 +444,7 @@ def _run_make(kdir, arch, target=None, jopt=None, silent=True, cc='gcc',
     return shell_cmd(cmd, True)
 
 
-def _make_defconfig(defconfig, kwargs, fragments):
+def _make_defconfig(defconfig, kwargs, fragments, verbose):
     kdir, output_path = (kwargs.get(k) for k in ('kdir', 'output'))
     result = True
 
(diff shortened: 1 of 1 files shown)


gctucker:
Now we can see the build error with clang in the log:
https://staging-storage.kernelci.org/kernelci/staging.kernelci.org/staging-20190826.1/arm64/defconfig+CONFIG_CPU_BIG_ENDIAN=y/clang-8/build.log
```
# cd /home/buildslave/staging/workspace/kernel-build@2/linux
export ARCH=arm64
export HOSTCC=clang
export CC=clang
export CROSS_COMPILE=aarch64-linux-gnu-
scripts/kconfig/merge_config.sh -O build 'build/.config' 'build/frag.config'  
#
Using build/.config as base
Merging build/frag.config
Value of CONFIG_CPU_BIG_ENDIAN is redefined by fragment build/frag.config:
Previous value: # CONFIG_CPU_BIG_ENDIAN is not set
New value: CONFIG_CPU_BIG_ENDIAN=y

make[1]: Entering directory '/home/buildslave/staging/workspace/kernel-build@2/linux/build'
  HOSTCC  scripts/basic/fixdep
/bin/sh: 1: gcc: not found
make[2]: *** [scripts/Makefile.host:90: scripts/basic/fixdep] Error 127
make[1]: *** [/home/buildslave/staging/workspace/kernel-build@2/linux/Makefile:505: scripts_basic] Error 2
make[1]: Leaving directory '/home/buildslave/staging/workspace/kernel-build@2/linux/build'
make: *** [Makefile:179: sub-make] Error 2
```

gctucker:
However, the gcc build also failed but it doesn't show any clear error:
https://staging-storage.kernelci.org/kernelci/staging.kernelci.org/staging-20190826.1/arm64/defconfig+CONFIG_CPU_BIG_ENDIAN=y/gcc-8/build.log
```
# cd /home/buildslave/workspace/kernel-build/linux
export ARCH=arm64
export HOSTCC=gcc
export CC=gcc
export CROSS_COMPILE=aarch64-linux-gnu-
scripts/kconfig/merge_config.sh -O build 'build/.config' 'build/frag.config'  
#
Using build/.config as base
Merging build/frag.config
Value of CONFIG_CPU_BIG_ENDIAN is redefined by fragment build/frag.config:
Previous value: # CONFIG_CPU_BIG_ENDIAN is not set
New value: CONFIG_CPU_BIG_ENDIAN=y

make[1]: Entering directory '/home/buildslave/workspace/kernel-build/linux/build'
  GEN     Makefile
scripts/kconfig/conf  --alldefconfig Kconfig
#
# configuration written to .config
#
make[1]: Leaving directory '/home/buildslave/workspace/kernel-build/linux/build'
```
It turns out this is because some config options are not present at all in the resulting `.config`, so this command returns an error and the script aborts:
```
ACTUAL_VAL=$(grep -w -e "$CFG" "$KCONFIG_CONFIG")
```
With a simple hack to ignore the error, the output becomes visible again:
```
Value requested for CONFIG_ARM64_ACPI_PARKING_PROTOCOL not in final .config
Requested value:  # CONFIG_ARM64_ACPI_PARKING_PROTOCOL is not set
Actual value:     
```
I believe this is because the `make` command run in `merge_config.sh` doesn't do the right thing for us, when we just want to enable a few config options rather than merge two defconfig files.

The bottom line is that `merge_config.sh` is essentially not well suited for our needs, and we should rather have a tailored solution using `scripts/config` with a Python function to verify the extra config options have been set as expected.

At least this PR helped getting to the bottom of the issues.

test-configs.yaml
```diff
@@ -41,15 +41,15 @@ file_systems:
 
   debian_buster_ramdisk:
     type: debian
-    ramdisk: 'buster/20190809.0/{arch}/rootfs.cpio.gz'
+    ramdisk: 'buster/20190823.0/{arch}/rootfs.cpio.gz'
```
gctucker: BTW This looks like a bug in Github as the rootfs URLs have already been updated on the master branch...

gctucker:
Please note that we should not merge this PR until this has also been applied:
https://lore.kernel.org/lkml/4f92e9b3a88e60c8b5962504d77bc596442b0a40.1567023309.git.guillaume.tucker@collabora.com/

As things currently stand, with merge_config.sh returning error status on linux-next, virtually all config fragments will fail regardless of the compiler being used.

The version currently in mainline never returns a non-zero status so it silently fails with clang but merges the configs fine when it works.  The version currently in -next almost always returns an error as it's very easy to get a config option not present in the resulting config file when it's set to its default value (saw this with `defconfig+CONFIG_CPU_BIG_ENDIAN` on arm64).  If the fix from the link above gets applied to the kbuild for-next branch then we may merge this as it should then start working again.
cc @broonie 

In any case, the patches for the relative log path may still be merged as that could be useful for any other build command.

Then we should have our own Python function to apply config fragments and stop using merge_config.sh as discussed previously, except maybe for `kselftest` fragments if they require extra care.

gctucker:
Note: The fix patch from the link  above is now being applied to the kernelci tree as well to test it on staging.

gctucker:
Pending patch v2: https://lore.kernel.org/lkml/a60fe77a1475ba960ad9fc851f1ace2196b661b8.1567436778.git.guillaume.tucker@collabora.com/

gctucker:
Mailing list thread: https://groups.io/g/kernelci/topic/kernelci_staging_kernelci_org/33124028

gctucker:
The patch has been applied to linux-kbuild, will check it works in linux-next and then this PR can be merged.

gctucker:
Tested OK with next-20190905 on staging, got the expected build pass and fail with big-endian:
```
--------------------------------------------------------------------------------
defconfig+CONFIG_CPU_BIG_ENDIAN=y (arm64, clang-8) — FAIL, 1 error, 0 warnings, 0 section mismatches

Errors:
    /bin/sh: 1: gcc: not found

--------------------------------------------------------------------------------
defconfig+CONFIG_CPU_BIG_ENDIAN=y (arm64, gcc-8) — PASS, 0 errors, 2 warnings, 0 section mismatches

Warnings:
    WARNING: "HYPERVISOR_platform_op" [vmlinux] is a static EXPORT_SYMBOL_GPL
    WARNING: "HYPERVISOR_platform_op" [vmlinux] is a static EXPORT_SYMBOL_GPL
```

So this is ready to be merged.

With linux-next and the kernel releases starting from v5.4, errors hit while running `make` such as the known one with `clang` will be propagated.  Config options checks will not cause the script to fail.

With older kernels, when merge_config.sh works everything is fine.  But `make` errors will be silently dropped causing config options to not be correctly merged by the script, such as in the known issue with `clang`.  Another solution will be needed to address this problem.

PR #160, opened 2019-08-27
test-configs.yaml: define qemu arch and cpu directly in context

gctucker:
test-configs.yaml: define qemu arch and cpu directly in context

Rather than having default values for arch and cpu defined in the qemu
boot template, define them in the YAML device configuration.


gctucker:
Tested OK on staging:
https://staging.kernelci.org/boot/all/job/kernelci/branch/staging.kernelci.org/kernel/staging-20190829.1/

PR #159, opened 2019-08-26
T16036 test-configs.yaml: clean up and consolidate device types

gctucker:
Clean up test-configs.yaml content and schema to make it more future-proof.

PR #161, opened 2019-08-27
Remove hard coded qemu context

montjoie:
This patch remove the hardcoded job context, and move all its content to
test-configs.yaml.
This will permit more customization of qemu machines templates.

### DIFF
--- a/templates/boot/generic-qemu-boot-template.jinja2
+++ b/templates/boot/generic-qemu-boot-template.jinja2
@@ -1,19 +1,14 @@
 {% extends 'base/kernel-ci-base.jinja2' %}
-{% set ctx_arch = arch %}
 {% if arch == "arm" %}
-{% set ctx_cpu = 'cortex-a15' %}
 {% set console_dev = 'ttyAMA0' %}
 {% endif %}
 {% if arch == "arm64" %}
-{% set ctx_cpu = 'cortex-a57' %}
 {% set console_dev = 'ttyAMA0' %}
 {% endif %}
 {% if arch == 'x86_64' %}
-{% set ctx_cpu = 'qemu64' %}
 {% set console_dev = 'ttyS0' %}
 {% endif %}
 {% if arch == 'i386' %}
-{% set ctx_cpu = 'qemu32' %}
 {% set console_dev = 'ttyS0' %}
 {% endif %}
 {% block metadata %}
@@ -23,13 +18,6 @@
 {{ super() }}
 {% endblock %}
 {% block actions %}
-context:
-  arch: {{ ctx_arch }}
-  cpu: {{ ctx_cpu }}
-  guestfs_interface: virtio
-{%- if qemu_machine %}
-  machine: {{ qemu_machine }}
-{%- endif %}
 
 actions:
 - deploy:
--- a/test-configs.yaml
+++ b/test-configs.yaml
@@ -139,32 +139,16 @@ test_plans:
   sleep:
     rootfs: debian_buster_ramdisk
 
-  uefi-arm-gicv2:
+  uefi-arm:
     rootfs: buildroot_ramdisk
     pattern: 'boot/generic-qemu-boot-template.jinja2'
     params:
-      qemu_machine: 'virt,gic-version=2'
       bios_url: 'https://storage.kernelci.org/images/uefi/111bbcf87621/QEMU_EFI.fd-ARM-RELEASE-111bbcf87621'
 
-  uefi-arm-gicv3:
+  uefi-arm64:
     rootfs: buildroot_ramdisk
     pattern: 'boot/generic-qemu-boot-template.jinja2'
     params:
-      qemu_machine: 'virt,gic-version=3'
-      bios_url: 'https://storage.kernelci.org/images/uefi/111bbcf87621/QEMU_EFI.fd-ARM-RELEASE-111bbcf87621'
-
-  uefi-arm64-gicv2:
-    rootfs: buildroot_ramdisk
-    pattern: 'boot/generic-qemu-boot-template.jinja2'
-    params:
-      qemu_machine: 'virt,gic-version=2'
-      bios_url: 'https://storage.kernelci.org/images/uefi/111bbcf87621/QEMU_EFI.fd-AARCH64-RELEASE-111bbcf87621'
-
-  uefi-arm64-gicv3:
-    rootfs: buildroot_ramdisk
-    pattern: 'boot/generic-qemu-boot-template.jinja2'
-    params:
-      qemu_machine: 'virt,gic-version=3'
       bios_url: 'https://storage.kernelci.org/images/uefi/111bbcf87621/QEMU_EFI.fd-AARCH64-RELEASE-111bbcf87621'
 
   uefi-x86_64:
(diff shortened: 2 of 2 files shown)


test-configs.yaml
```diff
@@ -831,35 +815,53 @@ device_types:
       - whitelist: {defconfig: ['defconfig']}
       - blacklist: {kernel: ['v3.']}
 
-  qemu_arm:
+  qemu_i386:
```
gctucker: Please keep the entries in alphabetical order.

test-configs.yaml
```diff
@@ -1324,22 +1326,32 @@ test_configs:
...
       - simple_qemu
-      - uefi-arm64-gicv2
-      - uefi-arm64-gicv3
+      - uefi-arm
       - v4l2_compliance_qemu_vivid
 
+  - device_type: qemu-arm64-virt-gicv2
+    test_plans:
+      - boot_qemu
+      - uefi-arm64
```
gctucker: Please also keep the test plans in alphabetical order.

test-configs.yaml
```diff
@@ -831,35 +815,53 @@ device_types:
...
-    arch: arm64
+    arch: x86_64
     boot_method: qemu
+    context: {arch: 'x86_64', cpu: 'qemu64', guestfs_interface: 'virtio'}
     filters:
       - whitelist: {defconfig: ['defconfig']}
 
-  qemu_i386:
-    name: qemu
+  qemu-arm-virt-gicv2:
```
gctucker: Missing `name: qemu`

The `name` is the actual device type name used when submitting jobs in LAVA, which has to be `qemu`.

test-configs.yaml
```diff
@@ -831,35 +815,53 @@ device_types:
...
+    arch: arm
     boot_method: qemu
+    context: {arch: 'arm', cpu: 'cortex-a15', guestfs_interface: 'virtio', machine: 'virt,gic-version=2'}
     filters:
-      - whitelist: {defconfig: ['i386_defconfig']}
+      - whitelist: {defconfig: ['multi_v7_defconfig', 'vexpress_defconfig']}
 
-  qemu_x86_64:
-    name: qemu
+  qemu-arm-virt-gicv3:
```
gctucker: Missing `name: qemu`

test-configs.yaml
```diff
@@ -831,35 +815,53 @@ device_types:
...
     mach: qemu
-    arch: x86_64
+    arch: arm
+    boot_method: qemu
+    context: {arch: 'arm', cpu: 'cortex-a15', guestfs_interface: 'virtio', machine: 'virt,gic-version=3'}
+    filters:
+      - whitelist: {defconfig: ['multi_v7_defconfig', 'vexpress_defconfig']}
+
+  qemu-arm64-virt-gicv2:
```
gctucker: `name: qemu`

test-configs.yaml
```diff
@@ -831,35 +815,53 @@ device_types:
...
+    mach: qemu
+    arch: arm64
+    boot_method: qemu
+    context: {arch: 'arm64', cpu: 'cortex-a57', guestfs_interface: 'virtio', machine: 'virt,gic-version=2'}
+    filters:
+      - whitelist: {defconfig: ['defconfig']}
+
+  qemu-arm64-virt-gicv3:
```
gctucker: `name: qemu`

test-configs.yaml
montjoie: ok

test-configs.yaml
montjoie: No, we need to get away from qemu, it shadow too many different machines.
We should now start to use device aliases since LAVA support it

test-configs.yaml
montjoie: ok

test-configs.yaml
montjoie: This comment answer also the previous comment (which I cannot answer, I dont know why). With device alias, we can distinguished different uefi-boot

gctucker:
So here's what I think we can do to solve this:
* merge #159 which starts differentiating the device type "logical name" listed in YAML from the device type "actual name" used in LAVA jobs
* make another patch to use the logical name as the `device.type` meta-data value and the actual name as the `device_type` value in LAVA job definitions
* update this PR which devices more qemu variants which will all be treated as different devices in the backend and everywhere

@montjoie How does that sound?

test-configs.yaml
gctucker: Yes it's confusing to have `qemu` as a single device type when we're actually running many different things.  However, LAVA aliases are not ideal because they need to be manually configured in each lab, kept the same across labs otherwise jobs won't run, and they are specific to LAVA so other types of labs would probably need to create some kind of hard-coded mapping somewhere.

Instead, we can take advantage of the fact that the LAVA device type name is independent from the device type as known to KernelCI.  See #159 which removes the unnecessary "name" attribute and replaces it with "type_name" which is what is meant to be used when submitting jobs.  The logical KernelCI device name is passed separately in the job meta-data.

test-configs.yaml
```diff
@@ -751,35 +735,57 @@ device_types:
       - whitelist: {defconfig: ['defconfig']}
       - blacklist: {kernel: ['v3.']}
 
-  qemu_arm:
-    type: qemu
+  qemu-arm-virt-gicv2:
+    type_name: qemu
```
gctucker: Sorry for the confusion with the recent changes - this should be `type: qemu`

The code has `type_name` mainly because `type()` is a built-in Python class name.

templates/boot/generic-qemu-boot-template.jinja2
```diff
@@ -1,19 +1,15 @@
 {% extends 'base/kernel-ci-base.jinja2' %}
 {% set ctx_arch = arch %}
```
gctucker: This line should be removed too if `arch` is defined in the context from YAML.

test-configs.yaml
```diff
@@ -751,35 +735,57 @@ device_types:
       - whitelist: {defconfig: ['defconfig']}
       - blacklist: {kernel: ['v3.']}
 
-  qemu_arm:
-    type: qemu
+  qemu-arm-virt-gicv2:
```
gctucker: I think it's a bit easier to read `qemu_arm-virt-gicv2` to make it clear that it's a variant of qemu, but that's quite subjective.

test-configs.yaml
```diff
@@ -1328,22 +1334,32 @@ test_configs:
...
 
-  - device_type: qemu_arm
+  - device_type: qemu-arm-virt-gicv2
     test_plans:
       - boot_qemu
       - simple_qemu
-      - uefi-arm-gicv2
-      - uefi-arm-gicv3
+      - uefi-arm
       - v4l2_compliance_qemu_vivid
```
gctucker: No need to run v4l2-compliance-vivid twice on arm and arm64, so probably only with `-gicv3` in both cases.

test-configs.yaml
```diff
@@ -1328,22 +1334,32 @@ test_configs:
...
+  - device_type: qemu-arm64-virt-gicv2
+    test_plans:
+      - boot_qemu
+      - uefi-arm64
+      - simple_qemu
+
+  - device_type: qemu-arm64-virt-gicv3
+    test_plans:
+      - boot_qemu
+      - uefi-arm64
```
gctucker: Alphabetical order problem

test-configs.yaml
```diff
@@ -1328,22 +1334,32 @@ test_configs:
...
+      - simple_qemu
+
```
gctucker: So, please add v4l2-compliance-vivid here to have it run with arm64 gicv3 (and arm gicv3, but no need to also run it with gicv2).

test-configs.yaml
montjoie: I dont see it

gctucker: `simple` comes before `uefi`

montjoie: the UI show me only the 4 last line... but the mail show me the whole diff

gctucker: Surely you have a local copy of the file anyway?

gctucker:
Thanks for the changes, looks good.

One more thing that needs to be updated is the `kernel-arch-complete.sh` script, as the name of the UEFI test plans has changed.

gctucker:
Testing this PR now on staging.

@montjoie Please add a commit to update `kernel-arch-complete.sh` with the new UEFI test plan names, otherwise the UEFI tests won't be run.

gctucker:
Tested OK on staging, you can show "all" the boot results and filter "qemu" to see all the variants with architecture and GIC versions:
https://staging.kernelci.org/boot/all/job/kernelci/branch/staging.kernelci.org/kernel/staging-20190828.2/

gctucker:
However we're still missing the UEFI tests as per my earlier comment.

gctucker:
The UEFI tests plans that haven't changed names were run, but that's only the x86 ones.  Here's an example:
https://lkft.validation.linaro.org/scheduler/job/892791

gctucker:
See #167 which is a cleaned-up version of this in separate patches to first use the context data and then refactor the UEFI test plans.

gctucker:
Replaced with #161 which has now been merged.

PR #166, opened 2019-08-28
Add support to stop boot testing Clang

mattface:
Add support for filtering test jobs by build environment.

Then use this support to stop boot testing clang builds until the boot emails issue is properly resolved.

### DIFF
--- a/test-configs.yaml
+++ b/test-configs.yaml
@@ -62,19 +62,27 @@ test_plan_default_filters:
         - ['i386', 'i386_defconfig']
         - ['x86_64', 'x86_64_defconfig']
 
+  - blacklist: &clang_environment_filter
+      build_environment: ['clang']
+
+  - blacklist: &kselftest_defconfig_filter
+      defconfig: ['kselftest']
+
+  - blacklist: &kselftest_defconfig_clang_environment_filter
+      <<: *kselftest_defconfig_filter
+      <<: *clang_environment_filter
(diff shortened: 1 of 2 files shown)


test-configs.yaml
```diff
@@ -62,19 +62,27 @@ test_plan_default_filters:
         - ['i386', 'i386_defconfig']
         - ['x86_64', 'x86_64_defconfig']
 
+  - blacklist: &clang_environment_filter
+      build_environment: ['clang']
+
+  - blacklist: &kselftest_defconfig_filter
+      defconfig: ['kselftest']
+
+  - blacklist: &kselftest_defconfig_clang_environment_filter
```
gctucker: We can also just have the 2 aliases above listed in each test plan.  I'm not sure if it's that helpful to have clang kselftest filters tied together into one alias.

mattface: Yeah I wasn't sure either, but having just learned about YAML anchoring I liked the less duplication

gctucker: Fair enough, it's easy to adjust again in the future to whatever appears most practical.

gctucker:
Tested OK on staging, all boards and kernel builds still appear to be booting as before except the clang ones:
https://staging.kernelci.org/boot/all/job/kernelci/branch/staging.kernelci.org/kernel/staging-20190829.2/

PR #94, opened 2019-06-13
Fastboot

touilkhouloud:
Adding new fastboot template.
As the sei510 support fastboot, use it as an examples to show the changes need to be added to each board to be used with fastboot.
For the sei510, the device-type has been already modified and merged to support fastboot.

PR #167, opened 2019-08-29
Device type params

gctucker:
Cherry-pick and rebase @touilkhouloud's patches from #94 to add parameters to device types.


gctucker:
Tested OK on staging, with no side effects as this isn't being used yet.

PR #168, opened 2019-08-29
QEMU UEFI rework

gctucker:
This depends on #160 to be merged first, and is a clean-up of #161.

### DIFF
--- a/test-configs.yaml
+++ b/test-configs.yaml
@@ -139,47 +139,31 @@ test_plans:
   sleep:
     rootfs: debian_buster_ramdisk
 
-  uefi-arm-gicv2:
+  uefi_arm:
     rootfs: buildroot_ramdisk
     pattern: 'boot/generic-qemu-boot-template.jinja2'
     params:
-      qemu_machine: 'virt,gic-version=2'
       bios_url: 'https://storage.kernelci.org/images/uefi/111bbcf87621/QEMU_EFI.fd-ARM-RELEASE-111bbcf87621'
 
-  uefi-arm-gicv3:
+  uefi_arm64:
     rootfs: buildroot_ramdisk
     pattern: 'boot/generic-qemu-boot-template.jinja2'
     params:
-      qemu_machine: 'virt,gic-version=3'
-      bios_url: 'https://storage.kernelci.org/images/uefi/111bbcf87621/QEMU_EFI.fd-ARM-RELEASE-111bbcf87621'
-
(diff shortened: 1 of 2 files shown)


test-configs.yaml
```diff
@@ -751,19 +735,55 @@ device_types:
       - whitelist: {defconfig: ['defconfig']}
       - blacklist: {kernel: ['v3.']}
 
-  qemu_arm:
+  qemu_arm-gicv2:
```
montjoie: I prefer to qualify better the machine, it is a virt-gicv2

gctucker: Sure.

mgalka:
I'm not entirely convinced with the "base_name" attribute. It's certainly better than a "type_name", maybe we could use something like "type_" ?
Apart from that LGTM.

gctucker:
> maybe we could use something like "type_" ?

How would you intuitively make the difference between `DeviceType.name` and `DeviceType.type_`?  That's essentially what we had before with `.type_name` and it was very vague.

The point of `.base_name` is to show that it's a "lower-level" name, something other device types are derived from in the YAML definitions.  Also we can have something similar for `TestPlan.base_name`, there's another PR coming for that.

mgalka:
> How would you intuitively make the difference between `DeviceType.name` and `DeviceType.type_`? That's essentially what we had before with `.type_name` and it was very vague.

Well, it somehow looked to me as a class-metaclass relationship in Python, hence "type_" seemed valid for me, but what seems clear to me doesn't necessarily be a good solution.

> The point of `.base_name` is to show that it's a "lower-level" name, something other device types are derived from in the YAML definitions. Also we can have something similar for `TestPlan.base_name`, there's another PR coming for that.

Got your point. The consistent naming looks good.

gctucker:
Tested OK on staging:
https://staging.kernelci.org/boot/all/job/kernelci/branch/staging.kernelci.org/kernel/staging-20190829.3/
https://staging-storage.kernelci.org/kernelci/staging.kernelci.org/staging-20190829.3/arm/multi_v7_defconfig/gcc-8/lab-baylibre-dev/boot-qemu_arm-virt-gicv2.html

Also got the UEFI jobs running as expected in the LKFT lab (7 variants, here are a couple):
https://lkft.validation.linaro.org/scheduler/alljobs?length=25&search=uefi#table
https://lkft.validation.linaro.org/scheduler/job/893837
https://lkft.validation.linaro.org/scheduler/job/893834

PR #169, opened 2019-08-29
Use yaml safe_load insteald of load

montjoie:
On my gentoo, using yaml.load now give:
  raise RuntimeError("Unsafe load() call disabled by Gentoo. See bug #659348")
  RuntimeError: Unsafe load() call disabled by Gentoo. See bug #659348

Note that on recent ubuntu, a warning appears also.
  YAMLLoadWarning: calling yaml.load() without Loader=... is deprecated, as the default Loader is unsafe. Please read https://msg.pyyaml.org/load for full details.

This is due to a security risk of using yaml.load()

Since kernelci-core does not rely on any behavour provided by load(), let's convert the call to safe_load().

### DIFF
--- a/kernelci/configs.py
+++ b/kernelci/configs.py
@@ -867,7 +867,7 @@ class TestConfig(YAMLObject):
 
 def builds_from_yaml(yaml_path):
     with open(yaml_path) as f:
-        data = yaml.load(f)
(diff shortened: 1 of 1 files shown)


gctucker:
Thanks, tested OK on staging.

PR #170, opened 2019-08-29
Device type fixup

gctucker:
As discussed in previous PRs, this is to fix rename the `DeviceType.type_name` attribute to clarify things.


gctucker:
Tested OK on staging: https://staging.kernelci.org/boot/all/job/kernelci/branch/staging.kernelci.org/kernel/staging-20190829.2/

PR #176, opened 2019-09-05
Remove vexpres_defconfig for qemu_arm-virt-gicv3

montjoie:
The vexpres_defconfig produces unbootable kernel for
qemu_arm-virt-gicv3, let's remove it from the qemu_arm-virt-gicv3's defconfig used.

### DIFF
--- a/test-configs.yaml
+++ b/test-configs.yaml
@@ -767,7 +767,7 @@ device_types:
       guestfs_interface: 'virtio'
       machine: 'virt,gic-version=2'
     filters:
-      - whitelist: {defconfig: ['multi_v7_defconfig', 'vexpress_defconfig']}
+      - whitelist: {defconfig: ['multi_v7_defconfig']}
 
   qemu_arm-virt-gicv3:
     base_name: qemu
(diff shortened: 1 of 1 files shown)


test-configs.yaml
```diff
@@ -780,7 +780,7 @@ device_types:
       guestfs_interface: 'virtio'
       machine: 'virt,gic-version=3'
     filters:
-      - whitelist: {defconfig: ['multi_v7_defconfig', 'vexpress_defconfig']}
+      - whitelist: {defconfig: ['multi_v7_defconfig']}
```
gctucker: Are we sure we still want to boot it on `qemu_arm-virt-gicv2` even though we don't currently know whether it's actually supposed to be booting?

montjoie: I have just checked, and vexpress_defconfig dont have ARCH_VIRT. Same for the final .config produced from it. So I will remove also gicv2

gctucker:
Tested OK on staging.  The `vexpress_defconfig` was built:
https://staging.kernelci.org/build/id/5d711779cf3a0f1385ad4506/

but there was no boot test as intended:
https://staging.kernelci.org/boot/all/job/kernelci/branch/staging.kernelci.org/kernel/staging-20190905.0/

@montjoie BTW it's better to start commits with the part of the code being changed (i.e. `test-configs.yaml: ...`) and add `Signed-off-by`.  There's no hard rule about that so not a blocker.

mattface:
Theres a typo in the commit (missing an S in vexpress) but otherwise LGTM as this is really needed to clean up boot results.

gctucker:
Also the commit message only mentions GIC v3 but it's actually disabling boots for both v2 and v3.

PR #177, opened 2019-09-05
T16420 baseline: add dmesg tests

gctucker:
Add test cases to the baseline test plan to check for kernel errors
using dmesg.  One test case is created for each log level starting
from "error" and up to "emergency".  Log messages at the "warning"
level are also checked but not reported as a test case.

All the kernel log lines matching these levels are then output again
in the log for debugging purposes.  The backend can't at the moment
store anything like arbitrary log strings to track individual errors,
so detecting whether there were any errors is a first step.

This relies on the latest version of the "baseline" KernelCI buildroot
file system which has the standard dmesg implementation from
util-linux rather than the busybox one.

### DIFF
--- a/templates/baseline/baseline.jinja2
+++ b/templates/baseline/baseline.jinja2
@@ -15,6 +15,42 @@
             - lava-test-shell
         run:
           steps:
+            - >
+                for level in warn err; do
+                  dmesg --level=$level --notime -x -k > dmesg.$level
+                done
+            - >
+                for level in crit alert emerg; do
+                  dmesg --level=$level --notime -x -k > dmesg.$level
+                  test -s dmesg.$level && res=fail || res=pass
+                  count=$(cat dmesg.$level | wc -l)
+                  lava-test-case $level \
+                    --result $res \
+                    --measurement $count \
+                    --units lines
+                done
+            - cat dmesg.emerg dmesg.alert dmesg.crit dmesg.err dmesg.warn
+      from: inline
+      name: dmesg
+      path: inline/dmesg.yaml
+
+- test:
+    timeout:
+      minutes: 10
+    definitions:
+    - repository:
+        metadata:
+          format: Lava-Test Test Definition 1.0
+          name: baseline
+          description: "baseline test plan"
+          os:
+            - debian
+          scope:
+            - functional
+          environment:
+            - lava-test-shell
+        run:
+          steps:
             - export PATH=/opt/bootrr/helpers:$PATH
             - cd /opt/bootrr && sh helpers/bootrr-auto
       lava-signal: kmsg


templates/baseline/baseline.jinja2
```diff
@@ -1,3 +1,32 @@
...
+          environment:
+            - lava-test-shell
+        run:
+          steps:
+            - dmesg --level=warn --notime -x -k > dmesg.warn
+            - >
+                for level in err crit alert emerg; do
+                  dmesg --level=$level --notime -x -k > dmesg.$level
+                  test -s dmesg.$level && res=fail || res=pass
+                  lava-test-case $level --result $res
```
khilman: Would be useful to have an count here as well for each level (probably just a `cat dmesg.$level |wc -l)
We could report that with the `--measurement` option to lava-test-case.  Not sure what to use for `--units` though.  Maybe just "messages"?

khilman: Related, the warnings could always be reported as a pass, but using `--measurement` you could at least report a count of warnings.  Could be useful for tracking increasing/decreasing warning count.

khilman: c.f. https://docs.lavasoftware.org/lava/writing-tests.html#recording-test-case-measurements-and-units

gctucker: True, counting the lines reduces the scope for missing new types of messages.  If there was one fixed / one new at the same time it would go unnoticed, but that's a bit better already than just a boolean pass/fail.  I'll give that a try.

gctucker:
> Appart from missing a skiplist, this is ok.
> But adding a skiplist need to change from inline to a real git repo.

We could define strings to skip in `test-configs.yaml` as parameters for `baseline`.  But, are there really any errors or higher level log messages that should be skipped?

montjoie:
In my personal test suite I have the same and the test is rarelly OK.
Example of I got just on my allwinner boards:
[    0.001356] /cpus/cpu@0 missing clock-frequency property
[    0.001375] /cpus/cpu@1 missing clock-frequency property
[    0.001389] /cpus/cpu@2 missing clock-frequency property
[    0.001403] /cpus/cpu@3 missing clock-frequency property
[    1.689549] sun4i-usb-phy 1c19400.phy: Couldn't request ID GPIO
[    2.415069] dwmac-sun8i 1c30000.ethernet: IRQ eth_wake_irq not found
[    2.421454] dwmac-sun8i 1c30000.ethernet: IRQ eth_lpi not found
[    7.112452] udevd[126]: specified group 'kvm' unknown
[   10.072559] sun50i-de2-bus 1000000.bus: Error couldn't map SRAM to device
[    0.091058] CPU0: Spectre v2: firmware did not set auxiliary control register IBE bit, system vulnerable
etc...

gctucker:
@montjoie So, how about getting all these fixed?

Also, detecting new errors is really what we should be doing.  So that's kind of a "skip list" but automatically managed based on previous results.  Until we can do that, we can at least detect when their number increases as Kevin suggested.

gctucker:
@khilman I've updated the test plan to report the number of lines as a measurement, testing that on staging today.

However I think we should first add support in the backend for detecting measurement changes as regressions before also reporting the number of warnings as otherwise I fear it might just cause a lot of noise and tend to hide the more serious errors.

Having the warnings printed in the log could help a bit with debugging some issues I suppose.

gctucker:
Forget to mention, if we always report the warnings as "pass" then they'll never appear in the email reports as only failures are shown there currently.

gctucker:
Got a report from staging, it's not showing the measurement values:
```
  dmesg - 4 tests: 3  PASS, 1 FAIL, 0 SKIP
    * err: never passed  
```
So while we can merge this test plan independently, the email reports would need to be improved in that respect to make use of the information.

khilman:
@touilkhouloud has done done some work to add measurement values to the test reports, but it's on a fork of ours that needs to be rebased since lots of email report changes have happened since.

Re: skiplists.  There are things like deferred probe messages that will always show up and are not really errors because the driver eventually loads.

templates/baseline/baseline.jinja2
```diff
@@ -1,3 +1,33 @@
...
+            - lava-test-shell
+        run:
+          steps:
+            - dmesg --level=warn --notime -x -k > dmesg.warn
+            - >
+                for level in err crit alert emerg; do
+                  dmesg --level=$level --notime -x -k > dmesg.$level
+                  test -s dmesg.$level && res=fail || res=pass
+                  count=$(cat dmesg.$level | wc -l)
+                  lava-test-case $level --result $res --measurement $count
```
khilman: We should probably have a `--unit` here.  The unit will be used by the frontend (and email reports) .  The unit could just be "messages"

Otherwise, LGTM

gctucker: Sure, I'll add that.

gctucker:
By grepping around the kernel, I can see deferred probe messages at random log levels.  Maybe they should not be higher than warnings.

Maybe we should also not treat errors as a test case until we're tracking the number of messages, so just print warn/err in the log and only report anything in crit/alert/emerg as test cases initially.

khilman:
Yes, the deferred probe messages should probably be unified to be warnings only.

I agree that starting with crit/alert/emerg as test cases that report failure is probably a good idea.

templates/baseline/baseline.jinja2
gctucker: I opted for "lines" as it seemed more accurate.  Some messages may span across several lines such as a kernel oops, but we're not parsing those to group all the lines as a single message (yet).

gctucker:
Tested OK on staging, here are some sample results from a LAVA job showing the measurement values;
https://lava.collabora.co.uk/results/1812980/0_dmesg

gctucker:
I think we can merge this now and add support in the email reports to show the measurement values as a follow-up.

PR #180, opened 2019-09-09
T9980 kernelci.config.test: optional plan name in TestConfig.match()

gctucker:
Make the plan argument optional in TestConfig.match() to be able to
use it to filter based on build meta-data only.

Update lava-v2-jobs-from-api.py accordingly.


gctucker:
Tested OK on staging, all test email reports were sent as expected.

PR #184, opened 2019-09-12
kernelci: build: add support for KCONFIG_* variables

roxell:
Make it possible to pass KCONFIG_* variable to configure a kernel.
One example that will be solved is when building an allmodconfig
kernel that can boot, we need to pass the defconfig file to the
variable KCONFIG_ALLCONFIG when configuring the kernel
with 'make allmodconfig'

### DIFF
--- a/kernelci/build.py
+++ b/kernelci/build.py
@@ -450,31 +450,41 @@ def _run_make(kdir, arch, target=None, jopt=None, silent=True, cc='gcc',
     return shell_cmd(cmd, True)
 
 
-def _make_defconfig(defconfig, kwargs, fragments, verbose, log_file):
+def _make_defconfig(defconfig, kwargs, extras, verbose, log_file):
     kdir, output_path = (kwargs.get(k) for k in ('kdir', 'output'))
     result = True
 
+    defconfig_kwargs = dict(kwargs)
+    defconfig_opts = dict(defconfig_kwargs['opts'])
+    defconfig_kwargs['opts'] = defconfig_opts
     tmpfile_fd, tmpfile_path = tempfile.mkstemp(prefix='kconfig-')
     tmpfile = os.fdopen(tmpfile_fd, 'w')
+    tmpfile_used = False
     defs = defconfig.split('+')
     target = defs.pop(0)
     for d in defs:
+        if d.startswith('KCONFIG_'):
+            config, value = d.split('=')
+            defconfig_opts[config] = value
+            extras.append(d)
         if d.startswith('CONFIG_'):
             tmpfile.write(d + '\n')
-            fragments.append(d)
+            extras.append(d)
+            tmpfile_used = True
         else:
             frag_path = os.path.join(kdir, d)
             if os.path.exists(frag_path):
                 with open(frag_path) as frag:
                     tmpfile.write("\n# fragment from : {}\n".format(d))
                     tmpfile.writelines(frag)
-                fragments.append(os.path.basename(os.path.splitext(d)[0]))
+                    tmpfile_used = True
+                extras.append(os.path.basename(os.path.splitext(d)[0]))
     tmpfile.flush()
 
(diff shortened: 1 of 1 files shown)


kernelci/build.py
```diff
@@ -450,15 +450,20 @@ def _run_make(kdir, arch, target=None, jopt=None, silent=True, cc='gcc',
     return shell_cmd(cmd, True)
 
 
-def _make_defconfig(defconfig, kwargs, fragments, verbose, log_file):
+def _make_defconfig(defconfig, kwargs, fragments, verbose, log_file, opts):
```
gctucker: No need to pass `opts` since a copy is made in the function.

(hrm, I decline all liability for this one ;-) )

kernelci/build.py
```diff
@@ -558,7 +563,7 @@ def build_kernel(build_env, kdir, arch, defconfig=None, jopt=None,
     fragments = []
     if defconfig:
         result = _make_defconfig(
-            defconfig, kwargs, fragments, verbose, log_file)
+            defconfig, kwargs, fragments, verbose, log_file, opts)
```
gctucker: Same thing, no need to pass `opts`.

kernelci/build.py
```diff
@@ -450,15 +450,20 @@ def _run_make(kdir, arch, target=None, jopt=None, silent=True, cc='gcc',
...
 
+    defconfig_kwargs = dict(kwargs)
+    defconfig_opts = dict(defconfig_kwargs['opts'])
+    defconfig_kwargs['opts'] = defconfig_opts
     tmpfile_fd, tmpfile_path = tempfile.mkstemp(prefix='kconfig-')
     tmpfile = os.fdopen(tmpfile_fd, 'w')
     defs = defconfig.split('+')
     target = defs.pop(0)
     for d in defs:
+        if d.startswith('KCONFIG_ALLCONFIG'):
```
gctucker: Looks good for this use-case.  If we see some value in passing other `KCONFIG_*` things in the full defconfig name then we may want to do something a bit more general such as:
```python
if d.startswith('KCONFIG_'):
    config, value = d.split('=')
    defconfig_opts[config] = value
```

gctucker:
I've found a couple of issues with the `publish_path`.  First of all, the KCONFIG_ALLCONFIG is not a fragment but it needs to be in the publish path.  So I've made a fix to get a list of `extras` to add to the defconfig name and use in the publish path, these could be fragments or other things like KCONFIG_* values.  Then there's a problem with having a path in that name, since it means the kernel builds get pushed to a sub-directory.  To solve that, I'm replacing the `/` characters with `-` in the publish path, but not in the `defconfig_full`.  While this works for the build, I'm not entirely sure it will work from a backend point of view or other things that might rely on the convention that the `publish_path` must contain exactly the defconfig_full value.  So this would need more testing, including some LAVA jobs automatically submitted with it to verify it works all the way.

Here's the fix I made, which I think can be applied to this PR so we can test it further on staging (you should be able to merge that into this PR and then squash):
https://github.com/roxell/kernelci-core/pull/1

I used the patch above to build one kernel on staging without any modules to save time, here's the result:
https://staging-storage.kernelci.org/kernelci/staging.kernelci.org/staging-20190913.1/arm64/allmodconfig+KCONFIG_ALLCONFIG=arch-arm64-configs-defconfig/gcc-8/

gctucker:
@roxell The commit message and subject will need to be reworked a bit as this patch is not just for KCONFIG_ALLCONFIG any more but any KCONFIG_* variable.

gctucker:
OK thanks the commit message is fine now :)

I'll build a kernel with this for arm64 and get it submitted on staging with my own Jenkins jobs for boot and baseline tests to check if that works.

If it turns out to be more complicated, it may be worth giving you access to staging Jenkins so you can do that yourself.

gctucker:
Replaced with #214 as this one doesn't apply on top of master any more.

PR #190, opened 2019-09-24
build-configs: fix path to the kvm_guest.config fragment

roxell:
x86 kernels with kvm guest support isn't built since the path to the
fragment isn't correct.
kvm_guest.config got moved from arch/x86/configs to kernel/configs.

### DIFF
--- a/build-configs.yaml
+++ b/build-configs.yaml
@@ -149,7 +149,7 @@ fragments:
       - 'CONFIG_VIDEO_VIVID_MAX_DEVS=64'
 
   x86_kvm_guest:
-    path: "arch/x86/configs/kvm_guest.config"
+    path: "kernel/configs/kvm_guest.config"
 
 
 build_environments:


mattface:
This frag is needed because the base x86 defconfig doesn't have support for VIRTIO devices built in. Since LAVA now uses a VIRTIO device for the overlay, none of the qemu jobs on x86 that have run any tests will actually work. They fail when LAVA tries to mount the overlay filesystem.

In a future PR, we will need to make x86 qemu jobs only run test plans when on builds with kvm_guest frag enabled.

gctucker:
So that might explain why baseline-uefi tests are failing on x86 qemu.

mattface:
> So that might explain why baseline-uefi tests are failing on x86 qemu.

Yep, plus any other test plans we run on x86

mattface:
Sorry, slip of the touchpad

gctucker:
Tested OK locally with this command (as this fragment is not built on staging):
```
 ./kci_build build_kernel --arch=x86_64 --defconfig=defconfig+kernel/configs/kvm_guest.config
```
I'm also adding an error case for when a specified fragment in the `--defconfig=` option is not in the tree as it's currently silently ignoring it.

PR #192, opened 2019-09-26
T9980 kernelci.build: install bmeta.json and simplify JSON publish format

gctucker:
Rename the build.json file created during the install phase as
bmeta.json and make it a superset of the bmeta.json file created
during the initial build.  Also simplify the builds.json created
during the publish step to have a list of data structures like
bmeta.json with the dtbs.json incorporated into them as dtb_dir_data
to match what gets stored in the backend.

This depends on https://github.com/kernelci/kernelci-backend/pull/163 to be merged at the same time.

### DIFF
--- a/kci_build
+++ b/kci_build
@@ -395,7 +395,7 @@ class cmd_build_kernel(Command):
 
 
 class cmd_install_kernel(Command):
-    help = "Install the kernel binaries and build.json locally"
+    help = "Install the kernel binaries and bmeta.json locally"
     args = [Args.kdir]
     opt_args = [Args.config, Args.tree_name, Args.tree_url, Args.branch,
                 Args.commit, Args.describe, Args.describe_verbose,
--- a/kernelci/build.py
+++ b/kernelci/build.py
@@ -599,7 +599,7 @@ def build_kernel(build_env, kdir, arch, defconfig=None, jopt=None,
     bmeta = {
         'build_threads': jopt,
         'build_time': round(build_time, 2),
-        'build_result': 'PASS' if result is True else 'FAIL',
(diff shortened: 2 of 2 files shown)


kci_build
```diff
@@ -395,7 +395,7 @@ class cmd_build_kernel(Command):
 
 
 class cmd_install_kernel(Command):
-    help = "Install the kernel binaries and build.json locally"
+    help = "Install the kernel binaries and bmeta.json locally"
```
alexandrasp: @gctucker which is the semantic behind `bmeta` ? Is it just `binary metadata` ?

kernelci/build.py
```diff
@@ -815,39 +815,34 @@ def publish_kernel(kdir, install='_install_', api=None, token=None,
...
+            'path': 'file_server_resource',
+            'file_server_resource': 'file_server_resource',
+            'job': 'job',
+            'git_branch': 'git_branch',
+            'arch': 'arch',
+            'kernel': 'git_describe',
+            'build_environment': 'build_environment',
+            'defconfig': 'defconfig',
+            'defconfig_full': 'defconfig_full',
+        }.iteritems()}
```
alexandrasp: what will we use that data dict for?

kci_build
gctucker: It's the kernel build meta-data.

kernelci/build.py
gctucker: This is the payload of the HTTP POST request sent to the backend to add a kernel build entry in the database.

gctucker:
Tested OK on staging today.

gctucker:
Tested again on staging today with https://github.com/kernelci/kernelci-backend/pull/163.

PR #196, opened 2019-10-01
kernelci.config.test: fix TestConfig.match()

gctucker:
Fix the logic in TestConfig.match() with brackets around the condition
where there is no test plan defined.  The "and" operator takes
precedence over the "or" operator, leading to wrong results in the
corner case when the plan is None.

Fixes: b301fc1e1888 ("kernelci.config.test: optional plan name in TestConfig.match()")


gctucker:
Tested OK on staging.

PR #198, opened 2019-10-01
kernelci.build: set none defconfig when there is no defconfig

gctucker:
When using the local .config file already present and no --defconfig
option is being used, set the defconfig name to "none" in the build
meta-data.  This should not happen with automated builds but can
happen with local builds with a .config manually created.


gctucker:
Tested OK on staging.

PR #200, opened 2019-10-01
T16583 kci_test

gctucker:
Add a `kci_test` command line utility to generate and submit all the tests for a given kernel build.  It can talk to LAVA labs directly, but other types of labs can be added with extra Python modules implementing `kernelci.lab.LabAPI`.


gctucker:
Tested OK on staging, i.e. nothing broke as a side effect of this PR.

PR #204, opened 2019-10-03
thermal testing branch

gctucker:
Add definition for the minimal set of builds with the main defconfig for each arch, and the "thermal" tree with testing branch which uses the minimal set of builds.


gctucker:
Tested OK on staging up to the point where a source tarball was created:
https://staging-storage.kernelci.org/thermal/testing/v5.4-rc1-6-g9c1b4e6b6c45/

PR #206, opened 2019-10-08
baseline: set 1min timeout for each sub-test

gctucker:
Baseline tests should be very quick, a few seconds at most.  Reduce
the timeout from 10min to 1min for both dmesg and bootrr tests to
avoid having failing jobs stuck for much longer than needed.


gctucker:
Tested OK on staging, got all the baseline test results as expected.

PR #209, opened 2019-10-08
arch-complete: fix agross email address

khilman:
Andy's email has been bouncing for awhile, updated it to his
kernel.org address.


khilman:
I confirmed with Andy (via IRC) that this is the correct updated email for his maintainer work.

gctucker:
LGTM, no visible side effects when testing on staging.

PR #210, opened 2019-10-09
Create 'variants' section in build_configs_results

mgalka:
The 'variants' section has been created under build_configs_defaults and
the current content has been moved to it. It allows to creating other
default values sections in future and keeping defaults in a common
place.

### DIFF
--- a/build-configs.yaml
+++ b/build-configs.yaml
@@ -196,63 +196,62 @@ build_environments:
 
 # Default config with full build coverage
 build_configs_defaults:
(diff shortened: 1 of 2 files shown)


mgalka:
> Please start the commit message summary with `build-configs.yaml: ...` to make it easier to know which part of the code is being changed.

Done

gctucker:
Tested OK on staging, trying a build-trigger now with mainline build config as it uses the defaults.

PR #212, opened 2019-10-11
test-configs.yaml: Add sleep to the test_plan for rk3399-gru-kevin

eballetbo:
Suspend/resume on kevin should just work but, specially, after a merge
window, breaks. In order to catch such regressions soon, enable the sleep
test plan for this board.

### DIFF
--- a/test-configs.yaml
+++ b/test-configs.yaml
@@ -1613,6 +1613,7 @@ test_configs:
       - baseline
       - boot
       - cros-ec
+      - sleep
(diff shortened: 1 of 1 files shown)


gctucker:
Tested OK on staging:
```
11  | rk3399-gru-kevin    | arm64 | lab-collabora | clang-8  | defconfig          | 1/1    
12  | rk3399-gru-kevin    | arm64 | lab-collabora | gcc-8    | defconfig          | 1/1    
```

PR #213, opened 2019-10-14
lab-configs.yaml: add lab-clabbe tree whitelist

gctucker:
Add missing tree whitelist filter for lab-clabbe.

Fixes: 893bea670a4e ("add lab-configs.yaml")

### DIFF
--- a/lab-configs.yaml
+++ b/lab-configs.yaml
@@ -24,6 +24,10 @@ labs:
             - boot-nfs
             - simple
             - sleep
+          tree:
+            - kernelci
+            - next
+            - stable-rc
 
   lab-collabora:
     lab_type: lava


gctucker:
Tested locally usnig `kci_test list_jobs`.

lab-configs.yaml
```diff
@@ -24,6 +24,11 @@ labs:
             - boot-nfs
             - simple
             - sleep
+      - whitelist:
```
roxell: Do you need to say "- whitelist:" again ?

gctucker: No it's not needed actually, I've just push a fix for that.  It works in any case, but this it's a bit simpler to have only one whitelist filter.

PR #216, opened 2019-10-21
Add pull_tarball subcommand and switch build job to use it

mgalka:
kci_build pull_tarball downloads kernel sources tarball from the specified url and untars it to the kdir. It was impelemented to replace the Jenkins pipeline implementation of this mechanism.
It's possible to set number of retries if HTTP error occurs as well as to set the interval between retries.

e.g.
```
./kci_build pull_tarball --kdir /home/buildslave/staging/workspace/mgalka/kernel-build@2/linux --url http://staging-storage.kernelci.org/mgalka/mgalka/kernelci-local-snapshot-032/linux-src_mgalka.tar.gz --retries=12 --retry-sleep=5
```

 Remove pipeline implementation of the kernel source tarball download and use kci_build instead.


mgalka:
Local testing done and passed:
[x] Downloading `kci_build pull_tarball `
[x] Retries in case of HTTP status other than 200
[x] Retries sleep adjustment


Ran a job in my directory on staging - `kci_build pull_tarball` works in a build.jpl job
https://staging-ci.kernelci.org/job/mgalka/job/kernel-build/12/console

```
Running in /home/buildslave/staging/workspace/mgalka/kernel-build@2/kernelci-core
[Pipeline] {
[Pipeline] sh
+ ./kci_build pull_tarball --kdir /home/buildslave/staging/workspace/mgalka/kernel-build@2/linux --url http://staging-storage.kernelci.org/mgalka/mgalka/kernelci-local-snapshot-032/linux-src_mgalka.tar.gz --retries=12 --retry-sleep=5
[Pipeline] }
[Pipeline] // dir
[Pipeline] }
[Pipeline] // timeout
[Pipeline] }
[Pipeline] // stage
[Pipeline] stage
[Pipeline] { (Build)
[Pipeline] timeout
Timeout set to expire in 3 hr 0 min
[Pipeline] {
[Pipeline] dir
Running in /home/buildslave/staging/workspace/mgalka/kernel-build@2/kernelci-core
``

File: src/org/kernelci/build/Kernel.groovy [hunk 1]
```diff
@@ -1,66 +0,0 @@
-/*
```
gctucker: The `bisect.jpl` and `build-trigger.jpl` jobs also import this file.  So please don't delete it yet, but update the other jobs accordingly to use the new Python function and drop the import, then it can be deleted.

mgalka: I checked other files for usage of  `downloadTarball` function as it was the only one that was provided by `src/org/kernelci/build/Kernel.groovy`. I missed `bisect.jpl` and `build-trigger.jpl` as these were just importing the `Kernel.groovy`, but not using any of its code.
I removed the unnecessary imports from these files.

gctucker: Oh I see, they were stale imports.  Then that's easier, nice.

mgalka:
Removed stale imports and re-run staging build in my directory. It looks that build-trigger.jpl works fine https://staging-ci.kernelci.org/job/mgalka/job/kernel-build-trigger/7/console

gctucker:
Yes it seems fine.  We're missing a lot of lab-baylibre results from staging, checking if this is just a lab issue or a side effect of something else.  Hopefully this should also be ready to be merged once we've got more positive results on staging.

kernelci/build.py
```diff
@@ -341,6 +356,20 @@ def push_tarball(config, kdir, storage, api, token):
     return tarball_url
 
 
+def pull_tarball(kdir, url, dest_filename, retries, retry_sleep):
+    recreate_dir(kdir)
+    while retries > 0:
```
danrue: Consider something like `for i in range(0, retry_sleep)`, so that you don't need to modify retry_sleep (you can just `sleep(i)`), which I think is a bad idea. It also means the sleep will get longer each retry, rather than shorter. It may be better to have even more backoff, such as `sleep(2**i)`.

danrue: Sorry, I now notice retries and retry_sleep which I conflated in my first message. My concrete suggestion is to drop retry_sleep, and do `for i in range(1, retries): sleep(2**i)`.

mgalka: It looks like a good idea. I modified the implementation to gradually increase the sleep time in case of consequent failures.

kernelci/build.py
```diff
@@ -286,6 +287,20 @@ def make_tarball(kdir, tarball_name):
     subprocess.check_output(cmd, shell=True)
 
 
+def extract_tarball(dest_path, tarball_name):
```
gctucker: Why not just do that as part of `pull_tarball()`?

kernelci/build.py
```diff
@@ -872,3 +900,28 @@ def publish_kernel(kdir, install='_install_', api=None, token=None,
         resp.raise_for_status()
 
     return True
+
+
+def recreate_dir(path):
```
gctucker: Same comment here, this is a trivial function used in only one place inside this module. I don't see any reason why it should be put as a separate function.

kernelci/build.py
```diff
@@ -872,3 +900,28 @@ def publish_kernel(kdir, install='_install_', api=None, token=None,
...
+    directories don't exist they will be created.
+
+    *path* is a directory name or a path
+    """
+    if os.path.exists(path):
+        shutil.rmtree(path)
+    os.makedirs(path)
+
+
+def download_file(url, dest_filename, chunk_size=1024):
```
gctucker: This is a helper function used within this module, so please call it `_download_file()`.

gctucker:
Other than that, it was tested OK on staging so it's good from a functional point of view.

kernelci/build.py
mgalka: removed and code included in pull_tarbal()

kernelci/build.py
mgalka: removed and code included in pull_tarbal()

gctucker:
Thanks @mgalka - I think you just need to rebase it to get the Travis CI checks to pass.

gctucker:
Ah it's passing now, fine :)

gctucker:
Tested OK on staging.

PR #217, opened 2019-10-21
kernelci.test: add missing plan filter check in match_configs()

gctucker:
Add missing call to TestPlan.match() for all the test plans in a test
config before adding it to the list of jobs to run.  This was causing
unwanted things such as v4l2-compliance-vivid being run on all kernel
builds rather than only those with +virtualvideo fragment.

Fixes: a1a08c3876cf ("add kernelci.test with functions to generate test jobs")


gctucker:
Tested OK on staging.

PR #218, opened 2019-10-22
kernelci.config.build: Fix "variants" section reading from configuration

mgalka:
Lack of "build_configs_defaults" section or lack of "variants" section
within it caused KeyError during the build-configs.yaml parsing. This
fix provides appropriate default values in case these sections are
missing.

### DIFF
--- a/kernelci/config/build.py
+++ b/kernelci/config/build.py
@@ -318,7 +318,8 @@ class BuildConfig(YAMLObject):
         kw.update(cls._kw_from_yaml(
             config, ['name', 'tree', 'branch']))
         kw['tree'] = trees[kw['tree']]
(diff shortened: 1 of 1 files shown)


gctucker:
OK thanks, I'm not entirely sure if kci_build can actually be used without defaults but this change is fine in principle.

gctucker:
The subject should start with `kernelci.config.build:` rather than `build.py`.

mgalka:
> The subject should start with `kernelci.config.build:` rather than `build.py`.

Commit message topic and PR topic changed accordingly.

gctucker:
Tested OK on staging and confirmed that the unit tests are now all passing.

PR #220, opened 2019-10-23
lab-configs.yaml: blacklist android jobs in lab-collabora

gctucker:
The android tree is causing a lot of builds from downstream kernel
branches.  As such, they are not part of the initial goal of
kernelci.org which is to test the upstream kernel.  The extra load
added to test labs is significant enough to justify blacklisting this
tree when needed.

For these reasons, blacklist the android tree in lab-collabora as it
is starting to hit the limit of the lab's capacity.


gctucker:
Tested OK on staging.

PR #221, opened 2019-10-24
Modify debos crush.sh to skip specific components whenever required

Lakshmipathi:
Instead of crushing all components from image, provide skip (-s) option to ignore
specific components from available list. By default, all components will be removed.

### DIFF
--- a/jenkins/debian/debos/scripts/crush.sh
+++ b/jenkins/debian/debos/scripts/crush.sh
@@ -6,151 +6,221 @@ set -e
 # IMPORTANT: The Debian system is not longer functional at this point,
 # for example, apt and dpkg will stop working
 
-UNNEEDED_PACKAGES="apt libapt-pkg5.0 "\
-"ncurses-bin ncurses-base libncursesw5 libncurses5 "\
-"perl-base "\
-"debconf libdebconfclient0 "\
-"libfdisk1 "\
-"insserv "\
-"init-system-helpers "\
-"cpio "\
-"passwd "\
-"libsemanage1 libsemanage-common "\
-"libsepol1 "\
-"gzip "\
-"gnupg "\
-"gpgv "\
-"hostname "\
-"adduser "\
-"debian-archive-keyring "\
+compress_tool() {
+   UNNEEDED_PACKAGES="\
+    cpio \
+    gzip \
+"
 
+   remove_packages "${UNNEEDED_PACKAGES}"
+}
+
+hostname_tool() {
+   UNNEEDED_PACKAGES="hostname "
+
+   remove_packages "${UNNEEDED_PACKAGES}"
+}
+
+filesystem()  {
+   UNNEEDED_PACKAGES="libfdisk1 "
+
+   remove_packages "${UNNEEDED_PACKAGES}"
+
+   # Partition and file system tools
+   rm -f usr/sbin/*fdisk
+   rm -f usr/sbin/mkfs*
+   rm -f usr/sbin/fsck*
+
+   # No need for fuse
+   find usr etc -name '*fuse*' -prune -exec rm -r {} \;
+
+}
+
+package_management()  {
+   UNNEEDED_PACKAGES="\
+    apt  \
+    libapt-pkg5.0  \
+    debconf \
+    libdebconfclient0 \
+    debian-archive-keyring \
+"
+
+   remove_packages "${UNNEEDED_PACKAGES}"
+
+   # Show what's left package-wise before dropping dpkg itself
+   COLUMNS=300 dpkg -l
+
+   # Drop dpkg
+   dpkg --purge --force-remove-essential --force-depends  dpkg
+
+   # No apt or dpkg, no need for its configuration archives
+   rm -rf etc/apt
+   rm -rf etc/dpkg
+}
+
+systemd()  {
+   # Unused systemd generators
+   rm -f lib/systemd/system-generators/systemd-cryptsetup-generator
+   rm -f lib/systemd/system-generators/systemd-debug-generator
+   rm -f lib/systemd/system-generators/systemd-gpt-auto-generator
+   rm -f lib/systemd/system-generators/systemd-hibernate-resume-generator
+   rm -f lib/systemd/system-generators/systemd-rc-local-generator
+   rm -f lib/systemd/system-generators/systemd-system-update-generator
+   rm -f lib/systemd/system-generators/systemd-sysv-generator
+
+   # Efi blobs
+   rm -rf usr/lib/systemd/boot
+
+   # Translation catalogs
+   rm -rf usr/lib/systemd/catalog
+
+   # hostnamectl
+   rm -f usr/bin/hostnamectl
+
+   # boot analyser
+   rm -f usr/bin/systemd-analyze
+
+   # Misc systemd utils
+   rm -f usr/bin/bootctl
+   rm -f usr/bin/busctl
+   rm -f usr/bin/localectl
+   rm -f usr/bin/systemd-cat
+   rm -f usr/bin/systemd-cgls
+   rm -f usr/bin/systemd-cgtop
+   rm -f usr/bin/systemd-delta
+   rm -f usr/bin/systemd-detect-virt
+   rm -f usr/bin/systemd-mount
+   rm -f usr/bin/systemd-path
+   rm -f usr/bin/systemd-run
+   rm -f usr/bin/systemd-socket-activate
+
+   # systemd hw database manager (connman is in use)
+   find usr etc -name '*systemd-hwdb*' -prune -exec rm -r {} \;
+}
+
+systemd_dns_nw_ntp_tools() {
+   # Systemd dns resolver
+   find usr etc -name '*systemd-resolve*' -prune -exec rm -r {} \;
+
+   # Systemd network configuration
+   find usr etc -name '*networkd*' -prune -exec rm -r {} \;
+
+   # systemd ntp client (connman is in use)
+   find usr etc -name '*timesyncd*' -prune -exec rm -r {} \;
+}
+
+ncurses()  {
+   UNNEEDED_PACKAGES="\
+    ncurses-bin \
+    ncurses-base \
+    libncursesw5 \
+    libncurses5 \
+"
+   remove_packages "${UNNEEDED_PACKAGES}"
+
+   # Utils using ncurses
+   rm -f usr/bin/pg
+   rm -f usr/bin/watch
+   rm -f usr/bin/slabtop
+}
+
+misc_packages()  {
+   UNNEEDED_PACKAGES="\
+    perl-base \
+    insserv \
+    init-system-helpers \
+    adduser \
+    passwd \
+    libsemanage1 \
+    libsemanage-common \
+    libsepol1 \
+    gnupg \
+    gpgv \
+"
+   remove_packages "${UNNEEDED_PACKAGES}"
+}
+
+misc_directories()  {
+    # Drop directories not part of ostree
+    # Note that /var needs to exist as ostree bind mounts the deployment /var over
+    # it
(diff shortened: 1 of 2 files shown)


Lakshmipathi:
Hi @gctucker , Please review these changes. thanks.

jenkins/debian/debos/scripts/crush.sh
```diff
@@ -1,156 +1,202 @@
-#!/bin/sh
+#!/bin/bash
```
gctucker: Why do you need `bash`?

If it's for the arrays when parsing the command line options, I think we can have a simpler implementation instead.

jenkins/debian/debos/scripts/crush.sh
```diff
@@ -1,156 +1,202 @@
...
+   do
+	echo "Forcing removal of ${PACKAGE}"
+	if ! dpkg --purge --force-remove-essential --force-depends "${PACKAGE}"
+	then
+		echo "WARNING: ${PACKAGE} isn't installed"
+	fi
+   done
+}
+
+# Either skip specific components or crush all
```
gctucker: How about a simpler approach by providing the list of functions to call on the command line, and call them all if none is provided?

Say, `./crush.sh` would call all the functions to do the same as it current does.  But `./crush.sh systemd misc_directories` would only call the `systemd()` and `misc_directories()` functions.  That way, it would be clearer which things are actually crushed and it would also simplify the implementation by removing the need for `getopt` and bash arrays.

jenkins/debian/debos/scripts/crush.sh
```diff
@@ -1,156 +1,202 @@
...
+    rm -f usr/lib/*/security/pam_userdb.so
+    rm -f usr/lib/*/libdb-5.3.so
+
+    # remove NSS support for nis, nisplus and hesiod
+    rm -f usr/lib/*/libnss_hesiod*
+    rm -f usr/lib/*/libnss_nis*
+}
+
+# Removing unneeded packages
+remove_packages()  {
```
gctucker: There's already a debos template parameter to provide a list of packages to remove `extra_packages_remove`.

gctucker:
Also please rebase on top of current master branch.

jenkins/debian/debos/scripts/crush.sh
Lakshmipathi: okay will make these changes.

jenkins/debian/debos/scripts/crush.sh
Lakshmipathi: Its not for arrays. After adding `function` Using `sh` reports error. Under `debos` container `/bin/sh` links to `dash` and running script as `sh crush.sh` fails. We need to run it as `bash crush.sh`  to avoid below error. 

`/tmp/script/crush.sh: Syntax error: "(" unexpected (expecting ";;")`

gctucker: The kernelci rootfs images are built with the KernelCI "base" Docker image, not the one from Debos.

Lakshmipathi: okay, then I'll use `sh` while sending the patch.

jenkins/debian/debos/scripts/crush.sh
Lakshmipathi: If I understand correctly, template used to remove non-essentials packages. With crush.sh critical packages like `apt` gets removed.  `crush.sh` invoked right after removing packages listed with `extra_packages_remove` .

gctucker: Oh I see, this function is only a local helper within the `crush.sh` script.  You made this function because we now have several places where we're calling that.  So yes, that's fine.

jenkins/debian/debos/scripts/crush.sh
Lakshmipathi: Added these changes and tested with debos.

Lakshmipathi:
please review the updates.

jenkins/debian/debos/scripts/crush.sh
```diff
@@ -6,151 +6,181 @@ set -e
...
+
+   # Partition and file system tools
+   rm -f usr/sbin/*fdisk
+   rm -f usr/sbin/mkfs*
+   rm -f usr/sbin/fsck*
+
+   # No need for fuse
+   find usr etc -name '*fuse*' -prune -exec rm -r {} \;
+
+   rm -f usr/bin/hostnamectl
```
gctucker: `hostnamectl` is not a file system utility.  Maybe remove it as part of `systemd()`?

jenkins/debian/debos/scripts/crush.sh
```diff
@@ -6,151 +6,181 @@ set -e
...
-"libsepol1 "\
-"gzip "\
-"gnupg "\
-"gpgv "\
-"hostname "\
-"adduser "\
-"debian-archive-keyring "\
 
+filesystem()  {
+   UNNEEDED_PACKAGES="libfdisk1 cpio hostname gzip "
```
gctucker: `hostname` and `gzip` are not file system tools. `cpio` isn't exactly one either even though it's often used for initrd.

jenkins/debian/debos/scripts/crush.sh
```diff
@@ -6,151 +6,181 @@ set -e
...
+   rm -f usr/bin/systemd-detect-virt
+   rm -f usr/bin/systemd-mount
+   rm -f usr/bin/systemd-path
+   rm -f usr/bin/systemd-run
+   rm -f usr/bin/systemd-socket-activate
 
-# No zsh, no need for comletions
-rm -rf usr/share/zsh/vendor-completions
+   # Systemd dns resolver
+   #find usr etc -name '*systemd-resolve*' -prune -exec rm -r {} \;
```
gctucker: Rather than keeping this as a comment, maybe it should be in a separate function that we can decide to call if these files should be removed.

jenkins/debian/debos/scripts/crush.sh
```diff
@@ -6,151 +6,181 @@ set -e
...
 
-# Partition and file system tools
-rm -f usr/sbin/*fdisk
-rm -f usr/sbin/mkfs*
-rm -f usr/sbin/fsck*
 
-# local compiler
-rm -f usr/bin/localedef
+misc_packages()  {
+   UNNEEDED_PACKAGES="perl-base insserv init-system-helpers adduser passwd libsemanage1 libsemanage-common libsepol1 gnupg gpgv "
```
gctucker: Please break long lists of packages so we have only one per line.

gctucker: Maybe this is where we could remove `hostname`?

jenkins/debian/debos/scripts/crush.sh
```diff
@@ -6,151 +6,181 @@ set -e
...
 
-# Unused systemd generators
-rm -f lib/systemd/system-generators/systemd-cryptsetup-generator
-rm -f lib/systemd/system-generators/systemd-debug-generator
-rm -f lib/systemd/system-generators/systemd-gpt-auto-generator
-rm -f lib/systemd/system-generators/systemd-hibernate-resume-generator
-rm -f lib/systemd/system-generators/systemd-rc-local-generator
-rm -f lib/systemd/system-generators/systemd-system-update-generator
-rm -f lib/systemd/system-generators/systemd-sysv-generator
+    # local compiler
```
gctucker: `locale` rather than `local`

jenkins/debian/debos/scripts/crush.sh
```diff
@@ -6,151 +6,181 @@ set -e
...
 
-# Efi blobs
-rm -rf usr/lib/systemd/boot
+    # lsb init function leftovers
+    rm -rf usr/lib/lsb
 
-# Translation catalogs
-rm -rf usr/lib/systemd/catalog
+    # boot analyser
+    rm -f usr/bin/systemd-analyze
```
gctucker: Move this to `systemd()` maybe?

jenkins/debian/debos/scripts/crush.sh
```diff
@@ -6,151 +6,181 @@ set -e
...
+    rm -f usr/lib/*/libdb-5.3.so
+
+    # remove NSS support for nis, nisplus and hesiod
+    rm -f usr/lib/*/libnss_hesiod*
+    rm -f usr/lib/*/libnss_nis*
+}
+
+# Removing unneeded packages
+remove_packages()  {
+   UNNEEDED_PACKAGES="$@"
```
gctucker: I think the scope of `UNNEEDED_PACKAGES` is not local to this function, only `bash` provides support for this using `local`.  So it's probably better to either use a different name or directly use `$@` in the for-loop below.

jenkins/debian/debos/scripts/crush.sh
Lakshmipathi: moved `gzip` and `cpio` under `compress_tools()` and `hostname` to `hostname_tool()`

jenkins/debian/debos/scripts/crush.sh
Lakshmipathi: moved them into separate function `systemd_dns_nw_ntp_tools`

Lakshmipathi:
@gctucker please review recent changes.

jenkins/debian/debos/scripts/crush.sh
```diff
@@ -6,151 +6,221 @@ set -e
...
-# remove NSS support for nis, nisplus and hesiod
-rm -f usr/lib/*/libnss_hesiod*
-rm -f usr/lib/*/libnss_nis*
+   done
+}
+
+## crush all by default or only given components.
+if [ $# -eq 0 ]; then
+  filesystem
+  package_management
```
gctucker: Should this not be lower on the list?  If you remove `dpkg` and `apt` then you can't uninstall packages in other functions anymore.

Lakshmipathi: ah, my bad. let me fix it.

gctucker:
@Lakshmipathi I've tried to run the `buster.jpl` job on staging Jenkins but it failed with errors like this:
```
2019/11/14 15:00:13 crush.sh | dpkg: error: --purge needs a valid package name but 'hostname ' is not: illegal package name in specifier 'hostname ': character ' ' not allowed (only letters, digits and characters '-+._')
```
Please try to reproduce this locally to fix `crush.sh`, this is the `debos` command that the Jenkins produced:
```
mkdir -p 20191114.0/amd64
debos -m 2G -t architecture:amd64 -t suite:buster -t basename:20191114.0/amd64 -t extra_packages: -t extra_packages_remove:"bash e2fslibs e2fsprogs klibc-utils libext2fs2 libgnutls30 libklibc libncursesw6 libp11-kit0 libunistring2 sensible-utils"  -t extra_files_remove:"*networkd* *resolved* tar patch dir partx find"  -t script: -t test_overlay: jenkins/debian/debos/rootfs.yaml
```

Lakshmipathi:
@gctucker  ok, will reproduce and resolve this issue .

Lakshmipathi:
Made required changes and tested it locally. Also ensured that `debian-buster`  job on staging Jenkins is successful.

gctucker:
Finally got a job working on staging:
https://staging-storage.kernelci.org/images/rootfs/debian/buster/20191129.1/?C=M&O=D

Now we should extract some rootfs archives and compare with the versions currently in production to check the same files are there.

gctucker:
Checked that the contents of `full.rootfs.tar.xz` and `initrd.cpio.gz` from staging and production were functionally identical (and actually exactly identical for the initrd).  I used these versions:
https://storage.kernelci.org/images/rootfs/debian/buster/20191202.1/arm64/
https://staging-storage.kernelci.org/images/rootfs/debian/buster/20191129.1/arm64/

PR #222, opened 2019-10-24
kci_test: don't generate jobs for builds that failed

gctucker:
If the build status is not PASS then don't generate any job
definitions.


gctucker:
Tested OK on staging.

PR #223, opened 2019-10-24
build-trigger.jpl: set special build job params allmodconfig

gctucker:
Add a NODE_LABEL parameter in build.jpl to dynamically pick a Jenkins
node on which to run the build.  Use this in build-trigger.jpl to
select a "big-config-builder" node with maximum number of parallel
build processes when the defconfig contains allmodconfig.


gctucker:
Tested OK on staging, the designated "big config" builds landed on the build server set up with the `big-config-builder` label.

PR #224, opened 2019-10-28
build-configs.yaml: add arm64 tree with for-kernelci branch

gctucker:
Add the arm64 tree and a build configuration using the for-kernelci
branch with only arm64 and the requested configs.

Link: https://groups.io/g/kernelci/topic/branch_for_kernelci/34551186


gctucker:
Tested OK on staging up to when the kernel builds were triggered (stopped before email reports):
https://staging-storage.kernelci.org/arm64/for-kernelci/v5.4-rc4-31-g7b16c1c76197/

PR #225, opened 2019-10-30
Add Travis CI

gctucker:
This PR depends on #226 to be merged first (to fix a unit test)

Add basic Travis CI definition to run existing unit tests and pep8 checks.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,15 @@
+dist: xenial
+
+language: python
+
+python:
+  - "2.7"
+
+install:
+  - pip install -r requirements.txt
+  - pip install pycodestyle
+
+script:
+  - make test
+  - pycodestyle kernelci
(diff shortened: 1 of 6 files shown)


mattface:
I'm all in favour of this, but I believe last time it was mentioned you were planning to do this on our own Jenkins?

gctucker:
@mattface We need both, there needs to be a Jenkins job to automatically schedule tests on staging as well as an automated way of running unit tests and static analysis.  Also it wasn't clear whether we would keep the code on Github or move to some Gitlab instance.

Now that we have a pseudo-automatic solution for staging with a script that does pretty much everything and is almost ready to be automated in Jenkins, and that there doesn't seem to be any incentive to move to something else than Github, it makes sense to run these tests on each pull request.

mattface:
@gctucker makes sense to me, sounds good

PR #226, opened 2019-10-30
kernelci.config.build: make reference optional

gctucker:
If for any reason no reference can be used with a build configuration,
and none is set either in the build configuration or in the defaults,
then set the reference to None.

Update kci_build accordingly to return False if there was no matchin
build configuration with a reference.

This also fixes the unit tests.

Fixes: 3b40138f4f4e ("kci_build: Add get_reference subcommand")


mgalka:
LGTM

PR #227, opened 2019-10-30
T16859 doc and clean up

gctucker:
Add docstrings and clean up a few loose ends in recently added modules and kci_test.


gctucker:
Tested OK on staging.

PR #230, opened 2019-11-04
WIP - Creation of buster-xfstests rootfs

Lakshmipathi:
Modified rootfs.yaml to accept crush_image_options,multiple_partition as optional parameters.
Added xfstests rootfs build script and jpl template file.

### DIFF
--- a/jenkins/debian/debos/scripts/buster-xfstests.sh
+++ b/jenkins/debian/debos/scripts/buster-xfstests.sh
@@ -1,0 +1,44 @@
+#!/bin/bash
+
+#####################################################
+# This script performs xfstests build
+#####################################################
+
+set -e
+
+# Build-depends needed to build the test suites, they'll be removed later
+BUILD_DEPS="\
+    xfslibs-dev \
+    uuid-dev \
+    libacl1-dev \
+    libaio-dev \
+    libattr1-dev \
+    git \
+    automake \
+    gcc \
+    make \
+    libtool-bin \
+"
+apt-get install --no-install-recommends -y  ${BUILD_DEPS}
+
+BUILDFILE=/test_suites.json
+echo '{  "tests_suites": [' >> $BUILDFILE
+XFSTESTS_URL=git://git.kernel.org/pub/scm/fs/xfs/xfstests-dev.git
+git clone --depth=1 $XFSTESTS_URL /xfstests && cd /xfstests
+
+echo '    {"name": "xfstests", "git_url": "'$XFSTESTS_URL'", "git_commit": ' \"`git rev-parse HEAD`\" '}' >> $BUILDFILE
+echo '  ]}' >> $BUILDFILE
+
+# Setup xfstests accounts, directory
+useradd fsgqa -m
+useradd 123456-fsgqa
+
+mkdir /test /scratch
+
+# Build xfstests
+cd /xfstests && make && make install
+
+########################################################################
+# Cleanup: remove files and packages we don't want in the images       #
+########################################################################
+apt-get remove --purge --allow-remove-essential -y  ${BUILD_DEPS}
--- a/jenkins/buster-xfstests.jpl
+++ b/jenkins/buster-xfstests.jpl
@@ -1,0 +1,48 @@
+@Library('kernelci') _
+import org.kernelci.debian.RootFS
+
+/* ----------------------------------------------------------------------------
+ * Jenkins parameters
+
+DOCKER_BASE
+  Dockerhub base address used for the build images
+
+*/
+
+def r = new RootFS()
+
+def config = [
+    'name':"buster-xfstests",
+    'arch_list': [
+        "i386",
+        "amd64",
+    ],
+    'debian_release':"buster",
+    'extra_packages':"\
+xfsprogs \
+e2fsprogs \
+uuid-runtime \
+libuuid1 \
+acl \
+fio \
+dbench \
+quota \
+attr \
+bsdmainutils \
+gawk \
+python \
+sqlite3 \
+bc \
+libgdbm-dev \
+perl \
+hostname \
+net-tools \
+",
+    'extra_packages_remove':"",
+    'crush_image_options':"package_management systemd ncurses misc_packages misc_directories",
+    'script':"scripts/buster-xfstests.sh",
+    'test_overlay': "overlays/xfstests",
+    'docker_image': "${params.DOCKER_BASE}debos",
+]
+
+r.buildImage(config)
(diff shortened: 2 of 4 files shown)


jenkins/debian/debos/rootfs.yaml
```diff
@@ -105,6 +107,29 @@ actions:
     chroot: false
     command: cd ${ROOTDIR} ; find -H  |  cpio -H newc -v -o | gzip -c - > ${ARTIFACTDIR}/{{ $basename -}}/full.rootfs.cpio.gz
 
+{{ if $multiple_partition }}
```
gctucker: The partitions should be on the storage device used to run the tests, not in the rootfs image.

jenkins/debian/debos/rootfs.yaml
```diff
@@ -6,6 +6,8 @@
 {{- $test_overlay := .test_overlay -}}
 {{- $extra_packages_remove := or .extra_packages_remove -}}
 {{- $extra_files_remove := or .extra_files_remove -}}
+{{- $crush_image_options := or .crush_image_options "" -}}
```
gctucker: Please keep changes to the common files such as `rootfs.yaml` in a separate commit from the one adding the new file system.

jenkins/debian/debos/scripts/buster-xfstests.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+
+#####################################################
+# This script create buster image for xfstests
+#####################################################
+
+set -e
+
+# Install required xfstests packages
+apt-get install --no-install-recommends xfslibs-dev uuid-dev libtool-bin \
```
gctucker: Please keep each package on a separate line, it makes it easier to see diffs when this changes.

gctucker: Also please follow the same approach as in `buster-v4l2.sh` by listing the packages in a variable i.e. `$BUILD_DEPS` so they can be uninstalled at the end of the script.  The build dependencies should not be needed as part of the rootfs at runtime, only to build the test suite.

jenkins/debian/debos/scripts/buster-xfstests.sh
```diff
@@ -0,0 +1,36 @@
...
+	        e2fsprogs automake gcc libuuid1 quota attr libattr1-dev make \
+	        libacl1-dev libaio-dev xfsprogs libgdbm-dev gawk fio dbench \
+	        uuid-runtime python sqlite3 git bc bsdmainutils acl -y
+
+apt-get install perl hostname net-tools -y
+
+# Clone repo
+git clone git://git.kernel.org/pub/scm/fs/xfs/xfstests-dev.git /xfstests
```
gctucker: Please store the URL and revision in a local manifest file so it can be added to the rootfs build-info.json.

jenkins/debian/debos/scripts/buster-xfstests.sh
```diff
@@ -0,0 +1,36 @@
...
+
+# Setup xfstests accounts, directory
+useradd fsgqa -m
+useradd 123456-fsgqa
+
+mkdir /test /scratch
+
+# Setup default xfstests config
+cat > /xfstests/local.config <<'EOF'
```
gctucker: What if we want to run tests on something else than `/dev/sda`?  I think this should be set at runtime by the LAVA job, with a template parameter define in `test-configs.yaml`.

jenkins/debian/debos/rootfs.yaml
Lakshmipathi: ah, okay I was under the assumption, its fixed within rootfs.  Will remove this option.

jenkins/debian/debos/rootfs.yaml
Lakshmipathi: okay will send a new PR request with changes only from `rootfs.yaml`

jenkins/debian/debos/scripts/buster-xfstests.sh
Lakshmipathi: okay, will apply these changes.

jenkins/debian/debos/scripts/buster-xfstests.sh
Lakshmipathi: Since partition used inside rootfs image itself this remained static. Now we are moving partition out of 'rootfs' image, yes , it should be set at runtime. Will check and add it  `test-configs.yaml`

jenkins/debian/debos/scripts/buster-xfstests.sh
Lakshmipathi: With this list - xfstests installs both dependencies for compilation and also for tests like `fio` or `e2fsprogs` etc. Will remove `lib` packages from this list and verify test results.

jenkins/debian/debos/scripts/buster-xfstests.sh
Lakshmipathi: I'm unable figure out this.  I looked into `test-configs.yaml` I assume, we will add something like `fstype: 'ext4'` which will be used in `templates/xfstests/file.jinja2` which runs the job?

jenkins/debian/debos/scripts/buster-xfstests.sh
Lakshmipathi: made this change and tested.

jenkins/debian/debos/scripts/buster-xfstests.sh
Lakshmipathi: Moved packages required for testing into `extra_packages` on `jenkins/buster-xfstests.jpl` . Also removed packages after compilation from rootfs.

jenkins/debian/debos/scripts/buster-xfstests.sh
gctucker: You can use the `params` attribute in a test definition, this is a dictionary passed to the template.

jenkins/debian/debos/rootfs.yaml
Lakshmipathi: Its a two line change related to `crush_image_options`, should it be sent with https://github.com/kernelci/kernelci-core/pull/221 PR instead this one? PR 221 provides better context to this change.

Lakshmipathi:
@gctucker please review recent changes.

gctucker:
This needs to be rebased now that #221 has been merged.

gctucker:
@Lakshmipathi We need only one PR, force-pushing to this one is fine and helps with keeping the review history.  What is #292 for?

Lakshmipathi:
okay reverted few commits and force-pushed required changes. Will close #292

gctucker:
Let's get the `kci_rootfs` changes merged and then revisit this PR to use the new YAML configuration.

gctucker:
A new PR should be made using `kci_rootfs`, with Debian Buster.

Issue #228, opened 2019-10-31
Add branch android-3.18-preview from Lee's Korg repo


lag-linaro:
Each git kernel branch is monitored every hour by kernelci.org.  Whenever a new
revision is detected, it will be built for a number of combinations of
architectures, defconfigs and compilers.  Then a build report will be sent,
some tests will be run and test reports will also be sent.

Please provide the information described below in order to add a new branch to
kernelci.org:

- **How much build coverage do you need for your branch?**

Generally speaking, a good rule is to build fewer variants for branches that
are "further" away from mainline and closer to individual developers.  This can
be fine-tuned with arbitrary filters, but essentially there are 3 main options:

1. Build everything, including allmodconfig, which amounts to about 220 builds.
This is we do with linux-next.

2. Skip a few things such as allmodconfig as it's very long to build and
doesn't really boot, and also architectures that are less useful such as MIPS
which saves 80 builds and doesn't have much test platforms in KernelCI.  This
is we do with some subsystems such as linux-media.

3. Build only the main defconfig for each architecture to save a lot of build
power, get the fastest results and highest boots/builds ratio.  This is what do
with some maintainer branches such as linusw' GPIO branch.

⇨ Choice: 1 (allmodconfig, but only for x86_64, arm and arm64 needed though)

- **How often do you expect this branch to be updated?**

If you push once a week or less, it's easier to allow for many build variants
as this reduces the build load on average.  Conversely, if you push several
times every day then a small set of builds should be used.

It's also possible to increase the build capacity if needed but this comes with
a cost.  Avoiding unnecessary builds is always a good way to reduce turnaround
time and not waste resources.

⇨ Estimated frequency: ~4 times a week

- **Who should the email reports be sent to?**

Standard email reports inlude builds and basic tests that are run on all
platforms. Please provide a list of email recipients for these.  Typical ones
are the regular KernelCI reports list, kernel mailing lists associated with the
changes going into the branch and related maintainers.

⇨ Recipients: lee <dot> jones <at> linaro <dot> org

lag-linaro:
Is there an E.T.A on this task?

gctucker:
@mattface Can you please take care of this one?

@lag-linaro One thing missing from the issue template is the git repo URL and branch name, could you please provide them?

I'll fix the template in the meantime...

lag-linaro:
Repo: git://git.kernel.org/pub/scm/linux/kernel/git/lee/linux.git
Branch: android-3.18-preview

[...]

PR #232, opened 2019-11-05
Add builds for 3.18 from Lees tree

mattface:
Add build configs and emails for the 3.18 branch on Lees tree.
Only arm, arm64 and x86_64.

Fixes issue #228


gctucker:
LGTM, we'll see if this doesn't cause any bad side-effects with a test run on staging.kernelci.org.

lag-linaro:
How did your testing go? Is there an ETA on when this will go live?

PR #240, opened 2019-11-11
build-configs: Add 'allmodconfig' builds to Armv7's default configuation

lag-linaro:
There shouldn't be any reason to have 'allmodconfig' builds for Arm64 and
x86_64 and to leave Armv7 out of the equation.

### DIFF
--- a/build-configs.yaml
+++ b/build-configs.yaml
@@ -229,6 +229,7 @@ build_configs_defaults:
         arm: &arm_arch
           base_defconfig: 'multi_v7_defconfig'
           extra_configs:
+            - 'allmodconfig'
             - 'allnoconfig'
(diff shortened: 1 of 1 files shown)


gctucker:
It used to never build, and now there are still some build failures in linux-next but it passed at least once last month (next-20191015): https://kernelci.org/build/id/5da56fe659b5140fa97525a0/

So yes I think we should enable it by default now.

gctucker:
Quick check that it does the right thing:
```
$ ./kci_build list_kernel_configs --kdir=linux --config=mainline | grep allmodconfig
x86_64 allmodconfig gcc-8
arm64 allmodconfig gcc-8
arm allmodconfig gcc-8
```

PR #241, opened 2019-11-11
T17441 recover git repo

gctucker:
Deal with failure to update a git repo in build-trigger.jpl by removing it and starting again.  Also improve error propagation in `kernelci.build`.


gctucker:
Tested OK on staging that it has no bad visible side effects.

PR #242, opened 2019-11-12
T17213 bisect.jpl: use kci_test with test plan variant parameter

gctucker:
Replace the PLAN job parameter with TEST_PLAN and TEST_PLAN_VARIANT,
and use it to run test jobs with kci_test.  Also store the lab info
JSON file and stash it along with the build meta-data and dtbs JSON
files produced by kci_build install_kernel.

This relies on the matching change in kernelci-backend to trigger
bisections with the new parameters including the test plan variant
name.


gctucker:
Tested OK on staging.

PR #243, opened 2019-11-12
bisect.jpl: tweak email subject

gctucker:
Remove the git describe to shorten the email subject and move the test
plan name after "bisection:" in preparation for test case bisections.

For example, rather than:
```
broonie-sound/for-next boot bisection: v5.4-rc6-292-gc1efaea10be0 on sun8i-h3-libretech-all-h3-cc
```

we would now have:
```
broonie-sound/for-next bisection: boot on sun8i-h3-libretech-all-h3-cc
```

or potentially:
```
broonie-sound/for-next bisection: v4l2-compliance on sun8i-h3-libretech-all-h3-cc
```

The git describe string was also a bit misleading as it's the "bad"
kernel revision initially tested, not the one found by the bisection.

To better connect this report with the commit it has found, we should
make it reply to the original thread when the patch was submitted on a
mailing list.

### DIFF
--- a/jenkins/bisect.jpl
+++ b/jenkins/bisect.jpl
@@ -466,7 +466,9 @@ def bisectNext(kdir, status) {
  */
 
 def pushResults(kci_core, kdir, checks, params_summary) {
-    def subject = "${params.KERNEL_TREE}/${params.KERNEL_BRANCH} ${params.PLAN} bisection: ${params.KERNEL_NAME} on ${params.TARGET}"
+    def subject = "\
+${params.KERNEL_TREE}/${params.KERNEL_BRANCH} bisection: \
+${params.TEST_PLAN} on ${params.TARGET}"
 
     dir(kci_core) {
         withCredentials([string(credentialsId: params.KCI_TOKEN_ID,


mattface:
@broonie is our resident email aficionado

gctucker:
Indeed, @broonie is now KernelCI's email reports editor-in-chief ;)

P.S. The `broonie-sound` example was entirely innocent :innocent:

gctucker:
> This makes sense to me, I might be tempted to add "boot" or something to the subject but that's a separate change.

There is "boot" when the bisection was for boot:
```
broonie-sound/for-next bisection: boot on sun8i-h3-libretech-all-h3-cc
```

broonie:
Yeah, I was thinking s/bisection: boot on/boot bisection on:/ or something - not 100% sure it's worth it though and like I say it's definitely a separate thing.

gctucker:
The current email subjects say "boot bisection" but the issue is that other test plans may not fit in very well with this scheme, say "tree/branch v4l2-compliance uvcvideo bisection on platform" is harder to read than "tree/branch bisection: v4l2-compliance uvcvideo on platform".  Also the other KernelCI email subjects start with "tree/branch build" or "tree/branch xyz: details".

But that's all rather subjective and easy enough to tweak again if needed.

gctucker:
Got this tested on staging as part of other bisection related tests, here's the subject from the email report:

```
kernelci/staging.kernelci.org bisection: boot on qemu_arm64-virt-gicv2
```

PR #244, opened 2019-11-12
T16848 build-trigger.jpl: build arm64 defconfig with medium-config-builder

gctucker:
Now that allmodconfig is being built on large builders with the
big-config-builder label, the arm64 defconfig is clearly the one that
takes the longest to complete with the default "make -j4" approach.
This can take up to 45min.  For this kind of medium-sized builds, we
can have dedicated smaller builders to run say at -j8 or -j12 and keep
each individual build well under 30min.


gctucker:
Testing this on staging with the `terbium` builder.

gctucker:
Tested OK on staging, the arm64 defconfig jobs landed on terbium.

Build time was 2205 sec and it's now 1503 sec, although `terbium` is probably the slowest builder on staging and the previous build was on `build01` which is faster for the same number of parallel build processes.

PR #245, opened 2019-11-14
jenkins: kernel-arch-complete: Send 'android-3.18' results to Lee Jones

lag-linaro:


gctucker:
@lag-linaro Do you only care about build results and not test runs?

gctucker:
It's fine to merge this now in any case as the test results will be sent to the regular recipients for Android branches.  It can be fine-tuned again for the next production update.

lag-linaro:
I don't look at test runs currently.

When I did try, they weren't helpful, as they seemed to fail for no reason.

It would be better if you could re-run (or run them initially?) with `earlycon` support.

PR #249, opened 2019-11-19
Remove unused files

gctucker:
Remove all the scripts and associated files that are not used anymore now that we have kci_test with YAML configuration and test-runner.jpl.


gctucker:
No, kernelci doesn't generate health checks.  It can generate arbitrary jobs though, and you can use that as a basis for your health checks if you want.  You might just as well copy a job definition from a kernelci job that has run in your lab.

Say you have a kernel build locally with the JSON files, you can use `kci_test generate` with the appropriate arguments to generate any LAVA job definition based on the KernelCI templates.

gctucker:
Tested OK on staging, all the builds and tests were run a a bisection completed as expected.

PR #250, opened 2019-11-19
debian: add new test rootfs with python for Chromebooks

eballetbo:
This new rootfs contains the python packages required to run some
Chromebooks tests written in python. The tests are still under
developement so this doesn't contain the chromebooks test suite itself
but the rootfs can be used as a base for a job to run those tests.

### DIFF
--- a/jenkins/buster-cros-ec.jpl
+++ b/jenkins/buster-cros-ec.jpl
@@ -1,0 +1,24 @@
+@Library('kernelci') _
+import org.kernelci.debian.RootFS
+
+/* ----------------------------------------------------------------------------
+ * Jenkins parameters
+
+DOCKER_BASE
+  Dockerhub base address used for the build images
+
+*/
+
+def r = new RootFS()
+
+def config = ['name':"buster-cros-ec",
+              'arch_list':["armhf", "arm64", "amd64"],
+              'debian_release':"buster",
+              'extra_packages':"python3-minimal python3-unittest2 python3-numpy",
+              'extra_packages_remove':"",
+              'script':"",
+              'test_overlay': "",
+              'docker_image': "${params.DOCKER_BASE}debos",
+             ]
+
+r.buildImage(config)


jenkins/buster-chromebooks.jpl
```diff
@@ -0,0 +1,24 @@
...
+ * Jenkins parameters
+
+DOCKER_BASE
+  Dockerhub base address used for the build images
+
+*/
+
+def r = new RootFS()
+
+def config = ['name':"buster-chromebooks",
```
gctucker: This rootfs doesn't actually have anything specific to Chromebooks.  Instead, it just has extra Python packages.  So I think we should call it `buster-python` instead to use it with any test suite that requires Python.

jenkins/buster-chromebooks.jpl
```diff
@@ -0,0 +1,24 @@
...
+              'arch_list':["armhf", "arm64", "amd64"],
+              'debian_release':"buster",
+              'extra_packages':"python3-minimal python3-unittest2 python3-numpy",
```
gctucker: What is the resulting uncompressed size of the ramdisk, compared to the plain`buster` rootfs?

Maybe some files in `/usr/share` or elsewhere could be removed with the `extra_files_remove` attribute to optimize the size.

eballetbo: Actually the size is 30M, I suppose there are some files that we can remove but I need to look at. Is it close to a non-working rootfs because of the size?

jenkins/buster-chromebooks.jpl
eballetbo: The purpose is to include the chromebooks test in this rootfs, hence the name includes chromebooks. But as you know these are still under development so our plan is to use that rootfs as a base and clone the test suite meanwhile. We will end with a buster-chromebooks rootfs so if nobody is interested in python right now I am not sure makes sense to rename. Anyway, both solutions work for me now, so just tell what do you prefer.

jenkins/buster-chromebooks.jpl
gctucker: OK I think it's actually the compressed size of the ramdisk that matters from a bootloader point of view.  Each device type has a different maximum size available for the ramdisk, bigger ramdisks are more likely to fail especially as the kernel image becomes bigger.  Also, smaller ramdisk images take less time to download and use less data.

The plain buster `rootfs.cpio.gz` is 15M, so that's basically twice the size.  I bet there are a lot of files that aren't really needed.  We can build one rootfs on staging as-is and try it anyway, especially since we already know that there will be a few iterations with this rootfs to then include the test suite.

jenkins/buster-chromebooks.jpl
gctucker: OK, then maybe it should be named with the test suite it will include.  If it's for cros-ec then maybe use that in the name (e.g. `buster-cros-ec`) rather than `buster-chromebook`.

eballetbo:
Updated the name to buster-cros-ec.

PR #253, opened 2019-11-19
kci_test: generate: support --mach

khilman:
Add support for generating only jobs for a specific 'mach'.

This is useful for arch/platform/SoC maintainer trees where they are
primarily concerned with tests on a specific SoC family.


gctucker:
Fair enough.  I wonder if there should be a more generic way of filtering which jobs to generate based on any device property, rather than adding one option specifically for `mach`.  Say, if we only want to generate jobs for devices with the big-endian flag or with a dtb or anything like that.

khilman:
Yes, there probably should be something more generic as well, but we can add those as needed.

gctucker:
OK fine for now.  Tested on staging that it didn't have any side-effects.

PR #254, opened 2019-11-19
kci_build: install_kernel: add publish_path

khilman:
Add the ability to specify the 'publish_path' which is used
to set the 'file_server_resource' in the build metadata.

This feature is useful when using kci_build + kci_test without a
kernelci-backend, and where the build output is desired in a specific
place.


gctucker:
Please fix the issues reported by the Travis CI checks.

gctucker:
Tested OK on staging (no bad side effects as this option isn't used in Jenkins jobs).

PR #255, opened 2019-11-19
kci_test: get_lab_info: ensure dir exists

khilman:
When creating the lab-json file, make sure that any directory
components exist before creating the file.

### DIFF
--- a/kci_test
+++ b/kci_test
@@ -92,6 +92,9 @@ class cmd_get_lab_info(Command):
             'url': lab.url,
             'devices': lab_api.devices,
         }
+        dir = os.path.dirname(args.lab_json)
+        if dir and not os.path.exists(dir):
+            os.makedirs(dir)
         with open(args.lab_json, 'w') as json_file:
             json.dump(data, json_file, indent=4, sort_keys=True)
         return True


kci_test
```diff
@@ -92,6 +92,8 @@ class cmd_get_lab_info(Command):
             'url': lab.url,
             'devices': lab_api.devices,
         }
+        dir = os.path.dirname(args.lab_json)
+        os.makedirs(dir)
```
gctucker: Please first check that the directory exists, otherwise this throws an exception:
```
  File "./kci_test", line 207, in <module>
    status = args.func(test_configs, lab_configs, args)
  File "./kci_test", line 96, in __call__
    os.makedirs(dir)
  File "/usr/lib/python2.7/os.py", line 157, in makedirs
    mkdir(name, mode)
OSError: [Errno 17] File exists: '/home/buildslave/build-trigger/workspace/kernel-build-trigger/labs'
```
https://staging-ci.kernelci.org/job/kernel-build-trigger/432/console

gctucker: Or maybe this really should be done outside of kci_test, by creating the directory before calling kci_test...

khilman: in Python >= 3.2, we can just `os.makedirs(path, exists_ok=True)`

gctucker: Maybe wait for #261 to get merged and then rebase?

khilman: More robust, I'll just add an os.path.exists() check.

gctucker: Fair enough.

kci_test
```diff
@@ -92,6 +92,9 @@ class cmd_get_lab_info(Command):
             'url': lab.url,
             'devices': lab_api.devices,
         }
+        dir = os.path.dirname(args.lab_json)
+        if not os.path.exists(dir):
+            os.makedirs(dir)
```
gctucker: This fails if the JSON file path is the current directory, because then `dir` is an empty string:
```
+ ./kci_test get_lab_info --lab=lab-baylibre --lab-json=lab-info.json --user=kernel-ci --token=****
 ​File "./kci_test", line 209, in <module>
   ​status = args.func(test_configs, lab_configs, args)
 ​File "./kci_test", line 98, in __call__
   ​os.makedirs(dir)
 ​File "/usr/lib/python3.7/os.py", line 221, in makedirs
   ​mkdir(name, mode)
FileNotFoundError: [Errno 2] No such file or directory: ''
```

khilman: Easy enough to check that `dir` is non-empty.

I know it's common practice for scripts to have to mkdir before calling, but I find that clumsy and error prone.  It's simple enough for the script to "do the right thing", so we should, IMO.

gctucker: Alrighty :)

PR #256, opened 2019-11-20
test-configs: meson8b-odroidc1: use amlogic for mach

khilman:
Convert the only 32-bit amlogic platform from mach=meson to
mach=amlogic to match the 64-bit platforms.


gctucker:
Tested OK on staging: https://staging.kernelci.org/boot/id/5ddd3d4de1abb11d8ced498e/

PR #260, opened 2019-11-20
kernelci.build: fix git_describe_v in build meta-data

gctucker:
The git describe "verbose" attribute in the build meta-data should be
called git_describe_v rather than git_describe_verbose to match what
kernelci-backend is expecting.  Fix this so the git_describe_v gets
stored correctly in the build documents.


gctucker:
Tested OK on staging.  This also fixed the full git describe in the data submitted with kcidb.

PR #261, opened 2019-11-20
Convert all Python files to Python 3

gctucker:
Convert all the kci_* utilities, the kernelci package and other
scripts to Python 3.


gctucker:
Tested OK several times on staging.

PR #262, opened 2019-11-21
Separate job name from job filename

mgalka:
This commit adds filename parameter that'll be used as a name of the
    LAVA job file. Dtb information has been removed from the job name
    and it's length has been reduced to 200 characters. It's meant to fix
    an issue with job names longer than 200 characters.

### DIFF
--- a/kernelci/test.py
+++ b/kernelci/test.py
@@ -79,8 +79,11 @@ def get_params(bmeta, target, plan_config, storage):
     file_server_resource = bmeta['file_server_resource']
     job_px = file_server_resource.replace('/', '-')
     url_px = file_server_resource
-    job_name = '-'.join([job_px, dtb or 'no-dtb',
-                         target.name, plan_config.name])
+    job_name = '-'.join([job_px, target.name, plan_config.name])
+    if len(job_name) > 200:
+        job_name = '{}...'.format(job_name[:197])
+    job_filename = '-'.join([job_px, dtb or 'no-dtb',
+                             target.name, plan_config.name])
     base_url = urlparse.urljoin(storage, '/'.join([url_px, '']))
     kernel_img = bmeta['kernel_image']
     kernel_url = urlparse.urljoin(storage, '/'.join([url_px, kernel_img]))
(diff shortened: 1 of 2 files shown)


mgalka:
Tested locally

kernelci/test.py
```diff
@@ -81,6 +81,7 @@ def get_params(bmeta, target, plan_config, storage):
     url_px = file_server_resource
     job_name = '-'.join([job_px, dtb or 'no-dtb',
                          target.name, plan_config.name])
+    job_filename = '-'.join([job_px, target.name, plan_config.name])[:200]
```
gctucker: The issue is not the length of the file name, but the length of the LAVA job name.

mgalka: Right, I misunderstood it. Corrected.

kernelci/test.py
```diff
@@ -104,6 +105,7 @@ def get_params(bmeta, target, plan_config, storage):
 
     params = {
         'name': job_name,
+        'filename': job_filename,
```
gctucker: Ah, except the file name isn't strictly an attribute of the job.  I think it's OK but slightly on the edge of what should really be in this dictionary.  The job file name used when generating LAVA job definition files is a LAVA-specific thing, whereas `kernelci.test` is only supposed to produce lab-independent data.  Also the 200-character limit is a LAVA thing.

So I think it would be better to leave this part as-is in `kernelci.test` and only truncate the job name in `kernelci.lab.lava` to ensure it's under 200 characters so it doesn't fail to get submitted.  It's OK to edit the dictionary returned from `get_params()` so the job name can be truncated there directly.

kernelci/test.py
```diff
@@ -79,8 +79,11 @@ def get_params(bmeta, target, plan_config, storage):
     file_server_resource = bmeta['file_server_resource']
     job_px = file_server_resource.replace('/', '-')
     url_px = file_server_resource
-    job_name = '-'.join([job_px, dtb or 'no-dtb',
```
gctucker: Again, there's no need to change anything in `kernelci.test`.  The job name length is a LAVA limitation and it can just be truncated in `kernelci.lab.lava`.

gctucker:
Replaced with #271

PR #267, opened 2019-11-25
lab/lava: fix alias handling

khilman:
The aliases dict key is called "device_types" (plural), and is
a list, not a string, so check if the device_type being searched for
is in the list, not matching a string.

### DIFF
--- a/kernelci/lab/lava.py
+++ b/kernelci/lab/lava.py
@@ -47,7 +47,7 @@ def get_device_type_by_name(name, device_types, aliases=[]):
     for alias in aliases:
         if alias["name"] == name:
             for device_type in device_types:
-                if alias["device_type"] == device_type:
+                if device_type in alias["device_types"]:
                     return device_type
     return None
 


khilman:
This fix is required to make the boards in the recently submitted lab-nxp work.

The official device-type name in the lab is "nxp-ls2088", but it has an alias to match the upstream DT name "fsl-ls2088a-rdb".

khilman:
In addition to the list vs. string problem, here's the traceback due to the wrong key name:
```
  File "./kci_test", line 203, in <module>
    status = args.func(test_configs, lab_configs, args)
  File "./kci_test", line 130, in __call__
    if not api.device_type_online(device_type):
  File "/work/kernel/ci/kernelci-core/kernelci/lab/lava.py", line 112, in device_type_online
    device_type.base_name, devices.keys(), self.devices['aliases'])
  File "/work/kernel/ci/kernelci-core/kernelci/lab/lava.py", line 50, in get_device_type_by_name
    if alias["device_type"] == device_type:
KeyError: 'device_type'
```

kernelci/lab/lava.py
```diff
@@ -47,7 +47,7 @@ def get_device_type_by_name(name, device_types, aliases=[]):
     for alias in aliases:
         if alias["name"] == name:
             for device_type in device_types:
-                if alias["device_type"] == device_type:
+                if device_type in alias["device_types"]:
```
gctucker: The `aliases` really have a `device_type` key in each entry, for example:
```json
        "aliases": [
            {
                "device_type": "x15", 
                "name": "am57xx-beagle-x15"
            }, 
            {
                "device_type": "", 
                "name": "db410c-uboot"
            }
        ], 
```
So maybe it's different if there are several aliases for the same type - it's all a bit confusing.

@khilman Any chance you could please share the `aliases` part of the JSON produced by `kci_test get_lab_info`?

khilman: Here's the lab json I get:
```

    "devices": {
        "aliases": [
            {
                "device_types": [
                    "nxp-ls2088"
                ],
                "name": "fsl-ls2088a-rdb"
            },
            {
                "device_types": [
                    "nxp-ls2088"
                ],
                "name": "nxp-ls2088"
            }
        ],
        "device_type_online": {
            "nxp-ls2088": true
        }
    },
    "lab": "lab-nxp",
    "lab_type": "lava",
    "url": "http://59.144.98.45/RPC2/"
}
```

gctucker: Well the latest code still seems to be producing it with a singular `device_type`:
https://git.lavasoftware.org/lava/lava/blob/master/lava_scheduler_app/api/aliases.py#L148
Digging a bit more...

gctucker: @khilman Found the problem, `lab-nxp` are using LAVA 2018.11 which did return `device_types`:
https://git.lavasoftware.org/lava/lava/blob/2018.11/lava_scheduler_app/api/aliases.py#L134

So I think they should upgrade to the latest LAVA release.

gctucker: This is the commit that "fixed" it, so apparently aliases are actually broken in 2018.11:
https://git.lavasoftware.org/lava/lava/commit/62c0e08eccd9ad3959703c966027a6e043422fd4

So I believe LAVA 2019.06 or more recent should have working aliases.

khilman: OK, this can be closed then.

PR #270, opened 2019-11-28
push-source: RootFS.groovy: Use explicitily Python 3

eballetbo:
Python community recommends against making the python name point to
Python 3. The PEP-394 states that:

 "If the python command is installed, it should invoke the same version of
  Python as the python2 command."

Now that we switched to Python 3 the symlink is not created anymore and
the following error appears pushing the source

  python push-source.py --token **** --api
  python: not found

To fix this, call explicitily python3.

### DIFF
--- a/push-source.py
+++ b/push-source.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python3
 #
 # Copyright (C) 2017 Linaro Limited
 # Author: Matt Hart <matthew.hart@linaro.org>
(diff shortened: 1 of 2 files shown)


gctucker:
Thanks, all the other scripts are calling `python3` in the recommended way and the debos jobs were not tested enough with the move to Python 3.  So these are oversights and they look fine, running a buster build job on staging now to verify it.

gctucker:
There seems to be a Python 3 encoding problem:
```
ERROR: failed to publish
'utf-8' codec can't decode byte 0x8b in position 1: invalid start byte
```
Taking a look...

gctucker:
Tested OK on staging with debian-buster:
https://staging-ci.kernelci.org/view/rootfs/job/debian-buster/16/

See also #273 about the issue with utf-8 decoding.

PR #274, opened 2019-11-29
Add kci_rootfs command line tool

mgalka:
Add basic initial implementation of the `kci_rootfs` tool which includes:
- configuration parser
- simple buster debos based rootfs configuration in rootfs-configs.yaml
- `kci_rootfs validate` subcommand that lists all configuration data based on the YAML file
- `kci_rootfs list_configs` subcommand that lists all rootfs configs available in the YAML file

### DIFF
--- a/kci_rootfs
+++ b/kci_rootfs
@@ -45,9 +45,19 @@ class cmd_validate(Command):
             return True
 
 
+class cmd_list_configs(Command):
+    help = "Lists all rootfs config names"
+
+    def __call__(self, config_data, *args, **kwargs):
+        rootfs_configs = config_data['rootfs_configs']
+        for config_name in rootfs_configs:
+            print(config_name)
+        return True
+
 # -----------------------------------------------------------------------------
 # Main
 #
+
 
 if __name__ == '__main__':
     args = parse_args("kci_rootfs", "rootfs-configs.yaml", globals())
--- a/kci_rootfs
+++ b/kci_rootfs
@@ -28,14 +28,21 @@ import kernelci.config.rootfs
 
 
 class cmd_validate(Command):
-    help = "Lists all rootfs configs"
+    help = "Validate the rootfs YAML configuration"
 
     def __call__(self, config_data, *args, **kwargs):
         rootfs_configs = config_data['rootfs_configs']
-        print(rootfs_configs)
         for config_name, config in rootfs_configs.items():
-            print(config.name)
-            print('\t{}'.format(config.rootfs_type))
+            print(config_name)
+            print('\trootfs_type: {}'.format(config.rootfs_type))
+            print('\tarch_list: {}'.format(config.arch_list))
+            print('\tdebian_release: {}'.format(config.debian_release))
+            print('\textra_packages: {}'.format(config.extra_packages))
+            print('\textra_packages_remove: {}'
+                  .format(config.extra_packages_remove))
+            print('\textra_files_remove: {}'.format(config.extra_files_remove))
+            print('\tscript: {}'.format(config.script))
+            return True
 
 
 # -----------------------------------------------------------------------------
--- a/kci_rootfs
+++ b/kci_rootfs
@@ -1,0 +1,49 @@
+#!/usr/bin/env python3
+#
+# Copyright (C) 2019 Collabora Limited
+# Author: Michal Galka <michal.galka@collabora.com>
+#
+# This module is free software; you can redistribute it and/or modify it under
+# the terms of the GNU Lesser General Public License as published by the Free
+# Software Foundation; either version 2.1 of the License, or (at your option)
+# any later version.
+#
+# This library is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public License for more
+# details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this library; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+
+import sys
+
+from kernelci.cli import Args, Command, parse_args
+import kernelci.config.rootfs
+
+# -----------------------------------------------------------------------------
+# Commands
+#
+
+
+class cmd_validate(Command):
+    help = "Lists all rootfs configs"
+
+    def __call__(self, config_data, *args, **kwargs):
+        rootfs_configs = config_data['rootfs_configs']
+        print(rootfs_configs)
+        for config_name, config in rootfs_configs.items():
+            print(config.name)
+            print('\t{}'.format(config.rootfs_type))
+
+
+# -----------------------------------------------------------------------------
+# Main
+#
+
+if __name__ == '__main__':
+    args = parse_args("kci_rootfs", "rootfs-configs.yaml", globals())
+    configs = kernelci.config.rootfs.from_yaml(args.yaml_configs)
+    status = args.func(configs, args)
+    sys.exit(0 if status is True else 1)
--- a/kernelci/config/rootfs.py
+++ b/kernelci/config/rootfs.py
@@ -39,12 +39,51 @@ class RootFS(YAMLObject):
 
 
 class RootFS_Debos(RootFS):
-    def __init__(self, *args, **kwargs):
-        super(RootFS_Debos, self).__init__(*args, **kwargs)
+    def __init__(self, name, rootfs_type, debian_release=None,
+                 arch_list=None, extra_packages=None,
+                 extra_packages_remove=None,
+                 extra_files_remove=None, script=""):
+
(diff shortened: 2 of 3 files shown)


kci_rootfs
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env python
```
gctucker: This should be Python 3.

kci_rootfs
```diff
@@ -0,0 +1,70 @@
...
+        rootfs_configs = config_data['rootfs_configs']
+        for config_name in rootfs_configs.keys():
+            print(config_name)
+
+# -----------------------------------------------------------------------------
+# Main
+#
+
+
+if __name__ == '__main__':
```
gctucker: You can use the default main function:
```
if __name__ == '__main__':
    args = parse_args("kci_build", "rootfs-configs.yaml", globals())
    configs = kernelci.config.rootfs.from_yaml(args.yaml_configs)
    status = args.func(configs, args)
    sys.exit(0 if status is True else 1)
```

kci_rootfs
```diff
@@ -0,0 +1,70 @@
...
+        for config_name, config in rootfs_configs.items():
+            print('\trootfs_type: {}'.format(config.rootfs_type))
+            print('\tarch_list: {}'.format(config.arch_list))
+            print('\tdebian_release: {}'.format(config.debian_release))
+            print('\textra_packages: {}'.format(config.extra_packages))
+            print('\textra_packages_remove: {}'
+                  .format(config.extra_packages_remove))
+            print('\textra_files_remove: {}'.format(config.extra_files_remove))
+            print('\tscript: {}'.format(config.script))
+            print('\tdocker_image: {}'.format(config.docker_image))
```
gctucker: Add:
```python
return True
```
otherwise the status code will be non-zero.

kci_rootfs
```diff
@@ -0,0 +1,70 @@
...
+from kernelci.cli import Args, Command
+import kernelci.config.rootfs
+
+# -----------------------------------------------------------------------------
+# Commands
+#
+
+
+class cmd_validate(Command):
+    help = "Lists all rootfs configs"
```
gctucker: This should be rather something like "Validate the rootfs YAML configuration"

kci_rootfs
```diff
@@ -0,0 +1,70 @@
...
+            print('\tdocker_image: {}'.format(config.docker_image))
+
+
+class cmd_list_configs(Command):
+    help = "Lists all rootfs config names"
+
+    def __call__(self, config_data, *args, **kwargs):
+        rootfs_configs = config_data['rootfs_configs']
+        for config_name in rootfs_configs.keys():
+            print(config_name)
```
gctucker: Same issue here:
```python
return True
```

kci_rootfs
```diff
@@ -0,0 +1,70 @@
...
+# Commands
+#
+
+
+class cmd_validate(Command):
+    help = "Lists all rootfs configs"
+
+    def __call__(self, config_data, *args, **kwargs):
+        rootfs_configs = config_data['rootfs_configs']
+        print(rootfs_configs)
```
gctucker: I think the `validate` command should be more about validating the YAML data than showing things, there can be other commands for that such as `list_configs`.  What's important is to return an error if the YAML doesn't follow the schema, so we can also run it in Travis CI.

kernelci/config/rootfs.py
```diff
@@ -0,0 +1,124 @@
...
+                 docker_image=None):
+
+        super(RootFS_Debos, self).__init__(name, rootfs_type)
+        self._debian_release = debian_release
+        self._arch_list = arch_list or list()
+        self._extra_packages = extra_packages or list()
+        self._extra_packages_remove = extra_packages_remove or list()
+        self._extra_files_remove = extra_files_remove or list()
+        self._script = script
+        self._docker_image = docker_image
```
gctucker: Please remove references to Docker from this config, the rootfs doesn't have anything to do with Docker.  Docker can be used to produce the rootfs but it's not a property of it.

kernelci/config/rootfs.py
```diff
@@ -0,0 +1,124 @@
...
+    def debian_release(self):
+        return self._debian_release
+
+    @property
+    def arch_list(self):
+        return list(self._arch_list)
+
+    @property
+    def extra_packages(self):
+        return ' '.join(list(self._extra_packages))
```
gctucker: Is there any value in converting this list into a string?  Maybe returning a list would be better and the caller can convert it to a string if needed.  The config in YAML is a list, this is a list of packages so I think a list makes more sense here.

kernelci/config/rootfs.py
```diff
@@ -0,0 +1,124 @@
...
+
+    @property
+    def extra_packages_remove(self):
```
gctucker: Same thing as `extra_packages`.

kernelci/config/rootfs.py
```diff
@@ -0,0 +1,124 @@
...
+        return ' '.join(list(self._extra_packages_remove))
+
+    @property
+    def extra_files_remove(self):
```
gctucker: Same thing as `extra_packages`.

kernelci/config/rootfs.py
```diff
@@ -0,0 +1,124 @@
...
+        return rootfs_cls.from_yaml(rootfs, kw)
+
+
+def from_yaml(yaml_path):
+    with open(yaml_path) as f:
+        data = yaml.safe_load(f)
+
+    rootfs_configs = {
+        name: RootFSFactory.from_yaml(name, rootfs)
+        for name, rootfs in data['rootfs_configs'].iteritems()
```
gctucker: `.iteritems()` is a Python 2 thing, we're using Python 3 now.

rootfs-configs.yaml
```diff
@@ -0,0 +1,35 @@
...
+    extra_files_remove:
+      - '*networkd*'
+      - '*resolved*'
+      - tar
+      - patch
+      - dir
+      - partx
+      - find
+    script: ""
+    docker_image: godebos/debos
```
gctucker: As mentioned before, please remove references to Docker from this config.  Also, we're not currently using this image in KernelCI but `kernelci/base-debos`.

mgalka: removed

kernelci/config/rootfs.py
mgalka: Obiously, for some reason I didn't take the Python3 migration into consideration. changed to `.items()`

kernelci/config/rootfs.py
mgalka: fixed

kernelci/config/rootfs.py
mgalka: fixed

kernelci/config/rootfs.py
mgalka: Absolutely, I missed this one while preparing the PR. We should stick to returning non-processed data as much as possible.

kci_rootfs
mgalka: fixed

kci_rootfs
mgalka: fixed

kci_rootfs
mgalka: fixed

kci_rootfs
mgalka: Changed

kci_rootfs
mgalka: Changed

kernelci/config/rootfs.py
```diff
@@ -0,0 +1,118 @@
...
+    def debian_release(self):
+        return self._debian_release
+
+    @property
+    def arch_list(self):
+        return list(self._arch_list)
+
+    @property
+    def extra_packages(self):
+        return self._extra_packages
```
gctucker: OK - except that by returning this list, it makes it possible to change it outside this class.  So maybe do like for `arch_list`:
```python
return list(self._extra_packages)
```

kernelci/config/rootfs.py
gctucker: Same as above, a copy of the list avoids having it changed by code outside this class.

kernelci/config/rootfs.py
gctucker: ditto

gctucker:
@mgalka OK the code looks fine now, thanks.

However, this commit was written by Laks but it's under your name now:
```
commit 9e5c8b97fc4c9646dc4bf9ce4840ef769e79d73a
Author: Michal Galka <michal.galka@collabora.com>
Date:   Fri Nov 29 15:38:16 2019 +0100

    kci_rootfs: Add configuration for buster debos based rootfs
    
    Add configuration for a buster rootfs based on Jenkins jobs.
    Extend configuration parser and "validate" subcommand to work with
    new configuration.
    
```

So I think the author should be reset to Laks, and he should be added to the list of authors in the copyright header.

Also some fixes were put in this commit but they should have been applied to the first commit in this PR `Add skeleton for kci_rootfs` instead.

gctucker:
Thanks, LGTM and tested locally.

PR #277, opened 2019-12-06
bisect.jpl: generate single job.yaml

gctucker:
Since the bisection only runs one job at a time, generate the job
definition in a single file to simplify the code.


gctucker:
Tested OK on staging.

PR #279, opened 2019-12-09
test-configs: fix at91-sama5dx name

montjoie:
There was a typo in both at91-sama5dx, preventing any job to be
generated in lab-baylibre for them.

### DIFF
--- a/test-configs.yaml
+++ b/test-configs.yaml
@@ -363,12 +363,12 @@ device_types:
       - blacklist: *allmodconfig_filter
       - blacklist: {defconfig: ['multi_v7_defconfig+CONFIG_SMP=n']}
 
-  at91-sama5d2-xplained:
+  at91-sama5d2_xplained:
     mach: at91
     class: arm-dtb
     boot_method: uboot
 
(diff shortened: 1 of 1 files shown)


gctucker:
So these boards were never tested before due to this typo?

If so, might be worth adding:
```
Fixes: b98116a396cb ("test-configs.yaml: clean up and consolidate device types")
```

Also please note that the device types used to have 2 underscores e.g. `at91_sama5d2_xplained`, not just one as in this PR which changes it to `at91-sama5d4_xplained`.

montjoie:
The LAVA device type AND the devicetree has thoses - then _
https://git.lavasoftware.org/lava/lava/blob/master/etc/dispatcher-config/device-types/at91-sama5d4_xplained.jinja2

gctucker:
Tested OK on staging: https://staging.kernelci.org/boot/id/5dee7d1ffb87d5a5ef1589f4/

There was no boot results for the `at91-sama5d2_xplained` but I guess it just happened to not be online at the time.

PR #280, opened 2019-12-10
test-configs: fix running of test on qemu-x86_64/i386

montjoie:
The basic x86_64_defconfig does not permit to use any VIRTIO devices.
So all baseline jobs fail to mount the virtual drive with LAVA tests.

Let's use the already present ide interface for qemu-x86_64.
Same problem and same solution for i386.

### DIFF
--- a/test-configs.yaml
+++ b/test-configs.yaml
@@ -924,7 +924,7 @@ device_types:
     context:
       arch: i386
       cpu: 'qemu32'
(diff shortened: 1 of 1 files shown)


gctucker:
Tested OK on staging:
https://staging.kernelci.org/boot/id/5df2c2f742c166025b3b29b0/
https://staging.kernelci.org/boot/id/5df2c215139890de353b299f/

PR #281, opened 2019-12-11
Pine h64 model b

montjoie:


montjoie:
Device-type is in LAVA upstream. the model B is ready in our lab.

test-configs.yaml
```diff
@@ -1040,11 +1040,32 @@ device_types:
...
+    mach: allwinner
+    class: arm64-dtb
+    boot_method: uboot
+
+#Pine H64 exists in two version A and B.
+#Before 5.5, sun50i-h6-pine-h64 was for model B but could be also used for model A.
+#Starting with 5.5, sun50i-h6-pine-h64 will be only for model A and a new
+#dtb will exists for model B.
+#And so a new LAVA device-type will be necessary.
+#But we need to test model B before 5.5 using the pre 5.5 sun50i-h6-pine-h64.dtb WITH the the new model B LAVA device-type.
```
gctucker: Maybe indent the comments to align with the device definition and add a space after the `#` to make it easier to read.

montjoie:
Yes it is online (and in "Good" state) in lab-baylibre.

gctucker:
@montjoie OK the dts hasn't been merged in mainline yet, it's in linux-next though, in `next-20191220`:
```
arch/arm64/boot/dts/allwinner/sun50i-h6-pine-h64-model-b.dts
```
Staging jobs are running on head of mainline master branch by default so that explains why it's not being tested here.

I can run today's staging job based on linux-next to completely confirm this.

gctucker:
Tested OK on staging using next-20191220:
https://staging.kernelci.org/boot/id/5dfcc4259d0264f6d94f03aa/

test-configs.yaml
```diff
@@ -1040,11 +1040,32 @@ device_types:
...
+  sun50i-h6-pine-h64-model-b_5.4:
+    base_name: "sun50i-h6-pine-h64-model-b"
+    dtb: "allwinner/sun50i-h6-pine-h64.dtb"
+    mach: allwinner
+    class: arm64-dtb
+    boot_method: uboot
+    filters:
+      - whitelist: {kernel: ['v4.9', 'v4.14', 'v4.19', 'v5.3', 'v5.4']}
```
gctucker: Actually I think there's a problem with this, as a `whitelist` filter needs the strings to match exactly.  So this won't work with stable branches.  Instead, a `regex` filter should be used to allow kernel version strings such as `v5\.4\..*`.

PR #282, opened 2019-12-12
Add kci_rootfs build option

Lakshmipathi:
Build rootfs images using debos command.
Usage: kci_rootfs build --config buster --debos-path /path/to/debos

Signed-off-by: Lakshmipathi.G <Lakshmipathi.Ganapathi@Collabora.com>

### DIFF
--- a/kernelci/cli.py
+++ b/kernelci/cli.py
@@ -229,6 +229,11 @@ class Args(object):
         'help': "Relative path where build artifacts are published",
     }
 
+    data_path = {
+        'name': '--data-path',
+        'help': "Path to the debos files",
+    }
+
 
 class Command(object):
     """A command helper class.
--- a/kernelci/config/rootfs.py
+++ b/kernelci/config/rootfs.py
@@ -15,6 +15,7 @@
 # along with this library; if not, write to the Free Software Foundation, Inc.,
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
 
+import sys
 import yaml
 
 from kernelci.config import YAMLObject
@@ -94,10 +95,15 @@ class RootFSFactory(YAMLObject):
     @classmethod
     def from_yaml(cls, name, rootfs):
         rootfs_type = rootfs.get('rootfs_type')
-        kw = {
-            'name': name,
-            'rootfs_type': rootfs_type,
-        }
+        if rootfs_type is None:
+            raise TypeError("rootfs_type cannot be Empty")
+        elif rootfs_type not in cls._rootfs_types:
+            raise ValueError("Unsupported value {}".format(rootfs_type))
+        else:
+            kw = {
+                'name': name,
+                'rootfs_type': rootfs_type,
+                }
         rootfs_cls = cls._rootfs_types[rootfs_type] if rootfs_type else RootFS
         return rootfs_cls.from_yaml(rootfs, kw)
 
--- a/kernelci/rootfs.py
+++ b/kernelci/rootfs.py
@@ -1,0 +1,52 @@
+# Copyright (C) 2019 Collabora Limited
+# Author: Lakshmipathi G <lakshmipathi.ganapathi@collabora.com>
+#
+# This module is free software; you can redistribute it and/or modify it under
+# the terms of the GNU Lesser General Public License as published by the Free
+# Software Foundation; either version 2.1 of the License, or (at your option)
+# any later version.
+#
+# This library is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public License for more
+# details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this library; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+
+from kernelci import shell_cmd
+import os
+
+
+def _build_debos(name, config, data_path):
+    for arch_type in config.arch_list:
+        cmd = 'cd {data_path} && debos \
+-t architecture:{arch} \
+-t suite:{release_name} \
+-t basename:{name}/{arch} \
+-t extra_packages:"{extra_packages}" \
+-t extra_packages_remove:"{extra_packages_remove}" \
+-t extra_files_remove:"{extra_files_remove}" \
+-t script:"{script}" \
+rootfs.yaml'.format(
+            name=name,
+            data_path=data_path,
+            arch=arch_type,
+            release_name=config.debian_release,
+            extra_packages=" ".join(config.extra_packages),
+            extra_packages_remove=" ".join(config.extra_packages_remove),
+            extra_files_remove=" ".join(config.extra_files_remove),
+            script=config.script
+            )
+        shell_cmd(cmd)
+
+    return True
+
+
+def build(name, config, data_path):
+    if config.rootfs_type == "debos":
+        return _build_debos(name, config, data_path)
+    else:
+        print("rootfs_type:{} not supported".format(config.rootfs_type))
+        return False
--- a/kci_rootfs
+++ b/kci_rootfs
@@ -20,6 +20,7 @@
 import sys
 
 from kernelci.cli import Args, Command, parse_args
+import kernelci.rootfs
 import kernelci.config.rootfs
 
 # -----------------------------------------------------------------------------
@@ -54,6 +55,21 @@ class cmd_list_configs(Command):
             print(config_name)
         return True
 
+
+class cmd_build(Command):
+    help = "Build rootfs image"
+    args = [Args.config, Args.data_path]
+
+    def __call__(self, config_data, args, **kwargs):
+        data_path = args.data_path
+        config_name = args.config
+        if config_name not in config_data['rootfs_configs']:
+            print("{} invalid. Check 'kci_rootfs list_configs' for valid \
+                    entries".format(config_name))
+            return False
+        config = config_data['rootfs_configs'][args.config]
+        kernelci.rootfs.build(config_name, config, data_path)
+        return True
 # -----------------------------------------------------------------------------
 # Main
 #


kernelci/rootfs.py
```diff
@@ -0,0 +1,41 @@
...
+# along with this library; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+
+from kernelci import shell_cmd
+
+
+def build(name, config, debos_path):
+
+    for arch_type in config.arch_list:
+        cmd = 'cd {debos_path} && \
```
mgalka: I think it'd be better to use Python `os.chdir()` to switch the directory instead of calling `cd` from underlying shell.

Lakshmipathi: Added this change and will update this PR.

kernelci/rootfs.py
```diff
@@ -0,0 +1,41 @@
...
+# details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this library; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+
+from kernelci import shell_cmd
+
+
+def build(name, config, debos_path):
```
gctucker: This function is specific to `debos`.  It should instead look at the rootfs type from the config, and call a `_build_debos()` function or similar if it's `debos` or return an error if it's anything else.  We'll be adding `_build_buildroot()` or something like that once the implementation around `debos` is complete.

Lakshmipathi: okay, added this change and updated the PR

kernelci/rootfs.py
```diff
@@ -0,0 +1,50 @@
...
+# You should have received a copy of the GNU Lesser General Public License
+# along with this library; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+
+from kernelci import shell_cmd
+import os
+
+
+def _build_debos(name, config, debos_path):
+    os.chdir(debos_path)
```
mgalka: I think we could store the cwd somewhere and restore it when the work is done.

kernelci/config/rootfs.py
```diff
@@ -94,6 +95,9 @@ class RootFSFactory(YAMLObject):
     @classmethod
     def from_yaml(cls, name, rootfs):
         rootfs_type = rootfs.get('rootfs_type')
+        if rootfs_type not in cls._rootfs_types:
+            print("Unsupported rootfs_type {}".format(rootfs_type))
+            sys.exit(1)
```
gctucker: If you want to abort the program then throw an exception instead of `print` and `sys.exit(1)`.

Also you can check `if rootfs_type is None` since that would lead to a different type of error (i.e. no rootfs type specificed).

kernelci/rootfs.py
```diff
@@ -0,0 +1,53 @@
...
+            name=name,
+            debos_path=debos_path,
+            arch=arch_type,
+            release_name=config.debian_release,
+            extra_packages=" ".join(config.extra_packages),
+            extra_packages_remove=" ".join(config.extra_packages_remove),
+            extra_files_remove=" ".join(config.extra_files_remove),
+            script=config.script
+            )
+        shell_cmd(cmd, ret_code=True)
```
gctucker: If you pass `ret_code=True` then you should get the result and return it.

kernelci/rootfs.py
```diff
@@ -0,0 +1,53 @@
...
+    os.chdir(cwd)
+    return True
+
+
+def build(name, config, debos_path):
+    if config.rootfs_type == "debos":
+        _build_debos(name, config, debos_path)
```
gctucker: Return the result from `_build_debos()` to propagate any errors.

kernelci/cli.py
```diff
@@ -229,6 +229,11 @@ class Args(object):
         'help': "Relative path where build artifacts are published",
     }
 
+    debos_path = {
+        'name': '--debos-path',
```
gctucker: This option name is a bit misleading as it looks like the path to the `debos` command.  Maybe something like `--data-path` would be better, also it would not be specific to debos.

kernelci/config/rootfs.py
```diff
@@ -93,13 +94,22 @@ class RootFSFactory(YAMLObject):
...
+                raise ValueError("Unsupported value {}".format(rootfs_type))
+            else:
+                kw = {
+                    'name': name,
+                    'rootfs_type': rootfs_type,
+                }
+                rootfs_cls = cls._rootfs_types[rootfs_type] if rootfs_type \
+                    else RootFS
+                return rootfs_cls.from_yaml(rootfs, kw)
+        except (TypeError, ValueError) as e:
```
gctucker: I don't see any reason why trying to catch exceptions if all we're doing is raising another one to abort the command.  I think we should just not have a try-except block here, this looks like diverting exception handling into flow control.

kernelci/rootfs.py
```diff
@@ -0,0 +1,56 @@
...
+from kernelci import shell_cmd
+import os
+
+
+def _build_debos(name, config, debos_path):
+    cwd = os.getcwd()
+    os.chdir(debos_path)
+    for arch_type in config.arch_list:
+        cmd = 'debos \
+-t architecture:{arch} -t suite:{release_name} -t basename:{name}/{arch} \
```
gctucker: Maybe keep one option per line, to make it easier to read and maintain.

kernelci/rootfs.py
```diff
@@ -0,0 +1,56 @@
...
+            arch=arch_type,
+            release_name=config.debian_release,
+            extra_packages=" ".join(config.extra_packages),
+            extra_packages_remove=" ".join(config.extra_packages_remove),
+            extra_files_remove=" ".join(config.extra_files_remove),
+            script=config.script
+            )
+        ret_code = shell_cmd(cmd, True)
+        if not ret_code:
+            return False
```
gctucker: Does this leave the current directory as `debos_path` when the command fails?

kernelci/rootfs.py
```diff
@@ -0,0 +1,56 @@
...
+    return True
+
+
+def build(name, config, debos_path):
+    if config.rootfs_type == "debos":
+        return _build_debos(name, config, debos_path)
+    else:
+        print("rootfs_type:{} not supported".format(config.rootfs_type))
+        return False
+    return True
```
gctucker: This line will never be reached.

kernelci/cli.py
Lakshmipathi: okay changed it.

kernelci/rootfs.py
Lakshmipathi: no, it doesn't leave current directory as `debos_path`

kernelci/config/rootfs.py
Lakshmipathi: okay removed try-except block - just used `raise`

kernelci/rootfs.py
```diff
@@ -0,0 +1,57 @@
...
+            extra_packages=" ".join(config.extra_packages),
+            extra_packages_remove=" ".join(config.extra_packages_remove),
+            extra_files_remove=" ".join(config.extra_files_remove),
+            script=config.script
+            )
+        ret_code = shell_cmd(cmd, True)
+        if not ret_code:
+            return False
+
+    os.chdir(cwd)
```
gctucker: If the current directory isn't changed, then why `chdir(cwd)` in this branch and not when the debos command fails?

Lakshmipathi: If we plan to support multiple rootfs build option in future (ex: `--config=build_all`), then we need change directory to original one after successfully creating first rootfs.

gctucker: I'm not asking about why we need to change the directory, I know we need to do that because the file paths in the YAML config are all relative.

My problem with this code is that it changes back to the original directory only when there was no error.  Why behave differently when there is an error?

Lakshmipathi: I assumed that  when there is an error, the kci_rootfs shouldn't proceed further and abort. If we are planning to proceed further with `next` config (ex: `--config=first_failed,next`), will add `chdir()` for that case too.

gctucker: I think we should just use "cd" in the shell command.

kernelci/rootfs.py
```diff
@@ -0,0 +1,54 @@
...
+            name=name,
+            data_path=data_path,
+            arch=arch_type,
+            release_name=config.debian_release,
+            extra_packages=" ".join(config.extra_packages),
+            extra_packages_remove=" ".join(config.extra_packages_remove),
+            extra_files_remove=" ".join(config.extra_files_remove),
+            script=config.script
+            )
+        ret_code = shell_cmd(cmd, True)
```
gctucker: Actually, if the command fails with `shell_cmd(cmd)`, then an exception is raised and the program is aborted.  So the manual error handling here isn't really necessary.

kernelci/rootfs.py
```diff
@@ -0,0 +1,52 @@
...
+            name=name,
+            data_path=data_path,
+            arch=arch_type,
+            release_name=config.debian_release,
+            extra_packages=" ".join(config.extra_packages),
+            extra_packages_remove=" ".join(config.extra_packages_remove),
+            extra_files_remove=" ".join(config.extra_files_remove),
+            script=config.script
+            )
+        shell_cmd(cmd, True)
```
gctucker: No I meant without `ret_code=True`, so `shell_cmd(cmd)`.  If you pass `ret_code=True` then it returns the status code.  If you don't then it raises an exception in case the command failed.

Lakshmipathi: okay dropped `ret_code=True` here.

gctucker:
OK thanks this looks good now.  @mgalka Can you please take over this review?  Trying to build a rootfs locally would seem like a good test until we can do this in Jenkins.

Lakshmipathi:
kci_rootfs test run: https://staging-ci.kernelci.org/view/My%20jobs/job/laks/job/kci_rootfs_test/124/

gctucker:
Yes but we need to try it on the command line first in any case, the Jenkins integration will come in a follow-up PR.

Lakshmipathi:
Just now realized that debos  command missing `test_overlay`  and newly introduced `crush_image_options` - will update the patch.

PR #283, opened 2019-12-13
Make install and module path configurable

jeromebrunet:
Hi,

I have been trying to use kernel-ci core with (yet) another CI tool: concourse
Being able to configure the install path make thing easier with it .. so here it my modest contribution

### DIFF
--- a/kci_build
+++ b/kci_build
@@ -267,7 +267,7 @@ class cmd_install_kernel(Command):
     args = [Args.kdir]
     opt_args = [Args.config, Args.tree_name, Args.tree_url, Args.branch,
                 Args.commit, Args.describe, Args.describe_verbose,
-                Args.output, Args.publish_path]
+                Args.output, Args.publish_path, Args.install_path]
 
     def __call__(self, configs, args):
         if args.config:
@@ -289,21 +289,23 @@ or
             return False
         return kernelci.build.install_kernel(
             args.kdir, tree_name, tree_url, branch, args.commit, args.describe,
-            args.describe_verbose, args.output, args.publish_path)
-
+            args.describe_verbose, args.output, args.publish_path,
+            args.install_path)
 
 class cmd_push_kernel(Command):
     help = "Push the kernel binaries"
     args = [Args.kdir, Args.api, Args.token]
+    opt_args = [Args.install_path]
 
     def __call__(self, configs, args):
-        return kernelci.build.push_kernel(args.kdir, args.api, args.token)
+        return kernelci.build.push_kernel(args.kdir, args.api, args.token,
+                                          args.install_path)
 
 
 class cmd_publish_kernel(Command):
     help = "Publish the kernel meta-data"
     args = [Args.kdir]
-    opt_args = [Args.api, Args.token, Args.json_path]
+    opt_args = [Args.api, Args.token, Args.json_path, Args.install_path]
 
     def __call__(self, configs, args):
         if not ((args.api and args.token) or args.json_path):
@@ -314,8 +316,7 @@ Invalid arguments, please provide at least one of these sets of options:
 """)
             return False
         return kernelci.build.publish_kernel(
-            args.kdir, api=args.api, token=args.token,
-            json_path=args.json_path)
+            args.kdir, args.install_path, args.api, args.token, args.json_path)
 
 
 class cmd_pull_tarball(Command):
--- a/kci_build
+++ b/kci_build
@@ -253,13 +253,14 @@ class cmd_list_kernel_configs(Command):
 class cmd_build_kernel(Command):
     help = "Build a kernel"
     args = [Args.build_env, Args.arch, Args.kdir]
-    opt_args = [Args.defconfig, Args.j, Args.verbose, Args.output]
+    opt_args = [Args.defconfig, Args.j, Args.verbose, Args.output,
+                Args.mod_path]
 
     def __call__(self, configs, args):
         build_env = configs['build_environments'][args.build_env]
         return kernelci.build.build_kernel(
             build_env, args.kdir, args.arch, args.defconfig, args.j,
-            args.verbose, args.output)
+            args.verbose, args.output, args.mod_path)
 
 
 class cmd_install_kernel(Command):
@@ -267,7 +268,8 @@ class cmd_install_kernel(Command):
     args = [Args.kdir]
     opt_args = [Args.config, Args.tree_name, Args.tree_url, Args.branch,
                 Args.commit, Args.describe, Args.describe_verbose,
-                Args.output, Args.publish_path, Args.install_path]
+                Args.output, Args.publish_path, Args.install_path,
+                Args.mod_path]
 
     def __call__(self, configs, args):
         if args.config:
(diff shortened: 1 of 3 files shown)


jeromebrunet:
Duhh ... apparently Travis does not like it.
I don't really get why but if you can help me understand what's wrong, I'd be happy to correct it

khilman:
pycodestyle found a few lines that were too long.

kernelci/cli.py
```diff
@@ -189,6 +189,11 @@ class Args(object):
         'help': "Path the output directory",
     }
 
+    install = {
+        'name': '--install',
+        'help': "Path the install directory",
```
khilman: nit: Path *of* the install directory

kernelci/cli.py
```diff
@@ -194,6 +194,11 @@ class Args(object):
         'help': "Path the install directory",
     }
 
+    mod_path = {
+        'name': '--mod-path',
+        'help': "Path the installed modules",
```
khilman: nit: Path *of* the installed modules

khilman:
Tested this in my gitlab-ci test job, and it's working great.  Thanks for this!

jeromebrunet:
I've seen it now - The link was easy to spot :)
I will fix the problems

gctucker:
> Tested this in my gitlab-ci test job, and it's working great. Thanks for this!

@khilman BTW, is there a public place where to find your Gitlab CI setup yet?

khilman:
@gctucker : it's a moving target, but here's the YAML: https://gitlab.com/khilman/linux-amlogic/blob/gitlab-ci/.gitlab-ci.yml

Right now, it's only building arm + arm64 and doing baseline tests on a single SoC family (amlogic)

Some pipelines that have run: https://gitlab.com/khilman/linux-amlogic/pipelines

gctucker:
Sorry @jeromebrunet was not on the list of trusted users so this PR wasn't tested on staging on Monday.  Doing another run now.

kernelci/build.py
```diff
@@ -633,11 +633,12 @@ def build_kernel(build_env, kdir, arch, defconfig=None, jopt=None,
             result = _run_make(target='dtbs', **kwargs)
     build_time = time.time() - start_time
 
-    if result and mods and mod_path:
-        abs_mod_path = os.path.join(output_path, mod_path)
-        if os.path.exists(abs_mod_path):
-            shutil.rmtree(abs_mod_path)
-        os.makedirs(abs_mod_path)
+    if result and mods:
+        if not mod_path:
```
gctucker: It's better to do this at the top of the function where the other arguments are assigned default values such as `output_path`.

kernelci/build.py
```diff
@@ -784,13 +785,13 @@ def install_kernel(kdir, tree_name, tree_url, git_branch, git_commit=None,
...
 
     modules_tarball = None
-    if mod_path:
-        abs_mod_path = os.path.join(output_path, mod_path)
-        if os.path.exists(abs_mod_path):
-            modules_tarball = 'modules.tar.xz'
-            modules_tarball_path = os.path.join(install_path, modules_tarball)
-            shell_cmd("tar -C{path} -cJf {tarball} .".format(
-                path=abs_mod_path, tarball=modules_tarball_path))
+    if not mod_path:
```
gctucker: Same comment, default values for the arguments should be set at the top of the function.

kernelci/build.py
```diff
@@ -832,12 +833,13 @@ def push_kernel(kdir, api, token, install='_install_'):
     *kdir* is the path to the kernel source directory
     *api* is the URL of the KernelCI backend API
     *token* is the token to use with the KernelCI backend API
-    *install* is the path to the installation directory inside kdir
+    *install* is the path to the installation directory
```
gctucker: This should be `install_path`.

kernelci/cli.py
```diff
@@ -189,6 +189,16 @@ class Args(object):
         'help': "Path the output directory",
     }
 
+    install = {
```
gctucker: Maybe call it `install_path` / `--install-path` to be consistent with `mod_path` and `json_path`, and also be clear it's a path.

kernelci/build.py
```diff
@@ -832,12 +833,13 @@ def push_kernel(kdir, api, token, install='_install_'):
...
 
     The returned value is True if it was done successfully or False if an error
     occurred.
     """
-    install_path = os.path.join(kdir, install)
+    if not install_path:
+        install_path = os.path.join(kdir, '_install_')
```
gctucker: So the default value is actually still inside `kdir`.  I think it's fine, but maybe the docstring should be clearer about the default behaviour.  Maybe something like this:
```
*install_path* is the path to the installation directory, or _install_ inside kdir by default
```

kernelci/build.py
gctucker: Also the docstring should be updated to explain the default value of `mod_path`, see below for a similar case with `install_path`.

kernelci/build.py
jeromebrunet: Yes, the default behavior should not have changed with these patches. Indeed, I did not do a great job at documenting this. I'll improve this

kernelci/cli.py
```diff
@@ -189,6 +189,20 @@ class Args(object):
         'help': "Path the output directory",
     }
 
+    install_path = {
+        'name': '--install-path',
+        'required': False,
```
gctucker: This works, but it's not really necessary to add here as arguments put in `opt_args` in the command class have required set to `False` automatically.

kci_build
```diff
@@ -289,21 +291,23 @@ or
...
             args.kdir, tree_name, tree_url, branch, args.commit, args.describe,
-            args.describe_verbose, args.output, args.publish_path)
+            args.describe_verbose, args.output, args.publish_path,
+            args.install_path, args.mod_path)
 
 
 class cmd_push_kernel(Command):
     help = "Push the kernel binaries"
-    args = [Args.kdir, Args.api, Args.token]
+    args = [Args.kdir, Args.api, Args.token, Args.install_path]
```
gctucker: Why not do the same as with other commands and use it as an optional argument?

```python
opt_args = [Args.install_path]
```

jeromebrunet: Thanks for pointing this out.
I'm still mostly hacking my way through ... I did not get that difference before

jeromebrunet: And sorry for noise, I'm pushing things to my branch for concourse to pick it up ... I forgot this it updates the PR each time

gctucker: No problem :)

kernelci/build.py
gctucker: OK I see you've explained the default path in the command line argument help, which is more useful for the user so that's fine.

The default path is however actually implemented in this function, and that's not explained in the docstring.  Maybe the default path should be all handled by `kci_build` and there should be no default value in this function.  Something to improve in another PR I guess, it's not a big deal.

gctucker:
@jeromebrunet By the way, what is "alien CI" as I see it's in the name of your branch?

jeromebrunet:
> @jeromebrunet By the way, what is "alien CI" as I see it's in the name of your branch?

Alien or "unfamiliar" ... IOW, not jenkins :wink: 
I'm testing concourse CI right now but the change are not really concourse related.

I'm trying to come up with something that monitor patchwork, apply and test kernel with the submitted patches. I'm using kernelci-core to build the kernel.

It used to work well but, ATM, I have a problem with `install` and `git describe`

gctucker:
Nice, I've never seen it used in production before.  Found the website: https://concourse-ci.org/

Feel free to ask any questions about the problems you've found on the `#kernelci` IRC channel or on the mailing list kernelci@groups.io.

PR #284, opened 2019-12-13
docker: build-base: add python3-pip

khilman:
Add python3-pip to the base images.  This enables build jobs to
install extra python dependencies without having to install pip first.

### DIFF
--- a/jenkins/dockerfiles/build-base/Dockerfile
+++ b/jenkins/dockerfiles/build-base/Dockerfile
@@ -38,6 +38,7 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
     python3.7 \
     python3-jinja2 \
     python3-keyring \
+    python3-pip \
     python3-pyelftools \
     python3-requests \
     python3-yaml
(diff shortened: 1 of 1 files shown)


gctucker:
If a Python dependency is needed, why not add it directly to the Docker image instead?

khilman:
Because we don't know what all the dependencies are. 

If we want people to start (re)using kernelci-core and our docker images (e.g. for gitlab-ci jobs, or other CI systems) we should make them more flexible.

Also, I just added a 2nd patch which adds kernelci-core to the base image as well.  In current expeirments (with gitlab-ci) this is the main need for python3-pip

gctucker:
I think adding kernelci-core should be a separate PR as it's quite a different topic.  Also it would be better to first add proper package setup support in kernelci-core using setuptools to install a `kernelci` Python package and the `kci_*` utilities in the standard path.  We also need to be sure we don't break the staging workflow which always needs the latest version from Github.

About python3-pip, well I guess that's because you apparently need it in your Gitlab CI setup, but every other alternative CI setup will most likely have their own different needs.  So I'm not entirely sure if it's really wise to add things in the upstream Dockerfile that are only used downstream.  Is it not flexible enough to patch the Dockerfile with an additional layer for your specific use-case, or use `FROM kernelci/build-base` in another Dockerfile?

khilman:
OK, dropping this for now until kernelci-core can be officially packaged for python.

PR #285, opened 2019-12-18
templates/sleep: increase rtcwake delay to 5s

gctucker:
Increase the rtcwake delay from 1s to 5s in the sleep test plan to
give enough time for the system to fully suspend before triggering
resume.

Also use multi-line YAML string to make it easier to read the shell
commands.

### DIFF
--- a/templates/sleep/sleep.jinja2
+++ b/templates/sleep/sleep.jinja2
@@ -13,8 +13,16 @@
           - functional
         run:
           steps:
-            - for i in $(seq 1 10); do lava-test-case rtcwake-mem-$i --shell rtcwake -d rtc0 -m mem -s 1; done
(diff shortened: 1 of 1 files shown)


gctucker:
Tested OK on staging, here's the row from the email report:
```
run | platform             | arch  | lab           | compiler | defconfig          | results
----+----------------------+-------+---------------+----------+--------------------+--------8   | rk3399-gru-kevin     | arm64 | lab-collabora | gcc-8    | defconfig          | 11/11
```

However, the "freeze" tests were not reported as failures,probably because rtcwake never returned.  That's something to improve with the handling of test results.

I wonder if we can first report all the test cases as "failed", and then if they complete report them again as "pass" so by the end of the job all the test cases will have a result.  That's a LAVA thing to investigate.

PR #287, opened 2019-12-21
build-configs: Add i386 builds to Lee's android-3.18 job

lag-linaro:


gctucker:
Fair enough.

We should however start a separate YAML file for Android builds at some point, maybe in a separate repo.  We probably should have YAML files in separate repositories anyway now that the schema has become stable enough.

PR #290, opened 2019-12-27
debian: add new buster-crostests rootfs

eballetbo:
This new rootfs contains the python packages required to run the
Chromebooks Test Suite written in python and the test suite itself. This
is a replacement of the current cros-ec tests.

### DIFF
--- a/jenkins/debian/debos/scripts/buster-cros-ec-tests.sh
+++ b/jenkins/debian/debos/scripts/buster-cros-ec-tests.sh
@@ -1,0 +1,44 @@
+#!/bin/bash
+set -x
+
+# Important: This script is run under QEMU
+
+set -e
+
+# Build-depends needed to build the test suites, they'll be removed later
+BUILD_DEPS="\
+    git \
+    python3-setuptools \
+    python3-pip \
+"
+
+apt-get install --no-install-recommends -y ${BUILD_DEPS}
+
+########################################################################
+# Build tests                                                          #
+########################################################################
+
+BUILDFILE=/test_suites.json
+echo '{  "tests_suites": [' >> $BUILDFILE
+
+########################################################################
+# Build and install tests                                              #
+########################################################################
+
+CROS_URL="https://git.kernel.org/pub/scm/linux/kernel/git/chrome-platform/cros-ec-tests.git"
+CROS_SHA=$(git ls-remote ${CROS_URL} | head -n 1 | cut -f 1)
+
+pip3 install git+${CROS_URL}@${CROS_SHA}
+
+echo '    {"name": "cros-ec-tests", "git_url": "'$CROS_URL'", "git_commit": "'$CROS_SHA'" }' >> $BUILDFILE
+echo '  ]}' >> $BUILDFILE
+
+########################################################################
+# Cleanup: remove files and packages we don't want in the images       #
+########################################################################
+
+rm -fr /src/cros-ec-tests
+apt-get remove --purge -y ${BUILD_DEPS} perl-modules-5.28
+apt-get autoremove --purge -y
+apt-get clean
+
--- a/jenkins/buster-cros-ec.jpl
+++ b/jenkins/buster-cros-ec.jpl
@@ -1,0 +1,24 @@
+@Library('kernelci') _
+import org.kernelci.debian.RootFS
+
+/* ----------------------------------------------------------------------------
+ * Jenkins parameters
+
+DOCKER_BASE
+  Dockerhub base address used for the build images
+
+*/
+
+def r = new RootFS()
+
+def config = ['name':"buster-cros-ec",
+              'arch_list':["armhf", "arm64", "amd64"],
+              'debian_release':"buster",
+              'extra_packages':"python3-minimal python3-unittest2",
+              'extra_packages_remove':"bash e2fsprogs e2fslibs libext2fs2",
+              'script':"scripts/buster-cros-ec-tests.sh",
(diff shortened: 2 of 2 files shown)


jenkins/debian/debos/scripts/buster-crostests.sh
```diff
@@ -0,0 +1,30 @@
...
+# Build and install tests                                              #
+########################################################################
+
+pip3 install git+https://gitlab.collabora.com/chromiumos/crostests.git
+
+########################################################################
+# Cleanup: remove files and packages we don't want in the images       #
+########################################################################
+
+rm -fr /src/crosstests
```
gctucker: Typo: `crostests` (one 's').

jenkins/debian/debos/scripts/buster-crostests.sh
```diff
@@ -0,0 +1,30 @@
...
+    python3-pip \
+"
+
+apt-get install --no-install-recommends -y ${BUILD_DEPS}
+
+########################################################################
+# Build and install tests                                              #
+########################################################################
+
+pip3 install git+https://gitlab.collabora.com/chromiumos/crostests.git
```
gctucker: Maybe this should have been added to the existing `https://gitlab.collabora.com/chromium` group?

gctucker:
This seems fine in principle as long as the scope for the tests is well defined.  All the tests added in this PR are using the cros-ec driver so I think it would still be better to keep the `cros-ec` test plan name, to be more specific about this type of hardware.  It would help clarify the fact that it's not about testing any actual Chrome OS features, or not any random hardware that happen to be used on Chromebooks, but really just the cros-ec (embedded controller) kernel driver and its related hardware.

For example, Chromebook hardware that is not controlled by the cros-ec such as the GPU should be covered with i-g-t, or video devices with the v4l2 test plan.  Also, Chrome OS features that are purely user-space are outside the scope of kernelci.org.  This test suite is obviously not trying to cover those, but it should be made clearer from the name and description to avoid any ambiguity and diversion in the future.

eballetbo:
@gtucker, new version uploaded.

I renamed all to cros-ec and tried to be more explicit by making clean that the test suite is for the Chrome OS Embedded Controller and its drivers. Note also that the official test suite is now in kernel.org/chrome-platform

Waiting for your feedback.

jenkins/debian/debos/scripts/buster-crostests.sh
gctucker: OK it's on kernel.org now, even better :)

gctucker:
Thanks, LGTM.  I'll make a Jenkins job in the main folder to give it a try on staging.

gctucker:
Watch this space: https://staging-ci.kernelci.org/view/rootfs/job/debian-buster-cros-ec/1/

gctucker:
I've started a new build since the previous rootfs image was not copied when redeploying staging.kernelci.org onto the new server.

@eballetbo Once this completes, please run a LAVA job with it to check it works as expected, at least for one CPU arch.

gctucker:
New rootfs build completed: https://staging-storage.kernelci.org/images/rootfs/debian/buster-cros-ec/20200121.1/

eballetbo:
Tested on 3 devices:

* https://lava.collabora.co.uk/results/2193710
* https://lava.collabora.co.uk/results/2193711
* https://lava.collabora.co.uk/results/2193712

gctucker:
OK thanks, that's good.

One thing that's missing is the manifest file, with the git revision of the test suite being used.  If you look at other rootfs such as buster-igt, you'll see that they create a `test_suites.json` file:
```
BUILDFILE=/test_suites.json
echo '{  "tests_suites": [' >> $BUILDFILE
```
https://github.com/kernelci/kernelci-core/blob/master/jenkins/debian/debos/scripts/buster-igt.sh#L41

It's not absolutely required but it helps a lot when investigating test case failures because it allows to trace the version of the test suite for each result.  This information also gets added to test email reports.

eballetbo:
Pushed with the requested change.

jenkins/debian/debos/scripts/buster-cros-ec-tests.sh
```diff
@@ -0,0 +1,37 @@
...
+"
+
+apt-get install --no-install-recommends -y ${BUILD_DEPS}
+
+########################################################################
+# Build tests                                                          #
+########################################################################
+
+BUILDFILE=/test_suites.json
+echo '{  "tests_suites": [' >> $BUILDFILE
```
gctucker: Err, I meant this is only to start creating the file but then the git revision of cros-ec-tests should be added there as well, like in buster-igt.sh.

gctucker:
Thanks, build tested OK on staging:
https://staging-storage.kernelci.org/images/rootfs/debian/buster-cros-ec/20200123.3/

gctucker:
The manifest part seems to have worked fine:
https://staging-storage.kernelci.org/images/rootfs/debian/buster-cros-ec/20200123.3/amd64/build_info.json
```json
{
    "date": "Thu Jan 23 10:54:24 UTC 2020",
    "distro": "debian",
    "distro_version": "10.2",
    "tests_suites": [
        {
            "name": "cros-ec-tests",
            "git_url": "https://git.kernel.org/pub/scm/linux/kernel/git/chrome-platform/cros-ec-tests.git",
            "git_commit": "6813e889c71c5c8c14ccacd6d11add8ac0bd99db"
        }
    ]
}
```

PR #291, opened 2019-12-27
bisect.jpl: remove kernel git clone if it appears broken

gctucker:
If for any reason the kernel git clone appears to be broken, remove it
and create a new one.  This is pretty rare but it can happen as it's
kept from one job to the next to avoid recreating one every time.


gctucker:
Tested OK on staging.

PR #294, opened 2020-01-03
T18197 disable sending bisection email reports to automatic recipients

gctucker:
Add `--no-auto-recipients` option to `push-bisection-results.py` to disable sending bisection email reports to the automatically generated list of recipients.

Use it in `bisect.jpl` to not send email reports initially as the new test case bisection is being rolled out.  If the reports appear to be fine with real use-cases then this can be reverted to enable them again.


gctucker:
Tested OK on staging:
https://staging-ci.kernelci.org/job/lava-bisect/8230/console

PR #295, opened 2020-01-10
kci_rootfs: Include debos test_overlay, crush_image_options options

Lakshmipathi:
Handle debos test_overlay and crush_image_options options


gctucker:
OK this is not causing and bad side-effects on staging so merging now.

PR #298, opened 2020-01-13
Add kci_rootfs upload option

Lakshmipathi:
Upload rootfs images to kernelci backend.
Usage: ./kci_rootfs upload --rootfs-dir=/path/to/rootfsdir
       --upload-path=path/to/uploaddir --api=ip-address --token <token>

### DIFF
--- a/kernelci/rootfs.py
+++ b/kernelci/rootfs.py
@@ -16,6 +16,7 @@
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
 
 from kernelci import shell_cmd
+from kernelci.storage import upload_files
 import os
 
 
@@ -50,3 +51,20 @@ def build(name, config, data_path):
     else:
         print("rootfs_type:{} not supported".format(config.rootfs_type))
         return False
+
+
+def upload(api, token, upload_path, input_dir):
+    """Upload rootfs to KernelCI backend.
+
+    *api* is the URL of the KernelCI backend API
+    *token* is the backend API token to use
+    *upload_path* is the target on KernelCI backend
+    *input_dir* is the local rootfs directory path to upload
+    """
+    artifacts = {}
+    for root, _, files in os.walk(input_dir):
+        for f in files:
+            px = os.path.relpath(root, input_dir)
+            artifacts[os.path.join(px, f)] = open(os.path.join(root, f), "rb")
+    upload_files(api, token, upload_path, artifacts)
+    return True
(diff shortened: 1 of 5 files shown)


kernelci/rootfs.py
```diff
@@ -50,3 +52,37 @@ def build(name, config, data_path):
     else:
         print("rootfs_type:{} not supported".format(config.rootfs_type))
         return False
+
+
+def _upload_files(api, token, path, input_files):
```
mgalka: There is a very similar function defined in `kernelci/build.py`. We should avoid this kind of a duplication. I think it's a good idea to create a `kernelci/storage.py` module and put the functions responsible for calling the /upload API and import it in both `rootfs.py` and `build.py`.

Lakshmipathi: okay added required changes to `rootfs.py` and `build.py`

kernelci/rootfs.py
```diff
@@ -16,7 +16,10 @@
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
 
 from kernelci import shell_cmd
+from kernelci.storage import upload_files
 import os
+import requests
+from urllib.parse import urljoin
```
mgalka: These imports seem to be unused

mgalka:
Tested locally, works fine.

kernelci/storage.py
```diff
@@ -0,0 +1,50 @@
...
+        'file{}'.format(i): (name, fobj)
+        for i, (name, fobj) in enumerate(input_files.items())
+    }
+    url = urljoin(api, 'upload')
+    resp = requests.post(url, headers=headers, data=data, files=files)
+    resp.raise_for_status()
+    return True
+
+
+def upload_files(api, token, upload_path, input_files):
```
mgalka: This function does nothing apart from calling the non-public API (_upload_files()). We can get rid of this function and rename `_upload_files` -> `upload_files`.
The docstring should be also applied to the renamed function.

kernelci/storage.py
```diff
@@ -0,0 +1,50 @@
...
+    """Upload rootfs to KernelCI backend.
+
+    *api* is the URL of the KernelCI backend API
+    *token* is the backend API token to use
+    *upload_path* is the target on KernelCI backend
+    *input_files* dictionary of input files
+    """
+    _upload_files(api, token, upload_path, input_files)
+    return True
```
mgalka: This function always returns `True` and doesn't really care about the result.

Lakshmipathi: `raise_for_status()` call should handle the negative case and should report any exception.

gctucker: Then, there's no need to return anything.  That's what the original `_upload_files()` did.

gctucker:
OK starting a Buster job on staging today as a one last check before merging.

gctucker:
Sorry, forgot this isn't integrated with Jenkins yet on the master branch.  So yes it seems all ready to go then.

PR #299, opened 2020-01-16
Drop done files

gctucker:
Drop `$ARCH.done` files in `kernel-arch-complete.sh` to avoid having to set up a Jenkins runner on the storage server.  Also this isn't needed any more since the build trigger job knows when all the builds are complete.

### DIFF
--- a/jenkins/build-trigger.jpl
+++ b/jenkins/build-trigger.jpl
@@ -260,10 +260,9 @@ def buildKernelStep(job, arch, defconfig, build_env, opts, labs, kci_core) {
     }
 }
 
-def buildsComplete(job, opts, arch) {
+def buildsComplete(job, opts) {
     def str_params = [
         'TREE_NAME': opts['tree'],
-        'ARCH': arch,
         'GIT_DESCRIBE': opts['describe'],
         'BRANCH': opts['branch'],
     ]
@@ -388,11 +387,7 @@ get_lab_info \
         }
 
         stage("Complete") {
-            /* ToDo: convert kernel-arch-complete as a stage in this job */
-            def arch_list = listArchitectures(kci_core, params.BUILD_CONFIG)
-            for (String arch: arch_list) {
-                buildsComplete("kernel-arch-complete", opts, arch)
-            }
+            buildsComplete("kernel-arch-complete", opts)
         }
     }
 }
--- a/jenkins/kernel-arch-complete.sh
+++ b/jenkins/kernel-arch-complete.sh
@@ -22,11 +22,6 @@ if [[ -z $GIT_DESCRIBE ]]; then
   exit 1
 fi
 
-if [[ -z $ARCH ]]; then
-  echo "ARCH not set.  Not publishing."
-  exit 1
-fi
-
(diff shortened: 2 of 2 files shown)


gctucker:
Tested OK on staging.

@khilman I think this might break your non-LAVA lab workflow.  We can do something about it if that's the case, so you get notified about completed builds somehow.  Other labs that don't have an API to submit tests will probably be in a similar situation.

khilman:
Yes, my lab will definitely stop testing builds that are missing $ARCH.done files.

I'm slowly moving labs over to my LAVA worker, but  I still have lots of boards in the non-LAVA side, so it would be really nice to have another way of detecting when builds are done.

gctucker:
How about querying the backend API for new builds?

It would be good to have a subscription mechanism too, but with the current API I think it's already possible to query the available builds and check when they were added to determine the new ones you want to test.

gctucker:
Also maybe `kci_build` could have some extra commands to find out about new builds from the backend API.

khilman:
@gctucker do you have any ideas (or better, examples) of how to use the current API to query for new builds?

khilman:
I've tried to call the `/trigger/boot` API using the same token I'm currently using to upload, and I"m getting `403 Operation not permitted: provided token is not authorized`

mgalka:
> @gctucker do you have any ideas (or better, examples) of how to use the current API to query for new builds?

@khilman There is no API call that is able to serve builds that has been recently done. Having said that it's possible to use `time_range` and `date_range` query parameters to get builds from a specific timespan. You can also use `job` and `branch` query parameters to get builds only for specific job and branch.

e.g.

Get builds for branch `my-local-branch` and job `test-job`
```
curl -H "Authorization: YOUR_BACKEND_TOKEN" http://api.kernelci.org/build?job=test-job\&branch=my-local-branch
```

Get builds for branch `my-local-branch` and job `test-job` from the last 30 minutes
```
curl -H "Authorization: YOUR_BACKEND_TOKEN" http://api.kernelci.org/build?job=test-job\&branch=my-local-branch\&time_range=30
```

Get builds for branch `my-local-branch` and job `test-job` from the last 2 days
```
curl -H "Authorization: YOUR_BACKEND_TOKEN" http://api.kernelci.org/build?job=test-job\&branch=my-local-branch\&date_range=2
```

Get builds for branch `my-local-branch` and job `test-job` from starting 2 days before 2020-01-01
```
curl -H "Authorization: YOUR_BACKEND_TOKEN" http://api.kernelci.org/build?job=test-job\&branch=my-local-branch\&date_range=2\&created_on=20200101
```

A few things about `time_range` and  `date_range`

`time_range` - Time range granularity is 1 minute. Minimum value is 10 and maximum is 1440 (24 hours). 

`date_range` - Date range granularity is 1 day. The number of days is subtracted from the current date unless `created_on` parameter is specified.

gctucker:
OK thanks, merging now.

I've created this issue about adding required features to support "generic" test labs:
https://github.com/kernelci/kernelci-core/issues/315

PR #303, opened 2020-01-24
Switch to use the new cros-ec test plan

eballetbo:
We have now a cros-ec test plan for all Chromebooks devices. So remove the
dummy and obsolete cros-ec test and use the new cros test plan. The
cros-ec test plan is using the `cros-ec-tests` python package composed of
a number of modules that will help you to do stand alone tests or setup a
fully automated tests.

### DIFF
--- a/templates/cros-ec/cros-ec.jinja2
+++ b/templates/cros-ec/cros-ec.jinja2
@@ -5,15 +5,15 @@
     - repository:
         metadata:
           format: Lava-Test Test Definition 1.0
-          name: cros-ec
-          description: "check /dev/cros_ec"
+          name: {{ plan }}
+          description: "cros-ec test plan"
           os:
           - oe
           scope:
           - functional
         run:
           steps:
-            - lava-test-case cros-ec-presence --shell test -c /dev/cros_ec
+            - python3 -m cros.runners.lava_runner
       from: inline
-      name: cros-ec
-      path: inline/cros-ec.yaml
+      name: {{ plan }}
+      path: inline/{{ plan }}.yaml
--- a/test-configs.yaml
+++ b/test-configs.yaml
@@ -43,6 +43,10 @@ file_systems:
     type: debian
     ramdisk: 'buster/20200127.0/{arch}/rootfs.cpio.gz'
 
+  debian_buster-cros-ec_ramdisk:
+    type: debian
+    ramdisk: 'buster-cros-ec/20200130.0/{arch}/rootfs.cpio.gz'
+
   debian_buster-igt_ramdisk:
     type: debian
     ramdisk: 'buster-igt/20200127.0/{arch}/rootfs.cpio.gz'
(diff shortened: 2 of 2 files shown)


test-configs.yaml
```diff
@@ -162,7 +162,7 @@ test_plans:
       - blacklist: *clang_environment_filter
 
   cros-ec:
-    rootfs: debian_buster_ramdisk
+    rootfs: debian_buster-cros-ec_ramdisk
```
gctucker: Please also add a rootfs entry with `debian_buster-cros-ec_ramdisk`, with the URL based on the last staging build i.e. 20200123.3.

eballetbo: Pushed a new version using that one from staging

templates/cros-ec/cros-ec.jinja2
```diff
@@ -5,15 +5,15 @@
     - repository:
         metadata:
           format: Lava-Test Test Definition 1.0
-          name: cros-ec
-          description: "check /dev/cros_ec"
+          name: {{ plan }}
+          description: "cros-ec test plan"
           os:
-          - oe
```
gctucker: We'll see if this works, but `oe` is actually also used with stripped-down versions of Debian such as this rootfs.  The igt tests are running with a small Debian rootfs and the LAVA job definition uses `oe`.  I guess you've used `debian` in your LAVA jobs so this should probably be fine.

gctucker:
We now have a first rootfs available in production:
https://storage.kernelci.org/images/rootfs/debian/buster-cros-ec/20200127.0/

So we can run this PR once with the staging binary but if it's all fine then we directly start using the production URL.

gctucker:
@eballetbo I've run a LAVA job with the new rootfs and it seems to be missing some Python modules:
https://lava.collabora.co.uk/scheduler/job/2205153#L1386

gctucker:
I'll make a temporary patch for staging.kernelci.org to not use the production version, so we can get this to work on staging first.

gctucker:
It still failed with the new rootfs built with the latest version of the test suite:
https://lava.collabora.co.uk/scheduler/job/2207746#L1389

gctucker:
@eballetbo It worked with the latest version of the test suite from today, thanks for fixing it:
https://lava.collabora.co.uk/scheduler/job/2209241

So please use this URL for the buster-cros-ec rootfs in `test-configs.yaml` in this PR:
https://storage.kernelci.org/images/rootfs/debian/buster-cros-ec/20200130.0/arm64/rootfs.cpio.gz

test-configs.yaml
gctucker: There's a rootfs available in production now, see my other comment.

templates/cros-ec/cros-ec.jinja2
gctucker: Actually this is fine, the `deploy` stage still has the OS set to `oe` which is what I think matters before running tests.

eballetbo:
Pushed a new version using the latest rootfs in test-configs.yaml and replacing 'debian' with 'oe' for os.

gctucker:
Tested OK on staging: https://lava.collabora.co.uk/results/2215038/0_cros-ec

PR #304, opened 2020-01-24
Delete the .gitreview file

mattface:
This is VERY legacy from ye olde Linaro days, might as well just remove it.


gctucker:
There was so much dust on top of it I never noticed that file.

PR #310, opened 2020-01-30
test-configs.yaml: add sun8i-r40-bananapi-m2-ultra

montjoie:
This patch adds sun8i-r40-bananapi-m2-ultra which is ready in my lab.
The LAVA device-type is upstream for a long time.


gctucker:
Tested OK on staging: https://staging.kernelci.org/boot/id/5e32df750eb0c22897230fa1/

PR #311, opened 2020-01-30
WIP: templates: baseline: dump loaded drivers list

montjoie:
This patch introduce a way to dump the list of all loaded drivers (along
with the compatible which matched if any).
The list is dumped in a yaml format which can be used directly by reading
the jobs logs.

The final goal is to have:
- a way to list all boot with a specific driver
- found regression when a driver is not loaded anymore

A patch on the backend will be necessary to use it.

### DIFF
--- a/templates/baseline/baseline.jinja2
+++ b/templates/baseline/baseline.jinja2
@@ -57,6 +57,45 @@
             - lava-test-shell
         run:
           steps:
+            - echo "==DRIVER_DUMP==" > driverlist
+            - echo "==VERSION=0==" >> driverlist
+            - echo "drivers:" >> driverlist
+            - >
+               for line in $(find /sys/devices/ -name uevent |xargs grep -l ^DRIVER); do
+                 echo "  - driver: $(grep ^DRIVER= "$line" |cut -d= -f2)" >> driverlist
+                 if [ ! -z $(grep ^OF_COMPATIBLE_N= "$line") ];then
+                   echo "    compatibles:" >> driverlist
+                   for compatible in $(grep ^OF_COMPATIBLE_[0-9]*= "$line" | cut -d= -f2); do
+                     echo "      - $compatible" >> driverlist
+                   done
+                 fi
+               done
+            - echo "==DRIVER_DUMP_END==" >> driverlist
+            - cat driverlist
+      from: inline
+      name: drivers
+      path: inline/drivers.yaml
+
+- test:
+{%- if test_namespace %}
+    namespace: {{ test_namespace }}
+{%- endif %}
+    timeout:
+      minutes: 1
+    definitions:
+    - repository:
+        metadata:
+          format: Lava-Test Test Definition 1.0
+          name: baseline
+          description: "baseline test plan"
+          os:
+            - debian
+          scope:
+            - functional
+          environment:
+            - lava-test-shell
+        run:
+          steps:
             - export PATH=/opt/bootrr/helpers:$PATH
             - cd /opt/bootrr && sh helpers/bootrr-auto
       lava-signal: kmsg


montjoie:
See http://kernel.montjoie.ovh/lava/local/20200131T08:38:37-42733.log.html#1339 for an arm boot
or http://kernel.montjoie.ovh/lava/local/20200201T19:58:05-42947.log.html#828 for an x86_64 qemu boot.

montjoie:
I need to rework a bit for working around driver loaded twice (uart for example)

gctucker:
Note that I'm not a big fan of having test-specific changes in the backend though, but there should be a way to support this kind of use-case in a generic way.  It's similar to benchmarks and performance tracking, where the pass/fail criteria depends on previous results.

gctucker:
This seems to be taking way too long to run, it's even causing some jobs to hit the timeout:
https://lava.collabora.co.uk/scheduler/job/2221999

montjoie:
Yes the default timeout of 1minute is too low.
Furthermore, the set+x of LAVA made this very verbose.
Anyway the new version cannot be kept in "LAVA inline".
You can find the new version (2) in https://github.com/montjoie/lava-tests/blob/master/boot.sh
The result could be found at http://kernel.montjoie.ovh/lava/local/drivers/dump-45289-2

gctucker:
No, the 1min timeout is way long enough.  The baseline test plan should only take a few seconds to run, because it's running on all the devices for all the builds and it is only intended to do very quick checks.

But since what you are proposing doesn't do much processing other than walking though sysfs, we can assume that it should be possible to do it fast enough.

Then, can you please explain how you intend to use the output in a KernelCI context?  How would you generate test cases with pass/fail results?

montjoie:
There are no pass/fail. This just permit to store in the backend which driver are probed and most of all binded .
The goal is to have in the UI a way to search for boot by selecting the hardware which was tested.
Example: I want all boot which have tested my cryptoengine.

Furthermore, this could be used as a metric of the % of kernel source which is tested by kernelCI.

We could have some pass/fail but this need to compare two run for detecting regression. Why driver x is not binded anymore ? But this will be more complicated since we need to detect a base boot on which we want to reference.

gctucker:
OK, I see.

If we want to add more hardware information in the backend, I think we would need to have a new collection of documents for each device type with this kind of information.  Hardware doesn't change between kernel revisions, so except in the case where we're tracking drivers that fail to probe, why keep running this in CI?

gctucker:
@montjoie @khilman Please see my latest comment with unanswered questions.

Is there still something you want to do about this PR?

montjoie:
The goal is:
- tracking driver which fail to probe
- permit someone to search all devices which run a specific driver

Example I work on driver x, i want to be able to check is works everywhere it is probed/loaded
Extra bonus to be able to launch jobs on thoses devices according to this information.
Permiting this job filter is really usefull for me.

gctucker:
OK, well that is what `bootrr` is for and it's already being run as part of the `baseline` test plan.  Maybe `bootrr` can be improved if you think there are a few more things it could do?

gctucker:
In the absence of more updates, I think we should close this PR.  There are some good ideas though such as being able to find devices that use a particular driver.  But this should be done differently, it needs to be device type attributes rather than test results.

Something similar to this is being done for `v4l2-compliance` tests, see how it's configured in `test-configs.yaml`.  Only devices that are known to use the `uvcvideo` driver are running the test suite, or QEMU with `vivid` enabled in the defconfig.  So it's not done dynamically, but it's a start.

We could add device type attributes in `test-configs.yaml` to do this in a semi-dynamic way: manually set the attributes, but then let some jobs get generated dynamically for all devices with a particular attribute.  For example, a test_config entry could have a filter based on attributes rather than a fixed device_type value.

If this is something worth investigating, please start a thread on the mailing list or create a GitHub issue.  I don't know of any actual use-cases for it right now so it seems like a lower-priority improvement.

PR #313, opened 2020-02-01
kci_test: fix dtb_dir

khilman:
The URL for DTB files is missing 'dtbs' for boot jobs coming from
LAVA.  This is because the platform.dtb field in the LAVA metadata is
missing the 'dtb_dir' prefix from the build metadata (bmeta).

Fix this by checking if dtb_dir exists in the build metatdata.  If so
prepend it to the DTB relative path.

Link: https://github.com/kernelci/kernelci-backend/issues/131


gctucker:
Tested OK on staging.

PR #314, opened 2020-02-03
kci_rootfs: Include --arch option

Lakshmipathi:
By default, rootfs will be created for all architecture listed under
rootfs-configs.yaml. With this option, if required, rootfs image for given
architecture can be created.

Signed-off-by: Lakshmipathi Ganapathi <lakshmipathi.ganapathi@collabora.co.uk>

### DIFF
--- a/kci_rootfs
+++ b/kci_rootfs
@@ -61,17 +61,18 @@ class cmd_list_configs(Command):
 
 class cmd_build(Command):
     help = "Build rootfs image"
-    args = [Args.config, Args.data_path]
+    args = [Args.config, Args.data_path, Args.arch]
 
     def __call__(self, config_data, args, **kwargs):
         data_path = args.data_path
         config_name = args.config
+        arch = args.arch
         if config_name not in config_data['rootfs_configs']:
             print("{} invalid. Check 'kci_rootfs list_configs' for valid \
                     entries".format(config_name))
             return False
         config = config_data['rootfs_configs'][args.config]
-        kernelci.rootfs.build(config_name, config, data_path)
+        kernelci.rootfs.build(config_name, config, data_path, arch)
         return True
 
 
--- a/kernelci/rootfs.py
+++ b/kernelci/rootfs.py
@@ -20,9 +20,8 @@ from kernelci.storage import upload_files
 import os
 
 
-def _build_debos(name, config, data_path):
-    for arch_type in config.arch_list:
-        cmd = 'cd {data_path} && debos \
+def _build_debos(name, config, data_path, arch):
+    cmd = 'cd {data_path} && debos \
 -t architecture:{arch} \
 -t suite:{release_name} \
 -t basename:{name}/{arch} \
@@ -35,7 +34,7 @@ def _build_debos(name, config, data_path):
 rootfs.yaml'.format(
             name=name,
             data_path=data_path,
-            arch=arch_type,
+            arch=arch,
             release_name=config.debian_release,
             extra_packages=" ".join(config.extra_packages),
             extra_packages_remove=" ".join(config.extra_packages_remove),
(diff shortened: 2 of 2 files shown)


Lakshmipathi:
Tested locally and also on staging.

gctucker:
This is failing to merge with other branches on staging at the moment, so skipping it for now.

kernelci/rootfs.py
```diff
@@ -19,8 +19,14 @@
...
-def _build_debos(name, config, data_path):
-    for arch_type in config.arch_list:
+def _build_debos(name, config, data_path, arch_type):
+    arch_list = []
+    if arch_type is None:
+        arch_list = config.arch_list
+    else:
+        arch_list.append(arch_type)
+
+    for arch in arch_list:
```
gctucker: The `_build_debos()` function should be building only one rootfs, so it should just take the `arch` parameter and pass it to `debos`.

gctucker:
Please rebase, and also rename to `--arch` as that's what is used in other places.

kernelci/rootfs.py
```diff
@@ -44,14 +43,29 @@ def _build_debos(name, config, data_path):
...
             crush_image_options=" ".join(config.crush_image_options)
             )
-        shell_cmd(cmd)
+    shell_cmd(cmd)
 
     return True
 
 
-def build(name, config, data_path):
+def build(name, config, data_path, arch=None):
```
gctucker: I think it's better to just require `arch` as an argument, so each call to `build()` only produces one rootfs.  It simplifies the code and makes the behaviour more deterministic.

gctucker:
@Lakshmipathi Also please update the commit subject with `--arch`.

mgalka:
Works fine with kernelci/debos docker image.

kernelci/rootfs.py
```diff
@@ -44,17 +43,24 @@ def _build_debos(name, config, data_path):
...
+    *data_path* points to debos location
+    *arch* required architecture
+    """
+
     if config.rootfs_type == "debos":
-        return _build_debos(name, config, data_path)
+        _build_debos(name, config, data_path, arch)
     else:
         print("rootfs_type:{} not supported".format(config.rootfs_type))
-        return False
```
gctucker: If not returning a status any more then an exception should be raised instead.

kci_rootfs
```diff
@@ -62,16 +62,22 @@ class cmd_list_configs(Command):
...
         data_path = args.data_path
         config_name = args.config
+        arch = args.arch
         if config_name not in config_data['rootfs_configs']:
             print("{} invalid. Check 'kci_rootfs list_configs' for valid \
                     entries".format(config_name))
             return False
         config = config_data['rootfs_configs'][args.config]
-        kernelci.rootfs.build(config_name, config, data_path)
+        if arch is not None:
```
gctucker: I still think that `--arch` should be a required argument to run `kci_rootfs build --config=config-name --arch=arch-name` and always build only one rootfs per command line call.

Lakshmipathi: If we make `--arch` mandatory argument, do we still need to keep  `arch_list` inside `rootfs-configs.yaml`?

gctucker: Yes, these are different things.  The `rootfs-configs.yaml` file contains the list of architectures that should be built for each rootfs type.  Then, each architecture name can be passed in a separate call to `kci_rootfs build`.  I guess what is missing is the list of rootfs and arch combinations, and that's what `kci_rootfs list_variants` would be about.

Lakshmipathi: updated the patch. Will send a new PR for `list_variants` option.

kernelci/rootfs.py
```diff
@@ -44,17 +43,25 @@ def _build_debos(name, config, data_path):
             test_overlay=config.test_overlay,
             crush_image_options=" ".join(config.crush_image_options)
             )
-        shell_cmd(cmd)
+    shell_cmd(cmd)
 
     return True
```
gctucker: Note that this return value is not used anywhere.  But this is fine as any error would cause an exception before reaching this line anyway.

PR #316, opened 2020-02-04
kernelci: adding --always flag to git describe

Heidifahim:
Currently, the KernelCI command line tool page assumes that the kernel
directory is initialized through the update_repo command
(https://github.com/kernelci/kernelci-doc/wiki/KernelCI-command-line) in
which case git describe would return a tag. However, when a user sets
the kdir to their own local checkout possibly on a branch that doesn't
have default tags, a "fatal: no tags can describe ..." error occurs.
This is resolved by adding the --always flag to the git
describe commands.

Signed-off-by: Heidi Fahim <heidifahim@google.com>

### DIFF
--- a/kernelci/build.py
+++ b/kernelci/build.py
@@ -237,7 +237,7 @@ def git_describe(tree_name, path):
     cmd = """
 set -e
 cd {path}
-git describe {describe_args}
+git describe --always {describe_args}
 """.format(path=path, describe_args=describe_args)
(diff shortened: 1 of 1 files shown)


gctucker:
OK, and when there are no tags it just returns a short sha1:
```
$ git describe
fatal: No names found, cannot describe anything.
$ git describe --always
4255da54fa48
```

LGTM, checking that it has no side effects on staging.kernelci.org.

PR #318, opened 2020-02-11
kernel-arch-complete.sh: report baseline and sleep tests for "chrome-platform" tree

eballetbo:
Add email recipients for the baseline and sleep tests in
"chrome-platform" tree, so we receive reports for such test plans.


gctucker:
LGTM - tested that it had no bad side effects on staging.

PR #322, opened 2020-02-12
kci_build: Add option to delete the tarball after downloading

mattface:
Slightly lower the required disk space for kernel builds, by removing
the tarball after it has been extracted.

Change-Id: I7d0f51906446e9a81786d2e92be7b263c12700c1

### DIFF
--- a/kernelci/build.py
+++ b/kernelci/build.py
@@ -349,7 +349,7 @@ def _download_file(url, dest_filename, chunk_size=1024):
         return False
 
 
-def pull_tarball(kdir, url, dest_filename, retries):
+def pull_tarball(kdir, url, dest_filename, retries, delete):
     if os.path.exists(kdir):
         shutil.rmtree(kdir)
     os.makedirs(kdir)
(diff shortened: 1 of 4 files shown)


kernelci/build.py
```diff
@@ -362,6 +362,9 @@ def pull_tarball(kdir, url, dest_filename, retries):
         return False
     with tarfile.open(dest_filename, 'r:*') as tarball:
         tarball.extractall(kdir)
+    if delete:
+        if os.path.isfile(dest_filename):
```
gctucker: I would be surprised to find a case where the tarball wouldn't exist given that it was created just before in the same function.  If that did ever happen then letting it throw an exception would probably actually be a bit better.

gctucker:
Tested OK on staging.

It would have been a bit better to separate the change in kci_build and the change in the Jenkins job, but that's not a blocker.  Also I'll squash the clean-up fix as it's not useful to have it in a separate commit.

PR #324, opened 2020-02-19
Add kci_rootfs list_variants command

Lakshmipathi:
Display rootfs configurations along with their supported architectures.

### DIFF
--- a/kci_rootfs
+++ b/kci_rootfs
@@ -59,6 +59,33 @@ class cmd_list_configs(Command):
         return True
 
 
+class cmd_list_variants(Command):
+    help = "Lists all rootfs variants"
+    opt_args = [Args.config, Args.arch]
+
+    def __call__(self, config_data, args, **kwargs):
+        rootfs_configs = config_data['rootfs_configs']
+        config_name = args.config
+        arch = args.arch
+
+        if config_name and config_name not in rootfs_configs:
+            print("{} : invalid config entry".format(config_name))
+            return False
+
+        if arch and config_name and \
+           arch not in rootfs_configs[config_name].arch_list:
+            print("{} : invalid arch entry".format(arch))
+            return False
+
+        configs = (config for config in rootfs_configs
+                   if config == config_name or config_name is None)
+        for config in configs:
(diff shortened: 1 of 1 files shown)


kci_rootfs
```diff
@@ -59,6 +59,43 @@ class cmd_list_configs(Command):
         return True
 
 
+class cmd_list_variants(Command):
+    help = "Lists all rootfs variants"
+    opt_args = [Args.config, Args.arch]
+
+    def __call__(self, config_data, args, **kwargs):
```
mgalka: Can we come up with something more concise. Maybe something like:

```python
        if config_name and config_name not in rootfs_configs:
            print("{} : invalid config entry".format(config_name))
            return False
        if arch and config_name and arch not in rootfs_configs[config_name].arch_list:
            print("{} : invalid arch entry".format(arch))
            return False

        configs = (config for config in rootfs_configs if config == config_name or config_name is None)
        for config in configs:
            for arch_type in rootfs_configs[config].arch_list:
                if arch_type == arch or arch is None:
                    print(config, arch_type)
        return True
```

Lakshmipathi: ok, modified the command and implemented as stated above and tested.

kci_rootfs
```diff
@@ -59,6 +59,33 @@ class cmd_list_configs(Command):
...
+           arch not in rootfs_configs[config_name].arch_list:
+            print("{} : invalid arch entry".format(arch))
+            return False
+
+        configs = (config for config in rootfs_configs
+                   if config == config_name or config_name is None)
+        for config in configs:
+            for arch_type in rootfs_configs[config].arch_list:
+                if arch_type == arch or arch is None:
+                    print(config, arch_type)
```
gctucker: I think this would be a bit safer:
```python
print(' '.join([config, arch_type]))
```
to ensure the output has a whitespace between the config and arch names.  The `print(config, arch_type)` syntax works with the current Python 3 convention, it wouldn't work with Python 2 and who knows if this will change again in the future.

Lakshmipathi: ok, made above changes.

PR #335, opened 2020-03-05
Qemu 3.1

montjoie:
We have upgraded our lab to LAVA 2020.02 and so upgraded to buster.
Buster came with a more recent qemu which breaks some test jobs.

### DIFF
--- a/test-configs.yaml
+++ b/test-configs.yaml
@@ -920,6 +920,7 @@ device_types:
     filters:
       - whitelist: {defconfig: ['vexpress_defconfig']}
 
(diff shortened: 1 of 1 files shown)


test-configs.yaml
```diff
@@ -1773,7 +1778,6 @@ test_configs:
 
   - device_type: qemu_arm-versatilepb
     test_plans:
-      - baseline_qemu
```
gctucker: Why remove baseline?  The boot tests will soon disappear and this platform won't be tested at all any more.

montjoie: I restore baseline. I have sent a defconfig change that should be merged soon.

Issue #331, opened 2020-02-28
kselftest


gctucker:
[kselftest](https://www.kernel.org/doc/html/latest/dev-tools/kselftest.html) is one of the main test suites that comes with the Linux kernel itself.  As such, it is an obvious one to cover on kernelci.org.  Doing so poses a few challenges, as it needs changes to the build system to build the tests from the kernel tree for each revision being tested and then install them on the target platform to run them.

- [x] #339 basic plumbing to get kselftest builds
- [x] #487 initial runtime coverage in LAVA labs
- [x] #499 first pass at getting as many tests to build
- [x] #636 fine-tuning of the build environment to separate kselftest errors from main kernel build errors
- [x] #655 first pass at fixing runtime dependencies

kees:
Can the kselftest tarball be included in the build artifacts?

`make -C tools/testing/selftests gen_tar FORMAT=.xz`

kees:
Oh, I see https://github.com/kernelci/kernelci-core/pull/339 already exists. :)

gctucker:
Yes, it's not enabled in production yet though but should get there soon as the PR you found is functional.

alexandrasp:
After check and try to improve kselftest in kernel-ci build we decided to split missing selftests errors in little tasks because they are more complicated or are not related to just installing some package.
 
The new summary is:

- [ ] Fix problems related to reader files not found in kselftest builds
- [ ] Fix bpftool error with missing vmlinux
- [ ] Fix bpf kselftest build issue due to missing numa.h
- [ ] Build with success: filesystems/binderfs, exec, kvm and pidfd selftests.

alexandrasp:
And a brief update, after changes #531, #528 and #525
The list of extra directories that we have started to build with gcc are:

On arm64:

Skipping non-existent dir: exec
+Emit Tests for filesystems/binderfs
Skipping non-existent dir: kexec
+Emit Tests for kvm
+Emit Tests for pidfd
Skipping non-existent dir: tc-testing
Skipping non-existent dir: vm

On x86_64:

Skipping non-existent dir: exec
+Emit Tests for filesystems/binderfs
Skipping non-existent dir: kexec
+Emit Tests for kvm
+Emit Tests for pidfd
Skipping non-existent dir: tc-testing
+Emit Tests for vm

On arm:

Skipping non-existent dir: exec
+Emit Tests for filesystems/binderfs
Skipping non-existent dir: kexec
Skipping non-existent dir: kvm
[...]

PR #339, opened 2020-03-12
build.py: build/install kselftests

khilman:
Add conditional step to build_kernel that builds the in-kernel
selftests.  Build is conditioned on "kselftest" being in
defconfig_extras.

If kselftests are build, install_kernel will create a
kselftests.tar.xz archive that will be published in the publish_step,
and added to bmeta.json (key: kselftests)

### DIFF
--- a/kernelci/build.py
+++ b/kernelci/build.py
@@ -467,7 +467,8 @@ def _run_make(kdir, arch, target=None, jopt=None, silent=True, cc='gcc',
         args.append('CC={}'.format(cc))
 
     if output:
-        args.append('O={}'.format(os.path.relpath(output, kdir)))
+        # due to kselftest Makefile issues, O= cannot be a relative path
+        args.append('O={}'.format(os.path.abspath(output)))
 
     if target:
         args.append(target)
@@ -640,6 +641,28 @@ def build_kernel(build_env, kdir, arch, defconfig=None, jopt=None,
         })
         result = _run_make(target='modules_install', **kwargs)
 
+    # kselftest
+    if result and "kselftest" in defconfig_extras:
+        kselftest_install_path = os.path.join(output_path, '_kselftest_')
+        if os.path.exists(kselftest_install_path):
+            shutil.rmtree(kselftest_install_path)
+        opts.update({
+            'INSTALL_PATH': kselftest_install_path,
+        })
+        #
+        # Ideally this should just be a 'make kselftest-install', but
+        # due to bugs with O= in kselftest Makefile, this has to be
+        # 'make -C tools/testing/selftests install'
+        #
+        kwargs.update({
+            'kdir': os.path.join(kdir, 'tools/testing/selftests')
+        })
+        opts.update({
+            'FORMAT': '.gz',
+        })
+        # 'gen_tar' target does 'make install' and creates tarball
+        result = _run_make(target='gen_tar', **kwargs)
+
     cc_version_cmd = "{}{} --version 2>&1".format(
         cross_compile if cross_compile and cc == 'gcc' else '', cc)
     cc_version_full = shell_cmd(cc_version_cmd).splitlines()[0]
@@ -787,6 +810,17 @@ def install_kernel(kdir, tree_name, tree_url, git_branch, git_commit=None,
         shell_cmd("tar -C{path} -cJf {tarball} .".format(
             path=mod_path, tarball=modules_tarball_path))
 
+    # 'make gen_tar' creates this tarball path
+    kselftest_tarball = 'kselftest-packages/kselftest.tar.gz'
+    kselftest_tarball_path = os.path.join(output_path, '_kselftest_',
+                                          kselftest_tarball)
+    if os.path.exists(kselftest_tarball_path):
+        kselftest_tarball = os.path.basename(kselftest_tarball_path)
+        shutil.copy(kselftest_tarball_path,
+                    os.path.join(install_path, kselftest_tarball))
+    else:
+        kselftest_tarball = kselftest_tarball_path = None
+
     build_env = bmeta['build_environment']
     defconfig_full = bmeta['defconfig_full']
     defconfig_dir = defconfig_full.replace('/', '-')
@@ -810,6 +844,7 @@ def install_kernel(kdir, tree_name, tree_url, git_branch, git_commit=None,
         'git_describe_v': describe_v,
         'git_commit': git_commit,
         'file_server_resource': publish_path,
+        'kselftests': kselftest_tarball,
     })
 
     with open(os.path.join(install_path, 'bmeta.json'), 'w') as json_file:


khilman:
hmm, I just realized that we don't build the kselftest fragment on staging, so this build will not be tested.

gctucker:
Right, I've just merged the other PR to add kselftest fragments on staging.  So this should become active in the next staging run.

gctucker:
OK got some builds, here's one:
https://storage.staging.kernelci.org/kernelci/staging.kernelci.org/staging-20200326.2/arm64/defconfig+kselftest/gcc-8/?C=M&O=D

I haven't checked whether it is as expected, @khilman can you please take a look?

khilman:
The kselftest-install step is failing due to
```
/bin/sh: 1: rsync: not found
```
Which is now addressed by https://github.com/kernelci/kernelci-core/pull/345

khilman:
@gctucker this one should be ready now, no?

gctucker:
It would be a lot better to keep the kselftest steps in separate functions, the `build_kernel()` and `install_kernel()` functions are already doing a lot and they should really be only dealing with kernel builds.  Ideally we should have a way to build kselftest with a command line argument.  The test for 'kselftest in defconfig_extras' is quite fragile, this logic should be determined at a higher level.

Then I'm not sure of the impact of converting the output path from relative to absolute, there might be some unwanted side effects.

gctucker:
From a functional point of view, it seems to be doing the right thing.  Here's a sample build:
https://storage.staging.kernelci.org/kernelci/staging.kernelci.org/staging-20200520.2/arm64/defconfig+kselftest/gcc-8/

which includes a kselftest tarball:
https://storage.staging.kernelci.org/kernelci/staging.kernelci.org/staging-20200520.2/arm64/defconfig+kselftest/gcc-8/kselftest.tar.xz

Taking a quick look, this seems fine:
```
$ ls
android      ftrace        kselftest  lkdtm           pstore            sysctl  zram
cpufreq      futex         lib        memory-hotplug  run_kselftest.sh  tpm2
cpu-hotplug  intel_pstate  livepatch  net             static_keys       user
```

and there doesn't seem to be any side-effects of this PR on other "regular" builds.

It would be nice to at least try to run a test job with the contents of this archive to completely verify it was built correctly, maybe with a manually modified kernelci rootfs to skip the actual LAVA integration part.

Then with a bit of code clean-up this should be ready to go.  We should probably build a very small set of kernels with kselftest until we can actually run tests, just enough to facilitate things a bit.

kernelci/build.py
```diff
@@ -633,6 +634,24 @@ def build_kernel(build_env, kdir, arch, defconfig=None, jopt=None,
...
+        kselftest_install_path = os.path.join(output_path, '_kselftest_')
+        if os.path.exists(kselftest_install_path):
+            shutil.rmtree(kselftest_install_path)
+        opts.update({
+            'INSTALL_PATH': kselftest_install_path,
+        })
+        #
+        # Ideally this should just be a 'make kselftest-install', but
+        # due to bugs with O= in kselftest Makefile, this has to be
+        # 'make -C tools/testing/selftests install'
```
gctucker, line 648: Any chance these bugs can be fixed?  Is there any email thread about this topic to follow what is going?

gctucker:
Let's focus on the k8s stuff merged first and then work on resolving this one.  I think it's 90% good, might just need rebasing and a couple of minor tweaks.

kernelci/build.py
khilman: There were some similar discussions, but nothing on this specifially.  I'll start an email thread on the kselftest list.

However, even if this gets fixed upstream, it's unlikely this would be backported, so I think we should stick with the workaround in this patch for awhile, especially if we expect to run kselftest on the stable kernels.

khilman:
I wasn't aware of the `gen_tar` make target in tools/testing/selftest until seeing @kees comment in #331, I've reworked this PR slightly to use that instead of manually creating the tarball in `kernelci/build.py`.

Also rebased onto current master.

khilman:
Removed a stray debug print.
Rebased to current master.

kernelci/build.py
kees: FWIW, I *think* this was fixed in v5.7 (see merge 397a97946798890b9bdaa6122fcfad7147690670), but perhaps I'm not understanding the specific bug as it related to running under kernelci's builder.

kernelci/build.py
```diff
@@ -787,6 +807,17 @@ def install_kernel(kdir, tree_name, tree_url, git_branch, git_commit=None,
         shell_cmd("tar -C{path} -cJf {tarball} .".format(
             path=mod_path, tarball=modules_tarball_path))
 
+    # 'make gen_tar' creates this tarball path
+    kselftest_tarball = 'kselftest-packages/kselftest.tar.gz'
```
kees, line 811: Commit says ".xz", code expects ".gz", but depends on the default behavior of "gen_tar" without a FORMAT= specified. Perhaps explicitly declare "FORMAT=xz" (or =gz) (and update the filename and commit message appropriately)?

khilman: Right, I'll make it more explicit and clean up the changelog.  Thanks for the review.

FWIW, I started with .xz for size reasons, but we're recycling a test-plan from another test project (Linaro LKFT) and their test wrappers etc. have a bunch of hard-coded assumptions for .gz, so I switched to that (but forgot to update commit message.)

kernelci/build.py
gctucker: @khilman Can you please add a link to the discussion in the kselftest mailing list archives?

khilman: There are several related issues around this:
- https://lore.kernel.org/lkml/158351957799.3363.15269768530697526765.stgit@devnote2/
- https://lore.kernel.org/linux-kselftest/MWHPR13MB0895B92C9B4807D94E1E6B04FDE30@MWHPR13MB0895.namprd13.prod.outlook.com/
- https://lore.kernel.org/bpf/cover.1583358715.git.skhan@linuxfoundation.org/

As well as some interesting tools to help sanity check:
- https://lore.kernel.org/linux-kselftest/86508e9d-4aff-d9a0-3783-971a4686ccd1@bird.org/

But again, even though most (maybe all?) of these issues are fixed upstream, it's unlikely these all be backported to stable, so the workaround in this series is still necessary if we want to build kselftest on stable.

khilman:
Updated PR: dropped patch which updated rootfs image since this is now updated in master branch.

gctucker (approved):
Overall, I think this is OK to get merged and rolled out in production.  A couple of things that should be kept in mind though:

* Adding test binaries to a kernel build is not a great precedent, as it blurs the definition of what a kernel build is for KernelCI.  Of course, kselftest is a bit special since it's tied to the kernel source.  Still, adding other things in the future like documentation or any other artifacts that can be built within a kernel tree is probably going to cause us some problems.  So we should try to deal with that by having a separate place for "kernel extras" that aren't actually the kernel, but built from the kernel tree.

* Building everything in the `build_kernel` function is not great, that's not new but it's not going in the right direction with `kselftest` being added there too.  As a result, we can't build a kernel with kselftest config enabled without also building the kselftest tarball.  That will need some slight redesign all the way to the `kci_build` command line interface, so it's probably best to tackle that once the settings file support is merged and when all the data-related operations go via `kernelci.data` (i.e. `publish_kernel`, `push_kernel` etc...).

gctucker:
Tested OK on staging, combined with #445: https://staging.kernelci.org/test/plan/id/5f45aff802c5b617e7b1458c/

PR #342, opened 2020-03-20
adding imx6q-sabresd and imx7d-sdb to kernelci.org

suram-nxp:
adding imx6q-sabresd and imx7d-sdb to kernelci.org

### DIFF
--- a/test-configs.yaml
+++ b/test-configs.yaml
@@ -600,6 +600,16 @@ device_types:
             - 'multi_v7_defconfig+kselftest'
           kernel: ['v3.18']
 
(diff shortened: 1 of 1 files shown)


khilman:
Tested OK.

I used a local copy of kernelci-core with this patch to build mainline and submit jobs to the NXP lab with kci_test.  LAVA jobs generated/submitted OK.
https://lavalab.nxp.com/scheduler/alljobs?length=25&search=arm-defconfig#table

gctucker:
OK thanks.  I'll double check with the "pipe cleaner" job after updating production.

PR #344, opened 2020-03-31
test-configs.yaml: align all URLs to use plain HTTP

gctucker:
Some URLs were using HTTPS while others HTTP, align them all to use
HTTP instead.  In particular, this makes it possible to have file
caching in test labs with Squid.


gctucker:
Tested OK on staging, sample LAVA job: https://lava.collabora.co.uk/scheduler/job/2318150

PR #345, opened 2020-03-31
jenkins/dockerfiles: build-base: add rsync

khilman:
kselftest build/install has (unexplained) dependency on rsync.  While
this may be removed eventually, let's add so we can build/install
current (and older) kselftest.


gctucker:
The only thing the automated update on staging.kernelci.org doesn't do is Docker images because there aren't any credentials installed to push the new images to the Docker hub.  So I'm updating the staging images now, they should take effect in the next runs.

PR #348, opened 2020-04-07
WIP get baseline dmesg definition from git repo with dmesg.sh

gctucker:
ToDo: create kernelci/test-definitions if we want to do that

Rather than defining the dmesg test inline, use a separate repository
with dmesg.yaml as well as a script dmesg.sh to control what is being
output in the LAVA log.  This way, commands inside the shell script
are not visible in the log and reduce noise.  It also helps to be able
to separate test definitions from kernelci-core which is aiming to be
a pure Python package.

### DIFF
--- a/templates/baseline/baseline.jinja2
+++ b/templates/baseline/baseline.jinja2
@@ -5,37 +5,11 @@
     timeout:
       minutes: 1
     definitions:
-    - repository:
-        metadata:
-          format: Lava-Test Test Definition 1.0
-          name: baseline
-          description: "baseline test plan"
-          os:
-            - debian
-          scope:
-            - functional
-          environment:
-            - lava-test-shell
-        run:
-          steps:
-            - >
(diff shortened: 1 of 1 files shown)


gctucker:
This is a WIP, we should not be using my personal repository on kernelci.org.  Before creating a repository, if we actually need one, I want to check that the log fragments look good on staging when using this `dmesg.sh` approach.

@montjoie What kind of filter are you thinking of?  And, in what way would this be harder to use than the previous inline definition?

gctucker:
Tested OK on staging: https://staging.kernelci.org/test/case/id/5e8d8cfa9cb1a40f620ab6fc/

Now I need to make this production-friendly and drop the dependency on my personal repo on gitlab.collabora.com before we can merge it.

gctucker:
Creating a new PR, the `dmesg.sh` script is being added to the rootfs rather than on a remote git repo.

PR #350, opened 2020-04-09
jenkins: remove build- from DOCKER_BASE

gctucker:
Remove the build- part of the DOCKER_BASE strings and add it in each
job that actually requires it.  This is to be able to have the same
DOCKER_BASE for jobs that require other types of Docker images such as
kernelci/debos.


gctucker:
Tested OK on staging with `kernelci-jenkins` deployment.

PR #351, opened 2020-04-09
dockerfiles: Add a container for clang-10 builds

broonie:

### DIFF
--- a/jenkins/dockerfiles/clang-10/Dockerfile
+++ b/jenkins/dockerfiles/clang-10/Dockerfile
@@ -1,0 +1,18 @@
+FROM kernelci/build-base
+
+RUN apt-get update && apt-get install --no-install-recommends -y \
+    software-properties-common \
+    gnupg2
+
+RUN wget -q -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add -
+RUN apt-add-repository 'deb http://apt.llvm.org/buster/ llvm-toolchain-buster-10 main'
+
+RUN apt-get update && apt-get install --no-install-recommends -y \
+    binutils-aarch64-linux-gnu \
+    binutils-arm-linux-gnueabihf \
+    binutils \
+    clang-10
+
+RUN update-alternatives --install /usr/bin/clang clang /usr/bin/clang-10 500
+
+RUN apt-get autoremove -y gcc


gctucker:
@broonie This will add a Docker image, but it won't be used until we also update `build-configs.yaml`.  Do we want to start building some kernels with clang-10 now?

broonie:
It would be nice to start building with clang-10 but given the amount of builders we have right now is probably not the time - I figure it's good to have the image there in advance of when it's needed though to simplify deployment.

gctucker:
If we want to build linux-next with clang-10 now instead of clang-9, that's fine.

@nickdesaulniers What are your views on this?

broonie:
It would be good to keep coverage of the oldest version we support if we can, though while it's not something that people get from their distros that's perhaps not that important.

gctucker:
As far as Clang builds are concerned, we're only doing linux-next/master with Clang 9 at the moment.  Clang 8 builds have stopped, and I don't think we've ever used earlier versions.

We can actually decide to add just a couple of builds, say arm64 defconfig and x86_64_defconfig with Clang 10 and gradually increase.  It's easy enough to configure in build-configs.yaml.

nickdesaulniers:
Thanks for setting up the new image!

Clang 10 has been released, but there will be a 10.1 release in a few months with bugfixes only that @tstellar is driving.  It would be good to stand up some testing wit clang 10 asap, to make sure we don't miss any fixes (we're still chasing some backports for PPC recently).

Clang 9 the only issue that has regressed that I know of is https://github.com/ClangBuiltLinux/linux/issues/961 though it sounds like we have a fix on the kernel side.

If you all don't mind, and have the capacity, I wouldn't mind continuing support for clang-9+.  We haven't committed to a minimum required version of clang yet for the kernel, though we're moving towards clang-9 as the most likely candidate.

Personally, I'm more interested in the latest releases not regressing anything, but I think we have bandwidth to support older releases (not that clang-9 is old) and it's not painful quite yet.

I probably didn't need to delete clang-8 support when I set up clang-9.

gctucker:
OK so I think we can safely add defconfig builds for arm64 and x86_64 with clang-10 now, and keep clang-9 as-is.  See #363.

I'll also enable this on staging.kernelci.org now temporarily to check everything works as expected.

gctucker:
Tested OK on staging with arm, arm64 and x86_64.

PR #352, opened 2020-04-09
dockerfiles: Add container for DT validation

broonie:
The kernel has started moving devicetree schemas to a format which
allows for automatic validation.  This requires a separate build target
and a toolset (currently not packaged for any distro) so create a new
docker container with the toolset installed that we can use to run these
tests.

### DIFF
--- a/jenkins/dockerfiles/build-and-push.sh
+++ b/jenkins/dockerfiles/build-and-push.sh
@@ -23,13 +23,14 @@
 # -p = push to dockerhub
 # -b = also rebuild build-base
 # -d = also rebuild debos
+# -i = also rebuild dt-validation
 # -q = make the builds quiet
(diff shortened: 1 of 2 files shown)


gctucker:
Sounds good, that would be something to add to the "static analysis" category of tests we're starting to add.

We should just need to run a DT validation in this Docker image manually to get this PR merged and then work on how to automate things.  It should get pushed to the Docker hub later today.

gctucker:
The image is there: https://hub.docker.com/r/kernelci/staging-dt-validation

@broonie Can you please try to run dt-validate using this Docker image with an actual kernel source tree (from git or a kernelci tarball) and post a comment with the results?

broonie:
Yes, it works for me.

PR #353, opened 2020-04-09
dockerfiles: Include arm toolchain in the arm64 GCC container

broonie:
arm64 supports 32 bit arm userspace, including a VDSO for that userspace
which needs a 32 bit toolchain to build, so add the arm toolchain so we
can provide coverage for it.

Signed-off-by: Mark Brown <broonie@kernel.org>

PR #354, opened 2020-04-09
Cross compile compat

broonie:
These two patches update kci_build to build the compat VDSO on arm64, giving us coverage of that.

Doing these builds will depend on having a docker container with the Arm compiler added as enabled by the separate pull request https://github.com/kernelci/kernelci-core/pull/353

### DIFF
--- a/kci_build
+++ b/kci_build
@@ -193,9 +193,8 @@ def _show_build_env(build_env, arch=None):
     print(build_env.cc_version)
     if arch:
         print(build_env.get_arch_name(arch) or '')
-        xc = build_env.get_cross_compile(arch)
-        if xc:
-            print(xc)
+        print(build_env.get_cross_compile(arch) or '')
+        print(build_env.get_cross_compile_compat(arch) or '')
 
 
 class cmd_get_build_env(Command):
--- a/build-configs.yaml
+++ b/build-configs.yaml
@@ -171,6 +171,8 @@ build_environments:
       arm64: 'aarch64-linux-gnu-'
       mips: 'mips-linux-gnu-'
       riscv: 'riscv64-linux-gnu-'
+    cross_compile_compat: &default_cross_compile_compat
+      arm64: 'arm-linux-gnueabihf-'
 
   clang-9:
     cc: clang
@@ -184,6 +186,7 @@ build_environments:
       arc:
       riscv:
     cross_compile: *default_cross_compile
+    cross_compile_compat: *default_cross_compile_compat
 
   clang-10:
     cc: clang
--- a/kernelci/build.py
+++ b/kernelci/build.py
@@ -434,7 +434,7 @@ def _output_to_file(cmd, log_file, rel_dir=None):
 
 def _run_make(kdir, arch, target=None, jopt=None, silent=True, cc='gcc',
               cross_compile=None, use_ccache=None, output=None, log_file=None,
-              opts=None):
+              opts=None, cross_compile_compat=None):
     args = ['make']
 
     if opts:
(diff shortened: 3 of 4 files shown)


khilman:
Added `staging-skip` since this breaks staging until this dockerfiles change is also pushed: https://github.com/kernelci/kernelci-core/pull/353

gctucker:
The Docker images have now been updated so I've removed the `staging-skip` label on this PR.  We'll see in the next staging build whether this worked as expected.

kci_build
```diff
@@ -196,6 +196,9 @@ def _show_build_env(build_env, arch=None):
         xc = build_env.get_cross_compile(args.arch)
         if xc:
             print(xc)
+        xc_compat = build_env.get_cross_compile_compat(args.arch)
```
gctucker, line 199: The only issue here is that if for any reason `cross_compile_compat` is define but not `cross_compile`, then the compat value will be printed on the line where one would normally expect to see the cross compiler.  This probably won't be a problem in practice as I guess you need to be cross-compiling first if you also want to have a compat toolchain, but there's nothing in the configuration to enforce it.

So it's probably better to print an empty string when a value is not there to always have the same number of lines being output.  Ideally these should be key-value pairs but then it makes it a bit harder to parse...

kci_build
```diff
@@ -196,6 +196,9 @@ def _show_build_env(build_env, arch=None):
...
+        if xc_compat:
+            print(xc_compat)
```
gctucker, line 201:
```suggestion
        xc = build_env.get_cross_compile(args.arch, '')
        print(xc)
        xc_compat = build_env.get_cross_compile_compat(args.arch, '')
        print(xc_compat)
```

gctucker:
Waiting for the next round of builds on staging to check this is all working as expected, the staging Docker images have been updated with 32-bit arm toolchains thanks to #353.

gctucker:
Tested OK on staging, at least the command line looks fine:
https://staging.kernelci.org/build/id/5ea238da209d5df6a0fda81b/
https://storage.staging.kernelci.org/kernelci/staging.kernelci.org/staging-20200424.0/arm64/defconfig/gcc-8/build.log
```
# make KBUILD_BUILD_USER=KernelCI -C/home/buildslave/workspace/kernel-build/linux ARCH=arm64 CROSS_COMPILE=aarch64-linux-gnu- CROSS_COMPILE_COMPAT=arm-linux-gnueabihf- HOSTCC=gcc CC="ccache aarch64-linux-gnu-gcc" O=/home/buildslave/workspace/kernel-build/linux/build defconfig
```

gctucker:
Here's the kernel config for the build in the link above:
https://storage.staging.kernelci.org/kernelci/staging.kernelci.org/staging-20200424.0/arm64/defconfig/gcc-8/kernel.config
which does have this option enabled:
```
CONFIG_COMPAT_VDSO=y
```
so that's another good sign, and the `vdso32` binaries show up in the log:
```
  CALL    ../scripts/checksyscalls.sh
  LDS     arch/arm64/kernel/vdso/vdso.lds
  AS      arch/arm64/kernel/vdso/note.o
  AS      arch/arm64/kernel/vdso/sigreturn.o
  CC      arch/arm64/kernel/vdso/vgettimeofday.o
  LD      arch/arm64/kernel/vdso/vdso.so.dbg
  VDSOSYM include/generated/vdso-offsets.h
  LDS     arch/arm64/kernel/vdso32/vdso.lds
  CC32    arch/arm64/kernel/vdso32/note.o
  CC32    arch/arm64/kernel/vdso32/vgettimeofday.o
  AS32    arch/arm64/kernel/vdso32/sigreturn.o
  HOSTCC  arch/arm64/kernel/vdso32/../../../arm/vdso/vdsomunge
  LD32    arch/arm64/kernel/vdso32/vdso.so.raw
  MUNGE   arch/arm64/kernel/vdso32/vdso.so.dbg
  VDSOSYM include/generated/vdso32-offsets.h
```

gctucker:
The previous comment was about `gcc-8` but this also worked with `clang-9`:
https://staging.kernelci.org/build/id/5ea23f06abd486facefda818/
https://storage.staging.kernelci.org/kernelci/staging.kernelci.org/staging-20200424.0/arm64/defconfig/clang-9/build.log

PR #361, opened 2020-04-15
kci_build: fix show_build_env and use it for Docker image names

gctucker:
Fix the output of `kci_build show_build_environment` and update the Jenkins library to use it rather than hard-coding the mapping between architecture names for the kernel and compilers.


gctucker:
Tested OK on staging with gcc and clang.

PR #364, opened 2020-04-17
RFC: demonstrate k8s builders

khilman:
NOT READY FOR MERGE

Intended to start review/conversation on how to best support build clusters from Kubernetes.

### DIFF
--- a/templates/k8s/job-build.jinja2
+++ b/templates/k8s/job-build.jinja2
@@ -1,0 +1,78 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ job_name }}
+spec:
+  backoffLimit: 3
+  template:
+    spec:
+      restartPolicy: OnFailure
+
+      # in-memory tmpfs
+      volumes:
+      - name: scratch-volume
+        emptyDir: { medium: "Memory" }
+
+      # only run on preemptible nods
+      #nodeSelector:
+      #  cloud.google.com/gke-preemptible: "true"
+
+      containers:
+      - name: kci-build
+        image: {{ "FIXME" | env_override('DOCKER_BASE') }}build-{{ "FIXME" | env_override('DOCKER_IMAGE') }}
+
+        volumeMounts:
+        - mountPath: "/scratch"
+          name: scratch-volume
+
+        command: ["/bin/bash", "-x", "-c"]
+        args: ["echo nproc=$(nproc) && df && free && \
+export KDIR=/scratch/linux && export OUTPUT=$HOME/build && export CCACHE_DISABLE=true &&\
+git clone --depth=1 --branch ${KCI_CORE_BRANCH} ${KCI_CORE_URL} && cd kernelci-core && \
+./kci_build pull_tarball --kdir ${KDIR} --url ${SRC_TARBALL} --retries 3 --delete && \
+./kci_build build_kernel --kdir ${KDIR} --output ${OUTPUT} --defconfig=${DEFCONFIG} --arch=${ARCH} --build-env=${BUILD_ENVIRONMENT} --verbose ${PARALLEL_JOPT} && \
+./kci_build install_kernel --kdir ${KDIR} --output ${OUTPUT} --config ${BUILD_CONFIG} --describe=${GIT_DESCRIBE} --describe-verbose=${GIT_DESCRIBE_VERBOSE} --commit=${COMMIT_ID} && \
+./kci_build push_kernel --kdir ${KDIR} --token=${KCI_API_TOKEN} --api=${KCI_API_URL} && \
+./kci_build publish_kernel --kdir ${KDIR} --token=${KCI_API_TOKEN} --api=${KCI_API_URL} \
+"]
+
+        env:
+        - name: ARCH
+          value: "{{ "FIXME" | env_override('ARCH') }}"
+        - name: BUILD_ENVIRONMENT
+          value: "{{ "FIXME" | env_override('BUILD_ENVIRONMENT') }}"
+        - name: DEFCONFIG
+          value: "{{ "FIXME" | env_override('DEFCONFIG') }}"
+        - name: KCI_CORE_URL
+          value: "{{ "FIXME" | env_override('KCI_CORE_URL') }}"
+        - name: KCI_CORE_BRANCH
+          value: "{{ "FIXME" | env_override('KCI_CORE_BRANCH') }}"
+        - name: SRC_TARBALL
+          value: "{{ "FIXME" | env_override('SRC_TARBALL') }}"
+        - name: GIT_DESCRIBE
+          value: "{{ "FIXME" | env_override('GIT_DESCRIBE') }}"
+        - name: GIT_DESCRIBE_VERBOSE
+          value: "{{ "FIXME" | env_override('GIT_DESCRIBE_VERBOSE') }}"
+        - name: COMMIT_ID
+          value: "{{ "FIXME" | env_override('COMMIT_ID') }}"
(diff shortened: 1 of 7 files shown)


gctucker:
Please recreate this PR with the destination branch set to `master`.

broonie:
Big picture this seems to make sense to me, especially as a first pass integration thing - I know even less about the Groovy bits than the Kubernetes bits but they look sensible too. I'll go through with some more comments but most of them are more random ideas or things we could do in future.

templates/k8s/job-build.jinja2
```diff
@@ -0,0 +1,71 @@
...
+      #nodeSelector:
+      #  cloud.google.com/gke-preemptible: "true"
+
+      containers:
+      - name: kci-build
+        image: {{ "FIXME" | env_override('DOCKER_BASE') }}build-{{ "FIXME" | env_override('DOCKER_IMAGE') }}
+
+        volumeMounts:
+        - mountPath: "/scratch"
+          name: scratch-volume
```
broonie, line 26: May be worth having a separate volume for the kernel source - again, more for building on in future than for right now. It'd be especially good if we could get to a point where we're sharing downloads between multiple builds on the same cluster, but that's definitely a separate step. I think that's working quite well for me in the experiments I'm doing. That's definitely a separate step in terms of KCI integration though.

mattface:
Does this mean there would have to be a builder with "k8s" label, which in turn launches nodes in kubernetes?
Can 1 builder launch many kubernetes builds or is it a 1:1 ratio?

templates/k8s/job-build.jinja2
```diff
@@ -0,0 +1,71 @@
...
+        - name: GIT_DESCRIBE_VERBOSE
+          value: "{{ "FIXME" | env_override('GIT_DESCRIBE_VERBOSE') }}"
+        - name: COMMIT_ID
+          value: "{{ "FIXME" | env_override('COMMIT_ID') }}"
+        - name: BUILD_CONFIG
+          value: "{{ "FIXME" | env_override('BUILD_CONFIG') }}"
+        - name: KCI_API_URL
+          value: "{{ "FIXME" | env_override('KCI_API_URL') }}"
+        - name: KCI_API_TOKEN
+          value: "{{ "FIXME" | env_override('KCI_API_TOKEN') }}"
```
broonie, line 63: There's a Kubernetes secret store (https://kubernetes.io/docs/concepts/configuration/secret/) but I've not had cause to try to use it yet.

templates/k8s/job-build.jinja2
```diff
@@ -0,0 +1,71 @@
...
+
+        resources:
+          limits:
+            cpu: "8"
```
broonie, line 67: This probably wants to be joined up with the small/medium/big build stuff we've got.

gctucker: Yes, looks like it would be worth having a template parameter for this.

khilman: Sure, I can parametrize this, but I think longer-term, we might want to rethink our small/medium big.  That's needed when we have a static pool of VM, and "big" builds would block others.  With k8s, we have clusters that can auto-scale based on demand, so we shouldn't need to manage the small/medium/large from the jenkins side, IMO.

templates/k8s/job-build.jinja2
khilman: Yeah, I saw that too and need to look into it.

templates/k8s/job-build.jinja2
khilman: AFAICT it looks like we can have a shared volume between all nodes in cluster.  If that's do-able, it's probably worth rethinking the whole tarball download, and having a way for a k8s cluster to just use a local git mirror on a shared volume.

But for now, this fits well into our current tarball method.

templates/k8s/job-build.jinja2
```diff
@@ -0,0 +1,71 @@
...
+
+        volumeMounts:
+        - mountPath: "/scratch"
+          name: scratch-volume
+
+        command: ["/bin/bash", "-x", "-c"]
+        args: ["echo nproc=$(nproc) && df && free && \
+export KDIR=/scratch/linux && export OUTPUT=$HOME/build && export CCACHE_DISABLE=true && env &&\
+git clone --depth=1 --branch ${KCI_CORE_BRANCH} ${KCI_CORE_URL} && cd kernelci-core && \
+./kci_build pull_tarball --kdir ${KDIR} --url ${SRC_TARBALL} --retries 3 --delete && \
```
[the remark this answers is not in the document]
khilman: > Does this mean there would have to be a builder with "k8s" label, which in turn launches nodes in kubernetes?
> Can 1 builder launch many kubernetes builds or is it a 1:1 ratio?

If you use the official k8s jenkins plugin, it's a 1:1 ration, but I'm not using that (for many lots of reasons.)

In this approach, the builder is just to drive k8s via the cmdline (kubectl), so the builder node can launch as many k8s builds as we want.

My idea was that we create a single `k8s` builder with the #executors set to the maximim number of builds we can support in the k8s cluster.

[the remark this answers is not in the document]
khilman: > It might be good to split the downloads into a Kubernetes init step. It looks like this lets us take advantage of Kubernetes' retry mechanisms - I know there's some in kci_build itself but it's a bit more idiomatic.

I'm new enough to k8s to not be sure how to do that.  AFAICT, each step in k8s is a separate container, and while you can then chain them with dependencies, I haven't looked closely at how to do that yet.  Hopefully someone who knows more about k8s can help us evolve this in an intelligent way.

templates/k8s/job-build.jinja2
broonie: I think it might still be worth it from the point of view of helping the scheduler place things effectively, especially if we manage to do stuff like sharing the fetch of the source between multiple nodes - though from my fiddling with my stuff it might end up being more that we have things like separate templates for some of the builds, especially the smaller ones like most of the arm configs and allnoconfig.

templates/k8s/job-build.jinja2
broonie: Yeah, we can definitely have a shared volume (though a lot of it seems annoyingly provider specific) - have to figure out if they're performant enough but hopefully. I definitely agree this is good for what we're doing with tarballs though, a lot of these were just ideas I'd had from my fiddling.

templates/k8s/job-build.jinja2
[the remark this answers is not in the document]
broonie: It's pretty noddy to use - the examples in the page linked below are good, it's basically just a list of containers/commands that it steps through in series in the order they're listed in the spec. I'm using it by putting the /scratch volume in the initContainers and doing the source fetches there.

https://kubernetes.io/docs/concepts/workloads/pods/init-containers/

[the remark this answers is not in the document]
mattface: > > Does this mean there would have to be a builder with "k8s" label, which in turn launches nodes in kubernetes?
> > Can 1 builder launch many kubernetes builds or is it a 1:1 ratio?
> 
> If you use the official k8s jenkins plugin, it's a 1:1 ration, but I'm not using that (for many lots of reasons.)
> 
> In this approach, the builder is just to drive k8s via the cmdline (kubectl), so the builder node can launch as many k8s builds as we want.
> 
> My idea was that we create a single `k8s` builder with the #executors set to the maximim number of builds we can support in the k8s cluster.

Ah cool, so really it's just a very low cpu k8s "monitor" job, gotcha

PR #365, opened 2020-04-21
test-configs: blacklist stm32_defconfig for stm32mp157c-dk

montjoie:
The stm32mp157c-dk does not boot with kernel produced by stm32_defconfig.
More generaly (quoting Alexandre Torgue):
For mpu boards you have to use multi_v7_defconfig.
The stm32_defconfig has to be used for MCU boards (f4/f7/h7) only.

### DIFF
--- a/test-configs.yaml
+++ b/test-configs.yaml
@@ -1084,6 +1084,8 @@ device_types:
     mach: stm32
     class: arm-dtb
     boot_method: uboot
+    filters:
+      - blacklist: {defconfig: ['stm32_defconfig']}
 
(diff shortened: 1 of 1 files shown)


gctucker:
This looks fine but it would be good to confirm that the platform is still getting tested on staging as a sanity check for the filter, however lab-baylibre is not available right now (investigating).

gctucker:
The lab-baylibre has just come back online with the new API tokens etc... and I can't see any baseline results for `stm32mp157c-dk2` so I think this will have to wait until the next production update.

PR #366, opened 2020-04-21
RFC v2: demonstrate k8s builders

khilman:
Updates from v1:
- rebase onto master branch

updated from last time
- CPU limit/request paramterized based on PARALLEL_BUILDS
- KCI_API_TOKEN stored as k8s secret
- new docker image (khilman/build-k8s) for driving k8s.  Based on build-base w/gcloud-sdk installed
- jenkins "k8s" nodes now have env variables to configure
  all gcloud secrets and environment

Issues
- kci_build needs a way to pass secrets (e.g tokens) not on the cmdline (otherwise they show on stdout from k8s)

### DIFF
--- a/templates/k8s/gen.py
+++ b/templates/k8s/gen.py
@@ -1,0 +1,66 @@
+#!/usr/bin/env python3
+
+import os
+from jinja2 import Environment, FileSystemLoader
+import re
+
+
+def env_override(value, key):
+    return os.getenv(key, value)
+
+
+d = os.getenv('DEFCONFIG').split('+')
+defconfig = d[0]
+frag = None
+if len(d) > 1:
+    frag = d[1]
+
+job_name = 'build-j{}-{}-{}-{}'.format(
+    os.getenv('BUILD_ID'),
+    os.getenv('ARCH'),
+    os.getenv('BUILD_ENVIRONMENT'),
+    defconfig,
+)
+
+if frag:
+    frag = os.path.splitext(os.path.basename(frag))[0]
+    job_name += "-{}".format(frag)
+
+# job name can only have '-'
+job_name = re.sub('[\./_+=]', '-', job_name).lower()
+
+# k8s limits job-name to max 63 chars (and be sure it doesn't end with '-')
+job_name = job_name[0:63].rstrip('-')
+
+# FIXME: needs to be tweaked according to k8s cluster VMs
+cpu_limit = int(os.getenv('K8S_CPU_LIMIT', 8))
+parallel_builds = os.getenv('PARALLEL_BUILDS')
+if parallel_builds:
+    parallel_builds = int(parallel_builds)
+    cpu_limit = min(cpu_limit, parallel_builds)
+    os.environ['PARALLEL_JOPT'] = "-j{}".format(parallel_builds)
+
+if (cpu_limit < 8):
+    cpu_request = cpu_limit * 0.875
+else:
+    cpu_request = cpu_limit - 0.9
+
+# VMs are generous, let's be greedy and ask for 1Gb per core :)
+mem_request = cpu_limit
+
+params = {
+    'job_name': job_name,
+    'cpu_limit': cpu_limit,
+    'cpu_request': cpu_request,
+    'mem_request': "{}Gi".format(mem_request)
+}
+env = Environment(loader=FileSystemLoader(['.', 'templates', 'templates/k8s']),
+                  extensions=["jinja2.ext.do"])
+env.filters['env_override'] = env_override
+template = env.get_template("job-build.jinja2")
+job_yaml = template.render(params)
+f = job_name + ".yaml"
+fp = open(f, "w")
+fp.write(job_yaml)
+fp.close()
+print(job_name)
--- a/templates/k8s/job-build.jinja2
+++ b/templates/k8s/job-build.jinja2
@@ -1,0 +1,100 @@
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: {{ job_name }}
+spec:
+  completions: 1
+  template:
+    spec:
+      restartPolicy: OnFailure
+
+      # in-memory tmpfs
+      volumes:
+      - name: scratch-volume
+        emptyDir: { medium: "Memory" }
+
+      # only run on preemptible nods
+      #nodeSelector:
+      #  cloud.google.com/gke-preemptible: "true"
+
+      # tolerate Azure spot instances
+      tolerations:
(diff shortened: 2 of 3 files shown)


khilman:
Actually, in v2, the runners don't need any gcloud credentials, only the gcloud and k8s utils installed (hence the new build-k8s docker image in https://github.com/kernelci/kernelci-core/pull/368)

The way it works now is the only secret is the GCP service-account key-file, which is configured as a jenkins secret, and then mounted into the build-k8s container when it runs.

The runner also has a few environment variables to specify the specific cluster (GCE project name, cluster name, region) but these aren't secrets.  This way we would just need one runner for each k8s cluster of VMs (these are typically region specific, at least in Azure and GCP).

About the 1:1 mapping.  It's not a direct mapping.

Right now its a 1:1 mapping between 1 kernel build (jenkins job) and one k8s **job** (not VM.).  k8s then take s the "job", creates a pod for it then finds a VM for it to run on.  If there's room on an exiting VM, it will use that, otherwise it will spin up another. 

k8s uses the cpu_request part of the yaml job description to determine if it can fit on an existing VM or not.  (I"m using our current PARALLEL_JOBS to map to k8s CPU request.)

In my current experiments, I have a cluster of 8-core VMs, so there's room for 1 medium/big job per VM and 2 "small" jobs per VM.

khilman:
Updated, now tested with Azure.

khilman:
Found a limitation with the  `kubectl` directly on builder approach.  To control multiple clusters (e.g. we can have more than one on GCE as well as on azure), `kubectl` needs to switch "context" between them.  It does this by updating a local config file, and it uses locking to updated this config file.

When running `kubectl` for all the clusters as the buildslave user, there will be multiple instances of the buildslave user trying to use `kubectl` from different contexts, causing races and lock contention.

The fix for this is to use a container for the k8s "control" jobs.  This has been started in https://github.com/kernelci/kernelci-core/pull/368, but now needs work to configure auth and credential for both GCP and Azure.

khilman:
@gctucker @broonie OK, this version is pretty stable now and has been working for a solid week on bot.kernelci.org

The main difference in previous versions is the new `wait.py` script that uses the python3-kubernetes library instead of the pile of calls to `kubectl` directly from build.jpl

I think we should test his for a while in staging, and then probably migrate my `gen.py` and `wait.py` into new sub-functions of kci_build.

gctucker:
Thanks, sounds good.

I guess we can have a k8s folder on staging with jobs using the branch from this draft PR, then when it becomes ready to be merged it will get used in the main staging jobs.

From an infrastructure point of view, this will also enable the new bot.kernelci.org instance to become production with all the current static builders to retire the "old" ci.kernelci.org until this gets merged.

gctucker:
I've created this temporary branch to try thing out on staging side-by-side with regular builds: https://github.com/kernelci/kernelci-core/tree/staging-k8s

broonie:
Looks good to me.

gctucker:
Builds are working fine on staging now, but tests can't be scheduled because the test-runner job gets the JSON files generated by the build job to get the meta-data.  They are archived in `build.jpl`:
```
    dir("linux/_install_") {
        archiveArtifacts("*.json")
    }
```
and then retrieved in `test-runner.jpl` using the given Jenkins build job number.  This is to avoid going via the storage server when the information is already present in Jenkins.

In this PR, the JSON files are left in the Kubernetes container.  So we need to either retrieve them from there in the Jenkins job and archive them to replicate the previous behaviour, or change test-runner to retrieve them from the storage server.  The latter approach is probably more viable.

Note that we should keep using the labs meta-data JSON from in the build-trigger job as that's not pushed anywhere outside Jenkins.

khilman:
D'oh, I was testing without scheduling tests so overlooked the artifacts passed to the test-runner job.

IIUC, what's missing is `bmeta.json` and `dtbs.json` from the build step, right?

I started looking at how to copy these back from the k8s build container, but then realized that the build step is already pushing them to the KCI storage server.  What if we modify test-runner so that it just fetches those files from `KCI_STORAGE_URL` instead of them being passed in the pipeline?  An added benefit to this approach is making it more portable to other pipeline systems.

gctucker:
Yes, like I wrote in my comment:

> So we need to either retrieve them from there in the Jenkins job and archive them to replicate the previous behaviour, or change test-runner to retrieve them from the storage server. The latter approach is probably more viable.

Downloading from the storage server seems fine.  One potentially tricky part though is to create the full URL for the build directory where the JSON files were uploaded.  That's essentially the `file-server-resource` value from the build meta-data which is set by `kci_build`.  It's best if we can reuse this value somehow rather than recreate it, as that would be bound to break at some point.

khilman:
heh, sorry.  didn't fully read your comment.

Re: `file_server_resource`, yes I was just looking at `kernelci/build.py` to see where `file_server_resource` was created.  Looks like we should separate that part of build.py out into something that `kci_test` could use if `test-runner.jpl` was passed all the same vars as `build.jpl`.

khilman:
Updated this to only have the support scripts/templates.  Jenkins job changes are now moved to kernelci-jenkins: https://github.com/kernelci/kernelci-jenkins/pull/20

gctucker:
@khilman From this PR's description:

> kci_build needs a way to pass secrets (e.g tokens) not on the cmdline (otherwise they show on stdout from k8s)

Is this still the case? And if so, are the k8s job logs available anywhere so that it makes it a blocking issue?

khilman:
> @khilman From this PR's description:
> 
> > kci_build needs a way to pass secrets (e.g tokens) not on the cmdline (otherwise they show on stdout from k8s)
>
> Is this still the case? And if so, are the k8s job logs available anywhere so that it makes it a blocking issue?

The k8s job logs are in the stdout of the Jenkins jobs.  If you search for `--token` you will see the tokens in plaintext.

I'm not sure if our jenkins logs are public, so not sure if this is a blocking issue.

gctucker:
Erm, the bot.staging.kernelci.org instance had read-only permissions for anonymous users but I've just turned that off due to the issue highlighted above.  So we need to finally add support for a local config file to kernelci-core, pretty much like kernelci-deploy already does.

This isn't really a blocker in practice right now as it's not public any more and production isn't public either, but it's pretty much as close to being a blocker issue as it can.  Also I think it would be good to replace the tokens anyway in case they were leaked, and even use different tokens for k8s jobs scheduled from either staging or production.

khilman:
> @khilman I think this is ready to be merged now, but can you please update the commit message?
> The parts about Jenkins should be added to the PR in kernelci-jenkins now, this PR is only about the Python core functions.

Updated.

PR #368, opened 2020-04-22
dockerfiles: add k8s build node

khilman:
Add new container capable of submitting build jobs to a k8s cluster.

The service-account key must be mounted at docker-run time for
'gcloud auth` to work correctly.

### DIFF
--- a/jenkins/dockerfiles/build-k8s/Dockerfile
+++ b/jenkins/dockerfiles/build-k8s/Dockerfile
@@ -1,0 +1,31 @@
+#
+# Container to drive Kubernetes cluster from KernelCI
+#
+FROM kernelci/build-base
+MAINTAINER "Kevin Hilman <khilmnan@baylibre.com>"
+
+#
+# Install gcloud-sdk
+# From: https://cloud.google.com/sdk/docs/downloads-apt-get
+#
+RUN apt-get update && apt-get install -y apt-transport-https ca-certificates gnupg curl
+RUN echo "deb [signed-by=/usr/share/keyrings/cloud.google.gpg] https://packages.cloud.google.com/apt cloud-sdk main" | tee -a /etc/apt/sources.list.d/google-cloud-sdk.list
+RUN curl https://packages.cloud.google.com/apt/doc/apt-key.gpg | apt-key --keyring /usr/share/keyrings/cloud.google.gpg add -
+RUN apt-get update && apt-get install -y google-cloud-sdk kubectl
+
+#
+# Azure CLI
+# From: https://docs.microsoft.com/en-us/cli/azure/install-azure-cli-apt?view=azure-cli-latest
+#
+RUN curl -sL https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor |  tee /etc/apt/trusted.gpg.d/microsoft.asc.gpg > /dev/null
+RUN AZ_REPO=buster && echo "deb [arch=amd64] https://packages.microsoft.com/repos/azure-cli/ $AZ_REPO main" | tee /etc/apt/sources.list.d/azure-cli.list
+RUN apt-get update && apt-get install -y azure-cli
+
+#
+# Kubernetes python lib (need latest from pip)
+# - remove existing deps (urllib3, requests) to be sure
+#   latest are installed from pip
+#
+RUN apt-get update && apt-get install -y python3-pip python3-setuptools
+RUN apt-get remove -y python3-urllib3 python3-requests && apt-get autoremove -y
+RUN python3 -m pip install --upgrade kubernetes


khilman:
Note that all that is needed in this container is the gcloud and kubectl tools installed, and then a "mount point" for the key-file.

I could not figure out how to bind-mount a single-file without first making sure it exists.  If I don't do the `touch` here, trying to bind-mount a file ends up creating an empty directory instead.

khilman:
I need to install the Azure CLI tools in this image also.  Will update.

khilman:
The way I have it working in my WIP K8S branch is this
- some jenkins build nodes are ansibled with GCP and Azure tools
- these nodes also have auth credentials setup/configured

Then, when a build runs on one of these "k8s" nodes, it runs in this `build-k8s` container, but bind mounts the config dirs needed for GCP, Azure and `kubectl`.

gctucker:
OK - do you still need to install the GCP and Azure tools on the builders with Ansible if the job is using a Docker container which also has them installed?

(not sure the Azure tools are installed in this PR, but I guess they could)

khilman:
Yes, I still need all the tools on the builders.  The builders are ansible'd with the tools as well as the auth/credentials.  Then when a k8s build is run, this container is started with bind-mounts for the config dirs (for gcloud, azure and kubectl.)

This allows the auth/credentials to only be configured once (at ansible time) instead of for every build.

and yes, Azure tools are now installed in this PR.

gctucker:
Sorry if I'm missing anything obvious, but why can't the tools be in the Docker image?

If the Docker container is started before doing any cluster authentication, and the credentials are bind-mounted, then it seems like it should be possible to use these tools in Docker too.

khilman:
Yes, the tools are in the docker image **in addition to** being installed on the builders directly (in order to setup the initial auth/credentials with ansible.)

jenkins/dockerfiles/build-k8s/Dockerfile
```diff
@@ -0,0 +1,31 @@
+#
+# Container to driver Kubernetes cluster from KernelCI
```
gctucker, line 2: s/driver/drive/ I guess?

jenkins/dockerfiles/build-k8s/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+#
+# Azure CLI
+# From: https://docs.microsoft.com/en-us/cli/azure/install-azure-cli-apt?view=azure-cli-latest
+#
+RUN curl -sL https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor |  tee /etc/apt/trusted.gpg.d/microsoft.asc.gpg > /dev/null
+RUN AZ_REPO=buster && echo "deb [arch=amd64] https://packages.microsoft.com/repos/azure-cli/ $AZ_REPO main" | tee /etc/apt/sources.list.d/azure-cli.list
+RUN apt-get update && apt-get install -y azure-cli
+
+#
+# Kubernetes python lib (need latest from pip)
```
gctucker, line 25: Do we really need the _latest_ version?  Having a numbered version would probably be a bit safer and also trigger image rebuilds when the number changes in the Dockerfile.

jenkins/dockerfiles/build-k8s/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+# - remove existing deps (urllib3, requests) to be sure
+#   latest are installed from pip
+#
+RUN apt-get install -y python3-pip python3-setuptools
```
gctucker, line 29: Please always do `apt-get update && apt-get install [...]` as Docker will only run this line again if the list of packages changes, and it really should run `apt update` first to get the latest meta-data.

jenkins/dockerfiles/build-k8s/Dockerfile
khilman: Yes

jenkins/dockerfiles/build-k8s/Dockerfile
khilman: The version deps seem buggy/incomplete in the original gcloud packages, so this was the only combination I was able to make work reliably.

jenkins/dockerfiles/build-k8s/Dockerfile
khilman: ok, updated.

jenkins/dockerfiles/build-k8s/Dockerfile
khilman: yes, updated.

khilman:
Yes, this exact Dockerfile was used to build images pushed to `khilman/`

jenkins/dockerfiles/build-k8s/Dockerfile
gctucker: OK fine.  If this breaks one day then we'll hopefully be able to go back to released versions that do work.  Maybe we should tag the Docker images too, to be able to roll back easily.

PR #370, opened 2020-04-23
lab-configs: use the new LAVA url to access lab-baylibre

montjoie:
Our lab has move and is now accessible only via https.


gctucker:
Let's see tomorrow if we get some results on staging.

gctucker:
Tested OK on staging https://staging.kernelci.org/test/job/kernelci/branch/staging.kernelci.org/kernel/staging-20200424.1/plan/baseline/

PR #329, opened 2020-02-21
Add buster-v4l2 and buster-igt configs

Lakshmipathi:
Include buster-v4l2 and buster-igt rootfs configurations

Signed-off-by: Lakshmipathi Ganapathi <lakshmipathi.ganapathi@collabora.com>

PR #371, opened 2020-04-23
STAGING user buster rootfs from staging 20200421.0

gctucker:
To test the output of #329


gctucker:
Tested OK on staging, the new rootfs images are now being deployed in production.

PR #373, opened 2020-04-24
build-trigger: pass-on params to build

khilman:
The build trigger should pass it's arguments on to the build job, in
particular those that affect kernelci-core repo usage and the kernelci
API.


khilman:
This is mostly useful when doing development on kernelci-core where you want the build-trigger and the build jobs to be using the same kernelci-core url/branch, same docker image base and same kernelci API.

gctucker:
Yes, I was hoping we would have switched to using config files for this before we had the need to pass all those arguments between jobs.  But that's fine for now.

gctucker:
Tested OK on staging.

PR #375, opened 2020-04-27
lab-configs.yaml: drop lab-drue as it's permanently offline

gctucker:
As discussed with Dan, lab-drue is now going to be offline until
further notice so remove it from the YAML config.


danrue:
:+1: Thanks @gctucker

gctucker:
Thanks for all the fish :)

PR #376, opened 2020-04-28
Handle vmlinux for MIPS build

montjoie:
Many of MIPS build have no kernel due to the fact they use vmlinux.
The case which I care are all malta defconfig.

So this patch adds vmlinux as valid kernel for MIPS and add the ability
for seek kernel images in the base output directory.

### DIFF
--- a/kernelci/build.py
+++ b/kernelci/build.py
@@ -49,7 +49,7 @@ KERNEL_IMAGE_NAMES = {
     'arm64': ['Image'],
     'arc': ['uImage'],
     'i386': ['bzImage'],
-    'mips': ['uImage.gz', 'vmlinux.gz.itb'],
+    'mips': ['uImage.gz', 'vmlinux.gz.itb', 'vmlinuz'],
     'riscv': ['Image', 'Image.gz'],
     'x86_64': ['bzImage'],
     'x86': ['bzImage'],
@@ -613,10 +613,12 @@ def build_kernel(build_env, kdir, arch, defconfig=None, jopt=None,
         print_flush("ERROR: Missing kernel config")
         result = False
     if result:
-        target = (
-            'xipImage' if _kernel_config_enabled(dot_config, 'XIP_KERNEL')
-            else MAKE_TARGETS.get(arch)
-        )
+        if _kernel_config_enabled(dot_config, 'XIP_KERNEL'):
+            target = 'xipImage'
+        elif _kernel_config_enabled(dot_config, 'SYS_SUPPORTS_ZBOOT'):
+            target = 'vmlinuz'
+        else:
+            target = MAKE_TARGETS.get(arch)
         result = _run_make(jopt=jopt, target=target, **kwargs)
     mods = _kernel_config_enabled(dot_config, 'MODULES')
     if result and mods:
@@ -753,6 +755,12 @@ def install_kernel(kdir, tree_name, tree_url, git_branch, git_commit=None,
             if name in files:
                 kimages.append(name)
                 image_path = os.path.join(root, name)
+                shutil.copy(image_path, install_path)
+    for files in os.listdir(output_path):
+        for name in kimage_names:
+            if name == files:
+                kimages.append(name)
+                image_path = os.path.join(output_path, name)
                 shutil.copy(image_path, install_path)
     if kimages:
         for name in kimage_names:


kernelci/build.py
```diff
@@ -754,6 +754,12 @@ def install_kernel(kdir, tree_name, tree_url, git_branch, git_commit=None,
                 kimages.append(name)
                 image_path = os.path.join(root, name)
                 shutil.copy(image_path, install_path)
+    for files in os.listdir(output_path):
+        for name in kimage_names:
+            if name == files:
+                kimages.append(name)
+                image_path = os.path.join(output_path, name)
+                shutil.copy(image_path, install_path)
```
gctucker, line 762: Isn't this already done by the block of code just above?

gctucker: Oh I see, `vmlinux` is not placed in the boot directory.  I think it would be better to change the loop above to iterate over the 2 paths instead of duplicating it.

montjoie: no, it seek only in arch/$arch/boot

montjoie: The first loop is a find, mine is just a "ls". I dont see how to combine both cleanly

gctucker:
The code could still be improved to avoid duplicating the logic to copy files, but that's not a blocker.

Note however that `vmlinux` images can be rather large, for example:
```
-rw-r--r-- 1 www-data www-data  64M May 20 08:49 vmlinux
-rw-r--r-- 1 www-data www-data 4.8M May 20 08:49 vmlinux.gz.itb
```
Is it actually practical to boot some MIPS board with such a `vmlinux` image over TFTP?

montjoie:
Unfortunatly we have no MIPS board to boot. The only one I could add is a CI20 but I am stuck trying to boot anything recent. So only the qemu malta will be booted and so no TFTP for it.
But qemu malta works also with vmlinuz which is smaller. So i will update this PR

kernelci/build.py
gctucker: OK I'll post a suggestion to improve this or a follow-up PR.

gctucker:
Thanks, waiting for the next staging build to complete to check the actual size of vmlinuz.

gctucker:
OK it built today, but there's no `vmlinuz` as far as I can tell:
https://storage.staging.kernelci.org/kernelci/staging.kernelci.org/staging-20200611.1/mips/32r2el_defconfig/gcc-8/

The build command line is:
```
make KBUILD_BUILD_USER=KernelCI -C/home/buildslave/workspace/kernel-build@2/linux -j4 ARCH=mips CROSS_COMPILE=mips-linux-gnu- HOSTCC=gcc CC="ccache mips-linux-gnu-gcc" O=build
```
Maybe `vmlinuz` is not built by default in the MIPS Makefile, so it should be added explicitly.

montjoie:
I think vmlinuz is built only with some targets. I have built today:
./kci_build build_kernel --kdir $HOME/linux-next/ --build-env gcc-8 --arch mips --defconfig malta_kvm_guest_defconfig --output build
and vmlinuz was found

gctucker:
That's right, it depends on `CONFIG_SYS_SUPPORTS_ZBOOT`:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/arch/mips/Makefile?h=v5.7#n86

which gets turned on in various places in `arch/mips/Kconfig` so this depends on the defconfig.

In practice, it's enabled in `malta_kvm_guest_defconfig`:
```
$ ARCH=mips make malta_kvm_guest_defconfig && grep ZBOOT .config
#
# No change to .config
#
CONFIG_SYS_SUPPORTS_ZBOOT=y
# CONFIG_DEBUG_ZBOOT is not set
```

but not in `32r2el_defconfig`:
```
$ ARCH=mips make 32r2el_defconfig && grep ZBOOT .config
Using ./arch/mips/configs/generic_defconfig as base
Merging arch/mips/configs/generic/32r2.config
Merging arch/mips/configs/generic/el.config
Merging ./arch/mips/configs/generic/board-sead-3.config
Merging ./arch/mips/configs/generic/board-ranchu.config
Merging ./arch/mips/configs/generic/board-ni169445.config
Merging ./arch/mips/configs/generic/board-boston.config
Merging ./arch/mips/configs/generic/board-ocelot.config
Merging ./arch/mips/configs/generic/board-xilfpga.config
scripts/kconfig/conf  --olddefconfig Kconfig
#
# configuration written to .config
#
```

gctucker:
So we should try to enable `malta_kvm_guest_defconfig` builds on staging to see if we can use `vmlinuz` to test this PR.

montjoie:
I have tested my PR against 32r2el_defconfig without any failure.
For this defconfig, no vmlinuz, but the build is not breaked.

gctucker:
Looks good, waiting to get staging back to work with Kubernetes and then that should be ready to be merged.  It will also be worth checking with a `malta_kvm_guest_defconfig` build on staging.

gctucker:
OK kicking a build with `malta_kvm_guest_defconfig` on staging now to finally check that it all works as expected.

gctucker:
Erm, kicking another build now.  Not sure what happened back then, probably got sidetracked.

gctucker:
Got a build with vmlinuz: https://staging.kernelci.org/build/id/5f909dffea1f21ccef99b284/

gctucker (approved):
OK this works fine on staging, and I guess it makes sense in practice if there aren't any boards able to boot with a `vmlinux.gz.itb` image.

PR #377, opened 2020-04-28
test-configs: add qemu malta

montjoie:
Since kernelCI builds all MIPS defconfig, we could now boot some MIPS qemu.
From all MIPS qemu, only malta are working.

### DIFF
--- a/test-configs.yaml
+++ b/test-configs.yaml
@@ -1015,6 +1015,21 @@ device_types:
     filters:
       - whitelist: {defconfig: ['i386_defconfig']}
 
+  qemu_mips-malta:
+    base_name: qemu
+    mach: qemu
+    arch: mips
+    boot_method: qemu
+    context:
+      arch: mipsel
+      guestfs_interface: 'ide'
+      machine: 'malta'
+      memory: 256
+      model: 'model=pcnet'
+      no_kvm: True
+    filters:
+      - whitelist: {defconfig: ['malta_kvm_guest_defconfig']}
+
   qemu_x86_64:
     base_name: qemu
     mach: qemu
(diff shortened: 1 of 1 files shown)


montjoie:
it seems that systemd does not like MIPS build. probably due to missing CGROUP

test-configs.yaml
```diff
@@ -1957,6 +1972,12 @@ test_configs:
       - boot_qemu
       - simple_qemu
 
+  - device_type: qemu_mips-malta
+    test_plans:
+      - baseline_qemu
+      - boot_qemu
+      - simple_qemu
```
gctucker, line 1979: Both boot and simple test plans have now been dropped, so please remove them.

FlyGoat:
malta_kvm_guest_defconfig was designed for trap and emul KVM, and it may not work with QEMU TCG.
I do think that malta_defconfig can be more practical.

FlyGoat:
The qemu cmdline that proofed to run malta with cgroup selected:
```
qemu-system-mipsel -nographic -net nic,model=pcnet \
        -machine malta -cpu M14Kc -m 2G \
        -kernel vmlinux \
        -append "root=/dev/sda rw console=ttyS0,38400n8r" \
        -drive file=/data/sysroots/debian-buster-mipsel.qcow2,if=ide
```

I'd suggest we add some virtio drivers to build config to help with build ability test.
Will #399 help?

gctucker:
@FlyGoat Hello, PR #399 is about determining which standard defconfigs were tested on some platforms, so it's not related to what needs to be done for QEMU with `malta`.

Enabling virtio drivers or any arbitrary kernel config options can be done with a config fragment, in the same way that the `virtualvideo` is being defined in `build-configs.yaml`.

gctucker:
Replaced with #522 which is this patch but with git conflicts resolved.

PR #374, opened 2020-04-27
T20434 baseline: use dmesg.sh script from the rootfs

gctucker:
The baseline rootfs images now have a dmesg.sh script which can be
called to run the dmesg tests and reduce the noise in the LAVA log.
Use it in the baseline job template rather than defining the steps
inline.

Signed-off-by: Guillaume Tucker <guillaume.tucker@collabora.com>

PR #379, opened 2020-04-29
T20434 test-configs.yaml: update buildroot with built-in dmesg.sh for baseline

gctucker:
Update to latest buildroot version kci-2019.02-10-g24600c76ce51 which
includes a KernelCI overlay with the dmesg.sh script for the baseline
test plan.


gctucker:
Tested OK on staging, see #374.

PR #381, opened 2020-04-30
build-configs: android: drop arc, mips builds

khilman:
Requested-by: Todd Kjos <tkjos@google.com>

### DIFF
--- a/build-configs.yaml
+++ b/build-configs.yaml
@@ -321,6 +321,15 @@ stable_variants: &stable_variants
         base_defconfig: 'x86_64_defconfig'
         extra_configs: ['allnoconfig']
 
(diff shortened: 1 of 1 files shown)


gctucker:
LGMT.

Side note: This also shows how it would be much simpler to have a separate file for the Android trees.  We should be able to get there soon enough, once we've separated the YAML config from kernelci-core.

gctucker:
Tested with this command:
```
$ ./kci_build list_kernel_configs --config=android_mainline --kdir=linux |  while read x; do echo $x | cut - -f1 -d\ ; done | sort -u
arm
arm64
i386
riscv
x86_64
```

PR #382, opened 2020-05-12
build-trigger: don't pass KCI_STORAGE_URL to build

khilman:
build.jpl doesn't use KCI_STORAGE_URL, don't pass it.

Fixes: 1a1c57041c61 (build-trigger: pass-on params to build)


khilman:
Don't understand the Travis fails.  `kci_rootfs` tests are failing, but this patch doesn't affect that at all.

gctucker:
Please rebase to clear the Travis CI errors that have now been fixed on master.

gctucker:
> Don't understand the Travis fails. `kci_rootfs` tests are failing, but this patch doesn't affect that at all.

I think that's because a new version of `pycodestyle` is now being used and it is stricter than before with short variable names.

PR #384, opened 2020-05-13
test-configs: handle rename of renesas boards

montjoie:
The DTB of all renesas board was renamed in v5.6, so renesas boards wont
get new jobs.
The simple fix is to duplicate them in kci and use LAVA aliases.

### DIFF
--- a/test-configs.yaml
+++ b/test-configs.yaml
@@ -678,12 +678,15 @@ device_types:
       - blacklist: {defconfig: ['allmodconfig', 'multi_v5_defconfig']}
       - blacklist: {kernel: ['v3.']}
 
-  r8a7796-m3ulcb:
+  r8a7796-m3ulcb: &r8a7796-m3ulcb
     mach: renesas
     class: arm64-dtb
     boot_method: uboot
     filters:
       - blacklist: {defconfig: ['allnoconfig', 'allmodconfig']}
+
+# this is the same device than r8a7796-m3ulcb
+  r8a77960-ulcb: *r8a7796-m3ulcb
 
   meson8b-odroidc1:
     mach: amlogic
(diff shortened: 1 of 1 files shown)


gctucker:
Please rebase to clear the Travis CI errors that have now been fixed on master.

test-configs.yaml
```diff
@@ -685,6 +685,14 @@ device_types:
     filters:
       - blacklist: {defconfig: ['allnoconfig', 'allmodconfig']}
 
+# this is the same device than r8a7796-m3ulcb
+  r8a77960-ulcb:
```
gctucker, line 689: How about creating an alias for the existing device entry and reusing it here?

```yaml
  r8a7796-m3ulcb: &r8a7796-m3ulcb
    mach: renesas
[...]
  r8a77960-ulcb: *r8a7796-m3ulcb
```
and maybe do the same thing for the list of test plans if we want the same coverage on both?

Out of interest, what is the actual difference between `r8a7796-m3ulcb` and `r8a77960-ulcb` apart from the name?  Is it a new PCB version or that the dtb name changed in the kernel?

test-configs.yaml
```diff
@@ -1093,6 +1101,15 @@ device_types:
     filters:
       - blacklist: {defconfig: ['allnoconfig', 'allmodconfig']}
 
+# this is the same device than r8a7795-salvator-x
```
gctucker, line 1104: Same comments as for r8a7796-m3ulcb.

test-configs.yaml
```diff
@@ -1981,13 +1998,30 @@ test_configs:
...
+  - device_type: r8a77950-salvator-x
+    test_plans:
+      - baseline
+      - boot
+      - boot-nfs
+      - kselftest
+      - simple
+
```
gctucker, line 2009: What I'm suggesting is:
```yaml
    test_plans: *r8a7795-salvator-x_test-plans
```

test-configs.yaml
montjoie: The DTB name changed in the kernel.
For the alias let me check it

gctucker:
@montjoie Also please rebase like I said a while ago, that's because the Travis CI checks have changed and there's a fix for that on the master branch.

gctucker:
Thanks, this should get tested on staging tomorrow.

gctucker:
It was tested while staging was frozen in preparation for the boot test deprecation in production, will check the next results when they're available.

gctucker:
Skipping temporarily while trying out conflicting changes with the new `lab-cip`.

gctucker:
There is a problem with dtb names, but it was already a problem before this PR since it's all due to the renaming that happened in the kernel.  So I'll merge this one now and make another PR to fix the issues.

PR #387, opened 2020-05-18
kernelci.build.install_kernel: replace slashes with dashes

gctucker:
Replace all slashes "/" with dashes "-" in each element of the path
where the build binaries are being uploaded to.  This can be done
because the actual values of each field is store separately in the
meta-data, the important thing is to keep the upload paths unique.

Doing this in particular enables slashes in git branch names.


gctucker:
While I think this works from a meta-data point of view, and it seems fine to store it in the database, there's still an issue with parts of the frontend not using the resource path.  Looking into getting this fully fixed end-to-end on staging before merging it.

gctucker:
Replaced with #492

PR #388, opened 2020-05-20
test-configs.yaml: Use max CPU for arm64 qemu targets

broonie:
The arm64 emulation in qemu supports architectural features which are not
present in currently shipping CPUs that are emulated by qemu. Many of
these have support in software and it would be good to get coverage for
them so rather than emulating cortex-a57 (which has plenty of physical
implementations) instead specify the CPU as "max" when invoking qemu which
will make the CPU as featureful as possible.

This does mean that the exact feature set of the CPU being emulated will
vary depending on the target system, the qemu version and when the host
is arm64 based the host hardware will control what gets selected. We will
need to pay attention to this going forwards.

### DIFF
--- a/test-configs.yaml
+++ b/test-configs.yaml
@@ -1004,7 +1004,7 @@ device_types:
     boot_method: qemu
     context:
       arch: arm64
-      cpu: 'cortex-a57'
+      cpu: 'max'
       guestfs_interface: 'virtio'
       machine: 'virt,gic-version=2'
     filters:
@@ -1017,7 +1017,7 @@ device_types:
     boot_method: qemu
     context:
       arch: arm64
-      cpu: 'cortex-a57'
+      cpu: 'max'
       guestfs_interface: 'virtio'
       machine: 'virt,gic-version=3'
     filters:


gctucker:
I think it's becoming really important to dump any details about the QEMU version and features of the CPU model in the test log.

@broonie Is there an easy way to get this kind of information at runtime, say from `proc`?  We could even add a test case that checks for some values in there to flag when some QEMU jobs start running unexpected variants.

broonie:
I'm not aware of anything we can see from the running kernel, sorry - we could run qemu with -version before invoking it for real but that's a bit more annoying for using directly in tests. It would then be possible to parse it and put it in the database though. That said what's more interesting for this stuff is going to be the hardware features that get exposed - like I said in the commit log those are going to also vary if we're running on an arm64 host and can run natively.

gctucker:
Right - of course the best thing would be to have even finer granularity over the QEMU variants we're testing so we positively run tests with them rather than have loose definitions and try to discover what was actually running.

For example, we could have more entries in `test-configs.yaml` with more parameters to pass to QEMU, and do something about QEMU versions.  However I think that can easily become quite hard to manage.

Maybe we should also look at what can be done in LAVA to get some information about QEMU (version, features) before booting it and have LAVA create extra test entries or meta-data fields automatically.  That's also not ideal as some labs may be running different LAVA versions, or not LAVA at all...

One last thought, could there be attributes added to the kernel in say, debugfs to expose this kind of information if it's already available internally?  Maybe that would take a bit of investigation and could only be done for some things but not everything...

broonie:
It looks like we're actually grabbing the version for the logs already:

https://storage.kernelci.org/broonie-regulator/for-next/v5.7-rc6-22-g95a220eb04f4/arm64/defconfig/gcc-8/lab-collabora/baseline-qemu_arm64-virt-gicv3.html

has it near the top of the logs (but not parsed usefully). 

I'm not aware of anything that's put in the emulated system to enable it to get version information at present, a bit of a poke around a running system doesn't show anything either.

gctucker:
Right, I see this from your log:
https://storage.kernelci.org/broonie-regulator/for-next/v5.7-rc6-22-g95a220eb04f4/arm64/defconfig/gcc-8/lab-collabora/baseline-qemu_arm64-virt-gicv3.html#L6
```
qemu-system-arm, installed at version: 1:3.1+dfsg-8+deb10u5, host architecture: amd64
```
I'll take a look to see if this gets sent in the meta-data or in the results as it would be much better to have it first-hand rather than trying to parse it from the log.

Then if we can get this information reliably, there are several ways to deal with it as it's not really a test result but more of an attribute of the test environment just like we have rootfs image versions etc...

gctucker:
Waiting for the next round of test results from staging to check if this causes any bad side-effects.  If not, then it may not be worth the extra effort of tracking the version in practice or at least not initially.

gctucker:
Tested OK on staging, here's a sample baseline result:
https://staging.kernelci.org/test/plan/id/5eeb57479b089d0931c68d09/

with the corresponding log showing the `max` parameter being used:
https://storage.staging.kernelci.org/kernelci/staging.kernelci.org/staging-20200618.2/arm64/defconfig/gcc-8/lab-baylibre/baseline-qemu_arm64-virt-gicv3.html#L161

gctucker:
Also, here's a sample baseline-uefi result, in a different lab:
https://staging.kernelci.org/test/plan/id/5eeb578f6e4824fb07c68d1c/

https://storage.staging.kernelci.org/kernelci/staging.kernelci.org/staging-20200618.2/arm64/defconfig/gcc-8/lab-cip/baseline-uefi-qemu_arm64-virt-gicv3.html#L191

gctucker:
So I think it's fine to merge this now, there aren't any regressions as far as I can tell.  If an issue appears in production, we can always revert this change and probably work on separating QEMU versions at the configuration level.

PR #390, opened 2020-05-29
T21524 drop boot emails

gctucker:
Clean-up `kernel-arch-complete.sh` and replace all remaining boot email reports with baseline equivalents.  Also add `kernelci-results@groups.io` to the list of recipients for public results.


gctucker:
Tested OK on staging.

PR #391, opened 2020-05-29
T21524 drop boot tests

gctucker:
Drop "boot" tests everywhere now that "baseline" is ready to replace it.  This also means replacing "boot-nfs" with "baseline-nfs" as a new test plan, and dropping the "simple" test plan which doesn't add any coverage compared to "baseline".


gctucker:
Tested OK on staging.

PR #392, opened 2020-06-02
test-configs: add D2500CC

montjoie:
D2500CC is an x86_64 board with an "embedded" intel CPU.
This board lacks PXE, so it is booted via a fake GRUB.
It is a bash init script which emulates GRUB commands and then kexec the final kernel.

### DIFF
--- a/test-configs.yaml
+++ b/test-configs.yaml
@@ -448,6 +448,11 @@ device_types:
     filters:
       - blacklist: {defconfig: ['allnoconfig', 'allmodconfig']}
 
+  d2500cc:
+    mach: x86
+    arch: x86_64
+    boot_method: grub
+
   da850-lcdk:
     mach: davinci
     class: arm-dtb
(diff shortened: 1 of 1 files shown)


montjoie:
I forgot to said that the LAVA device-type is upstream https://git.lavasoftware.org/lava/lava/-/merge_requests/1197

test-configs.yaml
```diff
@@ -1511,6 +1516,11 @@ test_configs:
       - boot
       - boot-nfs
 
+  - device_type: d2500cc
+    test_plans:
+      - baseline
+      - boot
```
gctucker, line 1522: Please remove `boot` as it has been dropped now.

gctucker:
Waiting for the next rounds of results from staging to check it's all working as expected.

gctucker:
Tested OK on staging.  The `dmesg` test cases appear to be missing, but I don't think this has anything to do with this PR:

https://staging.kernelci.org/test/plan/id/5eeb506eaa09c38205c68d0e/

PR #394, opened 2020-06-03
test-configs: Add imx6qp-wandboard-revd1

mgrzeschik:
This patch adds support for imx6qp-wandboard-revd1
LAVA device-type is upstreamed: https://git.lavasoftware.org/lava/lava/-/merge_requests/1206

### DIFF
--- a/test-configs.yaml
+++ b/test-configs.yaml
@@ -589,6 +589,18 @@ device_types:
             - 'multi_v5_defconfig'
             - 'allmodconfig'
 
+  imx6qp-wandboard-revd1:
+    mach: imx
+    class: arm-dtb
+    boot_method: barebox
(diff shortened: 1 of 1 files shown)


gctucker:
Replaced by #418.

mgrzeschik:
You are absolutely right, I have added the test configuration. Thanks for pointing it out.

gctucker:
Well I've already created #418 to fix this problem.  Now I've realised this board is in lab-pengutronix which isn't connected to staging.kernelci.org, we said before it was a good idea to do that but we need to create some new API tokens.  Let's chase this up on IRC, it would be much better to verify the board is working well with KernelCI tests before enabling it in production.

PR #398, opened 2020-06-08
build-configs.yaml: Add a branch in my misc tree

broonie:
Add a for-kernelci branch in my misc tree for testing of changes without
putting them into a production branch - this will be infrequently used,
the intention is to allow me to check changes that I'm worried might be
disruptive on a wide range of hardware without publishing them in my normal
development branches where they can cause trouble in trees like -next.

The goal here is to get runtime testing so ideally the built configurations
could be limited to those that will usefully boot, unfortunately we don't
really have a way of specifying that at present.

### DIFF
--- a/build-configs.yaml
+++ b/build-configs.yaml
@@ -25,6 +25,9 @@ trees:
   arnd:
     url: "https://git.kernel.org/pub/scm/linux/kernel/git/arnd/playground.git"
 
+  broonie-misc:
+    url: "https://git.kernel.org/pub/scm/linux/kernel/git/broonie/misc.git"
+
   broonie-regmap:
     url: "https://git.kernel.org/pub/scm/linux/kernel/git/broonie/regmap.git"
 
@@ -528,6 +531,10 @@ build_configs:
   arnd:
     tree: arnd
     branch: 'to-build'
+
+  broonie-misc:
+    tree: broonie-misc
+    branch: 'for-kernelci'
 
   broonie-regmap:
     tree: broonie-regmap


gctucker:
@broonie Using the main defconfigs in each arch is a simple rule to get a very high tests/build ratio.  For example, see the `minimal_variants` used on other branches:
```yaml
  linusw_fixes:
    tree: linusw
    branch: 'fixes'
    variants: *minimal_variants
```
which got this amount of test coverage with only 6 kernel builds (and no allmodconfig):
https://kernelci.org/test/job/linusw/branch/fixes/kernel/v5.7-rc6-9-ge9bdf7e655b9/

broonie:
I'm definitely looking for more than just the defconfigs - especially on Arm the platform specific defconfigs seem to catch corner cases for some of my subsystems.

gctucker:
Until we have an automated way of doing this, we could already take a look at the arm defconfigs that have been used in tests at least once over a period of time.

broonie:
This seems stuck - not sure what you're expecting at this point?

gctucker:
@broonie Either you're OK with the `minimal_variants`, or please provide the list of defconfigs you want to build.

Maybe take a look here to get an idea of which ones are being tested:
https://kernelci.org/test/job/next/branch/master/kernel/next-20200702/plan/baseline/

Anything more advanced that that would need to be done with an API token or directly in the Mongo DB shell.

broonie:
That's more of a blacklist of configurations I don't want (broadly all*config) which I'm not sure how to reasonably write (it looks like filters get applied per architecture which isn't awesome).

gctucker:
See https://github.com/broonie/kernelci-core/pull/1 for a basic set of filters, it will produce 108 builds without the special all*config ones.  It's still more than what is actually being tested in labs but it's a quick way to get something close to it until we have something better.

```
$ ./kci_build list_kernel_configs --kdir=linux --config=broonie-misc | wc
    108     324    3012
```

broonie:
My point is not that I can't work out how to list some configurations but that it seems like more effort than it's worth to do that, especially for something intermittently used.

gctucker:
So, are you happy with the change in the incoming PR from the comment above?

If so please feel free to merge it, it will then appear in this PR.

broonie:
No, ideally we'd get all the configs that can boot - some of the board-specific Arm configs especially are good at turning things up. I'll do something but it's not just a quick tweak.

gctucker:
> That's more of a blacklist of configurations I don't want (broadly all*config) which I'm not sure how to reasonably write (it looks like filters get applied per architecture which isn't awesome).

Sorry I thought you wanted a filter to not build the all*config, which is what the incoming PR does.

I'll leave it up to you then.

broonie:
Ah, sorry - when I looked at in my e-mail that looked like a pointer to your first comment on here rather than an incremental pull request (the github email notifications aren't great and I've not seen that style before). My mistake. That's 90% of the way there, just needs a few more configs adding I think. How about we just go with what you've got now and I'll come back later and add in the extra configs I'm missing?

gctucker:
Right :)  Yes that sounds good.  It's going to be a moving target anyway, so it will need to be adjusted manually with the current scheme.

Feel free to merge the PR I made on top of your branch and the extra patch should appear here.

PR #404, opened 2020-06-17
Added install commands that I had to execute for "make dt-binding-check" to work

isaiahduncan:
I was having trouble making the "make dt-binding-check" command properly work so I tried using this docker file as suggested. It turned out that the docker file was broken and I believe what I added will fix it so that this problem doesn't occur for others.


broonie:
LGTM

jenkins/dockerfiles/dt-validation/Dockerfile
```diff
@@ -9,4 +9,10 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
 
 RUN update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-8 500
 
-RUN pip3 install git+https://github.com/devicetree-org/dt-schema.git@master
+RUN sudo apt install flex bison
+RUN sudo apt install pkg-config python-yaml device-tree-compiler
+RUN sudo apt install libyaml-dev
```
gctucker, line 14: Please always do `apt-get update && apt-get install` in Dockerfiles, in the same `RUN` command.  This is because Docker will be caching layers based on each command, and if you have the list of packages in a separate command it won't run the "update" command which can cause some issues if the index of packages has changed.

See this example:
https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#sort-multi-line-arguments

Also I think it's best to list all the packages alphabetically like in that example, with `\` to have one package per line but still only one actual 'RUN' command.  This is also what was done with the other `RUN` command on line 3.  I think however that it makes sense to keep a separate command for these packages, to clearly show they are dependencies of `dt-schema` (i.e. if these change, Docker won't have to re-install gcc-8 and vise-versa).

jenkins/dockerfiles/dt-validation/Dockerfile
```diff
@@ -9,4 +9,10 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
...
+RUN make defconfig
```
gctucker, line 15: There is no Linux kernel source tree created in this Dockerfile, so I think this should be left out and run by the test steps.

jenkins/dockerfiles/dt-validation/Dockerfile
```diff
@@ -9,4 +9,10 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
...
+
+RUN pip3 install git+https://github.com/devicetree-org/dt-schema.git@master 
+RUN export PATH=$HOME/.local/bin/:$PATH
```
gctucker, line 18: It's probably a bit better to install the package globally as the tools will then be in the standard path.  This Dockerfile is all about instally the dt-schema tools, so it's expected.

See this `--system` option in `pip3 install --help`:
```
  --system                    Install using the system scheme (overrides --user on Debian systems)
```

jenkins/dockerfiles/dt-validation/Dockerfile
```diff
@@ -9,4 +9,10 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
 
 RUN update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-8 500
 
-RUN pip3 install git+https://github.com/devicetree-org/dt-schema.git@master
```
mgalka, line 12: Is it necessary to move this `install` command to the bottom?
It may result in unnecessary layer creation during image update from older version

jenkins/dockerfiles/dt-validation/Dockerfile
mgalka: +1

isaiahduncan:
Changes made in the last 2 commits: 
- Alphabetized installs and put them all into the same "RUN" command as suggested by gtucker
- Moved the pip3 install command up from the bottom of the file as suggested by mgalka
- Added --system to the pip3 install command to install the package globally as suggested by gtucker

jenkins/dockerfiles/dt-validation/Dockerfile
```diff
@@ -9,4 +9,15 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
...
+RUN sudo apt-get update && install \
+    bison \ 
+    device-tree-compiler \
+    flex \
+    libyaml-dev \
+    pkg-config \
+    python-yaml
+    
+ 
+RUN export PATH=$HOME/.local/bin/:$PATH
```
gctucker, line 23: Is this still needed now that the pip3 package is installed system-wide?

jenkins/dockerfiles/dt-validation/Dockerfile
```diff
@@ -9,4 +9,15 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
 
 RUN update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-8 500
 
-RUN pip3 install git+https://github.com/devicetree-org/dt-schema.git@master
+RUN pip3 install --system git+https://github.com/devicetree-org/dt-schema.git@master
+
+RUN sudo apt-get update && install \
```
gctucker, line 14:
```suggestion
RUN apt-get update && apt-get install \
```

gctucker: No need to run `sudo` as this is already run as `root` in the Docker container, but `apt-get` needs to be run as `install` is not a command on its own.

gctucker: Also `apt-install -y` is needed to run non-interactively and `--no-install-recommends` is to avoid installing a bunch of things not explicitly required in the Dockerfile.

jenkins/dockerfiles/dt-validation/Dockerfile
isaiahduncan: Right. I'll take that out!

jenkins/dockerfiles/dt-validation/Dockerfile
```diff
@@ -9,4 +9,13 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
...
+RUN pip3 install --system git+https://github.com/devicetree-org/dt-schema.git@master
+
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    bison \ 
+    device-tree-compiler \
+    flex \
+    libyaml-dev \
+    pkg-config \
+    python-yaml
+
```
gctucker, line 21: I think there's a trailing blank line here, which causes noise in the git diff.

gctucker:
@isaiahduncan Can you please remove the trailing blank line and confirm this Dockerfile was used when you produced the results shown on #397 ?  Then I think this will be ready to be merged.

PR #409, opened 2020-06-25
Add kci_data commandline tool

mgalka:
kci_data is a tool to feed test data to different types of result
storages. This commit adds basic functionality of the tool that allows
to submit data to /test API of kernelci-backend.

### DIFF
--- a/kci_data
+++ b/kci_data
@@ -1,0 +1,56 @@
+#!/usr/bin/env python3
+#
+# Copyright (C) 2020 Collabora Limited
+# Author: Michal Galka <michal.galka@collabora.com>
+#
+# This module is free software; you can redistribute it and/or modify it under
+# the terms of the GNU Lesser General Public License as published by the Free
+# Software Foundation; either version 2.1 of the License, or (at your option)
+# any later version.
+#
+# This library is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public License for more
+# details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this library; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+
+import sys
+
+import kernelci.config.data
+from kernelci.data import submit
+from kernelci.cli import Args, Command, parse_args
+
+
+class cmd_list_configs(Command):
+    help = "List all database configurations"
+
+    def __call__(self, config_data, *args, **kwargs):
+        db_configs = config_data['db_configs']
+        for config_name in db_configs:
+            print(config_name)
+        return True
+
+
+class cmd_submit(Command):
+    help = "Submit data to the specified database"
+    args = [Args.config, Args.data_file]
+    opt_args = [Args.token]
+
+    def __call__(self, config_data, args):
+        config = config_data['db_configs'][args.config]
+        if args.data_file == '-':
+            data = sys.stdin.read()
+        else:
+            with open(args.data_file, 'r') as f:
+                data = f.read()
+        return submit(args.config, config, data, args.token)
+
+
+if __name__ == '__main__':
+    args = parse_args("kci_data", "db-configs.yaml", globals())
+    configs = kernelci.config.data.from_yaml(args.yaml_configs)
+    status = args.func(configs, args)
+    sys.exit(0 if status is True else 1)
--- a/kernelci/cli.py
+++ b/kernelci/cli.py
@@ -262,6 +262,11 @@ class Args(object):
         'action': 'store_true'
     }
 
+    data_file = {
+        'name': '--data-file',
+        'help': "Path to the file with data to be submitted to storage",
+    }
+
 
 class Command(object):
     """A command helper class.
--- a/kernelci/config/data.py
+++ b/kernelci/config/data.py
@@ -1,0 +1,91 @@
+# Copyright (C) 2020 Collabora Limited
+# Author: Michal Galka <michal.galka@collabora.com>
+#
+# This module is free software; you can redistribute it and/or modify it under
+# the terms of the GNU Lesser General Public License as published by the Free
+# Software Foundation; either version 2.1 of the License, or (at your option)
+# any later version.
+#
+# This library is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public License for more
+# details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this library; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+
+import yaml
+
+from kernelci.config import YAMLObject
+
+
+class Database(YAMLObject):
+    def __init__(self, name, db_type):
+        self._name = name
+        self._db_type = db_type
+
+    @classmethod
+    def from_yaml(cls, db, kw):
+        return cls(**kw)
+
+    @property
+    def name(self):
+        return self._name
+
+    @property
+    def db_type(self):
+        return self._db_type
+
+
+class Backend(Database):
+    def __init__(self, name, db_type, url):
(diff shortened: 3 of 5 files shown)


kci_data
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env python3
+#
+# Copyright (C) 2018, 2019 Collabora Limited
```
gctucker, line 3:
```suggestion
# Copyright (C) 2020 Collabora Limited
```

kernelci/data.py
```diff
@@ -0,0 +1,21 @@
+import json
```
gctucker, line 1: Missing copyright and license statement.

kernelci/config/data.py
```diff
@@ -0,0 +1,74 @@
+import yaml
```
gctucker, line 1: Missing copyright and license statement.

kci_data
```diff
@@ -0,0 +1,59 @@
...
+
+import sys
+
+import kernelci.config.data
+from kernelci.data import submit
+from kernelci.cli import Args, Command, parse_args
+
+
+class cmd_list_configs(Command):
+    help = "list all data endpoint names"
```
gctucker, line 28:
```suggestion
    help = "List all database configurations"
```

kci_data
```diff
@@ -0,0 +1,59 @@
...
+
+    def __call__(self, config_data, *args, **kwargs):
+        db_configs = config_data['db_configs']
+        for config_name in db_configs:
+            print(config_name)
+        return True
+
+
+class cmd_submit(Command):
+    help = "Submit data to specified storage"
```
gctucker, line 38:
```suggestion
    help = "Submit data to the specified database"
```

gctucker: "storage" sounds more like files or objects, we're dealing with databases here.

kernelci/cli.py
```diff
@@ -262,6 +262,18 @@ class Args(object):
...
+        'name': '--data',
+        'help': "Data to be submitted to storage",
+        'required': False
+    }
+
+    data_file = {
+        'name': '--data-file',
+        'help': "Path to the file with data to be submitted to storage",
+        'required': False
+    }
```
gctucker, line 275: How about just having `--data-file` and reading from `stdin` when using `-`?  I think that would work a bit better than having data part of the command line, so tools that output some test results can directly be piped into kci_data.

Reading from a file is the main use-case so this version is already fine though.

spbnick: Seconding this. Actually, why not have a positional argument for this instead of the option?

I'm not sure if `--data` is required either. If you're using a shell, it's easy enough to submit the data inline with just `kci_data --data-file - <<<'{"foo": "bar"}'`, or better yet just `kci_data <<<'{"foo": "bar"}'`

Or am I missing some use case?

gctucker: I did think of a positional argument, but all the other kci_* command only have options so I thought it was probably best to be consistent.  Maybe we could change that as a follow-up, but being explicit is good for a first stab.

And yes using stdin is useful to do quick experiments on the command line.  It can also be useful for test suites that produce compatible data on their stdout, something like:
```
some-test-suite | kci_data --data-file -
```
But I guess many tests would need to first have the data stored in a file.  Also kci_data is likely to have some sub-commands, and the config needs to be passed as an argument to tell where the data is being sent to.  So a real-world command line would actually look more like this:
```
some-test-suite --output=results.json
kci_data submit --config=api.kernelci.org --token=SECRET --data-file=results.json
```
Past this point, an etherpad probably makes more sense ;)

mgalka: I created mandatory `--data-file` argument to stay consistent with other `kci_*` tools. It expects a path to the data file or `-` to parse `stdin`.

kci_data
mgalka: Good point. corrected

kci_data
mgalka: Corrected

kernelci/config/data.py
mgalka: Corrected

kernelci/data.py
mgalka: Corrected

kci_data
mgalka: Corrected

kernelci/cli.py
```diff
@@ -262,6 +262,12 @@ class Args(object):
         'action': 'store_true'
     }
 
+    data_file = {
+        'name': '--data-file',
+        'help': "Path to the file with data to be submitted to storage",
+        'required': False
```
gctucker, line 268: If it's required, than this line should be dropped.

mgalka: Nice catch. Corrected.

khilman:
Could you add attach/link an example data file and add an example cmdline for how you are testing this?

mgalka:
Steps to test the tool

1. Add your backend config to `db-configs.yaml`
e.g.
```yaml
db_configs:
  local-kernelci-org:
    db_type: kernelci_backend
    url: http://kernelci:5001/test

```

2. Prepare test data JSON. 
You can reuse the JSON used in the example usage of the /test API
https://github.com/kernelci/kernelci-doc/wiki/Test-API#example

3. Call `kci_data` to feed your database with test JSON.

```
./kci_data  submit  --token API_TOKEN --config local-kernelci-org --data-file /path/to/test_data.json
```

or

```
cat /path/to/test_data.json | ./kci_data  submit  --token API_TOKEN --config local-kernelci-org --data-file -
```

gctucker:
Tested OK locally:
```
$ ./kci_data submit --config=local --token=XXXXXXXX-XXXX-XXXX-XXXXXXXXXXXXXXXXX --data-file=results.json
Status: 201
Status: 201
{"reason":"Test group 'v4l2-compliance-fake' created","code":201,"result":[{"_id":{"$oid":"5f0eecd3d27138a08f81ffae"}}]}
```
with this dummy data in `results.json`:
```
{
    "board": "qemu",
    "build_environment": "gcc-8",
    "git_commit": "42a5a774ef093e1ebbe4df290405900c7debe7af",
    "name": "v4l2-compliance-fake",
    "git_branch": "kernelci-local",
    "job": "gtucker",
    "arch": "x86_64",
    "kernel" : "kernelci-local-snapshot-032",
    "defconfig": "x86_64_defconfig",
    "defconfig_full": "x86_64_defconfig",
    "lab_name": "gtucker-lava-testing",
    "test_cases": [
        {
            "name": "FAKE_VIDIOC_QUERYCAP",
            "time": 1.0,
            "status": "PASS"
        },
        {
            "name": "FAKE_MC-information-see-Media-Driver-Info-above",
            "status": "PASS"
        },
        {
            "name": "USERPTR-select",
            "time": 1.345,
            "status": "FAIL"
        }
    ],
    "time": 1.0,
    "log": "Fake LOG\nJust some text\nNo parsing"
}
```

gctucker:
Tested OK again but with staging this time:
```
$ ./kci_data submit --config=staging.kernelci.org --token=XXXXXXXX-XXXX-XXXX-XXXXXXXXXXXXXXXXX --data-file=results-staging.json
Status: 201
{"code":201,"result":[{"_id":{"$oid":"5f0eed8f62e0a05495a3d9de"}}],"reason":"Test group 'v4l2-compliance-fake' created"}
```
and this fake data in `results-staging.json`:
```
{
    "board": "qemu",
    "build_environment": "gcc-8",
    "git_commit": "42a5a774ef093e1ebbe4df290405900c7debe7af",
    "name": "v4l2-compliance-fake",
    "git_branch": "staging.kernelci.org",
    "job": "kernelci",
    "arch": "x86_64",
    "kernel" : "staging-20200715.1",
    "defconfig": "x86_64_defconfig",
    "defconfig_full": "x86_64_defconfig",
    "lab_name": "gtucker-lava-testing",
    "test_cases": [
        {
            "name": "FAKE_VIDIOC_QUERYCAP",
            "time": 1.0,
            "status": "PASS"
        },
        {
            "name": "FAKE_MC-information-see-Media-Driver-Info-above",
            "status": "PASS"
        },
        {
            "name": "USERPTR-select",
            "time": 1.345,
            "status": "FAIL"
        }
    ],
    "time": 1.0,
    "log": "Fake LOG\nJust some text\nNo parsing"
}
```
However the results don't quite appear as expected on the frontend:
https://staging.kernelci.org/test/job/kernelci/branch/staging.kernelci.org/kernel/staging-20200715.1/plan/v4l2-compliance-fake/

but this isn't an issue with `kci_data`, more something to do with the backend API that needs to be cleaned up.

PR #414, opened 2020-07-02
kci_rootfs: add debian riscv64 support for sid / debian-ports

khilman:
riscv64 support is available in debian, but only in unstable/sid.  It
also requires using debian-ports.

In order to use debian-ports, the debian mirror needs to be a config
parameter.

### DIFF
--- a/rootfs-configs.yaml
+++ b/rootfs-configs.yaml
@@ -11,7 +11,7 @@ rootfs_configs:
       - mips
       - mipsel
       - mips64el
-    extra_packages_remove:
+    extra_packages_remove: &extra_packages_remove_buster
       - bash
       - e2fslibs
       - e2fsprogs
@@ -23,7 +23,7 @@ rootfs_configs:
       - libp11-kit0
       - libunistring2
       - sensible-utils
-    extra_files_remove:
+    extra_files_remove: &extra_files_remove_buster
       - '*networkd*'
       - '*resolved*'
       - tar
@@ -107,3 +107,24 @@ rootfs_configs:
       - libext2fs2
     script: "scripts/buster-cros-ec-tests.sh"
     test_overlay: ""
+
+  sid:
+    rootfs_type: debos
+    debian_release: sid
+    debian_mirror: http://deb.debian.org/debian-ports
+    keyring_package: debian-ports-archive-keyring
+    keyring_file: /usr/share/keyrings/debian-ports-archive-keyring.gpg
+    arch_list:
+      - riscv64
+    extra_packages_remove: &extra_packages_remove_sid
+      - bash
+      - e2fslibs
+      - e2fsprogs
+      - klibc-utils
+      - libext2fs2
+      - libgnutls30
+      - libklibc
+      - libncursesw6
+      - libp11-kit0
+      - sensible-utils
+    extra_files_remove: *extra_files_remove_buster
--- a/jenkins/debian/debos/rootfs.yaml
+++ b/jenkins/debian/debos/rootfs.yaml
@@ -7,6 +7,9 @@
 {{- $extra_packages_remove := or .extra_packages_remove -}}
 {{- $extra_files_remove := or .extra_files_remove -}}
 {{- $crush_image_options := or .crush_image_options "" -}}
+{{- $debian_mirror := or .debian_mirror "http://deb.debian.org/debian" -}}
+{{- $keyring_package := or .keyring_package "" -}}
+{{- $keyring_file := or .keyring_file "" -}}
 
 architecture: {{ $architecture }}
 
@@ -15,8 +18,10 @@ actions:
     suite: {{ $suite }}
     components:
       - main
-    mirror: http://deb.debian.org/debian
+    mirror: {{ $debian_mirror }}
     variant: minbase
+    keyring-package: {{ $keyring_package }}
+    keyring-file: {{ $keyring_file }}
 
   - action: apt
     recommends: false
--- a/kci_rootfs
+++ b/kci_rootfs
@@ -46,6 +46,9 @@ class cmd_validate(Command):
             print('\ttest_overlay: {}'.format(config.test_overlay))
             print('\tcrush_image_options: {}'
                   .format(config.crush_image_options))
+            print('\tdebian_mirror: {}'.format(config.debian_mirror))
+            print('\tkeyring_package: {}'.format(config.keyring_package))
+            print('\tkeyring_file: {}'.format(config.keyring_file))
         return True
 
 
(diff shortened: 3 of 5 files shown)


khilman:
@Lakshmipathi @mgalka I was able to add `debian_mirror` as a config parameter, but the same needs to be done for the keyring, since a different mirror (in this case, debian-ports) is going to have a different keyring.  Since I'm not familiar with debos, could you help me figure out how to add a custom keyring that gets passed to debos?

Lakshmipathi:
Hi @khilman, Let me check about keyring with debos and update.

Lakshmipathi:
@khilman debos has `keyring-package` and `keyring-file` options as part of debootstrap action  https://godoc.org/github.com/go-debos/debos/actions#hdr-Debootstrap_Action

khilman:
Thanks @Lakshmipathi , I updated with support for a custom keyring now.  

Tested with `./kci_rootfs build --arch riscv64 --config sid --data-path jenkins/debian/debos`

khilman:
Also, I don't quite follow how these builds are triggered/automated in Jenkins.  @Lakshmipathi could you trigger a test build in staging for this?

Lakshmipathi:
@khilman ,  On `rootfs-builder` job pass the custom options for repo, branch, rootfs_arch and rootfs_config
https://bot.staging.kernelci.org/job/rootfs-builder/23/parameters/

The job fails with message https://bot.staging.kernelci.org/job/rootfs-builder/23/console
`Debootstrap | E: specified keyring file (/usr/share/keyrings/debian-ports-archive-keyring.gpg) not found`

We currently have two options for different requirement.  `test_overlay` - provide ability to add files to final rootfs image.
`script` - provides option to execute custom scripts inside rootfs image.

Here the requirement is to set-up environment which runs the debos command.  May be keep keyring inside the `jenkins/debian/debos/` will work here @gtucker, @mgalka ?

gctucker:
@khilman About Jenkins integration, it's similar to the kernel builds now: the rootfs-trigger job looks at all the rootfs variatns that need to be built (distros, variants with added test suites, architectures...) then schedules one rootfs-builder job for each rootfs to be built.  The rootfs-trigger job is not currently started automatically though, we could have a timer but it hasn't been really needed until now and would require a fair amount of extra CPU usage and disk storage.  I think that doing it once a week automatically would be good though, since that's the frequency at which they are normally being updated in production.

Lakshmipathi:
thanks @gtucker for the pointer.  Installed `debian-ports-archive-keyring` on docker debos image now build passed. (For testing I used laks/staging-debos)

https://bot.staging.kernelci.org/job/rootfs-builder/24/console
https://bot.staging.kernelci.org/job/rootfs-builder/24/parameters/

gctucker:
Yes the rootfs images can be found here: https://storage.staging.kernelci.org/images/rootfs/debian/sid/2020070601/riscv64/ :tada: 

Now it would be really nice to be able to smoke test them, maybe QEMU would be enough to prove they're valid.

montjoie:
I have rerun my qemu job with the rootfs.cpio.gz and got:
/init: error while loading shared libraries: libunistring.so.2: cannot open shared object file: No such file or director

gctucker:
@montjoie Can you try again with the full rootfs image?  I wonder if this file gets removed by accident in the small ramdisk image.

khilman:
I just copied the remove packages/files from buster, and that list has the unistring package removed:
https://github.com/kernelci/kernelci-core/blob/master/rootfs-configs.yaml#L24

I'm not sure how the remove files/packages was decided for buster, but maybe we need the same exercise for sid.  @Lakshmipathi @sjoerd-ccu ?

khilman:
I locally (re)build a sid rootfs without stripping unistring (patch below), and it now boots with qemu-5.0:
`/usr/local/bin/qemu-system-riscv64 -nographic -M virt -smp 8 -m 1G -display none -serial mon:stdio -bios default -kernel ~/work/kernel/linux/build-riscv/arch/riscv/boot/Image -D log -append 'rw console=ttyS0 earlycon=uart8250,mmio,0x10000000' -initrd ~/work/kernel/ci/kernelci-core/jenkins/debian/debos/sid/riscv64/rootfs.cpio.gz`

patch to rootfs-configs:

```
diff --git a/rootfs-configs.yaml b/rootfs-configs.yaml
index 78892b94715d..840a7ecebc1f 100644
--- a/rootfs-configs.yaml
+++ b/rootfs-configs.yaml
@@ -21,7 +21,6 @@ rootfs_configs:
       - libklibc
       - libncursesw6
       - libp11-kit0
-      - libunistring2
       - sensible-utils
     extra_files_remove: &extra_files_remove_buster
       - '*networkd*'
```

rootfs-configs.yaml
```diff
@@ -107,3 +107,14 @@ rootfs_configs:
...
+
+  sid:
+    rootfs_type: debos
+    debian_release: sid
+    debian_mirror: http://deb.debian.org/debian-ports
+    keyring_package: debian-ports-archive-keyring
+    keyring_file: /usr/share/keyrings/debian-ports-archive-keyring.gpg
+    arch_list:
+      - riscv64
+    extra_packages_remove: *extra_packages_remove_buster
```
gctucker, line 119: If we need to remove a different list of packages for `sid` then please make a copy of the list used for `buster` here and adjust it accordingly.  It's likely that this will keep evolving due to the nature of `sid`.  When `bullseye` gets released we'll bump all the rootfs variants to use it and replace the list of packages to remove.

gctucker:
> I just copied the remove packages/files from buster, and that list has the unistring package removed:
> https://github.com/kernelci/kernelci-core/blob/master/rootfs-configs.yaml#L24
> 
> I'm not sure how the remove files/packages was decided for buster, but maybe we need the same exercise for sid. @Lakshmipathi @sjoerd-ccu ?

This was largely trial-and-error, first by removing obvious ones then by manually optimizing the list of packages and files to remove in order to reduce the ramdisk image size as much as possible but still be able to login and run basic tests.

rootfs-configs.yaml
khilman: OK, will update with new list for sid.

khilman:
Updated with custom `extra_packages_remove` for sid.  Tested locally with riscv qemu-5.0 with initrd and `rootfs.cpio.gz`

khilman:
Can someone trigger a Jenkins build?  @gctucker @Lakshmipathi

Lakshmipathi:
Tested with `laks/staging-debos` . The debos dockerfile https://github.com/kernelci/kernelci-core/blob/master/jenkins/dockerfiles/debos/Dockerfile needs to be updated. It should include `debian-ports-archive-keyring` package after that `kernelci/staging-debos` should work too. 

https://bot.staging.kernelci.org/job/rootfs-builder/26/console
https://bot.staging.kernelci.org/job/rootfs-builder/26/parameters/

khilman:
Thanks for testing @Lakshmipathi 

Updated debos Dockerfile here: https://github.com/kernelci/kernelci-core/pull/443

PR #416, opened 2020-07-03
test-configs: Enable clang boot testing

broonie:
Currently we only build test with clang, we don't boot the resulting
binaries as there were some issue with the boot-specific reporting.  As
that is being phased out there is now no reason not to include clang
binaries in testing so do so, existing use of clang by developers and
results on staging do not suggest this is likely to result in many
additional failures.

### DIFF
--- a/test-configs.yaml
+++ b/test-configs.yaml
@@ -66,29 +66,22 @@ test_plan_default_filters:
         - ['i386', 'i386_defconfig']
         - ['x86_64', 'x86_64_defconfig']
 
-  - blacklist: &clang_environment_filter
-      build_environment: ['clang']
-
   - blacklist: &kselftest_defconfig_filter
       defconfig: ['kselftest']
-
-  - blacklist: &kselftest_defconfig_clang_environment_filter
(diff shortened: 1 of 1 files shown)


broonie:
@nickdesaulniers

gctucker:
The blacklist is still used for all the other trees, this is only enabling clang kernels to be tested on linux-next.

However, since clang kernels are only built with next, we should be able to drop it indeed since I don't see a reason for another tree to be built with kernel but not tested.

broonie:
Removed the blacklist

test-configs.yaml
```diff
@@ -66,9 +66,6 @@ test_plan_default_filters:
         - ['i386', 'i386_defconfig']
         - ['x86_64', 'x86_64_defconfig']
 
-  - blacklist: &clang_environment_filter
-      build_environment: ['clang']
-
   - blacklist: &kselftest_defconfig_filter
       defconfig: ['kselftest']
```
gctucker, line 71: Please also remove the entry below `kselftest_defconfig_clang_environment_filter`.

PR #418, opened 2020-07-06
wandboard fixup

gctucker:
Fixes #394.


test-configs.yaml
```diff
@@ -571,6 +571,18 @@ device_types:
             - 'multi_v5_defconfig'
             - 'allmodconfig'
 
+  imx6qp-wandboard-revd1:
```
gctucker, line 574: Tested OK on staging: https://staging.kernelci.org/test/plan/id/5fa0354506fbd0ea810145e5/

PR #421, opened 2020-07-07
build.py: merge_config needs CROSS_COMPILE_COMPAT

khilman:
The merge_config.sh script needs CROSS_COMPILE_COMPAT env set in order
to properly merge configs, otherwise some settings that depend on
that (e.g. VDSO-related) will get disabled during the merge.

### DIFF
--- a/kernelci/build.py
+++ b/kernelci/build.py
@@ -530,9 +530,11 @@ export ARCH={arch}
 export HOSTCC={cc}
 export CC={cc}
 export CROSS_COMPILE={cross}
+export CROSS_COMPILE_COMPAT={cross_compat}
 scripts/kconfig/merge_config.sh -O {output} '{base}' '{frag}' {redir}
 """.format(kdir=kdir, arch=kwargs['arch'], cc=kwargs['cc'],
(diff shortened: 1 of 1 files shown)


broonie:
Makes sense.

khilman:
I forgot to add that without this, builds for certain fragments (e.g. kselftest) are broken since the make step will (re)launch the interactive config, and never get any input.

broonie:
Separately to this we should probably redirect /dev/null to the input to stop it hanging.

khilman:
Mark Brown <notifications@github.com> writes:

> Separately to this we should probably redirect /dev/null to the input to stop it hanging.

I think that could actually mask issues and lead to unexpected results.

In previous iterations of these tools, we did: `yes "" | make oldconfig`
which had a similar effect of just accepting the defaults for any
questions posed by make config.

broonie:
That actually matches my interactive use of this stuff so not unexpected for me but I guess others might work differently.

khilman:
Yeah, that's how I use interactively too.  Anyways, unrelated to this specific PR.  We can fix that with a separate patch if/when needed.

Issue #423, opened 2020-07-09
Please upgrade QEMU/kvm firmware images to latest EDK2 stable tag


ardbiesheuvel:
Linux recently introduced a new protocol for loading the initrd from UEFI firmware, and QEMU has been updated to take advantage of it. Subsequently, the old method of passing it via the command line has been deprecated, which means that kernel configs may decide to disable it.

Since QEMU supports the new protocol, we should switch to it, not only for testing coverage but also because it will be the only method that is guaranteed to be available going forward.

Therefore, please upgrade all firmware images used for QEMU EFI boot to versions that are built from EDK2 git tag 'edk2-stable202005' (this applies to i386, x86_64, ARM and arm64)

gctucker:
The current UEFI binaries<sup>[1]</sup> were supplied to us and we just had to install them on the storage server.  This was the preferred way to ensure they were built according to your expectations.

See the original commit from @roxell before we copied the binaries to the storage server: https://github.com/roxell/kernelci-core/commit/5282995a6abcc50c826826c0e8ffae131eb4b805
and the meeting minutes: https://groups.io/g/kernelci/message/480

Are there new binaries for `edk2-stable202005` available anywhere?  If not, would you prefer to make them for us again or provide some instructions to have them made by KernelCI?

[1] https://storage.kernelci.org/images/uefi/111bbcf87621/

ardbiesheuvel:
If I remember that discussion correctly, you insisted on me providing instructions so you could build your own binaries. 

I can build the binaries for you, or provide instructions again, whichever you prefer.

ardbiesheuvel:
Prebuilt binaries can be found here:
https://people.linaro.org/~ard.biesheuvel/qemu-efi-stable202005.tgz

gctucker:
Thanks!  I'll upload that to the storage server and make a PR to start using them.

Build instructions are really important to be able to reproduce the builds but especially so that anyone looking into the results can find out what is actually being used in the tests.  The meta-data provided with the binaries was good enough in this respect last time.

gctucker:
I've unpacked the new binaries here:
https://storage.kernelci.org/images/uefi/edk2-stable202005/

@ardbiesheuvel What about the "mixed" 32/64 build for x86?  The current binaries include this one:
https://storage.kernelci.org/images/uefi/111bbcf87621/OVMF.fd-IA32X64-RELEASE-111bbcf87621

ardbiesheuvel:
[...]

PR #424, opened 2020-07-09
test-configs.yaml: update UEFI binaries with edk2-stable202005

gctucker:
Update the URLs for the UEFI firmware binaries provided by Ard to use
edk2-stable202005.  Also switch to using the same 32-bit i386 UEFI
binary with the "mixed" variant for x86_64 kernel.

### DIFF
--- a/test-configs.yaml
+++ b/test-configs.yaml
@@ -115,35 +115,35 @@ test_plans:
     rootfs: buildroot_baseline_ramdisk
     pattern: 'baseline/generic-qemu-baseline-template.jinja2'
     params:
-      bios_url: 'http://storage.kernelci.org/images/uefi/111bbcf87621/QEMU_EFI.fd-ARM-RELEASE-111bbcf87621'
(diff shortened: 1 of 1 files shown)


gctucker:
See also the discussion on #423 about the origin of the binaries and the `readme.txt` file.

If the test results are looking fine on staging then this should be good to go.

gctucker:
There is a regression with x86_64 but that seems to be a kernel problem in linux-next:
https://staging.kernelci.org/test/case/id/5f0e4dab648908480635c463/

since the "plain" QEMU job without the specific UEFI binary shows the same regression:
https://staging.kernelci.org/test/case/id/5f0e4da9cdadd65e6935c451/

So I think the binaries are OK, the other architectures are not regressing.

One issue though is that both the 64/64 and 32/64 x86 variants end up in the database as the same result, overwriting each other.  But I believe this was always the case, something worth fixing separately from this PR though.

gctucker:
Created #432 for the issue described above with i386 UEFI and x86_64 kernel.

PR #428, opened 2020-07-13
Rename blacklist/whitelist filters to blocklist/passlist

gctucker:
Replace the blacklist/whitelist filter names with their new standard equivalents blocklist/passlist and update the YAML config files accordingly.


gctucker:
Tested OK on staging.

PR #429, opened 2020-07-13
build-configs: android branch renames

khilman:
Request from Google (Todd Kjos):

> We renamed a couple of Android kernel branches, so the old names
> should be removed from kernelci and the new ones added (the old names
> still work, but will go away in a few weeks). Also, the android-4.19
> branch is deprecated (other android-4.19-* branches should still be
> tested):
>
> Repo: repo: https://android.googlesource.com/kernel/common
> Remove: android-5.4, android-5.4-stable, android-4.19
> Add: android11-5.4, android12-5.4

### DIFF
--- a/build-configs.yaml
+++ b/build-configs.yaml
@@ -471,19 +471,14 @@ build_configs:
     branch: 'android-4.19-stable'
     variants: *android_variants
 
-  android_4.19:
+  android11_5.4:
     tree: android
-    branch: 'android-4.19'
+    branch: 'android11-5.4'
     variants: *android_variants
 
-  android_5.4-stable:
+  android12_5.4:
     tree: android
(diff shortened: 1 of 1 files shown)


build-configs.yaml
```diff
@@ -471,19 +471,14 @@ build_configs:
     branch: 'android-4.19-stable'
     variants: *android_variants
 
-  android_4.19:
+  android11_5.4:
```
gctucker, line 474: The convention for all the other entries is `tree-name_branch-or-short-name`.  So `android_11-5.4` would be more consistent.  Not a big deal though, but I thought it was worth mentioning.

gctucker: Fixed in #434.

PR #430, opened 2020-07-14
Fix lab-configs.yaml by removing simple and kselftest from filters

gctucker:
The "simple" test plan has been dropped, and "kselftest" is not ready to be run yet.  Drop both from filters in lab-configs.yaml to only  run baseline in lab-nxp and lab-theobrama-systems.

### DIFF
--- a/lab-configs.yaml
+++ b/lab-configs.yaml
@@ -28,7 +28,6 @@ labs:
           plan:
             - baseline
             - baseline-nfs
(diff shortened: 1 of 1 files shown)


gctucker:
Tested OK on staging https://staging.kernelci.org/test/job/kernelci/branch/staging.kernelci.org/kernel/staging-20200715.0/

PR #433, opened 2020-07-15
Fix QEMU UEFI to use plain baseline test plan

### LINKED ISSUE #432: baseline-uefi on x86 QEMU results overwrite each other
The `baseline-uefi` test plan runs on QEMU with a particular UEFI binary.  For x86, it runs twice: first with x86_64 kernel and UEFI, then with x86_64 kernel but i386 UEFI.  Since the test plan name is the same `baseline-uefi` and the device type is the same `qemu_x86_64`, the results are stored in the same location in the backend.  It wasn't noticeable until a kernel regression happened, and now we have the same regression tracked twice:

https://staging.kernelci.org/test/plan/id/5f0e4dab648908480635c462/

I believe the best way to solve this given the current scheme of things is to have a separate QEMU device type variant, as changing UEFI from a x86_64 build to an i386 one does mean it is not really the same test platform any more.

gctucker:
Fixes #432, by running QEMU UEFI baseline tests using QEMU device variants rather than a separate test plan.


gctucker:
Tested OK on staging with arm, i386 and x86_64, here are some sample results:
https://staging.kernelci.org/test/plan/id/5f0f9dcee4a7c647ec3aa84e/
https://staging.kernelci.org/test/plan/id/5f0f9dca468ca7b0463aa84e/
https://staging.kernelci.org/test/plan/id/5f0fa0c6e519bf01e03aa88a/

The arm64 defconfig build is currently broken in linux-next so there were no test results.  I'll check the results from the daily mainline build a bit later.

gctucker:
The arm64 defconfig builds got fixed on linux-next, here's a sample test result with UEFI:
https://staging.kernelci.org/test/plan/id/5f14758ec385a7143e940979/

PR #436, opened 2020-07-15
Update baseline-nfs to use debian/buster ramdisk + full rootfs

khilman:


gctucker:
Looks good at first glance.

Please also add the `dmesg.sh` script which will look for kernel errors.  It's in the buildroot baseline overlay directory:
https://github.com/kernelci/buildroot/blob/kernelci/latest/kernelci/baseline/rootfs_overlay/opt/kernelci/dmesg.sh

khilman:
OK, updated to include the a baseline overlay (same contents as buildroot overlay)

gctucker:
I think `buildroot_baseline_nfs` can be dropped from test-configs.yaml now.

khilman:
> I think `buildroot_baseline_nfs` can be dropped from test-configs.yaml now.

OK, looks like it's unused, so I  added one more patch to do that.

khilman:
@gctucker this should be ready to merge now.

PR #438, opened 2020-07-15
kernelci: use Python 3 class and super()

gctucker:
Use the Python 3 class definition style without (object) for base
classes, and plain super() for calling direct parent method.


gctucker:
Tested OK on staging.

PR #443, opened 2020-07-16
dockerfiles/debos: add keyring for debian-ports

khilman:
In order for debootstrap to work from sid/unstable with debian-ports,
we need the keyring debian-ports.


Lakshmipathi:
Looks good :+1:

PR #445, opened 2020-07-17
update kselftest, use Linaro test-defintions

khilman:
Update the kselftest test-plan:

1) use the kselftest tarball that is now part of kernelci builds.
2) use debian NFS root (ramdisk + modules, then pivot-root to NFS)
3) use Linaro test-definitions repo for running kselftests

Sharing test-defintions with LInaro/LKFT will allow better collaboration on keeping kselftests running/reliable.

NOTE: this draft PR is based on staging due to dependencies on existing PRs not yet merged into master.

### DIFF
--- a/lab-configs.yaml
+++ b/lab-configs.yaml
@@ -10,6 +10,7 @@ labs:
           plan:
             - baseline
             - baseline-fastboot
+            - kselftest
 
   lab-broonie:
     lab_type: lava
--- a/templates/kselftest/kselftest.jinja2
+++ b/templates/kselftest/kselftest.jinja2
@@ -1,0 +1,25 @@
+- test:
+    timeout:
+      minutes: 55
+
+    definitions:
+    - from: inline
+      repository:
+        metadata:
+          format: Lava-Test Test Definition 1.0
+          name: timesync-off
+          description: Disable systemd time sync services
+        run:
+          steps:
+          - systemctl stop systemd-timesyncd || true
+      name: timesync-off
+      path: inline/timesync-off.yaml
+
+    - repository: https://github.com/Linaro/test-definitions.git
+      from: git
+      revision: master
+      path: automated/linux/kselftest/kselftest.yaml
+      name: kselftest
+      parameters:
+        TESTPROG_URL: {{ kselftests_url }}
+        SKIPFILE: skipfile-lkft.yaml
--- a/test-configs.yaml
+++ b/test-configs.yaml
@@ -161,9 +161,12 @@ test_plans:
         panfrost_submit
 
   kselftest:
-    rootfs: buildroot_kselftest_ramdisk
+    rootfs: debian_buster_nfs
+    pattern: 'kselftest/{category}-{method}-{protocol}-nfs-kselftest-template.jinja2'
     filters:
       - passlist: {defconfig: ['kselftest']}
+    params:
+      job_timeout: '60'
 
   sleep:
     rootfs: debian_buster_ramdisk
--- a/test-configs.yaml
+++ b/test-configs.yaml
@@ -1731,6 +1731,7 @@ test_configs:
   - device_type: meson-g12b-odroid-n2
     test_plans:
       - baseline
+      - kselftest
 
   - device_type: meson-g12b-a311d-khadas-vim3
     test_plans:
--- a/test-configs.yaml
+++ b/test-configs.yaml
@@ -1421,13 +1421,11 @@ test_configs:
   - device_type: apq8016-sbc
     test_plans:
       - baseline
-      - kselftest
 
   - device_type: ar9331-dpt-module
     test_plans:
       - baseline
       - baseline-nfs
-      - kselftest
       - sleep
 
   - device_type: armada-370-db
@@ -1490,7 +1488,6 @@ test_configs:
   - device_type: arndale
     test_plans:
       - baseline
-      - kselftest
 
   - device_type: at91-sama5d2_xplained
     test_plans:
@@ -1513,24 +1510,20 @@ test_configs:
   - device_type: bcm2837-rpi-3-b
     test_plans:
       - baseline
-      - kselftest
       - usb
 
   - device_type: bcm2837-rpi-3-b-32
     test_plans:
       - baseline
-      - kselftest
 
   - device_type: beaglebone-black
     test_plans:
       - baseline
       - baseline-nfs
-      - kselftest
 
   - device_type: beagle-xm
     test_plans:
       - baseline
-      - kselftest
 
   - device_type: cubietruck
     test_plans:
@@ -1553,54 +1546,44 @@ test_configs:
     test_plans:
       - baseline
       - baseline-nfs
-      - kselftest
       - sleep
 
   - device_type: dra7-evm
     test_plans:
       - baseline
-      - kselftest
 
   - device_type: fsl-ls1012a-rdb
     test_plans:
       - baseline
-      - kselftest
 
   - device_type: fsl-ls1028a-rdb
     test_plans:
       - baseline
-      - kselftest
 
   - device_type: fsl-ls1043a-rdb
     test_plans:
       - baseline
-      - kselftest
 
   - device_type: fsl-ls1046a-rdb
     test_plans:
       - baseline
-      - kselftest
 
   - device_type: fsl-ls1088a-rdb
     test_plans:
       - baseline
-      - kselftest
 
   - device_type: fsl-ls2088a-rdb
     test_plans:
       - baseline
-      - kselftest
 
   - device_type: fsl-lx2160a-rdb
     test_plans:
       - baseline
-      - kselftest
 
   - device_type: hi6220-hikey
     test_plans:
       - baseline
       - baseline-nfs
-      - kselftest
 
   - device_type: hip07-d05
     test_plans:
@@ -1619,35 +1602,30 @@ test_configs:
     test_plans:
       - baseline
       - baseline-nfs
-      - kselftest
       - sleep
 
   - device_type: imx27-phytec-phycard-s-rdk
     test_plans:
       - baseline
       - baseline-nfs
-      - kselftest
       - sleep
 
   - device_type: imx28-duckbill
     test_plans:
       - baseline
       - baseline-nfs
-      - kselftest
       - sleep
 
   - device_type: imx53-qsrb
     test_plans:
       - baseline
       - baseline-nfs
-      - kselftest
       - sleep
 
   - device_type: imx6dl-riotboard
     test_plans:
       - baseline
       - baseline-nfs
-      - kselftest
       - sleep
 
   - device_type: imx6q-nitrogen6x
@@ -1662,14 +1640,12 @@ test_configs:
     test_plans:
       - baseline
       - baseline-nfs
-      - kselftest
       - sleep
 
   - device_type: imx6q-sabrelite
     test_plans:
       - baseline
       - baseline-nfs
-      - kselftest
       - sleep
       - usb
 
@@ -1721,7 +1697,6 @@ test_configs:
     test_plans:
       - baseline
       - baseline-nfs
-      - kselftest
 
   - device_type: kirkwood-db-88f6282
     test_plans:
@@ -1736,7 +1711,6 @@ test_configs:
   - device_type: meson8b-odroidc1
     test_plans:
       - baseline
-      - kselftest
 
   - device_type: meson-g12a-sei510
     test_plans:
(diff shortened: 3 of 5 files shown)


khilman:
FYI: still using `staging-skip` because this depends on the fix here for a working rootfs: https://github.com/kernelci/kernelci-core/pull/463

khilman:
Update includes:
- rootfs image update for: https://github.com/kernelci/kernelci-core/pull/463
- rebase onto latest master

Successful LAVA job: https://lava.baylibre.com/scheduler/job/1757#bottom

Removing `staging-skip` now that rootfs is available.

khilman:
FYI @mgalka

gctucker:
The backend exception has been fixed: https://github.com/kernelci/kernelci-backend/pull/251

templates/kselftest/kselftest.jinja2
```diff
@@ -0,0 +1,25 @@
...
+          description: Disable systemd time sync services
+        run:
+          steps:
+          - systemctl stop systemd-timesyncd || true
+      name: timesync-off
+      path: inline/timesync-off.yaml
+
+    - repository: https://github.com/Linaro/test-definitions.git
+      from: git
+      revision: master
```
gctucker, line 20: One issue with using the head of the master branch is that it makes it harder to reproduce results.  There doesn't seem to be any regular release tags in the `test-definitions` repository so we should either ask for tags to be created or keep a mirror and add KernelCI tags for each production update.

templates/kselftest/kselftest.jinja2
```diff
@@ -0,0 +1,25 @@
...
+      path: automated/linux/kselftest/kselftest.yaml
+      name: kselftest
+      parameters:
+        TESTPROG_URL: {{ kselftests_url }}
+        SKIPFILE: skipfile-lkft.yaml
```
gctucker, line 25: Some skips will be platform-dependent, and there are many more platforms being tested in KernelCI than LKFT.  So we'll need to either have a common file for everyone, or `skipfile-kernelci.yaml` added to the repo, or have a separate file (probably in a fork).

gctucker: Also worth mentioning that KernelCI tracks and reports regressions only, not test cases that have never passed.  So we may be OK without maintaining a full list of tests that won't be fixed, only the broken ones that really don't work or cause some bad side-effects and prevent other tests from running.

test-configs.yaml
```diff
@@ -161,9 +161,12 @@ test_plans:
         panfrost_submit
 
   kselftest:
-    rootfs: buildroot_kselftest_ramdisk
+    rootfs: debian_buster_nfs
+    pattern: 'kselftest/{category}-{method}-{protocol}-nfs-kselftest-template.jinja2'
```
gctucker, line 165: Ah yes that pattern thing should be fixed in the code.  When both a ramdisk and nfs file systems are provided, nfs should take precedence.  It's OK to work around the issue with a special `pattern` here for now.

test-configs.yaml
```diff
@@ -161,9 +161,12 @@ test_plans:
...
     filters:
       - passlist: {defconfig: ['kselftest']}
+    params:
+      job_timeout: '60'
```
gctucker, line 169: This reminds me, we should change the parameter name to make it clear the timeout is a number of minutes...

gctucker (commented):
What about the other template variants though (`barebox`, `grub`)?  Even if they can't all be tested right now, I think they should be updated to also include `kselftest/kselftest.jinja2`.  And the `ipxe` one should probably be dropped altogether as nothing is using that any more.

gctucker:
Other than the comments above, this was tested OK on staging:
https://staging.kernelci.org/test/plan/id/5f45aff802c5b617e7b1458c/

gctucker:
@khilman Please also drop the "WIP" from the PR title if/when you think it's ready to be merged.

kees:
Yay; this looks like great progress! Selftest tarball built, saved, fetched, and run! :)

Next seems to be fixing the execution environment? It seems like something is mangling the log output (and likely interfering with the tests):

```
TAP version 13
13: _ TAP
1..72
: _ 1..72
# selftests lkdtm PANIC.sh
lkdtm: PANIC.sh_ #
# Skipping PANIC crashes entire system
PANIC: crashes_entire #
[SKIP] 1 selftests lkdtm PANIC.sh # SKIP
selftests: lkdtm_PANIC.sh [SKIP]
# selftests lkdtm BUG.sh
lkdtm: BUG.sh_ #
# BUG missing 'kernel BUG at' [FAIL]
missing: 'kernel_BUG #
[FAIL] 2 selftests lkdtm BUG.sh # exit=1
selftests: lkdtm_BUG.sh [FAIL]
```

All the weird "$word: _ $truncated_later_words" lines are coming from somewhere external to the kselftest run. Is that Lava getting in the way? e.g. "13: _ TAP", etc.

And maybe some portion of the test's stdout/stderr is missing? This lkdtm "BUG" test can't actually fail: something central seems to be missing.

gctucker:
@kees Thanks a lot for checking this out!  Yes LAVA adds some noise because it uses the serial console to communicate with the platform.  Then there is some logic in KernelCI backend to reduce the noise but maybe it's not tuned correctly.  Let's dig out the raw log and what the filters produced to see what's going on.

This is actually not a kselftest specific issue, the same logic is used for all LAVA tests.  Maybe it's just more visible with kselftest.

khilman:
> @khilman Please also drop the "WIP" from the PR title if/when you think it's ready to be merged.

Removed the WIP.  I think the issues with the output mangling are not directly related to this PR.  Some tweaking of the test-defintion itself to be sure that the results are parsed from a log file and not from stdout should probably be explored.

khilman:
I wonder if @danrue might have any feedback on the output mangling issues above.  We're using the LKFT LAVA test-definition un-modified.

kees:
Should I open a separate bug for the "log mangling"? I can't quite figure out yet how to reproduce the runtime environment the tests are in. (As in, how do I set up Lava locally to try to debug this myself, etc.)

khilman:
@kees setting up LAVA might be a bit overkill, but what you could try is running the wrapper scripts that the LInaro test-defintion runs.  

Here's the test-definition: https://github.com/Linaro/test-definitions/tree/master/automated/linux/kselftest
and the import part is the shell script that it runs: https://github.com/Linaro/test-definitions/blob/master/automated/linux/kselftest/kselftest.yaml#L50
That's the script that calls `./run_kselftest.sh` but it appears that does some of its own sed magic which helps LAVA parse results.

It does `tee` the output to a logfile, which the `kselftest.sh` script then uses to parse and collect LAVA results (which are then sent to KernelCI), so in theory, even with some mess on stdout, the result parsing should not be affected since it's using the logfile and not stdtout.

gctucker:
@kees @khilman I've captured the raw log produced by LAVA and dumped all the "target" messages in a plain text file:
https://people.collabora.com/~gtucker/tmp/kselftest-raw-lava-target-log.txt

The only difference between that and what is stored by KernelCI is some LAVA signals, here's the diff with what the filters did:
```diff
--- kselftest-raw-lava-target-log.txt	2020-08-28 09:29:10.661301920 +0100
+++ kselftest-meson-g12b-odroid-n2.txt	2020-08-28 09:14:01.000000000 +0100
@@ -1975,17 +1975,14 @@
 + cat uuid
 + UUID=31179_1.6.2.4.1
 + set +x
-<LAVA_SIGNAL_STARTRUN 0_timesync-off 31179_1.6.2.4.1>
 + systemctl stop systemd-timesyncd
 [0;1;31mWarning:[0m The unit file, source configuration file or drop-ins of systemd-timesyncd.service changed on disk. Run 'systemctl daemon-reload' to reload units.
 + set +x
-<LAVA_SIGNAL_ENDRUN 0_timesync-off 31179_1.6.2.4.1>
 + export TESTRUN_ID=1_kselftest
 + cd /lava-31179/0/tests/1_kselftest
 + cat uuid
 + UUID=31179_1.6.2.4.5
 + set +x
-<LAVA_SIGNAL_STARTRUN 1_kselftest 31179_1.6.2.4.5>
 + cd ./automated/linux/kselftest/
 + ./kselftest.sh -t kselftest_armhf.tar.gz -s false -u http://storage.staging.kernelci.org/kernelci/staging.kernelci.org/staging-20200825.2/arm64/defconfig+kselftest/gcc-8/kselftest.tar.gz -L  -S skipfile-lkft.yaml -b  -g  -e  -p /opt/kselftests/mainline/
 INFO: Generating a skipfile based on /lava-31179/0/tests/1_kselftest/automated/linux/kselftest/skipfile-lkft.yaml
```

So if the kselftest log output doesn't look right, it's indeed a problem with how the tests are being run in LAVA.  So it's not a KernelCI issue per se but rather something with either kselftest or the Linaro test definitions.  Let's see if @danrue has some thoughts on that; we can also check what the LKFT logs and results look like.

gctucker:
FYI, some LKFT results and sample log:
https://qa-reports.linaro.org/lkft/linux-stable-rc-5.8-oe/build/v5.8.5/testrun/3125177/suite/kselftest/tests/
https://qa-reports.linaro.org/lkft/linux-stable-rc-5.8-oe/build/v5.8.5/testrun/3125177/suite/kselftest/test/bpf_get_cgroup_id_user/log

Looks similar to what we get (with apparently an extra bug when calling `diff`...):
```
TAP version 13
13: _ TAP
1..71
: _ 1..71
# selftests lkdtm PANIC.sh
lkdtm: PANIC.sh_ #
# Skipping PANIC crashes entire system
PANIC: crashes_entire #
[SKIP] 1 selftests lkdtm PANIC.sh # SKIP
selftests: lkdtm_PANIC.sh [SKIP]
# selftests lkdtm BUG.sh
lkdtm: BUG.sh_ #
# diff unrecognized option '--changed-group-format=%>'
unrecognized: option_'--changed-group-format=%>' #
# BusyBox v1.27.2 (2020-08-04 221300 UTC) multi-call binary.
v1.27.2: (2020-08-04_221300 #
# 
: _ #
# Usage diff [-abBdiNqrTstw] [-L LABEL] [-S FILE] [-U LINES] FILE1 FILE2
diff: [-abBdiNqrTstw]_[-L #
# BUG missing 'kernel BUG at' [FAIL]
```

danrue:
I haven't looked at LKFT's kselftest implementation details in a while, but I know that we often have problems getting results from kselftest runs. The diff thing looks like a busybox vs gnu diff type problem.

lab-configs.yaml
```diff
@@ -10,6 +10,7 @@ labs:
           plan:
             - baseline
             - baseline-fastboot
```
gctucker, line 12: I believe it's worth adding `baseline-nfs` since `kselftest` uses nfsroot, to check it's working with regular kernel configs:
```suggestion
            - baseline-nfs
            - baseline-fastboot
```

khilman: I left that out on purpose because it's an extra test that doesn't add any value, other than sanity checking NFS.  

Since there's no way to check the status of `baseline-nfs` before running `kselftest`, I don't see any extra value here.

gctucker: Well it can be used to verify manually that nfsroot works as expected, when kselftest jobs fail to run.

khilman: Sure, but that's orthogonal to this patch, and should be up to the lab owner to decide.  Let's not hold this series up for unrelated issues.

gctucker: Sure, this is not what's holding this PR. It just seemed like an obvious thing to me when I started enabling kselftests on more devices.

gctucker (approved):
Tested OK in combination with #487, which will need to be rebased and tested again once this is is merged.

templates/kselftest/kselftest.jinja2
gctucker: As discussed, we will need a fork of `test-definitions` for KernelCI in any case to have a staging branch.  We could use that fork with a `kernelci.org` branch used in production too.

test-configs.yaml
gctucker: Fixed in #487.

kees:
> @kees setting up LAVA might be a bit overkill, but what you could try is running the wrapper scripts that the LInaro test-defintion runs.
> 
> Here's the test-definition: https://github.com/Linaro/test-definitions/tree/master/automated/linux/kselftest
> and the import part is the shell script that it runs: https://github.com/Linaro/test-definitions/blob/master/automated/linux/kselftest/kselftest.yaml#L50
> That's the script that calls `./run_kselftest.sh` but it appears that does some of its own sed magic which helps LAVA parse results.

Well this looks like exactly the problem:

```
./run_kselftest.sh 2>&1 | tee "${LOGFILE}" | sed 's/^not ok/[FAIL]/'|sed 's/^ok/[PASS]/'|sed 's/://g'|awk '{if ($0 ~ "# SKIP") {$1 = "[SKIP]"} print $0"\n"$3 ": " $4"_"$5 " " $1}'
```

It assume LAVA doesn't actually parse TAP? Hmmm

gctucker:
Right, so that's an issue with the [Linaro test definitions](https://github.com/Linaro/test-definitions) rather than KernelCI which runs them.

I've created a fork in [kernelci/test-definitions](https://github.com/kernelci/test-definitions) so we can work on things like that, and a first issue to follow up from this PR: https://github.com/kernelci/test-definitions/issues/1

PR #446, opened 2020-07-17
Format DTV Output

isaiahduncan:


dtvout
```diff
@@ -0,0 +1,2693 @@
+  CHKDT   Documentation/devicetree/bindings/extcon/wlf,arizona.yaml
```
gctucker: Please don't include this file, it should be generated when running the test.  Instead we should have some documentation about how to run it, and maybe a snippet from the output to show what it should look like.

format.py
```diff
@@ -0,0 +1,83 @@
+import re
+import sys
+from collections import defaultdict
+
+### BEFORE YOU RUN THIS SCRIPT, READ THE BELOW LINES
```
gctucker: It's good to have some notes about how to use the test.  Having a local `README.md` would be better though, to make it actual documentation.  Comments in the code are good when they really are just about the implementation.

This isn't a blocker for this PR, documentation can be added as a follow-up.

format.py
```diff
@@ -0,0 +1,83 @@
...
+Things Done: 
+	Separated Files with Errors or warnings from the rest
+
+"""
+
+# if len(sys.argv) == 1: #we need an argument to identify the name of the file
+# 	print("no arguments")
+# 	exit()
+
+dtoutput = open("dtvout", "r")# the second argument (aka argv[1]) will be the file that we look at ## sys.argv[1]
```
gctucker: Erm, yes as your comment says it would be better to parse command line arguments.  I think it's fine for this first version to have hard-coded `dtvout` as the file name, but another follow-up would be to clean up the script using a `main()` function and `argparse` to parse the command line.

dtvout
gctucker: Sorry I meant the whole `dtvout` file - it needs to be generated during each test run.  But that's OK, we can keep this as an example and remove it when the test gets automated in KernelCI as we'll then be producing lots of test logs.

PR #451, opened 2020-07-23
kernelci.test: Allow filtering on branches

broonie:
Currently we do not supply the branch name in the filter set we use to
determine which tests to run on a given lab. This is useful for trees
like stable with many branches and currently causes the filters set up
for the LKFT lab which include the branches to not match anything
resulting in no jobs being scheduled there. Add the branch name.

### DIFF
--- a/kernelci/test.py
+++ b/kernelci/test.py
@@ -39,6 +39,7 @@ def match_configs(configs, bmeta, dtbs, lab):
         'kernel': bmeta['git_describe'],
         'build_environment': bmeta['build_environment'],
         'tree': bmeta['job'],
+        'branch': bmeta['git_branch']
     }
 
     flags = {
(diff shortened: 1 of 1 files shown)


broonie:
@danrue This was what your config was running into.

kernelci/test.py
```diff
@@ -39,6 +39,7 @@ def match_configs(configs, bmeta, dtbs, lab):
         'kernel': bmeta['git_describe'],
         'build_environment': bmeta['build_environment'],
         'tree': bmeta['job'],
+        'branch': bmeta['git_branch']
```
gctucker, line 42: Nit: it's better to keep a trailing comma so if we add another line it keeps the diff shorter like in this patch.

gctucker:
```suggestion
        'branch': bmeta['git_branch'],
```

PR #452, opened 2020-07-23
lab-configs: Set priority for jobs submitted to my lab to low

broonie:
Prioritize my own personal use!

### DIFF
--- a/lab-configs.yaml
+++ b/lab-configs.yaml
@@ -14,6 +14,7 @@ labs:
   lab-broonie:
     lab_type: lava
     url: 'https://lava.sirena.org.uk/RPC2/'
+    priority: low
 
   lab-cip:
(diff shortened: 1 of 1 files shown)


gctucker:
What I've found to work well is to use `high` priority for my own jobs, as KernelCI ones run at `medium` priority.  You can also use any number between 0 and 100 now, medium being 50.  But yes this `priority` parameter should also work if that fits better with your workflow.

gctucker:
Tested OK on staging, sample job at low priority: https://lava.sirena.org.uk/scheduler/job/1388

broonie:
My plan is to have both interactive and non-interactive jobs of my own so I was going to set the interactive stuff to high, the non-interactive stuff to medium and push other people's stuff down to low.

PR #453, opened 2020-07-23
lab-configs: Accept all stable trees on LKFT lab

broonie:
Currently the LKFT lab has a specific list of branches that are accepted,
mainly used to filter exactly which stable trees are covered. Since there
have been issues with submitting jobs for a considerable time this now
only covers very old stable branches so let's update to include newer ones.

Do this by replacing the explicit list of branches with just accepting
trees as a whole, comparing the list of stable trees that KernelCI covers
with those covered in LKFT's native testing they look to be the same at
the minute. For the other trees covered, -next and kernelci, all the
branches in those trees are included in the explicit list already so there
should be no change. Doing things by tree will reduce the maintenance
overhead going forwards.

### DIFF
--- a/lab-configs.yaml
+++ b/lab-configs.yaml
@@ -64,23 +64,15 @@ labs:
     url: 'https://lkft.validation.linaro.org/RPC2/'
     priority: low
     filters:
-      - combination:
-          keys: ['tree', 'branch']
-          values:
-            - ['stable', 'linux-4.4.y']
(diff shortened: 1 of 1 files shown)


broonie:
@danrue As discussed on IRC this should add coverage of newer stables - let me know if this is OK or if you'd prefer to keep the specific list of branches to include.

danrue:
:+1: changes look good to me. Thanks @broonie

PR #455, opened 2020-07-24
TEMP: k8s: lower CPU request for 32-CPU VMs

khilman:
Azure nodes with 32 vCPUs refuse jobs with CPU request > 30.  This is
causing all "big" jobs stay stuck in a Pending state.

Support ticket open with Azure.

### DIFF
--- a/templates/k8s/gen.py
+++ b/templates/k8s/gen.py
@@ -42,6 +42,11 @@ if parallel_builds:
 
 if (cpu_limit < 8):
     cpu_request = cpu_limit * 0.875
+# HACK: Azure nodes with 32 vCPUs refuse jobs with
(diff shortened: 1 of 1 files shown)


khilman:
@gctucker still no (useful) response from Azure support on this.  Can we merge this so that at least the `aks-kci-uk-south` cluster can be unstuck?

gctucker:
@khilman Thanks for the update.  I've cherry-picked it on the `kernelci.org` branch which is already used in production, and it seems to be working fine.  I'll merge it later today if we still don't hear anything back as it means the workaround would have to stay for a little while.

gctucker:
OK it sounds like this is going to stay with us for a little while, so merging now until we can revert it.

PR #456, opened 2020-07-27
test-configs: Add ThunderX

broonie:
The Cavium ThunderX is an arm64 server class system, generally booting
using ACPI based firmware.

### DIFF
--- a/test-configs.yaml
+++ b/test-configs.yaml
@@ -2141,6 +2141,10 @@ test_configs:
       - sleep
       - usb
 
+  - device_type: thunderx
+    test_plans:
+      - baseline
+
   - device_type: x86-atom330
     test_plans:
       - baseline
--- a/test-configs.yaml
+++ b/test-configs.yaml
@@ -1316,6 +1316,14 @@ device_types:
     boot_method: depthcharge
     flags: ['lpae']
 
+  thunderx:
+    mach: sbsa
+    arch: arm64
+    boot_method: grub
+    filters:
+      - passlist: {defconfig: ['defconfig']}
+      - blocklist: {kernel: ['v3.']}
+
   x86-atom330:
     mach: x86
     arch: x86


gctucker:
Is there any test lab currently in KernelCI with a ThunderX?

broonie:
LKFT has one.

gctucker:
Great, I see: https://lkft.validation.linaro.org/scheduler/device_type/thunderx

@broonie Please also add an entry in the `test_configs` section to tell which test plans to run on `thunderx`.  Then we should start seeing staging jobs being scheduled on it.

danrue:
Just an FYI - the thunderx in LKFT has not been reliable due to how it's hosted (we don't have api access to the PDU and the bmc gets itself in a sad state apparently), so we're not using it in LKFT in production yet.

broonie:
So long as problems show up as detectable infra errors it shouldn't get in the way.

gctucker:
Are there any false positives on staging, or are any infra errors detected as such?

broonie:
It's detecting infra errors (eg https://lkft.validation.linaro.org/scheduler/job/1643913) - the health check fails sometimes so it doesn't always manage to schedule jobs at all.

gctucker:
And was that detected as such by KernelCI?

broonie:
As far as I can see, yes.

test-configs.yaml
```diff
@@ -2132,6 +2140,11 @@ test_configs:
       - sleep
       - usb
 
+  - device_type: thunderx
+    test_plans:
+      - baseline
+      - kselftest
```
gctucker, line 2146: Please drop `kselftest` from this list, it's not operational.
```suggestion
```

broonie:
It's probably worth adding that default test list...

gctucker:
> It's probably worth adding that default test list...

Feel free :)

PR #457, opened 2020-07-27
kernelci.cli: replace --token with --db-token and --lab-token

gctucker:
Replace Args.token with Args.db_token and Args.lab_token, with
argument names --db-token and --lab-token to avoid any ambiguity.

### DIFF
--- a/doc/kci_build.md
+++ b/doc/kci_build.md
@@ -110,10 +110,10 @@ binaries and also the meta-data with these commands:
 
 ```
 ./kci_build push_kernel \
-  --kdir=linux --api=https://localhost:12345 --token=1234-5678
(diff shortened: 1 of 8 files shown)


khilman:
Can we please keep `--token` as an alias for its previous behavior, at least for a little while.  There's no need to break existing users without a bit of time to migrate.

gctucker:
There is a need, for supporting a settings file with an upcoming PR.  Users can stick to the v1.0 of the `kernelci` Python package until they're ready to use the new version with the renamed arguments.

gctucker:
Tested OK on staging, and agreed this was worth doing with minimal impact on current users.

PR #458, opened 2020-07-27
WIP T22068 settings

gctucker:
Add support for a settings file to provide default values for command line arguments.

### DIFF
--- a/kci_build
+++ b/kci_build
@@ -19,7 +19,7 @@
 
 import sys
 
-from kernelci.cli import Args, Command, parse_args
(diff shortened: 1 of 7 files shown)


gctucker:
Note: while this essentially works it's not complete yet.  There needs to be support for sections that match separate components, for example to hold different `--lab-token` values for each lab.  Also the non-optional CLI arguments cause the parser to report an error even if a value is provided in the settings file - so that needs to be resolved somehow.  Some extra patches are on their way to fix that but it'll take a little longer.

gctucker:
Replacing with a better implementation.

PR #459, opened 2020-07-29
Add Lee's en-route to Mainline repos

lag-linaro:


khilman:
New trees/branches means increased build capacity.   Could you summarize how often these trees are getting updated

Related, are you OK with a reduced set of builds on the new trees?  We have a "minimal" setup which basically that means gcc only and a single defconfig for each arch. (c.f. `minimal_variants` in the the build-configs.yaml.

lag-linaro:
MFD probably gets updated once a week.  Backlight, maybe once a fortnight.

With regards to kernel configuration, defconfigs probably aren't that useful.

It would be better if we could build every file for every appropriate architecture in a given subsystem's directory.

So like `allmodconfig` and `allyesconfig`, but only for `drivers/mfd` and `drivers/video/backlight`.

That should require very little bandwidth, and would also help with other subsystem specific builds you have currently.

Is that something KernelCI can offer presently?

khilman:
I like the idea behind driver/subdir specific builds.  It's not something we do currently, but we would look into.  If you can give the exact make cmdline that gets at what would be most interesting to you, that would help us get started looking at that.

broonie:
You can probably get a long way by having a config that's for example "Enable CONFIG_MFD*" instead of just +CONFIG_SPECIFIC - that's probably easier than trying to only build certain directories and catches issues like failure to link which wouldn't be caught if we only enable certain directories.

lag-linaro:
When I do this locally, I have a script to scrape and enable all of the MFD_* configs listed in `drivers/mfd/Kconfig`.

Then I add `drivers/mfd` directly to the compile line, like:

```
ccache make -f Makefile -j24 W=1 \
ARCH=arm CROSS_COMPILE=arm-linux-gnueabihf- KBUILD_OUTPUT=../builds/build-multi-v7  \
drivers/mfd/
```

I have the same set-up for Backlight and PWM.

As @broonie says, this does not test linking.

gctucker:
It's probably easiest to pick a small arm defconfig (say, `exynos_defconfig`) or even just `multi_v7_defconfig` and define a fragment to enable all the mfd drivers.  Also you can get some runtime test coverage that way, with a complete kernel build.

lag-linaro:
Makes sense to me.

Is there any way to make the fragment dynamic i.e. would pick-up any new {MFD_*, BACKLIGHT_*} symbols as they appear?

gctucker:
It's not possible at the present time but it could be implemented with a regular expression to define the fragment and some logic in Python.  Otherwise, you can list all the configs available across all kernel revisions and only those that exist in the kernel being tested will be enabled, as I believe there's no check to enforce all the configs in a fragment to be actually in the resulting `.config`.  It's done using `merge_config.sh` which will print some warnings but not fail if it can't enable some configs.

There was an idea about rewriting how this is done, to do it in Python rather than with `merge_config.sh`.  That way, we could do more custom things.  But it'll take a bit of work to get it done, and it hasn't been a high priority issue so far.

gctucker:
There's still some value in enabling these trees, I guess someone just needs to make the MFD fragment for them.

aliceinwire:
@lag-linaro please rebase the patches and add make the MFD fragment

gctucker:
Replaced with #1671 and https://github.com/kernelci/kernelci-jenkins/pull/122.

PR #460, opened 2020-07-29
debos: sid: systemd-networking fixups

khilman:
systemd networking appears different on sid and this setup script is
failing.  Add some sanity checks before attempting to start services
that may not be installed on a basic debootstrap'd system.

### DIFF
--- a/jenkins/debian/debos/scripts/setup-networking.sh
+++ b/jenkins/debian/debos/scripts/setup-networking.sh
@@ -3,10 +3,16 @@
 set -e
 
 # Network management
-systemctl enable systemd-networkd
+if [ -e /lib/systemd/system/systemd-networkd.service ]; then
+    systemctl enable systemd-networkd
+fi
 # DNS resolving
-systemctl enable systemd-resolved
+if [ -e /lib/systemd/system/systemd-resolved.service ]; then
+    systemctl enable systemd-resolved
+fi
 # NTP client
-systemctl enable systemd-timesyncd
+if [ -e /lib/systemd/system/systemd-timesyncd.service ]; then
+    systemctl enable systemd-timesyncd
+fi
 
 ln -sf /lib/systemd/resolv.conf /etc/resolv.conf


khilman:
See failed rootfs build: https://bot.kernelci.org/job/rootfs-builder/230/console

Not sure what happened in sid since this was initially added, but `setup-networking.sh` started failing.

I tested this change locally and it builds fine, but I'm not sure this is the right solution because we'll end up with rootfs without networking support.  Maybe we need more systemd packages installed?   @Lakshmipathi @gctucker ?

gctucker:
That's going to be a recurring issue with debian/sid, by nature it's evolving all the time and the rootfs builds will break.  I think that as soon as we get one build to work on debian/sid we should stick to it for as long as we can.  Production updates normally include updating all the Debian rootfs images, but each rootfs has a particular URL in `test-configs.yaml` so we can just leave it unchanged for a while when start using riscv64 in production.

gctucker:
Starting a rootfs build on staging to double check that.  The change in itself makes the script a bit more resilient so I think it's fine.

gctucker:
Well, I think it got garbage collected as many other builds were run after that.  Starting a rootfs-trigger job with the `sid` config.

khilman:
ping

gctucker:
> ping

If you have the time, it would be great to build some Debian rootfs images on staging and do a smoke test run with it.  I guess `baseline-nfs` would be a good one to ensure networking is not broken.

gctucker:
There were issues with building the Buster rootfs images on Monday on production, so this will need a bit of attention.

mgalka:
I created a few buster rootfs'es with changes from this PR

https://storage.staging.kernelci.org/images/rootfs/debian/mgalka/buster/20201015.1/

I ran a few tests that used a buster rootfs for the baseline-nfs test plan.

https://lava.collabora.co.uk/scheduler/job/2733346
https://lava.collabora.co.uk/scheduler/job/2733337

mgalka:
There seem to be a problem with one test job, where auto-login-action times out.

https://lava.collabora.co.uk/scheduler/job/2733342

The problem doesn't occurr on a current buster rootfs. 
I'll investigate it further.

gctucker:
> There seem to be a problem with one test job, where auto-login-action times out.
> 
> https://lava.collabora.co.uk/scheduler/job/2733342
> 
> The problem doesn't occurr on a current buster rootfs.
> I'll investigate it further.

That's a fun one, it suspends before reaching login prompt:
```
<6>[   86.778309] PM: suspend entry (deep)
```
probably something triggered by mistake from user-side.

@khilman Have you seen such things with `sid` images on your side?

khilman:
>
> @khilman Have you seen such things with `sid` images on your side?
>

No, but I've only done minimal testing with riscv for sid.

Kevin

mgalka:
It looks that the suspend during boot issue was a temporary one. I cannot reproduce it anymore:

https://lava.collabora.co.uk/scheduler/job/2741032

gctucker:
OK, I'll merge it after the production update tonight and use new rootfs images from storage.kernelci.org on staging for a week to test it more widely.

gctucker (approved):

PR #461, opened 2020-07-31
setup: add all dependencies to setup.py

makohoek:
When installing kernelci via pip, all the dependencies are
not properly installed:
```
~ ϟ kci_build --help
  File "/home/mkorpershoek/.local/bin/kci_build", line 23, in <module>
    import kernelci.build
  File "/home/mkorpershoek/.local/lib/python3.8/site-packages/kernelci/build.py", line 32, in <module>
    import kernelci.elf
  File "/home/mkorpershoek/.local/lib/python3.8/site-packages/kernelci/elf.py", line 24, in <module>
    import elftools.elf.constants as elfconst
ModuleNotFoundError: No module named 'elftools'
```

```
~ ϟ pip3 show kernelci
Name: kernelci
Version: 1.0
Summary: KernelCI core functions
Home-page: https://github.com/kernelci/kernelci-core
Author: kernelci.org
Author-email: kernelci@groups.io
License: UNKNOWN
Location: /home/mkorpershoek/.local/lib/python3.8/site-packages
Requires: pyyaml, jinja2
Required-by:
```

To fix this, sync back all the dependencies from requirements.txt
into setup.py

Signed-off-by: Mattijs Korpershoek <mkorpershoek@baylibre.com>

Note: for users re-installing via pip, you should consider bumping up the minor version `version='1.1'`, as well, otherwise the package won't be upgraded.

Since I did not know how you release on pip I left that part up to the maintainers.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -50,7 +50,11 @@ setuptools.setup(
     ],
     python_requires='>=3.6',
     install_requires=[
-        "pyyaml",
         "jinja2",
+        "keyring",
+        "pyelftools",
+        "pytest",
+        "pyyaml",
+        "requests",
     ]
 )


gctucker:
Thanks, looks good.  I'll test it locally and indeed the package version should be bumped to 1.1 when uploading to PyPI.

setup.py
```diff
@@ -50,7 +50,11 @@
     ],
     python_requires='>=3.6',
     install_requires=[
-        "pyyaml",
         "jinja2",
+        "keyring",
```
gctucker, line 54: `keyring` actually doesn't appear to be required as it's not used anywhere.  I think it should also be dropped from `requirements.txt`.

gctucker:
OK merging now and removing the `keyring` requirement in a follow-up PR.

makohoek:
Thank you for merging this as-is.

For future reference, the `keyring` removal has been done here:
https://github.com/kernelci/kernelci-core/pull/471

PR #462, opened 2020-08-11
test-configs: add support for qemu RISCV

montjoie:
This patch adds suppport for RISCV qemu virt machine.

The device type is LAVA upstreamed:
https://git.lavasoftware.org/lava/lava/-/merge_requests/1255

But probably only my personal lab (LAVA patched with RISCV patch and recent qemu with riscV support installed) could ran it.
So I have set a flags qemu-riscv64 since qemu docker images are not yet ready.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -1197,6 +1197,22 @@ device_types:
     params: &uefi_i386
       bios_url: 'http://storage.kernelci.org/images/uefi/edk2-stable202005/i386/OVMF.fd'
 
+  qemu_riscv64:
+    base_name: qemu
+    mach: qemu
+    arch: riscv
+    boot_method: qemu
+    context:
+      arch: riscv64
+      cpu: 'rv64'
+      guestfs_interface: 'virtio'
+      machine: 'virt'
+      no_kvm: True
+      memory: 1024
+      extra_options: ['-bios default']
+    filters:
+      - passlist: {defconfig: ['defconfig']}
+
   qemu_x86_64: &qemu_x86_64
     base_name: qemu
     mach: qemu
@@ -2151,6 +2167,10 @@ test_configs:
   - device_type: qemu_i386-uefi
     test_plans:
       - baseline_qemu
+
+  - device_type: qemu_riscv64
+    test_plans:
+      - baseline-qemu-docker
 
   - device_type: qemu_x86_64
     test_plans:
--- a/config/lava/boot/generic-qemu-boot-template.jinja2
+++ b/config/lava/boot/generic-qemu-boot-template.jinja2
@@ -7,6 +7,10 @@
 {% set console_dev = console_dev|default('ttyAMA0') %}
 {% set qemu_binary = 'qemu-system-aarch64' %}
 {% endif %}
+{% if arch == 'riscv' %}
+{% set console_dev = console_dev|default('ttyS0') %}
+{% set qemu_binary = 'qemu-system-riscv64' %}
+{% endif %}
 {% if arch == 'x86_64' %}
 {% set console_dev = console_dev|default('ttyS0') %}
 {% set qemu_binary = 'qemu-system-x86_64' %}


test-configs.yaml
```diff
@@ -1016,6 +1016,23 @@ device_types:
...
+      arch: riscv64
+      cpu: 'rv64'
+      guestfs_interface: 'virtio'
+      machine: 'virt'
+      no_kvm: True
+      memory: 1024
+      extra_options: ['-bios default']
+    filters:
+      - passlist: {defconfig: ['defconfig']}
+    flags: ['qemu-riscv']
```
khilman, line 1034: This flag should probably be `qemu-riscv64` since you use "riscv64" in the rest of the naming.

gctucker: What is this flag for? As far as I can tell, it's not used anywhere.

test-configs.yaml
```diff
@@ -1016,6 +1016,23 @@ device_types:
...
+    mach: qemu
+    arch: riscv
+    boot_method: qemu
+    context:
+      arch: riscv64
+      cpu: 'rv64'
+      guestfs_interface: 'virtio'
+      machine: 'virt'
+      no_kvm: True
+      memory: 1024
```
gctucker, line 1030: These things are usually configured in each lab.  Why is it different for RISC-V?

montjoie: qemu RISCV, does not work at all with kvm enabled ("qemu-system-riscv64: invalid accelerator kvm")

For the flags I believed it was added directly to "job/tags" but seems not in fact.
So I need to find another way.

test-configs.yaml
montjoie: I wrongly understood flags? I have removed it

test-configs.yaml
gctucker: What about the memory size though?

montjoie: I didnt knew that memory was per lab. that's dangerous for reproductability

gctucker: It's set in the device dictionary in LAVA, so each QEMU instance can have a different amount.  And who knows what non-LAVA labs are doing. But then, every physical board may have a different PSU, a different firmware, a different network cable... We can't ensure everything is the same in all labs, but we can detect differences if a test fails in one lab but passes in another one.

gctucker:
@montjoie Please check in the results on staging.kernelci.org to see if this got tested as expected.

montjoie:
I forgot that qemu was private in my lab, so no job for it...
I removed this, so qemu should start to have job now.

montjoie:
Enabling qemu in my lab was interesting, I got some fail from non-riscV qemu (due to qemu too recent ?, debug in progress)
Anyway I got riscV jobs and "defconfig" ones works. Other fail.
So I need to update my PR for filtering defconfigs.

gctucker:
OK, can you please provide some links from staging.kernelci.org with qemu risc-v results?

montjoie:
A working one
https://staging.kernelci.org/test/plan/id/5f3ab2214525dea639355bf4/
And a not working one (due to unbootable defconfig)
https://staging.kernelci.org/test/plan/id/5f3ab1f071f7eb7678355da9/
The nommu defconfig should not have landed since it is not in passlist

montjoie:
I have tried to run the working job in lab-baylibre after our upgrade to 2020.07 (and LAVA riscV patch backported) but buster qemu is too old.
Buster has qemu riscV but didnt know the rv64 CPU.

templates/boot/generic-qemu-boot-template.jinja2
```diff
@@ -19,6 +22,11 @@
 {% endblock %}
 {% block actions %}
 
+{% if arch == 'riscv' %}
+tags:
+    - qemu-riscv64
```
gctucker, line 27: This is breaking things:
```
￼ERROR /data/workspace/bot.staging.kernelci.org/test-runner/workspace/test-runner/jobs/lab-baylibre/kernelci-staging.kernelci.org-staging-20200902.2-riscv-rv32_defconfig-gcc-8-no-dtb-qemu_riscv64-baseline_qemu.yaml: <Fault 400: "Invalid job definition: Device tag 'qemu-riscv64' does not exist in the database..">
```
I don't think using tags is going to work.  We should instead just be filtering which labs get to run RISC-V jobs on QEMU.

gctucker (changes requested):

gctucker (commented):
> Adding support for RISCV qemu is done in 3 part:
> - adding a base docker image for running qemu
> - adding a new testplan for qemu+docker
> - adding the qemu device-type for riscV

Please make 3 commits, it's all entangled at the moment:

* one commit to add the new Dockerfile and also update the script that will build and push the Docker image
* one commit to add support for Docker images in QEMU jobs, using a parameter from test-configs.yaml for the Docker image name rather than hard-coded in the templates
* one commit for the RISC-V specific stuff (or even a separate PR once the basic QEMU Docker support has been merged and tested with more common architectures like x86_64)

gctucker (approved):
Tested OK on staging: https://staging.kernelci.org/test/plan/id/60b903bb3119e59436902fd1/

PR #465, opened 2020-08-12
T23120 config args

gctucker:
This is in preparation for adding support for settings files.  Some commands may be using multiple configurations, such as a database and a build configuration or a test lab.  For this reason, it's necessary to be able to distinguish between them so each config argument should be specific i.e. `--build-config`, `--lab-config`, `--db-config` rather than a generic `--config`.


gctucker:
Tested OK on staging.

PR #467, opened 2020-08-13
kci_rootfs: align implementation of the validate command

gctucker:
Do not print anything by default when running kci_rootfs, only if
--verbose is provided as the main purpose is to validate the YAML
configuration.


gctucker:
Tested OK on staging.

PR #468, opened 2020-08-16
Add rootfs for Linux Test Project (LTP)

Lakshmipathi:
Include build and install script for LTP and update rootfs-config.yaml

### DIFF
--- a/rootfs-configs.yaml
+++ b/rootfs-configs.yaml
@@ -131,3 +131,11 @@ rootfs_configs:
       - libp11-kit0
       - sensible-utils
     extra_files_remove: *extra_files_remove_buster
+
+  buster-ltp:
+    rootfs_type: debos
+    debian_release: buster
+    arch_list:
+      - amd64
+      - arm64
+    script: "scripts/buster-ltp.sh"
--- a/jenkins/debian/debos/scripts/buster-ltp.sh
+++ b/jenkins/debian/debos/scripts/buster-ltp.sh
@@ -1,0 +1,55 @@
+#!/bin/bash
+
+# Important: This script is run under QEMU
+
+set -e
+
+# Build-depends needed to build the test suites, they'll be removed later
+BUILD_DEPS="\
+    gcc \
+    git \
+    make \
+    pkgconf \
+    autoconf \
+    automake \
+    bison \
+    flex \
+    m4 \
+    libc6-dev \
+"
+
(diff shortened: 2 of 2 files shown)


Lakshmipathi:
rootfs job: https://bot.staging.kernelci.org/job/rootfs-builder/135/

gctucker:
I think the first step is to produce a rootfs and have it manually tested in a LAVA lab, to show it can boot and has the test binaries installed.  Then the next step is to add a job template, it may be done in a separate PR.

rootfs-configs.yaml
```diff
@@ -110,7 +110,6 @@ rootfs_configs:
       - libext2fs2
     script: "scripts/buster-cros-ec-tests.sh"
     test_overlay: ""
-
```
gctucker, line 113: Please don't remove blank lines unnecessarily.  It causes noise in the diffs which can lead to git conflicts, and also blank lines help with readability of the file.

Lakshmipathi:
This patch is for rootfs creation, in the process of creating test templates, which will be send as another PR.

gctucker:
@Lakshmipathi please change the commit title to `rootfs-configs.yaml: add debian-ltp` to be clear that this commit and this PR only adds the rootfs images.

Lakshmipathi:
okay changed title. Tested with `arm64` https://bot.staging.kernelci.org/job/rootfs-builder/143/

gctucker:
Great, thanks.  Can you please just run a LAVA job that will boot it using nfsroot and list the contents of the directory where the LTP binaries are installed?

That should be enough to check this rootfs-only PR is valid.  Then the next PR will be about adding the test plan with actual test cases and all that.

Lakshmipathi:
Ok, I ran this job yesterday: https://lava.collabora.co.uk/scheduler/job/2553971/definition It ran a single syscall test.  Does it look good?

gctucker (approved):
Yes, that looks like a good nfsroot boot test.  There were a couple of minor things though in the LTP test case run, not sure they're blockers:
```
/opt/ltp/ver_linux: 115: /opt/ltp/ver_linux: ethtool: not found
```
```
/opt/ltp/ver_linux: 135: /opt/ltp/ver_linux: free: not found
```
I think those can also be addressed in a follow-up as we're likely to be missing other subtle dependencies in the rootfs and we'll find out when running more tests.

PR #471, opened 2020-08-19
setup.py: drop keyring from requirements

gctucker:
Drop they "keyring" package from the list of requirements as it's not
used anywhere.  Update requirements.txt accordingly.


gctucker:
Tested OK locally.

PR #469, opened 2020-08-17
dockerfiles: Add clang-11 docker container

broonie:
The next release of clang is due at the end of August (it's scheduled
for the 26th) so get a docker container ready for it.

Signed-off-by: Mark Brown <broonie@kernel.org>

PR #480, opened 2020-08-27
dockerfiles: add clang-latest

khilman:
Rather than keep versioning this file for clang versions, just build
the latest version from the apt.llvm.org repo.

In addition, this version also install lld and llvm packages so that
kernel builds can be built using 'make LLVM=1'.

The resulting docker image can then be pushed to the docker hub as a
specific version.

For example, as of today, latest version is 12, so this an be built
with the version in the image name:

   docker build -t kernelci/build-clang-12 .

### DIFF
--- a/build-configs.yaml
+++ b/build-configs.yaml
@@ -199,6 +199,7 @@ build_environments:
     cc_version: 10
     arch_map: *clang_arch_map
     cross_compile: *default_cross_compile
+    cross_compile_compat: *default_cross_compile_compat
 
 
 # Default config with full build coverage
--- a/build-configs.yaml
+++ b/build-configs.yaml
@@ -201,6 +201,13 @@ build_environments:
     cross_compile: *default_cross_compile
     cross_compile_compat: *default_cross_compile_compat
 
+  clang-12:
+    cc: clang
+    cc_version: 12
+    arch_map: *clang_arch_map
+    cross_compile: *default_cross_compile
+    cross_compile_compat: *default_cross_compile_compat
+
 
 # Default config with full build coverage
 build_configs_defaults:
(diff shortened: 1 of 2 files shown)


khilman:
Depends on building/pushing updated docker images for clang-12: https://github.com/kernelci/kernelci-core/pull/480

gctucker:
Using `lld` to be able to build with `LLVM=1` is one thing, it's useful because it's now the standard way to build with clang and also it fixes the `merge_config.sh` issue #136.  But that can be done in #469 with clang-11.

Using the latest version of clang is a very different thing though, as it means we're not just testing that kernels build with the stable clang releases but also that the latest clang works.  At the moment, we're not versioning the Docker containers other than the major compiler version (i.e. clang-10, clang-11...) so it will be very difficult to track down issues that are caused by a change in clang.

If we want to do that, we probably should be building both with the stable release _and_ the latest one to compare.  Build failures introduced in clang-latest due to a clang problem would not appear with say, clang-11.  Are we really sure we want to do that?

build-configs.yaml
```diff
@@ -768,6 +776,13 @@ build_configs:
           arm64: *arm64_defconfig
           x86_64: *x86_64_defconfig
 
+      clang-12:
```
gctucker, line 779: If we want to build with the latest clang, we should probably have `clang-latest` instead or just `clang`.  That way, we know when we actually start using clang 12 when it's released.

build-configs.yaml
```diff
@@ -768,6 +776,13 @@ build_configs:
...
+        build_environment: clang-10
```
gctucker, line 780: `clang-10` doesn't seem right, whether it's `clang-12`, `clang`, `clang-latest` or anything else.

gctucker (commented):
Assuming the idea is to keep building with both the current stable release (`clang-10`) _and_ the latest clang snapshot, we probably still should tweak a couple of things as per the review comments.

khilman:
Closing in favor of sticking with released versions.

PR #481, opened 2020-08-27
kci_build: clang builds use LLVM=1

khilman:
If building with clang, also set LLVM=1 which will enable all the
other LLVM tools as well, including linker.  When doing so, do not set
HOSTCC since that is already handled by LLVM=1

Also note that 'make LLVM=1' only works in v5.4+, but currently, we're
only doing clang builds in linux-next, so enabling this as the default
should be fine.

### DIFF
--- a/kernelci/build.py
+++ b/kernelci/build.py
@@ -456,7 +456,10 @@ def _run_make(kdir, arch, target=None, jopt=None, silent=True, cc='gcc',
     if cross_compile_compat:
         args.append('CROSS_COMPILE_COMPAT={}'.format(cross_compile_compat))
 
-    args.append('HOSTCC={}'.format(cc))
+    if cc.startswith('clang'):
+        args.append('LLVM=1')
+    else:
+        args.append('HOSTCC={}'.format(cc))
 
     if use_ccache:
         px = cross_compile if cc == 'gcc' and cross_compile else ''


khilman:
Depends on updated docker images for clang: c.f. https://github.com/kernelci/kernelci-core/pull/480

gctucker:
LGTM, let's see what the results are on staging.

About the kernel version, if we want to build stable branches prior to 5.4 with clang we would probably need to have this set as a parameter of the build variants in `build-configs.yaml`.

gctucker:
Clang builds are currently missing on staging:
https://staging.kernelci.org/build/kernelci/branch/staging.kernelci.org/kernel/staging-20200828.1/

khilman:
Updated to *not* set HOSTCC since that is already taken care of by the LLVM=1 handling.

khilman:
> Clang builds are currently missing on staging:

I'm guessing it's because the docker images for clang were not quite ready for LLVM=1 yet.

gctucker:
For some more background on this topic:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=a0d1c951ef08ed24f35129267e3595d86f57f5d3

gctucker (approved):
Tested OK on staging.

Just worth noting, that this bit isn't strictly required with clang any more since `LLVM=1` will set `CC=clang`:
```python
    elif cc != 'gcc':
        args.append('CC={}'.format(cc))
```
But since it's benign and should only help with any hypothetical new compiler in the future, we can leave it as-is.

PR #482, opened 2020-08-28
dockerfiles: Add lld and llvm to the clang Docker images

broonie:
In order to support testing of LLVM=1 builds include lld and llvm in the
images for the existing versions of clang.

### DIFF
--- a/jenkins/dockerfiles/clang-10/Dockerfile
+++ b/jenkins/dockerfiles/clang-10/Dockerfile
@@ -11,8 +11,8 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
     binutils-aarch64-linux-gnu \
     binutils-arm-linux-gnueabihf \
     binutils \
-    clang-10
+    clang-10 lld-10 llvm-10
 
-RUN update-alternatives --install /usr/bin/clang clang /usr/bin/clang-10 500
+ENV PATH=/usr/lib/llvm-10/bin:${PATH}
 
 RUN apt-get autoremove -y gcc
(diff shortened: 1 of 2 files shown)


gctucker (commented):
Tested OK on staging in combination with #481, the x86_64 build passed but not the arm64 one though:
https://staging.kernelci.org/build/kernelci/branch/staging.kernelci.org/kernel/staging-20200831.1/
https://storage.staging.kernelci.org/kernelci/staging.kernelci.org/staging-20200831.1/x86_64/x86_64_defconfig/clang-9/build.log
https://storage.staging.kernelci.org/kernelci/staging.kernelci.org/staging-20200831.1/arm64/defconfig/clang-9/build.log
```
  LD [M]  drivers/gpu/drm/nouveau/nouveau.o
  MODPOST Module.symvers
ERROR: modpost: "__compiletime_assert_611" [drivers/gpu/drm/vc4/vc4.ko] undefined!
make[2]: *** [/scratch/linux/scripts/Makefile.modpost:111: Module.symvers] Error 1
```
This doesn't seem to be an issue with the containers but rather something to fix in the kernel.  Also, note this was done with clang-9 which is being used on staging.  Maybe we should change that to use clang-10 if it's better supported in linux-next.  Both are built in production at the moment.

broonie:
We should probably just move the clang version forwards unless we can find someone shipping clang-9 in production, the kernel should still not break but newer versions are generally more interesting.

gctucker (approved):
Tested OK on staging.

PR #483, opened 2020-08-28
kernelci.build: make dtbs.json optional in .load_json()

gctucker:
Some platforms don't use any dtb file, and while the dtbs.json file is
normally always generated by the build_kernel() command, it's not
strictly required to keep it and use it again for platforms that don't
need it.  So make dtbs_json an optional argument in .load_json() and
update kci_test accordingly when generating test jobs.


gctucker:
Tested OK on staging.

PR #484, opened 2020-08-28
kernelci.cli: put lab_json option in lab section

gctucker:
Put the lab_json option in the per-lab section in the settings file.
Each lab should have a separate JSON file with a locally saved copy of
its particular data (i.e. online device types...).


gctucker:
Tested OK on staging and local command line.

PR #485, opened 2020-08-28
build-configs.yaml: Set compat cross compilers for clang-10

broonie:
The clang-10 build configuration did not provide compat cross compilers
meaning that we weren't getting coverage of the arm64 compat vDSO. While
clang can build arm without any compat tools the kernel build system
currently requires that a compat cross toolchain be explicitly configured
to enable build of the arm64 compat vDSO.

Reported-by: Kevin Hilman <khilman@baylibre.com>

### DIFF
--- a/build-configs.yaml
+++ b/build-configs.yaml
@@ -199,6 +199,7 @@ build_environments:
     cc_version: 10
     arch_map: *clang_arch_map
     cross_compile: *default_cross_compile
+    cross_compile_compat: *default_cross_compile_compat
 
 
 # Default config with full build coverage


khilman:
Tested locally, and confirmed that `CONFIG_COMPAT_VDSO=y` is now being set.

gctucker:
Fun fact, now that we can build clang with big-endian it looks like there's a vDSO build breakage:
https://storage.staging.kernelci.org/kernelci/staging.kernelci.org/staging-20200901.1/arm64/defconfig+CONFIG_CPU_BIG_ENDIAN=y/clang-10/build.log

I guess this is purely a kernel build problem and this PR is fine, but please confirm.

broonie:
Yes, that looks like a build issue in the stuff under test rather than a KernelCI problem.

gctucker (approved):
OK thanks.

gctucker (approved):
Tested OK on staging.

PR #486, opened 2020-08-31
kernelci.build: use LLVM=1 with merge_config.sh and clang

gctucker:
When using LLVM/Clang, also pass LLVM=1 when calling merge_config.sh
as this gets propagated to make sub-processes unlike CC and HOSTCC.


gctucker:
Tested OK on staging:
https://storage.staging.kernelci.org/kernelci/staging.kernelci.org/staging-20200901.0/x86_64/x86_64_defconfig+x86-chromebook/clang-9/build.log

PR #492, opened 2020-09-03
T21323 kernelci.build: use - instead of / in resource path components

gctucker:
Replace any instance of a slash character "/" in each component making
up the file server resource path with a "-" so they don't end up
creating sub-directories.  This was already done for the
defconfig_full component, but it needs to be done for all of them as
it's a common issue with branch names in particular.

Depends on https://github.com/kernelci/kernelci-backend/pull/254
Depends on https://github.com/kernelci/kernelci-frontend/pull/124


mattface:
You legend!

mattface (approved):
Looks very good to me!

gctucker:
np :) I'm also working on some frontend fixes, and a few tweaks in the backend too.

PR #499, opened 2020-09-16
T23362 docker: add kselftest build dependencies to gcc-8 and clang-10

gctucker:
Add build dependencies for x86, arm and arm64 to the gcc-8 and
clang-10 Docker images.  This causes some duplication, but there is no
easy way around it at the moment.

PR #500, opened 2020-09-16
Add clang-11 build environment

gctucker:
Add build environment definition for clang-11.

Depends on #469.

### DIFF
--- a/build-configs.yaml
+++ b/build-configs.yaml
@@ -201,6 +201,13 @@ build_environments:
     cross_compile: *default_cross_compile
     cross_compile_compat: *default_cross_compile_compat
 
+  clang-11:
+    cc: clang
+    cc_version: 11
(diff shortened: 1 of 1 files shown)


khilman:
Doesn't this need the kselfttest build dependencies too as in: https://github.com/kernelci/kernelci-core/pull/499 ?

gctucker:
Yes but I couldn't do it in #499 because this isn't merged yet.  It could be added as a follow-up PR.

In practice, kselftests aren't built at all with clang anyway at the moment.  So that would need to be updated once the current PRs are merged.

gctucker:
Tested OK on staging:
https://staging.kernelci.org/build/kernelci/branch/staging.kernelci.org/kernel/staging-20200923.0/

There are kernel build errors, but they're not due to this PR.

PR #504, opened 2020-09-18
test-configs: Add imx8mq-zii-ultra-zest

mgrzeschik:
This patch adds support for imx8mq-zii-ultra-zest:
LAVA device-type is upstreamed: https://git.lavasoftware.org/lava/lava/-/merge_requests/????

### DIFF
--- a/test-configs.yaml
+++ b/test-configs.yaml
@@ -679,6 +679,12 @@ device_types:
     dtb: 'freescale/imx8mp-evk.dtb'
     boot_method: uboot
 
+  imx8mq-zii-ultra-zest:
+    mach: imx
+    arch: arm64
+    dtb: 'freescale/imx8mq-zii-ultra-zest.dtb'
+    boot_method: barebox
+
   jetson-tk1:
     mach: tegra
     class: arm-dtb
@@ -1730,6 +1736,10 @@ test_configs:
       - baseline
 
   - device_type: imx8mm-ddr4-evk
+    test_plans:
+      - baseline
+
+  - device_type: imx8mq-zii-ultra-zest
     test_plans:
       - baseline
 


mgrzeschik:
Ok, This board is still not mainline. I forgot about the update on lava side:
https://git.lavasoftware.org/lava/lava/-/merge_requests/1349

I will let you know when it is ready to be merged.

mgrzeschik:
OK, This was fast. Now it was merged! So this is no show stopper any more.

gctucker:
Erm, looks like we didn't quite fix the `lab-pengutronix` setup on staging.kernelci.org... Something to chase up again, it's probably just a glitch in the API tokens.

@mgrzeschik Also, it looks like the only device of this type is not operational yet:
https://hekla.openlab.pengutronix.de/scheduler/device/imx8mq-zii-ultra-zest-0

test-configs.yaml
```diff
@@ -665,6 +665,11 @@ device_types:
     dtb: 'freescale/imx8mp-evk.dtb'
     boot_method: uboot
 
+  imx8mq-zii-ultra-zest:
+    mach: imx
+    class: arm64-dtb
```
gctucker, line 670: The `imx` boards in `arm64` are all under `freescale` and don't follow the convention defined by the `arm64-dtb` class, which uses the mach name (`imx` in this case).  This fixes the prolem:
```suggestion
    arch: arm64
    dtb: 'freescale/imx8mq-zii-ultra-zest.dtb'
```

gctucker: With the change above, I could get a staging job running fine:
https://hekla.openlab.pengutronix.de/scheduler/job/387286
https://staging.kernelci.org/test/plan/id/5fa07d66723bf044100145a8/

test-configs.yaml
```diff
@@ -1715,6 +1720,12 @@ test_configs:
     test_plans:
       - baseline
 
+  - device_type: imx8mq-zii-ultra-zest
+    test_plans:
+      - baseline
+      - baseline-nfs
+      - kselftest
```
gctucker, line 1727: Please drop `kselftest` for now as it's not quite usable. We're still fixing issues with the builds, so tests are not running correctly yet.

About `baseline-nfs`, well it doesn't seem to be working so might be better to leave it out initially:
https://hekla.openlab.pengutronix.de/scheduler/job/387288

gctucker (changes requested):
LGTM except a small issue with the dtb name and some test plans that don't work.

test-configs.yaml
mgrzeschik: baseline-nfs should actually work.
But someshow for the used defconfig the eth0 device does not seem to show up:

[    1.671854] IP-Config: Failed to open eth0
[    1.675980] IP-Config: Device `eth0' not found

test-configs.yaml
```diff
@@ -1733,6 +1738,10 @@ test_configs:
     test_plans:
       - baseline
 
+  - device_type: imx8mq-zii-ultra-zest
```
gctucker, line 1741: It appears to be failing to boot with all the staging jobs:
https://hekla.openlab.pengutronix.de/scheduler/job/392159
The health checks are looking fine though.  Can you please take a look?

test-configs.yaml
```diff
@@ -679,6 +679,11 @@ device_types:
     dtb: 'freescale/imx8mp-evk.dtb'
     boot_method: uboot
 
+  imx8mq-zii-ultra-zest:
+    mach: imx
+    arch: arm64
```
gctucker, line 684: You need to define the `dtb` name, especially as it's in a directory different from the `mach` name:

```suggestion
    arch: arm64
    dtb: freescale/imx8mq-zii-ultra-zest.dtb
```

mgrzeschik: updated!

gctucker: Great, let's see tomorrow if staging jobs worked as expected.

gctucker (approved):
Tested OK on staging:
https://staging.kernelci.org/test/plan/id/5faa51e4114f438902b24628/
and also with Clang:
https://staging.kernelci.org/test/plan/id/5faa52e8d5d4e4929cb24609/

PR #427, opened 2020-07-13
T22500 grunt

gctucker:
Add x86-chromebook fragment and hp-11A-G6-EE-grunt device type which uses it.

This is pending Depthchage changes to be merged upstream for loading a ramdisk over TFTP on x86 platforms, and the device type to be added to LAVA upstream.

LAVA prerequisite changes for x86 Depthcharge: https://git.lavasoftware.org/lava/lava/-/merge_requests/1283
LAVA device type: https://git.lavasoftware.org/lava/lava/-/merge_requests/1395
Depends on #560 and #561 to be merged first.

PR #507, opened 2020-09-21
T23766 octopus

gctucker:
Add the `hp-x360-12b-ca0004na-octopus` device using the `x86-chromebook` config fragment.

Depends on #427.


gctucker:
I'll be making several separate ones instead (enabling the fragment, adding igt-i915...).

PR #511, opened 2020-10-06
Add crypto test plan

montjoie:
This patch adds the first steps to test crypto in kernelCI
First a crypto fragment is added for enabling Linux crypto selftest.
Then a crypto test plan is added.
This test plan is enabled on various device with hardware crypto accelerator.
This test plan is enabled i both lab-baylibre and lab-clabbe where
thoses devices are present.

### DIFF
--- a/build-configs.yaml
+++ b/build-configs.yaml
@@ -139,6 +139,13 @@ trees:
 
 fragments:
 
+  crypto:
+    path: "kernel/configs/crypto.config"
+    configs:
+      - 'CONFIG_CRYPTO_TEST=m'
+      - 'CONFIG_CRYPTO_MANAGER_EXTRA_TESTS=y'
+      - '# CONFIG_CRYPTO_MANAGER_DISABLE_TESTS is not set'
+
   debug:
     path: "kernel/configs/debug.config"
 
--- a/templates/crypto/crypto.jinja2
+++ b/templates/crypto/crypto.jinja2
@@ -1,0 +1,9 @@
+- test:
+    timeout:
+      minutes: 10
+    definitions:
+    - repository: https://github.com/linaro/test-definitions
+      from: git
+      history: False
+      path: automated/linux/crypto/crypto.yaml
+      name: crypto
--- a/lab-configs.yaml
+++ b/lab-configs.yaml
@@ -12,6 +12,7 @@ labs:
             - baseline-fastboot
             - kselftest
             - preempt-rt
+            - crypto
 
   lab-broonie:
     lab_type: lava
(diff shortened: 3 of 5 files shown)


gctucker:
It would have been a bit better to separate this into different commits, but that's not a big issue.

@montjoie Can you please look on staging.kernelci.org or in lab-baylibre to see if tests have been run as expected?

build-configs.yaml
```diff
@@ -139,6 +139,12 @@ trees:
 
 fragments:
 
+  crypto:
+    path: "kernel/configs/crypto.config"
+    configs:
+      - 'CONFIG_CRYPTO_TEST=m'
+      - 'CONFIG_CRYPTO_MANAGER_EXTRA_TESTS=y'
+      - '# CONFIG_CRYPTO_MANAGER_DISABLE_TESTS is not set'
   debug:
```
gctucker, line 148: Please leave a blank line to make it easier to read and to keep the formatting consistent.

build-configs.yaml
```diff
@@ -139,6 +139,12 @@ trees:
 
 fragments:
 
+  crypto:
+    path: "kernel/configs/crypto.config"
+    configs:
+      - 'CONFIG_CRYPTO_TEST=m'
```
gctucker, line 145: Is there any advantage in having this built as a module rather than built-in?

montjoie: It is a symbol which can only be set as module.

montjoie:
On my lab the run is perfect. On lab-baylibre, some worker was too loaded and jobs timeouted in deploy.
But this is unrelated to crypto. I have reduced number of job in paralell and I will send  a Pr for upgrade timeout in deploy.
But generally it miss a try with the crypto fragment.

templates/crypto/crypto.jinja2
```diff
@@ -0,0 +1,9 @@
+- test:
+    timeout:
+      minutes: 10
+    definitions:
+    - repository: https://github.com/linaro/test-definitions
```
gctucker, line 5: Please use the `kernelci` repository, which tracks the `linaro` upstream:
https://github.com/kernelci/test-definitions.git
with the `kernelci.org` branch.

gctucker:
See the latest staging-next results: https://staging.kernelci.org/test/job/kernelci/branch/staging-next/kernel/staging-next-20201103.0/plan/crypto/
Several devices don't appear there, can you please check what is going on?

If the jobs are running but hitting a timeout, it should be fine to just extend the timeout in the job template.

If some device types are failing to run the job then it might be better to drop them from this PR and add them as separate PRs when they're working.

gctucker:
Setting `skip` tag now that we have some results to be able to test the LTP PR #521 which conflicts with this one.

mgalka:
@montjoie Are you planning to move forward with it?

PR #512, opened 2020-10-07
add selftests fragment to stable_5.8.y branch in build-configs

alexandrasp:


gctucker:
We need to fix the build issues first.  Then when enabling kselftest builds on stable, we could inherit the stable variants and add the fragments rather than create a new entry in YAML from scratch.

gctucker:
Let's revisit that later on when kselftest support is more mature both upstream and with the KernelCI integration.

PR #514, opened 2020-10-09
add libc6-dev-amd64-cross in arm64 gcc-8 Dockerfile

alexandrasp:
Add dependency for arm64 build to gcc-8 Docker image. It fixes: filesystems/binderfs, pidfd and vm selftests build.

### DIFF
--- a/jenkins/dockerfiles/gcc-8_arm64/Dockerfile
+++ b/jenkins/dockerfiles/gcc-8_arm64/Dockerfile
@@ -31,4 +31,5 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
    libcap-dev:arm64 \
    libcap-ng-dev:arm64 \
    libelf-dev:arm64 \
-   libpopt-dev:arm64
+   libpopt-dev:arm64 \ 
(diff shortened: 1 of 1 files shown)


jenkins/dockerfiles/gcc-8_arm64/Dockerfile
```diff
@@ -31,4 +31,5 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
    libcap-dev:arm64 \
    libcap-ng-dev:arm64 \
    libelf-dev:arm64 \
-   libpopt-dev:arm64
+   libpopt-dev:arm64 \ 
+   libc6-dev-amd64-cross:amd64
```
gctucker, line 35: This looks strange, it's an x86 package to cross-compile x86.  If you need to cross-compile x86 (amd64) on arm64, then you need `libc6-dev-amd64-cross:amd64`.  If it's for building host tools, then you need `libc6-dev` (without `:arm64`).

gctucker (commented):
There seems to be a bit of confusion around arm64 and amd64.  Can you please show an example of a build error fixed by this PR?

PR #520, opened 2020-10-20
 test-configs.yaml: Enable ltp-ipc tests on rk3288-veyron-jaq

Lakshmipathi:
Signed-off-by: Lakshmipathi <lakshmipathi.ganapathi@collabora.com>

PR #523, opened 2020-10-22
rootfs-configs.yaml: Include armel,armhf architecture

Lakshmipathi:

### DIFF
--- a/rootfs-configs.yaml
+++ b/rootfs-configs.yaml
@@ -143,4 +143,5 @@ rootfs_configs:
     arch_list:
       - amd64
       - arm64
+      - armhf
     script: "scripts/buster-ltp.sh"


gctucker (commented):
The change looks fine, but the commit message is not descriptive enough.

The summary should mention buster-ltp, and having a more detailed commit message to explain why we're doing this (or why armel and armhf were not built before) would be helpful.

rootfs-configs.yaml
```diff
@@ -143,4 +143,6 @@ rootfs_configs:
     arch_list:
       - amd64
       - arm64
+      - armel
```
gctucker, line 146: Actually, `armel` is not really needed.  We can run LTP with `armhf` on all the 32-bit arm platforms we have.  That's what the other Buster rootfs images do.

Lakshmipathi: okay  updated the patch.

Lakshmipathi:
Now added more details to commit message.

gctucker:
Thanks @Lakshmipathi. Have you built the `armhf` variant on staging already?

gctucker:
I see this was tested as described on #520, here's a sample job: https://lava.collabora.co.uk/results/2741325/0_ltp

PR #524, opened 2020-10-26
Fix renesas rename

gctucker:
There are on-going issues due to some Renesas platform being renamed, this is a WIP to fix it.

### DIFF
--- a/test-configs.yaml
+++ b/test-configs.yaml
@@ -1148,26 +1148,15 @@ device_types:
     class: arm-dtb
     boot_method: uboot
 
-  r8a7795-salvator-x: &r8a7795-salvator-x
-    mach: renesas
-    class: arm64-dtb
-    boot_method: uboot
-    flags: ['big_endian']
-    filters:
-      - blocklist: {defconfig: ['allnoconfig', 'allmodconfig']}
-
-# this is the same device than r8a7795-salvator-x
-  r8a77950-salvator-x: *r8a7795-salvator-x
-
   r8a7796-m3ulcb: &r8a7796-m3ulcb
+    base_name: r8a7796-m3ulcb
     mach: renesas
     class: arm64-dtb
     boot_method: uboot
     filters:
       - blocklist: {defconfig: ['allnoconfig', 'allmodconfig']}
 
-# this is the same device than r8a7796-m3ulcb
-  r8a77960-ulcb: *r8a7796-m3ulcb
+  r8a77960-ulcb: *r8a7796-m3ulcb # same device as r8a7796-m3ulcb
 
   rk3399-gru-kevin:
     mach: rockchip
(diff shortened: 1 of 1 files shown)


patersonc:
There are also variants like r8a77951-ulcb.dts, r8a77961-ulcb.dts, r8a77960-salvator-xs.dts etc. but I guess these aren't present in the board farms so not much point in adding them.

gctucker:
@patersonc Thanks for mentioning those, and yes there's little point in adding devices that aren't covered by any labs on kernelci.org.

test-configs.yaml
```diff
@@ -1117,26 +1117,15 @@ device_types:
     class: arm-dtb
     boot_method: uboot
 
-  r8a7795-salvator-x: &r8a7795-salvator-x
+  r8a77960-ulcb: &r8a77960-ulcb
```
gctucker, line 1120: Tested OK with stable 5.8: https://lava.collabora.co.uk/scheduler/job/2776404
However I believe the dtb name needs to be manually overridden for `r8a7796-m3ulcb` in older kernels.

mgalka (approved):

gctucker:
Tested OK on staging,

stable with `renesas/r8a7796-m3ulcb.dtb `:
https://staging.kernelci.org/test/plan/id/5fbe87bd49f3efeffed870b7/
https://staging.kernelci.org/test/case/id/5fbe87bd49f3efeffed870b8/

mainline with `renesas/r8a77960-ulcb.dtb `: 
https://staging.kernelci.org/test/plan/id/5fbe18f1a1eb5c25e7a91b47/
https://staging.kernelci.org/test/case/id/5fbe18f1a1eb5c25e7a91b48/

gctucker:
Note that these boards are currently failing to boot with the gcc-8 builds in mainline.  The 5.4 builds boot fine.  This may be a problem with the bootloader, in any case it doesn't seem related to the KernelCI configuration.

gctucker:
For salvator-x, it worked fine with stable 5.4:
https://staging.kernelci.org/test/plan/id/5fbe87fd49f3efeffed870fe/
https://staging.kernelci.org/test/case/id/5fbe87fd49f3efeffed870ff/

but for mainline, although the job was submitted with the correct dtb URL it failed to boot:
http://lava.baylibre.com:10080/scheduler/job/409824

gctucker:
@montjoie See the problem above - please confirm whether this is expected to fail or if there is a problem with the job definition.

montjoie:
I dont see anything, the job seems ok. I tried to move memory address of kernel/dtb in case of memory overlap, but no change.
I fear some hardcoded test in the uboot.

gctucker:
Thanks.  I think we can reasonably say that the issues are due to the bootloaders or related to the size of the kernel image, but the job definitions are correct with the matching dtb file for each board both up to 5.4 and after the rename.  So this can be merged and deployed now.

PR #525, opened 2020-10-28
T24394 - fix headers files problems in arm, arm64 and x86

alexandrasp:

### DIFF
--- a/jenkins/dockerfiles/gcc-8_arm/Dockerfile
+++ b/jenkins/dockerfiles/gcc-8_arm/Dockerfile
@@ -20,4 +20,6 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
    libcap-dev:armhf \
    libcap-ng-dev:armhf \
    libelf-dev:armhf \
+   libfuse-dev:armhf \
+   libnuma-dev:armhf \
    libpopt-dev:armhf
--- a/jenkins/dockerfiles/gcc-8_arm64/Dockerfile
+++ b/jenkins/dockerfiles/gcc-8_arm64/Dockerfile
@@ -31,4 +31,6 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
    libcap-dev:arm64 \
    libcap-ng-dev:arm64 \
    libelf-dev:arm64 \
+   libfuse-dev:arm64 \
+   libnuma-dev:arm64 \
    libpopt-dev:arm64
--- a/jenkins/dockerfiles/gcc-8_x86/Dockerfile
+++ b/jenkins/dockerfiles/gcc-8_x86/Dockerfile
@@ -12,4 +12,6 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
    libcap-dev \
    libcap-ng-dev \
    libelf-dev \
+   libfuse-dev \
+   libnuma-dev \
    libpopt-dev


jenkins/dockerfiles/gcc-8_arm/Dockerfile
```diff
@@ -20,4 +20,6 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
    libcap-dev:armhf \
    libcap-ng-dev:armhf \
    libelf-dev:armhf \
-   libpopt-dev:armhf
+   libpopt-dev:armhf \
```
gctucker, line 23: Please keep the package names in alphabetical order.

gctucker (commented):
I thought these header files were already installed from the kernel source tree, just that the kselftest build was not able to find them?

jenkins/dockerfiles/gcc-8_arm/Dockerfile
alexandrasp: I will keep, sorry about it! I forgot about your other comment.

alexandrasp:
> I thought these header files were already installed from the kernel source tree, just that the kselftest build was not able to find them?

Yes, these files suppose to be there, but after some locally tests installing it ʻagain` seems to solve almost all that first problems. I pushed it here, so you and others can see. Well, but according to new logs from staging..the only problem that have disappeared is related to `fuse.h`. 

[Log for amr64](https://storage.staging.kernelci.org/kernelci/staging-stable/staging-stable-20201029.0/arm64/defconfig+kselftest/gcc-8/build.log)
[Log for x86](https://storage.staging.kernelci.org/kernelci/staging-stable/staging-stable-20201029.0/x86_64/x86_64_defconfig+kselftest/gcc-8/build.log)
[Log for arm](https://storage.staging.kernelci.org/kernelci/staging-stable/staging-stable-20201029.0/arm/multi_v7_defconfig+kselftest/gcc-8/build.log)

Today I am running new experiments on jenkins to see how it goes, I pushed my own images there. Once I accomplish new things I will push here in this WIP branch to see if it works well on staging. Ok? :)

gctucker:
Yes, it's fine to have this WIP branch to try things out on staging.

Tip: Next time you make a WIP PR you can make it a "draft", which is for things that aren't ready to be merged.

alexandrasp:
@gctucker new update here. 

Well, it works: 
- [old logs from arm64](https://storage.staging.kernelci.org/kernelci/staging-stable/staging-stable-20201031.0/arm64/defconfig+kselftest/gcc-8/build.log) 
- [new log from arm64](https://storage.staging.kernelci.org/kernelci/staging-stable/staging-stable-20201102.0/arm64/defconfig+kselftest/gcc-8/build.log)

Now `drm.h`, `udmabuf.h` and more problems related to header file errors have gone. Also `pidfd`, `filesystems/binderfs`, `kvm` have started to build..

kernelci/build.py
```diff
@@ -435,7 +435,7 @@ def _output_to_file(cmd, log_file, rel_dir=None):
 
 def _run_make(kdir, arch, target=None, jopt=None, silent=True, cc='gcc',
               cross_compile=None, use_ccache=None, output=None, log_file=None,
-              opts=None, cross_compile_compat=None):
+              opts=None, cross_compile_compat=None, kselftest_kdir=None):
```
gctucker, line 438: I believe we can try to build with the output set to the local directory, like "normal" builds since this is what some kselftests expect. It's better to do that by changing the `build.jpl` Jenkins job rather than adding an extra argument with more complexity in the code.

gctucker (commented):
Are there any packages that can be dropped from the Docker image now that we know how to resolve the include path issue?

(I think it only happens when building with a separate output directory)

kernelci/build.py
alexandrasp: good input! I will update the code there. You have more xp in kernel-ci than me, when we know what the problem is, you and others can suggest other ways to solve, etc! Thanks @gctucker :)

gctucker:
OK thanks. Please update the commit message to something like `docker: install kselftest dependencies` and explain why this is needed.  The fix for the output build directory can be a separate PR.

gctucker:
Nice - also feel free to drop the `WIP` in the PR title when you think this is ready to be merged.

alexandrasp:
> Nice - also feel free to drop the `WIP` in the PR title when you think this is ready to be merged.

Ok! Tomorrow I will check if there is anything else that could be added here to fix build dependencies related to user-space, etc.

alexandrasp:
Full log with this PR applied:
https://storage.staging.kernelci.org/kernelci/staging-mainline/staging-mainline-20201105.2/arm64/defconfig+kselftest/gcc-8/build.log

Errors related to `#include <fuse.h>` and  `#include <numa.h>` have gone. It was necessary to install both libraries because are from user-space.

gctucker (approved):

PR #526, opened 2020-10-30
kci_build: Modify the build directory structure

wangmingyu84:
The original build directory is nested in two layers /KDIR/build/KDIR/build/_modules_,  modify the directory to /KDIR/build/_modules_, so that there is only one /KDIR/build left.

Signed-off-by: Wang Mingyu <wangmy@cn.fujitsu.com>

### DIFF
--- a/kernelci/build.py
+++ b/kernelci/build.py
@@ -645,7 +645,7 @@ def build_kernel(build_env, kdir, arch, defconfig=None, jopt=None,
             shutil.rmtree(mod_path)
         os.makedirs(mod_path)
         opts.update({
-            'INSTALL_MOD_PATH': mod_path,
+            'INSTALL_MOD_PATH': '_modules_',
             'INSTALL_MOD_STRIP': '1',
             'STRIP': "{}strip".format(cross_compile),
         })


gctucker:
Hello, I haven't seen this issue so maybe it only happens with some specific cases.  Could you please share the `kci_build` commands you've been using or maybe anything else needed to reproduce the problem?

wangmingyu84:
The steps are as follows:
$ ./kci_build update_mirror --build-config=mainline --mirror=linux-mirror.git
$ ./kci_build update_repo --build-config=mainline --kdir=linux --mirror=linux-mirror.git
$ ./kci_build build_kernel  --defconfig=defconfig --arch=x86_64 --build-env=gcc-8 --kdir=linux
make KBUILD_BUILD_USER=KernelCI -Clinux -s ARCH=x86_64 HOSTCC=gcc O=/home/wangmy/kernelci-core/linux/build defconfig
make KBUILD_BUILD_USER=KernelCI -Clinux -j6 -s ARCH=x86_64 HOSTCC=gcc O=/home/wangmy/kernelci-core/linux/build
/home/wangmy/kernelci-core/linux/tools/objtool
make KBUILD_BUILD_USER=KernelCI -Clinux -j6 -s ARCH=x86_64 HOSTCC=gcc O=/home/wangmy/kernelci-core/linux/build modules
make KBUILD_BUILD_USER=KernelCI INSTALL_MOD_PATH=linux/build/_modules_ INSTALL_MOD_STRIP=1 STRIP=strip -Clinux -s ARCH=x86_64 HOSTCC=gcc O=/home/wangmy/kernelci-core/linux/build modules_install

After executing the above command, the generated build directory is as follows:
./linux/build/linux/build/_modules_/lib/modules/
The following directory is empty:
./linux/build/_modules_/

gctucker:
OK I can see this now.  I've made an alternative fix by using an absolute path for the modules directory: #543.  This is more reliable because `_modules_` is only a default value, it's possible for users to change it.  So it shouldn't be hard-coded in the modules installation path.

gctucker:
Thanks for reporting this, it's now fixed with #543 instead.

PR #527, opened 2020-10-30
T24131 test-configs.yaml: update Debian rootfs URLs with new buster-ltp

gctucker:
Update all the Debian rootfs URLs which include a new buster-ltp
variant for armhf.


gctucker:
Replaced with today's Debian builds.

PR #529, opened 2020-11-05
build-configs.yaml: enable clang-10 builds on mainline

### LINKED ISSUE #518: clang builds of mainline
IIRC, we only ever turned on clang coverage for linux-next.  We should add builds of mainline with clang.

gctucker:
Enable clang-10 builds on mainline using the same set of architectures
and configs as on linux-next.

Fixes #518

### DIFF
--- a/build-configs.yaml
+++ b/build-configs.yaml
@@ -215,7 +215,7 @@ build_environments:
 # Default config with full build coverage
 build_configs_defaults:
   variants:
-    gcc-8:
+    gcc-8: &default_gcc-8
       build_environment: gcc-8
 
       fragments: &default_fragments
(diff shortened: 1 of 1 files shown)


gctucker:
Started a build on staging with this PR (`staging-mainline-20201105.1`), now adding the `staging-skip` label as it will conflict with the patch to build a subset of mainline kernels every day on staging.

gctucker:
Actually this had to be a true `mainline` build config, so here it is: https://staging.kernelci.org/build/mainline/branch/master/kernel/v5.10-rc2-140-g521b619acdc8/

broonie:
Looks OK to me.

PR #530, opened 2020-11-09
T24749 Add checks for alphabetical order of device types

gctucker:
Add checks to `kci_test validate` to detect when device types are not in alphabetical order.  Fixed any issues so the checks pass.


gctucker:
Tested OK on staging.

PR #532, opened 2020-11-09
T24749 build config sort check

gctucker:
Add checks for alphabetical order of trees, fragments and build configs in `build-configs.yaml`.  Also fix a few alphabetical issues found.

Depends on #530.


gctucker:
Tested OK on staging.

PR #533, opened 2020-11-10
kernelci.build: add missing kernel image targets for x86

gctucker:
Add missing kernel image target definitions for x86_64 and i386.  This
is needed to build only the kernel image, separately from modules or
anything else.

Fixes: 723f653f00e6 ("kernelci.build: add functions to build and publish kernels")


gctucker:
Tested OK on staging:
https://staging.kernelci.org/build/id/5fb475d7f3a42467ec033539/

PR #543, opened 2020-11-18
kernelci.build: use absolute path for INSTALL_MOD_PATH

gctucker:
Use an absolute path based on the current working directory when
setting the value of INSTALL_MOD_PATH to avoid creating a
sub-directory relative to the root of the kernel source tree.

Reported-by: Wang Mingyu <wangmy@cn.fujitsu.com>


khilman (approved):

gctucker:
Tested OK on staging.  This replaces #526.

PR #545, opened 2020-11-19
test_lava.py: Modify example in function comment of get_device_type_by_name.

wangmingyu84:
The example of the function does not match the code.

Signed-off-by: Wang Mingyu <wangmy@cn.fujitsu.com>

### DIFF
--- a/kernelci/lab/lava.py
+++ b/kernelci/lab/lava.py
@@ -32,13 +32,13 @@ def get_device_type_by_name(name, device_types, aliases=[]):
 
         Example:
         IN:
-            name = "x15"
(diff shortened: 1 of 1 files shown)


gctucker:
Sorry this got closed by accident when the `master` branch was renamed to `main`.  I think it was because the incoming repository for this PR is unknown.

gctucker:
@wangmingyu84 I've now simplified the code: #590. That function was from a legacy implementation before the `kernelci.lab.lava` module was created.  There's now a method in the `LAVA` class using a plain Python dictionary to translate aliases into device types.  Thanks for reporting the issue, it made it clear that there was a problem in the code.

PR #248, opened 2019-11-18
Allmodconfig

roxell:
enable allmodconfig build and boots with the KCONFIG_ALLCONFIG variable.

PR #548, opened 2020-11-19
T24951 roxell allmodconfig

gctucker:
This is to build `allmodconfig` for `arm64` with default values based on `defconfig` in an attempt at producing a kernel that can boot, and get tests to run on QEMU.

Replaces #248, rebased and adjusted with changes in kernelci-core made since then.

### DIFF
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -148,6 +148,13 @@ trees:
 
 fragments:
 
+  # from roxell, to build an allmodconfig kernel that boots on arm64
+  allmodconfig-arm64-boot:
+    path: "kernel/configs/allmodconfig-fixup.config"
+    configs:
+      - '# CONFIG_PROVE_RAW_LOCK_NESTING is not set'
+      - '# CONFIG_RCU_STRICT_GRACE_PERIOD is not set'
+
   amdgpu:
     path: "kernel/configs/amdgpu.config"
     configs:
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -1271,8 +1271,7 @@ build_configs:
           arm64:
             <<: *arm64_arch
             extra_configs:
-# Disabling to reduce load
-#              - 'allmodconfig'
+              - 'allmodconfig+KCONFIG_ALLCONFIG=arch/arm64/configs/defconfig'
               - 'allnoconfig'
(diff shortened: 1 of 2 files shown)


gctucker:
Tested manually with this extra fragment as per @roxell's instructions, and with no modules:
https://lava.collabora.co.uk/scheduler/job/2934555 (QEMU)
https://lava.collabora.co.uk/scheduler/job/2934556 (hip07-d05)

Config used:
```
./kci_build make_config --defconfig=defconfig+KCONFIG_DEFCONFIG=arch/arm64/configs/defconfig+kernel/configs/allmodfixup.config
```
With the `allmodfixup.config` fragment containing:
```
# CONFIG_PROVE_RAW_LOCK_NESTING is not set
# CONFIG_RCU_STRICT_GRACE_PERIOD is not set
```
Built with gcc-8.

aliceinwire:
@gctucker any update on this ?

gctucker:
Yes, it's at the top of my archaeological ToDo list for resolving old PRs.  I think it'll get resolved by the end of the year, if not we'll just close it.  There is some added value in booting allmodconfig at least on some devices (i.e. arm64 qemu natively on arm64 hardware in LKFT's lab).

gctucker:
Boot-tested an image I've built locally, without the modules:
https://lkft.validation.linaro.org/scheduler/job/6096526

That seems to be working well for a starting point.  So the next step is to get that tested on staging with an actual full allmodconfig build including the modules.

gctucker:

gctucker:
OK it's been over 3 years now and this won't get fixed with the legacy system.  It might be considered again with the new API & Pipeline so I've created an issue to keep track of this: #2142

gctucker (commented):
Basically, this means using this config for the kernel build:
```
allmodconfig+KCONFIG_ALLCONFIG=arch/arm64/configs/defconfig
```

There's also this fragment to disable things that break booting, which might need to be regularly updated:
```

  # from roxell, to build an allmodconfig kernel that boots on arm64
  allmodconfig-arm64-boot:
    path: "kernel/configs/allmodconfig-fixup.config"
    configs:
      - '# CONFIG_PROVE_RAW_LOCK_NESTING is not set'
      - '# CONFIG_RCU_STRICT_GRACE_PERIOD is not set'
```

PR #538, opened 2020-11-17
test-configs.yaml: Enable LTP-crypto tests on devices in lab-collabora

Lakshmipathi:
Signed-off-by: Lakshmipathi <lakshmipathi.ganapathi@collabora.com>

PR #552, opened 2020-11-24
test-configs.yaml: Add more LTP components with appropriate timeout v…

Lakshmipathi:
…alue

### DIFF
--- a/test-configs.yaml
+++ b/test-configs.yaml
@@ -170,16 +170,232 @@ test_plans:
     params:
       job_timeout: '120'
 
-  ltp-ipc:
+  ltp-opts: &ltp-rootfs
     rootfs: debian_buster-ltp_nfs
     pattern: 'ltp/{category}-{method}-{protocol}-{rootfs}-ltp-template.jinja2'
     filters:
       - blocklist: *kselftest_defconfig_filter
-    params:
+    params: &ltp-params
       skip_install: "true"
-      tst_cmdfiles: "ipc"
+      tst_cmdfiles: "ltp"
       job_timeout: '15'
       skipfile: skipfile-lkft.yaml
+
+  ltp-admin_tools:
+    <<: *ltp-rootfs
+    params:
+      <<: *ltp-params
+      tst_cmdfiles: "admin_tools"
+
+  ltp-connectors:
+    <<: *ltp-rootfs
+    params:
+      <<: *ltp-params
+      tst_cmdfiles: "connectors"
+
+  ltp-containers:
+    <<: *ltp-rootfs
+    params:
+      <<: *ltp-params
+      tst_cmdfiles: "containers"
+
+  ltp-controllers:
+    <<: *ltp-rootfs
+    params:
+      <<: *ltp-params
+      tst_cmdfiles: "controllers"
+      job_timeout: '45'
+
+  ltp-crypto:
+    <<: *ltp-rootfs
+    params:
+      <<: *ltp-params
+      tst_cmdfiles: "crypto"
+
+  ltp-device-drivers:
+    <<: *ltp-rootfs
+    params:
+      <<: *ltp-params
+      tst_cmdfiles: "device-drivers"
+
+  ltp-fcntl-locktests:
+    <<: *ltp-rootfs
+    params:
+      <<: *ltp-params
+      tst_cmdfiles: "fcntl-locktests"
+
+  ltp-filecaps:
+    <<: *ltp-rootfs
+    params:
+      <<: *ltp-params
+      tst_cmdfiles: "filecaps"
+
+  ltp-firmware:
+    <<: *ltp-rootfs
+    params:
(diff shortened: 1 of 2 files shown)


test-configs.yaml
```diff
@@ -1566,7 +1940,41 @@ test_configs:
   - device_type: bcm2836-rpi-2-b
     test_plans:
       - baseline
+      - ltp-admin_tools
```
gctucker, line 1943: For test configs, we can easily have a YAML anchor as we can have multiple test config entries with the same device type.  For example:
```yaml
ltp-test-plans: &ltp_test_plans
  - ltp-admin_tools
  - ltp-connectors
  - ltp-containers
# etc...

test_configs:
  - device_type: bcm2836-rpi-2-b
    test_plans:
      - baseline
      - sleep
      - usb

  - device_type: bcm2836-rpi-2-b
    test_plans: *ltp_test_plans
```

gctucker: Also we'll see how things go when running everything on all the test devices, maybe we'll want to also have anchors with some LTP subsets.  For example, `ltp-fs-plans`, `ltp-mem-plans` etc... with just a handful of things to run in order to optimise how labs' capacity is being used.  But that's for a bit later, once we have started collecting results.  It's probably going to be an on-going process in fact, with new labs and new devices and new versions of LTP.

gctucker (commented):
Looks like a good starting point to start running LTP more widely and get more results on staging.  We should however at least use YAML anchors to make this easier to maintain, and later on improve how the configuration is managed while this keeps growing.

gctucker:
This PR conflicts with earlier PRs and as such it's not currently being deployed on staging.  That's why the LTP results don't appear on the web frontend yet and no jobs are being submitted.  We should try to get this run at least once this week.

test-configs.yaml
```diff
@@ -170,6 +170,193 @@ test_plans:
     params:
       job_timeout: '120'
 
+  ltp-admin_tools:
```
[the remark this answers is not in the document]
Lakshmipathi: okay added this change and verified config files.

test-configs.yaml
Lakshmipathi: > For test configs, we can easily have a YAML anchor as we can have multiple test config entries with the same device type. For example:
> 
> ```yaml
> ltp-test-plans: &ltp_test_plans
>   - ltp-admin_tools
>   - ltp-connectors
>   - ltp-containers
> # etc...
> 
> test_configs:
>   - device_type: bcm2836-rpi-2-b
>     test_plans:
>       - baseline
>       - sleep
>       - usb
> 
>   - device_type: bcm2836-rpi-2-b
>     test_plans: *ltp_test_plans
> ```

updated the patch with this change.

test-configs.yaml
```diff
@@ -1468,6 +1721,7 @@ device_types:
       - blocklist: {defconfig: ['multi_v7_defconfig+linaro-base+distribution']}
 
 
+
```
gctucker, line 1724: Superfluous blank line...

test-configs.yaml
```diff
@@ -224,6 +440,43 @@ device_default_filters:
...
+      - ltp-pty
+      - ltp-sched
+      - ltp-security
+      - ltp-sound
+      - ltp-syscalls
+      - ltp-timers
+      - ltp-tracing
+      - ltp-uevents
+
 device_types:
```
gctucker, line 480: This is where having 2 blank lines helps :)

test-configs.yaml
```diff
@@ -170,17 +170,233 @@ test_plans:
...
+  ltp-opts: &ltp-rootfs
     rootfs: debian_buster-ltp_nfs
     pattern: 'ltp/{category}-{method}-{protocol}-{rootfs}-ltp-template.jinja2'
     filters:
       - blocklist: *kselftest_defconfig_filter
-    params:
+    params: &ltp-params
       skip_install: "true"
-      tst_cmdfiles: "ipc"
+      tst_cmdfiles: "ltp"
```
gctucker, line 180: Maybe use an empty string here with a comment rather than `ltp` as it's not a real subset name.

gctucker (commented):
Thanks, that's much better as there's a lot less duplication.  We'll probably need to split this config file into separate ones at some point, but for now it's fine.

Next step is to get it to run on staging, I think we have to merge the `ltp-crypto` PR #538 first as this one seems to be conflicting with it.

gctucker:
Also the summary in the git commit seems to be a bit too long.  Maybe reformat it like this:
```
test-configs.yaml: define more LTP test plans

[describe in more detail what this commit does]

Signed-off-by: ...
```

gctucker:
@Lakshmipathi Maybe this PR should be closed if you're now going to be adding each test plan one at a time?

Lakshmipathi:
Yes @gctucker will be adding each test plan one at a time. Closing this now.

PR #553, opened 2020-11-27
kernel-arch-complete.sh: Update email reporting for CIP and Renesas

patersonc:
Add email notifications for CIP.

For now I've added a notification for each test that is currently run on CIP trees, e.g. https://kernelci.org/test/job/cip/branch/linux-4.4.y-cip-rt/kernel/v4.4.244-cip51-rt32/

I've copied the `delay` time from similar entries.

I've also included `kernel-build-reports@lists.linaro.org` as it seems to be included everywhere else, happy to drop this though if required.


gctucker:
> I've also included `kernel-build-reports@lists.linaro.org` as it seems to be included everywhere else, happy to drop this though if required.

Thanks, yes that's fine.  It used to be the default list for KernelCI results, until we added kernelci-results@groups.io.  At some point we should drop the linaro.org one, when enough people have joined the new one.

gctucker (approved):
LGTM

PR #554, opened 2020-11-27
T23360 Refactor kci_build build_kernel with separate steps

gctucker:
This is a first step to redesign how the builds are run in `kci_build`.  It adds the concept of a build step, with the state between steps being kept in `bmeta.json`.  This defines a number of new `kci_build` commands without affecting how `kci_build build_kernel` currently works, until the new commands can be used instead.

Some changes may then be required with the integration in Jenkins and the Kubernetes build job definitions, but it is not yet required as part of this PR.

- [x] demonstrated how this can be used manually (see below)
- [x] docstrings added to the new classes
- [x] at least one resulting kernel tested in LAVA, outside of KernelCI pipeline https://lava.collabora.co.uk/scheduler/job/2931143

## Steps to use locally

In this example, we're going to build the mainline kernel with `gcc-8` for `arm64` with the generic `defconfig` and `KCONFIG_KASAN=y` on top of it using the new commands provided by this PR.

### Config file

First, let's use a local `kernelci.conf` file to simplify the command line arguments:
```ini
[kci_build]
kdir: linux
build_env: gcc-8
j: 4
output: linux/local-build
log: kernel-build.log
verbose: true
```

### Docker (optional)

You will need to either install a GCC toolchain for `aarch64` with all the kernel build dependencies or use the KernelCI Docker image:

```
$ docker run -it -v $PWD:/root/kernelci-core kernelci/build-gcc-8_arm64 /bin/bash
root@76ad3035ee88:/# cd /root/kernelci-core
root@76ad3035ee88:~/kernelci-core# 
```

### Kernel tree

Then, get the kernel source code.  This hasn't changed:
```
./kci_build update_repo --build-config=mainline
```

You should now have a `linux` directory with the latest mainline kernel source tree.

### The build

The following commands will build the kernel step by step, starting with the initial `bmeta.json` meta-data file which contains some information about the kernel revision and the build environment.  This will keep getting used between each build step, and you can open the file between each command to see what is being added to it.  Then the subsequent commands are to generate the kernel config, build the kernel image, then the modules and the dtbs:
```
./kci_build init_bmeta --build-config=mainline --arch=arm64
./kci_build make_config --defconfig=defconfig+CONFIG_KASAN=y 
./kci_build make_kernel
./kci_build make_modules
./kci_build make_dtbs
```

Once the build is complete, you should have everything in the output directory as specified in the config file, including the build log:
```
root@76ad3035ee88:~/kernelci-core# head linux/local-build/kernel.log
#
# make KBUILD_BUILD_USER=KernelCI -Clinux -j4 ARCH=arm64 CROSS_COMPILE=aarch64-linux-gnu- HOSTCC=gcc CC="ccache aarch64-linux-gnu-gcc" O=/root/kernelci-core/linux/local-build defconfig
#
make: Entering directory '/root/kernelci-core/linux'
```

### JSON meta-data

The `linux/local-build/bmeta.json` file should now be complete with all the information about the build:
```json
{
    "environment": {
        "arch": "arm64",
        "compiler": "gcc",
        "compiler_version": "8",
        "compiler_version_full": "aarch64-linux-gnu-gcc (Debian 8.3.0-2) 8.3.0",
        "cross_compile": "aarch64-linux-gnu-",
        "cross_compile_compat": "arm-linux-gnueabihf-",
        "make_opts": {
            "KBUILD_BUILD_USER": "KernelCI"
        },
        "name": "gcc-8",
        "platform": {
            "cpus": {
                "Intel Core Processor (Skylake, IBRS)": 3
            },
            "uname": [
                "Linux",
                "cros-build",
                "4.19.0-12-amd64",
                "#1 SMP Debian 4.19.152-1 (2020-10-18)",
                "x86_64",
                ""
            ]
        },
        "use_ccache": true
    },
    "kernel": {
        "defconfig": "defconfig",
        "defconfig_extras": [
            "CONFIG_KASAN=y"
        ],
        "defconfig_full": "defconfig+CONFIG_KASAN=y",
        "fragments": [
            "kernelci.config"
        ],
        "image": "Image",
        "vmlinux_bss_size": 9193632,
        "vmlinux_data_size": 13861696,
        "vmlinux_file_size": 508192704,
        "vmlinux_text_size": 21295464
    },
    "revision": {
        "branch": "master",
        "commit": "418baf2c28f3473039f2f7377760bd8f6897ae18",
        "describe": "v5.10-rc5",
        "describe_v": "v5.10-rc5",
        "tree": "mainline",
        "url": "https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git"
    },
    "steps": [
        {
            "duration": 0.12980270385742188,
            "name": "revision",
            "start_time": 1606494404.139411,
            "status": "PASS"
        },
        {
            "duration": 0.006347179412841797,
            "name": "environment",
            "start_time": 1606494404.269703,
            "status": "PASS"
        },
        {
            "duration": 17.760863304138184,
            "log_file": "kernel.log",
            "name": "config",
            "start_time": 1606494406.8110063,
            "status": "PASS",
            "threads": "4"
        },
        {
            "duration": 298.7760851383209,
            "log_file": "kernel.log",
            "name": "kernel",
            "start_time": 1606495414.587251,
            "status": "PASS",
            "threads": "4"
        },
        {
            "duration": 1197.6061565876007,
            "log_file": "kernel.log",
            "name": "modules",
            "start_time": 1606495747.7965348,
            "status": "PASS",
            "threads": "4"
        },
        {
            "duration": 12.293006896972656,
            "log_file": "kernel.log",
            "name": "dtbs",
            "start_time": 1606507046.7938302,
            "status": "PASS",
            "threads": "4"
        }
    ]
}
```
One of the new features compared to the current `build_kernel` command is that we can now have a different log file when running each step, by passing a different value to the `--log` option on the command line.  We also know the duration and start time of each step.  Also, the CPU information is more detailed (this example is on a VM with 3 vCPUs...).

### Next steps

Before this can be used in production for kernelci.org, a few more things will need to be done in follow-up PRs:
* create the modules tarball and install all the files to publish as per the `install_kernel` command
* produce the meta-data in the format expected by `api.kernelci.org` for publishing a new kernel build
* update the Kubernetes build job template to use the new commands
* potentially, keep `build_kernel` but turn it into a super-command to run all these things in one go

### DIFF
--- a/kci_build
+++ b/kci_build
@@ -336,6 +336,19 @@ class cmd_make_modules(Command):
         return step.run(args.mod_path, args.j, args.verbose)
 
 
+class cmd_make_dtbs(Command):
+    help = "Build device trees"
+    args = [Args.kdir]
+    opt_args = [Args.verbose, Args.output, Args.j, Args.log]
+
+    def __call__(self, configs, args):
+        step = kernelci.build.MakeDeviceTrees(args.kdir, args.output, args.log)
+        if not step.dt_enabled():
+            print("Device tree support disabled, not building.")
+            return True
+        return step.run(args.j, args.verbose)
+
+
 class cmd_build_kernel(Command):
     help = "Build a kernel"
     args = [Args.build_env, Args.arch, Args.kdir]
--- a/kci_build
+++ b/kci_build
@@ -323,6 +323,19 @@ class cmd_make_kernel(Command):
         return step.run(args.j, args.verbose)
 
 
+class cmd_make_modules(Command):
+    help = "Build kernel modules"
+    args = [Args.kdir]
+    opt_args = [Args.verbose, Args.output, Args.j, Args.log, Args.mod_path]
+
+    def __call__(self, configs, args):
+        step = kernelci.build.MakeModules(args.kdir, args.output, args.log)
+        if not step.modules_enabled():
+            print("Kernel modules disabled, not building.")
+            return True
+        return step.run(args.mod_path, args.j, args.verbose)
+
+
 class cmd_build_kernel(Command):
     help = "Build a kernel"
     args = [Args.build_env, Args.arch, Args.kdir]
--- a/kci_build
+++ b/kci_build
@@ -313,6 +313,16 @@ class cmd_make_config(Command):
         return step.run(args.defconfig, args.j, args.verbose)
 
 
+class cmd_make_kernel(Command):
+    help = "Make a kernel image"
+    args = [Args.kdir]
+    opt_args = [Args.verbose, Args.output, Args.j, Args.log]
+
+    def __call__(self, configs, args):
+        step = kernelci.build.MakeKernel(args.kdir, args.output, args.log)
+        return step.run(args.j, args.verbose)
+
+
 class cmd_build_kernel(Command):
     help = "Build a kernel"
     args = [Args.build_env, Args.arch, Args.kdir]
--- a/kci_build
+++ b/kci_build
@@ -303,6 +303,16 @@ Invalid arguments, --build-env and --arch need to be used together.""")
         return res
 
 
+class cmd_make_config(Command):
+    help = "Make kernel config"
+    args = [Args.kdir, Args.defconfig]
+    opt_args = [Args.verbose, Args.output, Args.j, Args.log]
+
+    def __call__(self, configs, args):
+        step = kernelci.build.MakeConfig(args.kdir, args.output, args.log)
+        return step.run(args.defconfig, args.j, args.verbose)
+
+
 class cmd_build_kernel(Command):
     help = "Build a kernel"
     args = [Args.build_env, Args.arch, Args.kdir]
--- a/kci_build
+++ b/kci_build
@@ -1,6 +1,6 @@
 #!/usr/bin/env python3
 #
-# Copyright (C) 2018, 2019 Collabora Limited
+# Copyright (C) 2018, 2019, 2020 Collabora Limited
 # Author: Guillaume Tucker <guillaume.tucker@collabora.com>
 #
 # This module is free software; you can redistribute it and/or modify it under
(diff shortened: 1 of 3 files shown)


khilman:
A couple things I noticed when converting my local scripts to use this.
There's no default for the build log, so if you leave it out, you get no log.  Previous version defaulted to `<output_dir>/build.log` which is also what the `install_kernel` step still expects.

The example json in the README includes the steps, but in my tests a separate `steps.json` is generated.

khilman:
The `install_kernel` step currently installs kernel, modules, dtbs, etc.  Are you planning to break that up into multiple steps also?

khilman:
A brief test of the build test with my local scripts, and everything works as expected (other than the minor things already mentioned.) 

However, the `install_kernel` step is still expecting the old bmeta.json format, so I wasn't able to test the rest of my scripts.

gctucker:
> A couple things I noticed when converting my local scripts to use this.
> There's no default for the build log, so if you leave it out, you get no log. Previous version defaulted to `<output_dir>/build.log` 

Yes, this is intentional and it should have been better explained in the commit messages.  The reasons for it are:
* When using a build log, it will persist between each step being run and new output will be added to it.  This is necessary so the `make_kernel` command doesn't wipe the log generated previous by `make_config` etc...  Having a `--keep-log` option would also work, to not have a growing log with each step being run, but that would complicate things a bit from a user point of view.
* When running the tool manually, it's likely the user will be running some steps many times.  The log will keep growing and that can end up causing a bad surprise when after building allmodconfig all day you get a huge file you don't need on your disk.
* Now we have support for the `kernelci.conf` settings file, we can specify a log file name as in the example show in this PR's description.  So if you do want to have the `build` directory and a `build.log` file there, you can just add it to your settings file and it will be like before.
* With automated builds, one of the reasons why this rework was needed was to be able to separate the logs for some steps (e.g. so dtbs and kselftest warnings and errors don't "pollute" the main kernel build).  Having a default log file name would still allow that to happen, but it doesn't seem that useful if we know we want to split the logs anyway.  So it also contributed to that decision although to a lesser extent.

If dropping the `build/build.log` defaults break some users' workflows with no easy alternative, then we should consider keeping it indeed.  Would it be OK to either add a `--log` option to each command or use a `kernelci.conf` file to provide your own default values?

gctucker:
> The `install_kernel` step currently installs kernel, modules, dtbs, etc. Are you planning to break that up into multiple steps also?

Yes, I'm currently thinking of adding `kci_build install_kernel` / `install_modules` etc... just like we have `make_kernel`, `make_modules`...  That's useful in particular if the install steps also produce some log output, so the `make` and `install` commands for a particular step can have their output in the same file.  I don't see any reason why that wouldn't work, both for command line usage and automation, but there can be subtle things on the way with unsuspected complications so we'll see how things go in practice.

gctucker:
> A brief test of the build test with my local scripts, and everything works as expected (other than the minor things already mentioned.)
> 
> However, the `install_kernel` step is still expecting the old bmeta.json format, so I wasn't able to test the rest of my scripts.

Thanks @khilman for testing this PR and for your comments.  As per the title, it is solely about replacing the `build_kernel` command with a series of steps. Now this is getting closer to being merged, I'll start working on `install_kernel` and `publish_kernel`, and then on `kci_test`.  For those things, the changes are mostly going to be about using the new `bmeta.json` format which still contains all the necessary data, so I think it should be fairly straightforward.  There will be follow-up PRs for that, and also in `kernelci-backend` since it also uses `bmeta.json`.

khilman (approved):

alexandrasp (approved):
I tested it locally and worked fine :)

PR #558, opened 2020-12-02
Docker qemu

montjoie:


MyleneJ (commented):
Hello,

Thank you for your patch!

I tested it and all is working fine for my test (running qemu/docker for an ARM architecture). The only issue I got was about the addition of the `tag` in the job (see coming comment in the lines).
I am new to KernelCI/LAVA but is it necessary? Without that, everything is working by adding this new `baseline-qemu-docker` test plan in my device-type.

Best regards,
Mylène

templates/boot/generic-qemu-boot-template.jinja2
```diff
@@ -19,6 +23,11 @@
 {% endblock %}
 {% block actions %}
 
+{% if qemu_docker %}
+tags:
+    - qemu-docker
+{% endif %}
+
```
MyleneJ, line 30: Is this tag is necessary?
thanks!

montjoie: Without this tag, labs which have workers with qemu AND does not support docker will get baseline-qemu-docker jobs and thoses jobs will fail.

gctucker: I think this should be done via the KernelCI YAML config, to filter jobs that get scheduled in labs, rather than rely on some arbitrary tags to be created in all the LAVA labs.

gctucker:
This is causing some errors on staging, with labs that don't have the `qemu-docker` tag:
```
ERROR /data/workspace/bot.staging.kernelci.org/test-runner/workspace/test-runner/jobs/lab-baylibre/kernelci-staging-stable-staging-stable-20201210.0-arm-multi_v7_defconfig-clang-10-no-dtb-qemu_arm-virt-gicv3-baseline-qemu-docker.yaml: <Fault 400: "Invalid job definition: Device tag 'qemu-docker' does not exist in the database..">
```

gctucker:
OK the tag has been removed from the job template so I think this should be fine now.

@montjoie Please rebase to resolve the conflicts, then we should be able to start testing this on staging.

templates/boot/generic-qemu-boot-template.jinja2
```diff
@@ -50,6 +54,11 @@ actions:
       minutes: 5
     method: qemu
     media: tmpfs
+{%- if qemu_docker %}
+    docker:
+      image: montjoie/qemu:bullseye
```
gctucker, line 59:
```suggestion
      image: kernelci/qemu
```
Same thing as for the `kernelci/fastboot` Docker image, this can be changed to `kernelci/staging-qemu` to run tests on staging.

templates/boot/generic-qemu-boot-template.jinja2
```diff
@@ -50,6 +54,11 @@ actions:
       minutes: 5
     method: qemu
     media: tmpfs
+{%- if qemu_docker %}
+    docker:
+      image: kernelci/staging-qemu
```
gctucker, line 59:
```suggestion
      image: kernelci/qemu
```

gctucker:
Thanks, I've built the image and I'm setting up the permissions in Docker Hub so the staging job can push it automatically.

jenkins/dockerfiles/build-and-push.sh
```diff
@@ -40,6 +40,7 @@ do
     d )  debos=true;;
     i )  dt_validation=true;;
     k )  k8s=true;;
+    Q )  qemu=true;;
```
gctucker, line 43: You need to add `Q` to the `options` variable and also in the comment describing each option.

gctucker (changes requested):
Please fix the issues with the new `Q` variable, this is preventing the image from being built with the `build-and-push.sh` script.

jenkins/dockerfiles/build-and-push.sh
gctucker: > and also in the comment describing each option.

jenkins/dockerfiles/build-and-push.sh
```diff
@@ -25,12 +25,13 @@
...
 # -i = also rebuild dt-validation
 # -q = make the builds quiet
+# -Q = also rebuild qemu docker image
 # -t = the prefix to use in docker tags (default is kernelci/)
 
 set -e
 tag_px='kernelci/'
 
-options='npbdikqt:'
+options='npbdikqtQ:'
```
gctucker, line 34:
```suggestion
options='npbdikqQt:'
```

gctucker (changes requested):
Still not quite correct.

gctucker:
@montjoie OK I think this isn't being used in any lab, so please enable it in `lab-configs.yaml` for at least one lab where you know it's supported with a recent enough version of LAVA.

gctucker:
@montjoie Hello, can you please rebase and enable your lab in `lab-configs.yaml` to run QEMU jobs in Docker?

gctucker:
Some results on staging:
https://staging.kernelci.org/test/job/kernelci/branch/staging-mainline/kernel/staging-mainline-20210326.0/plan/baseline-qemu-docker/
And here's a sample job in `lab-baylibre` which shows it's using the Docker image in the log:
http://lava.baylibre.com:10080/scheduler/job/37810

However, some jobs were also run in `lab-collabora` even though the `baseline-qemu-docker` test plan isn't enabled for it in `lab-configs.yaml`.  So this is something to investigate, it may not be an issue with this particular PR.  I'll try to reproduce this locally.

gctucker:
OK this is because of how the `passlist` filter works, I've made #627 to fix that.  If it works as expected in the next staging jobs then both PRs should be merged.

gctucker (approved):

gctucker (approved):
Tested OK on staging - see comments on #627.

PR #559, opened 2020-12-03
baseline: use kmsg for test shell messages

mwalle:
Sometimes the output is interrupted by the kernel messages, for example:
  https://lavalab.kontron.com/scheduler/job/1174#L1000

Use /dev/kmsg to mitigate that; as it is already used in the bootrr test
case.

Signed-off-by: Michael Walle <michael@walle.cc>


gctucker (commented):
LGTM, let's wait for some staging jobs to run to confirm it's working as intended.

gctucker:
Tested OK on staging, the `dmesg` results are still working as expected.

PR #560, opened 2020-12-07
kernelci.build: generate arch fragments

gctucker:
Generate config fragments included in a subset architectures as well
as full build variants.


gctucker:
Tested OK on staging.

PR #561, opened 2020-12-07
T22500 x86 chromebook fragment

gctucker:
Add an `x86-chromebook` config fragment to enable the serial console on most x86 Chromebooks (AMD and Intel), as well as the USB-Ethernet adapter found in the [Servo v4](https://chromium.googlesource.com/chromiumos/third_party/hdctools/+/master/docs/servo_v4.md#Type_C-Version) debug boards.


gctucker:
Tested OK on staging with #427.

PR #562, opened 2020-12-07
T23360 kci_build make_kselftest

gctucker:
Add the `kci_build make_kselftest` command to build kselftest as a separate step.

Depends on #554 

To test this locally, first create a `kernelci.conf` file in the current kernelci-core directory (to override any user or system-wide one):
```ini
[kci_build]
kdir: linux
build_env: gcc-8
j: 4
output: linux/build-kselftest
log: build.log
verbose: true
```
Then run these commands to generate the kselftest fragment and build in a Docker container:
```
$ ./kci_build generate_fragments --build-config=next
$ docker run -it -v $PWD:/root/kernelci-core kernelci/build-gcc-8_x86 /bin/bash
# ./kci_build init_bmeta --arch=x86_64 --build-config=next
# ./kci_build make_config --defconfig=x86_64_defconfig+kernel/configs/kselftest.config
# ./kci_build make_kernel
# ./kci_build make_kselftest
```
(the kselftest fragment can also be generated from within the container)

### DIFF
--- a/kci_build
+++ b/kci_build
@@ -349,6 +349,19 @@ class cmd_make_dtbs(Command):
         return step.run(args.j, args.verbose)
 
 
+class cmd_make_kselftest(Command):
+    help = "Build kselftest"
+    args = [Args.kdir]
+    opt_args = [Args.verbose, Args.output, Args.j, Args.log]
+
+    def __call__(self, configs, args):
+        step = kernelci.build.MakeSelftests(args.kdir, args.output, args.log)
+        if not step.fragment_enabled():
+            print("No kselftest fragment being used, not building.")
+            return True
+        return step.run(args.j, args.verbose)
+
+
 class cmd_build_kernel(Command):
     help = "Build a kernel"
     args = [Args.build_env, Args.arch, Args.kdir]
--- a/kernelci/build.py
+++ b/kernelci/build.py
@@ -978,6 +978,35 @@ class MakeDeviceTrees(Step):
         return res
 
 
+class MakeSelftests(Step):
+
+    def fragment_enabled(self):
+        """Check whether the kselftest fragment is enabled
+
+        Return True if the kselftest config fragment is enabled in the build
+        meta-data, or False otherwise.
+        """
+        return 'kselftest' in self._bmeta['kernel']['defconfig_extras']
+
+    def run(self, jopt=None, verbose=False):
+        """Make the kernel selftests
+
+        Make the kernel selftests or "kselftest" and produce a tarball so they
+        can be installed on a separate test platform.
+
+        *jopt* is the `make -j` option which will default to `nproc + 2`
+        *verbose* is whether the build output should be shown
+        """
+        opts = {
+            'FORMAT': '.xz',
+        }
+        res = self._make('gen_tar', jopt, verbose, opts,
+                         'tools/testing/selftests')
+        self._add_run_step('kselftest', jopt, res)
+        self._save_bmeta()
+        return res
+
+
 def _make_defconfig(defconfig, kwargs, extras, verbose, log_file):
     kdir, output_path = (kwargs.get(k) for k in ('kdir', 'output'))
     result = True
--- a/kernelci/build.py
+++ b/kernelci/build.py
@@ -515,30 +515,45 @@ class Step:
         self._kdir = kdir
         self._output_path = output_path or os.path.join(kdir, 'build')
         self._bmeta_path = os.path.join(self._output_path, 'bmeta.json')
+        self._steps_path = os.path.join(self._output_path, 'steps.json')
         self._log_file = log
         self._log_path = os.path.join(self._output_path, log) if log else None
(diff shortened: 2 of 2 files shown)


kernelci/build.py
```diff
@@ -492,6 +492,512 @@ def _run_make(kdir, arch, target=None, jopt=None, silent=True, cc='gcc',
...
+        self._output_path = output_path or os.path.join(kdir, 'build')
+        self._bmeta_path = os.path.join(self._output_path, 'bmeta.json')
+        self._steps_path = os.path.join(self._output_path, 'steps.json')
+        self._log_path = log
+        self._bmeta = dict()
+        self._steps = list()
+        self._dot_config = None
+        self._start_time = time.time()
+        if not os.path.exists(self._output_path):
+            os.mkdir(self._output_path)
```
mgalka, line 525: Is it intentional to create only the top level directory if it doesn't exist and fail if other parts of the path are missing?

kernelci/build.py
```diff
@@ -492,6 +492,512 @@ def _run_make(kdir, arch, target=None, jopt=None, silent=True, cc='gcc',
...
+    def _get_cpus(self):
+        cpus = {}
+        if os.path.exists('/proc/cpuinfo'):
+            cpu_list = []
+            with open('/proc/cpuinfo') as f:
+                for line in f:
+                    if 'model name' in line:
+                        cpu_list.append(line.split(':')[1].strip())
+            for cpu in cpu_list:
+                cpu_entry = cpus.setdefault(cpu, 0)
```
mgalka, line 587: cpu_entry is never used

kernelci/build.py
```diff
@@ -492,6 +492,512 @@ def _run_make(kdir, arch, target=None, jopt=None, silent=True, cc='gcc',
...
+        return cmd
+
+    def _get_make_opts(self, opts, make_path):
+        env = self._bmeta['environment']
+        make_opts = env['make_opts'].copy()
+        if opts:
+            make_opts.update(opts)
+
+        arch = env['arch']
+        make_opts['ARCH'] = arch
```
mgalka, line 611:
```suggestion
        make_opts['ARCH'] = env['arch']
```

mgalka (changes requested):

kci_build
```diff
@@ -261,6 +261,107 @@ class cmd_list_kernel_configs(Command):
...
+            conf = configs['build_configs'][args.build_config]
+            tree_name = conf.tree.name
+            tree_url = conf.tree.url
+            branch = conf.branch
+        else:
+            tree_name = args.tree_name
+            tree_url = args.tree_url
+            branch = args.branch
+        if not all((tree_name, tree_url, branch)):
+            print("""\
```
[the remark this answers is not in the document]
gctucker: We do need it to avoid having an extra new line.

kci_build
```diff
@@ -261,6 +261,107 @@ class cmd_list_kernel_configs(Command):
...
+Invalid arguments, either of these 2 sets are possible:
+   --tree-name, --tree-url and --branch
+or
+   --build-config (the tree and branch are read from the YAML config)\
```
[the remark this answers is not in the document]
gctucker: With this code, I get:
```
$ ./kci_build init_bmeta
Invalid arguments, either of these 2 sets are possible:
   --tree-name, --tree-url and --branch
or
   --build-config (the tree and branch are read from the YAML config)
```
If I remove the backslash:
```
$ ./kci_build init_bmeta

Invalid arguments, either of these 2 sets are possible:
   --tree-name, --tree-url and --branch
or
   --build-config (the tree and branch are read from the YAML config)
```
So I put it there to avoid having this superfluous blank line on the output, but still make the multi-line string appear in the code like it would appear on the output.

kci_build
```diff
@@ -261,6 +261,107 @@ class cmd_list_kernel_configs(Command):
...
+""")
+            return False
+
+        step = kernelci.build.RevisionData(args.kdir, args.output, reset=True)
+        print("Initialising build meta-data in {}".format(step.bmeta_path))
+        res = step.run(tree_name, tree_url, branch)
+
+        if res and (args.build_env or args.arch):
+            if not (args.build_env and args.arch):
+                print("""\
```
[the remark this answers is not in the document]
gctucker: ditto too

kernelci/build.py
gctucker: fixed

kernelci/build.py
gctucker: Yes, only the output directory has to be there because that's where the files created by this module will be stored, with no sub-directories.  Anything added in sub-directories within there will be created by the kernel build system.

kernelci/build.py
gctucker: The `arch` variable is used further down for `ccache`.

alexandrasp (approved):
@gctucker tested locally, it is working fine in x86_64 and arm64.

PR #568, opened 2020-12-11
test-configs.yaml: Enable LTP-lib tests on devices in lab-collabora 

Lakshmipathi:
This PR depends on https://github.com/kernelci/kernelci-core/pull/567

PR #569, opened 2020-12-11
test-configs.yaml: Enable LTP-mm tests on devices in lab-collabora

Lakshmipathi:
This depends on https://github.com/kernelci/kernelci-core/pull/568

### DIFF
--- a/test-configs.yaml
+++ b/test-configs.yaml
@@ -213,6 +213,12 @@ test_plans:
       <<: *ltp-params
       tst_cmdfiles: "ipc"
 
+  ltp-mm:
+    <<: *ltp
+    params:
+      <<: *ltp-params
+      tst_cmdfiles: "mm"
+      job_timeout: '30'
 
   preempt-rt:
     rootfs: debian_buster_nfs
@@ -1610,6 +1616,7 @@ test_configs:
       - baseline
       - ltp-crypto
       - ltp-ipc
+      - ltp-mm
       - sleep
       - usb
 
@@ -1632,6 +1639,7 @@ test_configs:
       - baseline-nfs
       - ltp-crypto
       - ltp-ipc
+      - ltp-mm
 
   - device_type: cubietruck
     test_plans:
@@ -1756,6 +1764,7 @@ test_configs:
       - baseline-nfs
       - ltp-crypto
       - ltp-ipc
+      - ltp-mm
       - sleep
       - usb
 
@@ -1827,6 +1836,7 @@ test_configs:
       - baseline-nfs
       - ltp-crypto
       - ltp-ipc
+      - ltp-mm
 
   - device_type: kirkwood-db-88f6282
     test_plans:
@@ -1957,6 +1967,7 @@ test_configs:
       - baseline-nfs
       - igt-kms-exynos
       - ltp-ipc
+      - ltp-mm
       - sleep
       - usb
 
@@ -2094,6 +2105,7 @@ test_configs:
       - baseline-nfs
       - ltp-crypto
       - ltp-ipc
+      - ltp-mm
       - sleep
       - usb
 
@@ -2108,6 +2120,7 @@ test_configs:
       - kselftest-lib
       - ltp-crypto
       - ltp-ipc
+      - ltp-mm
       - sleep
       - usb
 
@@ -2123,6 +2136,7 @@ test_configs:
       - kselftest
       - ltp-crypto
       - ltp-ipc
+      - ltp-mm
       - sleep
       - v4l2-compliance-uvc
 
@@ -2186,6 +2200,7 @@ test_configs:
       - baseline
       - ltp-crypto
       - ltp-ipc
+      - ltp-mm
 
   - device_type: sun50i-h6-pine-h64-model-b
     test_plans:
--- a/test-configs.yaml
+++ b/test-configs.yaml
@@ -190,27 +190,29 @@ test_plans:
       job_timeout: '10'
       kselftest_collections: "lib"
 
-  ltp-crypto:
+  ltp: &ltp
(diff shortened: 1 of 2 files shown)


test-configs.yaml
```diff
@@ -183,27 +183,47 @@ test_plans:
     params:
       job_timeout: '120'
 
-  ltp-crypto:
+  ltp-opts: &ltp-rootfs
```
gctucker, line 186:
```suggestion
  ltp: &ltp
```
This could be used as the "base" test plan definition, which is not a rootfs definition.

test-configs.yaml
```diff
@@ -183,27 +183,47 @@ test_plans:
...
-    params:
+    params: &ltp-params
       skip_install: "true"
-      tst_cmdfiles: "crypto"
+      tst_cmdfiles: ""
       job_timeout: '15'
       skipfile: skipfile-lkft.yaml
 
+  ltp-crypto:
+    <<: *ltp-rootfs
```
gctucker, line 198:
```suggestion
    <<: *ltp
```
following the suggestion above.

gctucker: This PR could have one patch with the 2 suggestions above, with the `&ltp` anchor.

Then one patch for `ltp-mm` which appears to be working fine.

test-configs.yaml
```diff
@@ -183,27 +183,47 @@ test_plans:
...
+    params:
+      <<: *ltp-params
+      tst_cmdfiles: "lib"
+
+  ltp-mm:
+    <<: *ltp-rootfs
+    params:
+      <<: *ltp-params
+      tst_cmdfiles: "mm"
+      job_timeout: '30'
```
gctucker, line 220: Patch number 2, to add `ltp-mm`.

```suggestion
  ltp-mm:
    <<: *ltp
    params:
      <<: *ltp-params
      tst_cmdfiles: "mm"
      job_timeout: '30'
```

test-configs.yaml
```diff
@@ -183,27 +183,47 @@ test_plans:
...
+
+  ltp-timers:
```
gctucker, line 222: There are known issues with `ltp-timers` so it could be added as a separate PR, to be merged after `ltp-mm`.

test-configs.yaml
```diff
@@ -183,27 +183,47 @@ test_plans:
...
-      - blocklist: *kselftest_defconfig_filter
+    <<: *ltp-rootfs
     params:
-      skip_install: "true"
+      <<: *ltp-params
       tst_cmdfiles: "ipc"
-      job_timeout: '15'
-      skipfile: skipfile-lkft.yaml
+
+  ltp-lib:
```
gctucker, line 209: Like `ltp-timers`, there are known issues because the test binaries appear to be missing from the rootfs (and the path to run the test with `runltp` may be wrong too).  So that could also be done with a separate PR after `ltp-mm`.

gctucker (commented):
This PR should be converted to only add the `&ltp` anchor definition and the `ltp-mm` test plan on top of the current `main` branch.  The `ltp-lib` and `ltp-timers` test plans have some issues so they should be added later as other PRs, once this one with `ltp-mm` has been merged.

gctucker:
Skipping until the comments above have been addressed.

Lakshmipathi:
@gtucker, I have too many merge conflicts while try to fix this PR.  I prefer to drop these (569, 568 and 567) draft version and prepare a new PR with 2 commits one for ltp yaml anchor setup and ltp-mm test.

gctucker:
@Lakshmipathi yes I thought you could do that and force-push to this branch

gctucker (commented):
For the first commit, it's better to use the present tense i.e.:
`test-configs.yaml: add LTP YAML anchors`
rather than `added`. That's a convention, to explain what the commit does when you apply it.

It's also good practice to include a small paragraph explaining what this change is doing.  It makes it easier to maintain the git history, without having to look at the diff or having to remember what the LTP anchors are about.

gctucker (commented):
The 2nd commit is in the present tense, so that's good.  There's no description, it would be good to mention the fact that the devices being enabled are known to successfully run the tests.  Also the commit adds the definition of the `ltp-mm` test plan, it doesn't just enable it on some devices.

gctucker (commented):
One last thing, the description of this PR needs to be updated as it doesn't depend on #568 any more, and I think it can be marked as "ready" rather than draft.

gctucker:
The `staging-skip` will be removed once #584 has been merged, then this PR will need to be rebased.

gctucker (approved):
Tested OK on staging:
https://staging.kernelci.org/test/job/kernelci/branch/staging-next/kernel/staging-next-20210127.1/plan/ltp-mm/

The rk3399-gru-kevin and rk3288-rock2-square boards are currently failing to boot with linux-next, so that's expected.

It's looking fine on mainline, where rk3399-gru-kevin is also broken but rk3288-rock2-square is booting:
https://staging.kernelci.org/test/job/kernelci/branch/staging-mainline/kernel/staging-mainline-20210127.0/plan/ltp-mm/

The "unknown" test results appear to all have been in the skiplist.

gctucker:
The rk3399-gru-kevin issue seems to be a recent regression with nfsroot:
https://staging.kernelci.org/test/case/id/60119fab4d29a765cc4f35fc/

But this might be an intermittent issue.  We have already enabled other LTP test plans on it so I think we should leave ltp-mm enabled on it as well for now, and see if the nfsroot issue can get resolved.

gctucker:
It passed 100% on stable linux-5.4.y (except skipped test cases of course):
https://staging.kernelci.org/test/job/kernelci/branch/staging-stable/kernel/staging-stable-20210128.0/plan/ltp-mm/

PR #570, opened 2020-12-15
templates: boot: generic-qemu: Add console params

MyleneJ:
Hello,

Here is a pull-request to add the possibility to set the console used by QEMU
using 'params' keyword in test-configs.yaml.
The default console will be the value of 'console_dev' entry, as done previously.

Thanks to that, if the emulated device has a different console than the one defined in the top of the file, it is possible to modify it using this `params: console:` entry.

Let me know what you think about it.
Best regards,
Mylène

### DIFF
--- a/templates/boot/generic-qemu-boot-template.jinja2
+++ b/templates/boot/generic-qemu-boot-template.jinja2
@@ -1,15 +1,15 @@
 {% extends 'base/kernel-ci-base.jinja2' %}
 {% if arch == "arm" %}
-{% set console_dev = 'ttyAMA0' %}
+{% set console_dev = console_dev|default('ttyAMA0') %}
 {% endif %}
 {% if arch == "arm64" %}
-{% set console_dev = 'ttyAMA0' %}
+{% set console_dev = console_dev|default('ttyAMA0') %}
 {% endif %}
 {% if arch == 'x86_64' %}
-{% set console_dev = 'ttyS0' %}
+{% set console_dev = console_dev|default('ttyS0') %}
 {% endif %}
 {% if arch == 'i386' %}
-{% set console_dev = 'ttyS0' %}
+{% set console_dev = console_dev|default('ttyS0') %}
 {% endif %}
 {% block metadata %}
 {{ super() }}


montjoie:
Do you have an example where this is needed ?

MyleneJ:
> Do you have an example where this is needed ?

Hello,
Yes, I needed it to boot on QEMU a Nanopi M1+ (Allwinner H3) image which is using `ttyS0` console and not `ttyAMA0`.

montjoie:
For testing, I have added a qemu cubieboard (I didnt found any nanopi in qemu machines).
I confirm console settings should be changed.
But, I think adding a new keyword "console" should be avoided.
Perhaps it is best to reuse console_dev and move the |default directly to the definitions:
 {% if arch == "arm" %}
-{% set console_dev = 'ttyAMA0' %}
+{% set console_dev = console_dev|default('ttyAMA0') %}

You could also add a second commit adding a machine which use it. (for testing purpose)
cubieboard is a good option since it is supported since a long time.
  qemu_arm-cubieboard:
    base_name: qemu
    mach: qemu
    arch: arm
    context:
      arch: 'arm'
      machine: 'cubieboard'
      cpu: 'cortex-a8'
      memory: 512
      model: 'model=allwinner-emac'
      guestfs_interface: 'none'
      extra_options: ['-device ide-hd,drive=lavatest']
    params:
      console_dev: 'ttyS0'
    dtb: 'sun4i-a10-cubieboard.dtb'
    boot_method: qemu

MyleneJ:
> For testing, I have added a qemu cubieboard (I didnt found any nanopi in qemu machines).
> I confirm console settings should be changed.
> But, I think adding a new keyword "console" should be avoided.
> Perhaps it is best to reuse console_dev and move the |default directly to the definitions:
> {% if arch == "arm" %}
> -{% set console_dev = 'ttyAMA0' %}
> +{% set console_dev = console_dev|default('ttyAMA0') %}
> 
> You could also add a second commit adding a machine which use it. (for testing purpose)
> cubieboard is a good option since it is supported since a long time.
> qemu_arm-cubieboard:
> base_name: qemu
> mach: qemu
> arch: arm
> context:
> arch: 'arm'
> machine: 'cubieboard'
> cpu: 'cortex-a8'
> memory: 512
> model: 'model=allwinner-emac'
> guestfs_interface: 'none'
> extra_options: ['-device ide-hd,drive=lavatest']
> params:
> console_dev: 'ttyS0'
> dtb: 'sun4i-a10-cubieboard.dtb'
> boot_method: qemu

Hello,
Thanks for the review.
Sure, I will add the cubieboard as a testing device and update my patch using `console_dev` overriding and not a new `console` params.
Best regards,
Mylène

gctucker:
LGTM, let's see if this works as expected on staging.

gctucker (commented):
Well it's working in `lab-clabbe`:
https://staging.kernelci.org/test/plan/id/5ffdcfa8aca52e5554a07594/
but failing in all the other labs (put `cubie` in the filter and show all results):
https://staging.kernelci.org/test/job/kernelci/branch/staging-next/kernel/staging-next-20210112.0/plan/baseline/

So I guess the change is fine, but we'll also need to be able to require Docker support in the labs to enable it, and provide a Docker image that has support for `cubieboard`.

montjoie:
No in fact, it does not work in my lab, (no public qemu in my lab).
qemu cubieboard fail on USB stuff.
So perhaps it is better to merge only the console stuff and not the cubieboard example.

gctucker:
Ah yes sorry, the results I found from your lab are for the real `sun7i-a20-cubieboard2` board, not `qemu_arm-cubieboard`.  So yes I think we can split this PR into 2, to only have the change with the serial console parameter and another PR later to enable models like `cubie` when we have a Docker image with a QEMU build that has support for it enabled.

MyleneJ:
Hello,
I updated the pull-request and removed the `qemu-cubieboard` commit.
Best regards,
Mylène

gctucker (approved):
OK this is fine now.  We can add board models later, when we're providing QEMU Docker images.

PR #571, opened 2020-12-16
jenkins: remove buildroot-builder.sh

gctucker:
Remove jenkins/buildroot-builder.sh and push-source.py as they are now being added to kernelci-jenkins instead.

Depends on https://github.com/kernelci/kernelci-jenkins/pull/42


gctucker:
Tested OK with hew buildroot build in production.

PR #573, opened 2020-12-22
T24749 kci rootfs alpha

gctucker:
Add checks for sorting configuration in `rootfs-configs.yaml` alphabetically and fix issues.


mgalka (approved):
Looks good to me

PR #581, opened 2021-01-08
T25447 scripts: create directory with standalone scripts for bisection

gctucker:
Create a "scripts" directory to keep the standalone Python scripts
used by the bisection job.  This is an intermediate step until the
functionality included in those scripts becomes part of the main
kernelci Python module.


gctucker:
Tested OK on staging, bisection jobs were run successfully with the new scripts directory.

PR #583, opened 2021-01-12
templates: add kselftest subset parameters

alexandrasp:
includes job template parameters to specify kselftest collections and tests to run.

### DIFF
--- a/templates/kselftest/kselftest.jinja2
+++ b/templates/kselftest/kselftest.jinja2
@@ -1,7 +1,6 @@
 - test:
     timeout:
-      minutes: 55
-
+      minutes: {{ job_timeout }}
     definitions:
     - from: inline
       repository:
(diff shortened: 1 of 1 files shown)


gctucker (commented):
The timeout is fixed to 55 minutes in `kselftest.jinja2`, I think that would need to become a parameter like in `ltp.jinja2`.  Also the name in `kselftest.jinja2` is hard-coded to `kselftest`, but it would need to use the test plan name instead (also like in `ltp.jinja2`).

gctucker:
Why did you remove the `{{ plan }}` parameter in the template?

gctucker:
This has now been deployed on staging, waiting for some results.

alexandrasp:
Tested OK on staging with some devices in collabora and baylibre labs, variables received right values and ran the right kselftest collection.
- [rk3288-veyron-jaq](https://staging.kernelci.org/test/plan/id/600be763e4f4a02a1674f907/) 
- [hip07-d05](https://staging.kernelci.org/test/plan/id/600be78e37db24a5c974f8fe/)

gctucker (approved):
LGTM, let's wait for a few more results today and then it should be good to get merged.

gctucker:
Tested OK on staging:
https://staging.kernelci.org/test/job/kernelci/branch/staging-next/kernel/staging-next-20210125.0/plan/kselftest-lib/

PR #586, opened 2021-01-22
T25969 config: move all LAVA job templates to config/lava

gctucker:
Move all the jinja2 LAVA job template files to a new directory in
config/lava to be able to better separate them from other types of
templates (i.e. k8s, or other lab types).

Update kernelci.lab.lava and sleep definition accordingly.

PR #589, opened 2021-01-23
T24646 sleep: add wakeup check and abort if checks fail

gctucker:
Move the check for the presence of the rtc0 device outside of the main
loop, and add a check for wakeup enabled.  Abort if any of these
checks fail before starting the main loop to keep the script easier to
maintain.

Depends on #586

### DIFF
--- a/config/lava/sleep/sleep.sh
+++ b/config/lava/sleep/sleep.sh
@@ -1,18 +1,30 @@
 #!/bin/sh
 
 modes=$*
-ls -l /dev/rtc0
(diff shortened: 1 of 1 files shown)


gctucker:
Tested OK on staging: https://staging.kernelci.org/test/plan/id/6011a18de0a07181104f364e/

gctucker:
Tested OK on staging again with templates in `config/lava`:
https://staging.kernelci.org/test/case/id/602480418e21f804054e6d21/

PR #590, opened 2021-01-25
T25987 kernelci.lab.lava: fix and simplify logic to handle LAVA aliases

gctucker:
The code to get the list of aliases did not work as expected, all the
entries looked like this:

        "aliases": [
            {
                "device_type": "",
                "name": "am57xx-beagle-x15"
            },

Fix it by using the scheduler.device_types.aliases XMLRPC entry point
for each device type present in the lab, and create a plain dictionary
to map the aliases with the standard device type names used by
KernelCI.

Drop the legacy standalone function get_device_type_by_name and use a
local method in the LAVA class instead to look up the device type used
in a lab if it's present in the aliases dictionary.


gctucker:
Tested OK on staging, for example with lab-linaro-lkft:
```json
    "devices": {
        "aliases": {
            "am57xx-beagle-x15": "x15",
            "i386": "x86",
            "qemu_arm": "qemu-arm",
            "qemu_arm64": "qemu-arm",
            "qemu_i386": "qemu",
            "qemu_x86_64": "qemu",
            "x86_64": "x86"
        },
        "device_type_online": {
            "dragonboard-410c": false,
            "dragonboard-845c": false,
            "hi6220-hikey-r2": false,
            "juno-r2": false,
            "qcom-qdf2400": false,
            "qemu": true,
            "qemu-arm": false,
            "x15": false,
            "x86": false
        }
    },
```

gctucker:
Tested also OK in LKFT lab: https://lkft.validation.linaro.org/scheduler/job/2213008

PR #591, opened 2021-01-26
Fastboot docker

montjoie:
We need to convert fastboot jobs from LXC to docker.
I have kept montjoie/kci-fastboot in patch for letting staging ran.
I will change it when kci docker images will be built.

### DIFF
--- a/config/docker/build-and-push.sh
+++ b/config/docker/build-and-push.sh
@@ -58,6 +58,7 @@ coccinelle \
 cvehound \
 debos \
 dt-validation \
+fastboot \
 kernelci \
 k8s \
 qemu \
--- a/config/docker/fastboot/Dockerfile
+++ b/config/docker/fastboot/Dockerfile
@@ -1,0 +1,3 @@
+FROM ubuntu:bionic
+
+RUN apt-get update && apt-get -y install cpio mkbootimg android-tools-fastboot android-tools-adb && apt-get clean
--- a/config/lava/baseline/generic-fastboot-baseline-template.jinja2
+++ b/config/lava/baseline/generic-fastboot-baseline-template.jinja2
@@ -2,8 +2,6 @@
 {% block actions %}
 {{ super () }}
 
-{% set test_namespace = "target" %}
-
 {% include 'baseline/baseline.jinja2' %}
 
 {% endblock %}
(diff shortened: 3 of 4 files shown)


gctucker:
Thanks, are there any fastboot tests already run on staging (with LXC or the new Docker)?

gctucker:
Meanwhile I'll update the periodic job on staging to also build this new Docker image and put a comment when it's available.

montjoie:
Production run lxc jobs already. (so staging also).
Note that docker jobs fail but for unrelated topic, a bug in network on next/5.10.

gctucker:
Skipping on staging for now as this is causing some issues with the automated job that pushes Docker images, because there's no permission set for the new image yet.  So an image needs to be built by hand and pushed, then permissions can be set, and this can be enabled again in the staging automated job.

jenkins/dockerfiles/build-and-push.sh
```diff
@@ -121,3 +122,15 @@ then
     docker push $tag
   fi
 fi
+
+if [ "x${fastboot_docker}" == "xtrue" ]
+then
+  tag=${tag_px}fastboot_docker
```
gctucker, line 128:
```suggestion
  tag=${tag_px}fastboot
```
It's obviously a Docker image, no need to add `docker` in the name.

gctucker: (same thing in other places)

templates/boot-fastboot/generic-fastboot-boot-template.jinja2
```diff
@@ -6,45 +6,27 @@
...
 
 - deploy:
     timeout:
       minutes: 40
-    namespace: target
-    to: download
+    to: downloads
+    postprocess:
+      docker:
+        image: montjoie/kci-fastboot
```
gctucker, line 19: Please set it to the definitive name, i.e. `kernelci/fastboot`.  We can have a patch on staging to use `kernelci/staging-fastboot` instead.

gctucker:
I've put the `staging-skip` label again because there is no `fastboot` tag in the LAVA lab and this is causing errors, leading to other jobs not being run either.

gctucker:
@montjoie Hello, please provide an update on the situation with `fastboot` LAVA tags and rebase the branch, otherwise I think we'll have to drop this PR.

montjoie:
DUT which can do fastboot are in bad condition and I cannot go to office before next week.
I will restore them and check this PR against them.

aliceinwire:
@montjoie can you update this pr ?

gctucker:
No updates for about 10 months and the Dockerfile would need to be rewritten with the new Jinja2 templates in `config/docker` if this was still relevant.

PR #593, opened 2021-01-26
build-configs.yaml: disable arc builds with clang

gctucker:
The arc architecture is not actually supported by clang, this was most
likely enabled by mistake.  Drop them for now until the architecture
can at least get built in theory so that KernelCI results become
useful.


gctucker:
Note: This change removes the parameters that would be used if any build was configured to be run for `arc` with clang.  However, this wasn't actually used anywhere as all the build configurations for clang on mainline or next only specified other architectures to be built.

PR #598, opened 2021-02-02
build.py: add HTTP user-agent

khilman:
Add User-Agent to HTTP requests.

This was requested by kernel.org sysadmins for experiments with
fetching kernel tar archives from kernel.org.

### DIFF
--- a/kernelci/build.py
+++ b/kernelci/build.py
@@ -347,7 +347,10 @@ def push_tarball(config, kdir, storage, api, token):
 
 
 def _download_file(url, dest_filename, chunk_size=1024):
-    resp = requests.get(url, stream=True)
+    headers = {
+        'User-Agent': 'kernelci 1.1'
+    }
+    resp = requests.get(url, stream=True, headers=headers)
     if resp.status_code == 200:
         with open(dest_filename, 'wb') as out_file:
             for chunk in resp.iter_content(chunk_size):


kernelci/build.py
```diff
@@ -347,7 +347,10 @@ def push_tarball(config, kdir, storage, api, token):
 
 
 def _download_file(url, dest_filename, chunk_size=1024):
-    resp = requests.get(url, stream=True)
+    headers = {
+        'User-Agent': 'kernelci-core build.py agent 1.0'
```
gctucker, line 351: We should probably be using the Python package name and version, i.e. `kernelci 1.1`.  At the moment that's all defined in `setup.py` but we could have the version number in `kernelci/__init__.py` and import it here.

gctucker:
@khilman Hello, please see my comment above.  Does that make sense?  I think it should be easy to merge this one but let's make sure we pick something that's as future-proof as we currently can.

khilman:
It make sense, I used the same name/version from `setup.py`, but I don't know exactly how this should be done cleaner to not duplicate strings.  Maybe you can do that as a follow-up patch?

gctucker:
Thanks, yes if this produced the right header it should be fine to merge it and then the version can be moved to a central place in `kernelci.__version__` as a follow-up.

gctucker:
Tested OK on staging:
```
storage.staging.kernelci.org-access.log:35.188.211.103 - - [26/Apr/2021:08:31:58 +0000] "GET /kernelci/staging-stable/staging-stable-20210426.0/linux-src_kernelci_staging-stable.tar.gz HTTP/1.1" 200 170809149 "-" "kernelci 1.1"
```
I'll make a follow-up PR to import the version number from `kernelci` later today.

gctucker:
See #646

PR #611, opened 2021-02-24
T25987 kernelci.lab: pass DeviceType object to .device_type_online()

gctucker:
Pass a DeviceType configuration object to Lab.device_type_online() and
let the implementation of the method decide how to look up the status
of the device type.  This is because the base name is used with LAVA
labs, so using DeviceType.name doesn't work.  Other types of lab may
need the specific name instead, and in any case this is not something
the API users should have to deal with.

This fixes issues with submitting jobs to qemu and r8a77960-ulcb
devices in particular.

Fixes: 9ba487d2f861 ("kernelci.lab.lava: fix and simplify logic to handle LAVA aliases")

### DIFF
--- a/kci_test
+++ b/kci_test
@@ -93,7 +93,7 @@ class cmd_list_jobs(Command):
         configs = kernelci.test.match_configs(
             test_configs['test_configs'], bmeta, dtbs, lab)
         for device_type, plan in configs:
-            if not api.device_type_online(device_type.name):
+            if not api.device_type_online(device_type):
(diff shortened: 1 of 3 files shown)


gctucker:
Tested OK on staging: https://staging.kernelci.org/test/job/kernelci/branch/staging-next/kernel/staging-next-20210226.0/plan/baseline/

The QEMU and `r8a77960-ulcb` devices were tested as expected.

PR #612, opened 2021-02-26
T25311 push kernel

gctucker:
Refactor the `kci_build push_kernel` implementation to use the new meta-data based on `kenrelci.build.Step`.

Note: Other changes for `kci_build publish_kernel` and in `kernelci-backend` are required before this can be enabled on staging, hence the `staging-skip` tag.


gctucker:
Tested OK locally.

PR #613, opened 2021-03-02
k8s: build kernels using normal storage

khilman:
Currently, we build everything in a memory-backed tmpfs using a volume
with the properties:

  emptyDir: { medium: "Memory" }

However, k8s limits the max size of tmpfs volumes to 50% of the memory
of the node.  The smallest nodes we use have 16G of memory, leading to
a max size tmpfs of 8Gb.

Some kernel builds (notably arm64 defconfig using CONFIG_DEBUG_INFO=y)
can have build artifacts that, combined with the kernel source, exceed
this limit.

To workaround that, don't build the kernel in tmpfs, instead just use
the default storage, backed by whatever storage is configured for the
cluster (typically an SSD disk.)

### DIFF
--- a/config/k8s/job-build.jinja2
+++ b/config/k8s/job-build.jinja2
@@ -33,7 +33,7 @@ spec:
           name: scratch-volume
 
         command: ["/bin/bash", "-x", "-c"]
-        args: ["cd /scratch && git clone --depth=1 --branch ${KCI_CORE_BRANCH} ${KCI_CORE_URL}"]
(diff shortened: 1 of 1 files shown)


khilman:
Manually tested with the same job/cluster that is failing here: https://bot.kernelci.org/job/kernel-build/385988/consoleFull
And it's  running to completion.

broonie:
Looks sensible to me.

gctucker (approved):
Tested OK on staging.

PR #614, opened 2021-03-02
T25311 publish kernel

gctucker:
Update implementation to submit the kernel build meta-data to the backend:
* move the "publish_kernel" from kci_build to kci_data since it's sent via the Database abstract interface.
* use the MetaStep class to load the build meta-data and pass it to the Database.submit_build() method
* generalise the Database.submit() method by including the "path" for the data in the payload (e.g. "test", "build"...)

This is still a work in progress as ideally it should be possible to just use the Database.submit() method.  Things may be simplified when all the components have been updated with the new meta-data format.

Depends on #612.

### DIFF
--- a/config/core/db-configs.yaml
+++ b/config/core/db-configs.yaml
@@ -2,12 +2,12 @@ db_configs:
 
   kernelci.org:
     db_type: kernelci_backend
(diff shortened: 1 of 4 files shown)


gctucker:
Tested OK locally.

PR #616, opened 2021-03-04
test-configs: Add stm32mp157c-lxa-mc1

mgrzeschik:
This patch adds support for stm32mp157c-lxa-mc1
LAVA device-type is upstreamed: https://git.lavasoftware.org/lava/lava/-/merge_requests/1438

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -1271,6 +1271,13 @@ device_types:
     filters:
       - blocklist: {defconfig: ['stm32_defconfig']}
 
+  stm32mp157c-lxa-mc1:
+    mach: stm32
+    class: arm-dtb
+    boot_method: barebox
+    filters:
+      - blocklist: {defconfig: ['stm32_defconfig']}
+
   sun4i-a10-olinuxino-lime:
     mach: allwinner
     class: arm-dtb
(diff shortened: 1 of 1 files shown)


config/core/test-configs.yaml
```diff
@@ -1271,6 +1271,13 @@ device_types:
     filters:
       - blocklist: {defconfig: ['stm32_defconfig']}
 
+  stm32mp157c-lxa-mc1:
+    mach: stm32
+    class: arm-dtb
+    boot_method: barebox
+    filters:
+      - blocklist: {defconfig: ['stm32_defconfig']}
```
gctucker, line 1279: Out of interest, why not test this defconfig on this platform?  (I see it's the same for `stm32mp157c-dk2`).

gctucker:
Tested OK on staging: https://staging.kernelci.org/test/plan/id/604501b3c791b74611448d16/

gctucker (approved):
OK this can be merged now, and the blocklist is the same as for the other stm32 board so no big deal.  Still it would be good to clarify why this is needed, @mgrzeschik ?

config/core/test-configs.yaml
mgrzeschik: I used the dk2 as the template. What is the reason the dk2 is having it in the blocklist?

PR #617, opened 2021-03-04
docker: Add clang-12 containers

broonie:
clang-12 is due for release any day now so let's add clang-12 containers
so we can use them to enable build testing with it.


gctucker:
@broonie How about updating `build-configs.yaml` to use this container?

khilman (approved):

gctucker:
Tested OK on staging alongside #625.

PR #620, opened 2021-03-12
T25311 artifacts

gctucker:
Rework how `artifacts.json` is created so it can be consumed by the backend.  Also send all the meta-data directly to the backend when publishing a new kernel build.

Depends on #614 and https://github.com/kernelci/kernelci-backend/pull/281


gctucker:
Tested OK locally.

PR #629, opened 2021-03-30
T25741: Enable kselftests on prod

alexandrasp:


config/core/test-configs.yaml
```diff
@@ -2078,15 +2095,24 @@ test_configs:
     test_plans:
       - baseline_qemu
       - baseline-qemu-docker
+      - kselftest-filesystems
+      - kselftest-futex
+      - kselftest-lib
```
gctucker, line 2100: A template is missing for running kselftest on QEMU, see this job error:
```
Template not found: config/lava/kselftest/generic-qemu-tftp-nfs-kselftest-template.jinja2
```
One issue here is that it's going to be a bit complicated to specify a QEMU template name for each kselftest collection.  So it might be better to drop QEMU from this PR and make another one just for that, so we can merge this one more easily.

gctucker (changes requested):
The QEMU jobs are causing some problems, I think we should drop it from this PR.

config/core/test-configs.yaml
alexandrasp: ok, qemu removed from this PR. I will open a new soon :)

gctucker (approved):

alexandrasp:
Tested OK on staging for some days:

hip07-d05 :
https://staging.kernelci.org/soc/hisilicon/job/kernelci/kernel/staging-mainline-20210422.0/
https://staging.kernelci.org/soc/hisilicon/job/kernelci/kernel/staging-mainline-20210421.0/

minnowboard:
https://staging.kernelci.org/soc/x86/job/kernelci/kernel/staging-mainline-20210426.0/
https://staging.kernelci.org/soc/x86/job/kernelci/kernel/staging-mainline-20210425.0/
https://staging.kernelci.org/soc/x86/job/kernelci/kernel/staging-mainline-20210424.0/

PR #630, opened 2021-04-01
T25311 bmeta fixup

gctucker:
Add extra optional arguments to `RevisionData.run()` for when building a kernel source tree that is not a Git repository (e.g. a extracted from a source tarball).


gctucker:
Tested OK locally.

PR #636, opened 2021-04-02
Use new build meta-data

gctucker:
Full integration of the new build meta-data with k8s builds and generated LAVA jobs.

Should be merged at the same time as the following PRs:
- [x] https://github.com/kernelci/kernelci-frontend/pull/140 "Build logs" on frontend
- [x] https://github.com/kernelci/kernelci-backend/pull/281 build meta-data parsing in backend
- [x] https://github.com/kernelci/kernelci-jenkins/pull/52 Jenkins jobs updates to match command line changes

### DIFF
--- a/kci_build
+++ b/kci_build
@@ -361,7 +361,8 @@ class cmd_make_config(MakeCommand):
     step_cls = kernelci.build.MakeConfig
 
     def _run_step(self, step, args, configs):
-        return step.run(args.defconfig, args.j, args.verbose)
+        frags_config = configs['fragments']
+        return step.run(args.defconfig, frags_config, args.j, args.verbose)
 
 
 class cmd_make_kernel(MakeCommand):
--- a/kci_build
+++ b/kci_build
@@ -306,7 +306,7 @@ or
             return False
 
         step = kernelci.build.RevisionData(args.kdir, args.output, reset=True)
-        print("Initialising build meta-data in {}".format(args.output))
+        print("Initialising build meta-data in {}".format(step.output_path))
         res = step.run(tree_name, tree_url, branch,
                        args.commit, args.describe, args.describe_verbose)
 
--- a/kci_build
+++ b/kci_build
@@ -440,9 +440,10 @@ class cmd_push_kernel(Command):
     opt_args = [Args.output, Args.db_config]
 
     def __call__(self, configs, args):
-        meta = kernelci.build.MetaStep(args.kdir, args.output)
-        publish_path = meta.get_value("kernel", "publish_path")
-        artifacts = kernelci.storage.discover_files(meta.install_path)
+        install = kernelci.build.Step.get_install_path(args.kdir, args.output)
+        meta = kernelci.build.Metadata(install)
+        publish_path = meta.get('bmeta', 'kernel', 'publish_path')
+        artifacts = kernelci.storage.discover_files(install)
         print("Upload path: {}".format(publish_path))
         kernelci.storage.upload_files(
             args.api, args.db_token, publish_path, artifacts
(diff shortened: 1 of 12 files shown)


kci_test
```diff
@@ -77,21 +77,31 @@ class cmd_validate(Command):
...
     def __call__(self, test_configs, lab_configs, args):
-        bmeta, dtbs = kernelci.build.load_json(args.bmeta_json, args.dtbs_json)
+        path_args = (args.build_output, args.install_path)
+        if not any(path_args) or all(path_args):
+            print("Either --build-output or --install-path is required")
+            return False
+
+        install = (
+            args.install_path
+            or kernelci.build.Step.get_install_path(None, args.build_output)
```
mgalka, line 92: I like how or is concise, but I guess that more pythonic way would be to use 
`args.install_path if args.install_path else kernelci.build.Step.get_install_path(None, args.build_output)`

kernelci/storage.py
```diff
@@ -21,6 +21,24 @@
...
+    file paths and open file objects which can then be passed directly to
+    upload_files().
+
+    *path* is the path to the file hierarchy where to look for files
+    """
+    artifacts = {}
+    for root, _, files in os.walk(path):
+        for fname in files:
+            px = os.path.relpath(root, path)
+            artifacts[os.path.join(px, fname)] = open(
```
mgalka, line 37: Does `requests.post()` call care internally about closing the file? If not maybe we should care about closing files in `upload_files()`?

mgalka (changes requested):

kci_build
```diff
@@ -304,8 +306,9 @@ or
             return False
 
         step = kernelci.build.RevisionData(args.kdir, args.output, reset=True)
-        print("Initialising build meta-data in {}".format(args.output))
-        res = step.run(tree_name, tree_url, branch)
+        print("Initialising build meta-data in {}".format(step.output_path))
+        res = step.run(tree_name, tree_url, branch,
```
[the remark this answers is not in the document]
gctucker: Well, I'm not sure how to do that in such a way that it would help.  Each build step takes a different set of arguments, so we could have a dictionary argument and document what goes inside but that would seem like more code complexity for handling something Python already does (i.e. keyword arguments). Simply having `(self, *args, **kwargs)` in the base class would not tell how derived class are used, and they would still have different method signatures.

kernelci/storage.py
gctucker: Well, they'll get closed when their refcount hits 0.  Maybe it should be clarified that the returned value from `discover_files()` should not be used more than once as the file data would have been consumed.

kci_test
gctucker: Fixed

kci_build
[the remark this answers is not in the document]
mgalka: At the moment the base class says clearly that the method takes no arguments. I'd say it'd be better to indicate that it actually can take arbitrary number of arguments.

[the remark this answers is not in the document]
gctucker: Fair enough, I'll update the docstring too.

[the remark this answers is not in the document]
gctucker: I've replaced the `Step.run()` signature and all its derived classes with this:
```python
def run(self, jopt=None, verbose=False, opts=None):
```
Now the `opts` argument is a dictionary with arbitrary options specific to each step implementation.

kernelci/storage.py
gctucker: The docstring already says that it should "return a dictionary with the file paths and open file objects".  So the caller of this function has to close the file (or let it close itself when it goes out of scope).

mgalka (approved):
Looks good

PR #637, opened 2021-04-08
kernelci.build: use short fragment name when listing builds

gctucker:
When listing all the combinations to build, use the short fragment
name rather than the expanded one.  The expanded one will be
determined later when doing the actual kernel build.


gctucker:
Tested OK on staging via the `bmeta-rework` branch.

PR #638, opened 2021-04-09
T25311 drop old build code

gctucker:
Delete code for the old build implementation to only leave the new one with separate build steps.

Depends on #636


mgalka (approved):

gctucker (commented):
Tested OK on staging.

PR #639, opened 2021-04-09
kernelci.build: fix MakeSelftests build result

gctucker:
Fix a mistake in the call to add the build step meta-data which
resulted in the wrong status to be saved.

Fixes: fa346b304840 ("kernelci.build: use Step.name in meta-data")


gctucker:
Tested OK on staging via the `bmeta-rework` branch.

PR #640, opened 2021-04-09
kernelci.cli: handle case where option is not defined

gctucker:
Handle the case where the requested option is not associated with the
command in Options.get().

This doesn't normally happen as the code using Options normally only
looks for options that were defined, but it may happen in some corner
cases when such things can't be determined.  Also, if it does happen
for any reason, it should return None rather than fail with an
exception.

Fixes: df17ad416551 ("kernelci.cli: add Options class to use a settings file")


gctucker:
Tested OK on staging.

PR #643, opened 2021-04-15
Enable kselftest on baylibre lab

alexandrasp:
Enable kselftest on devices present in lab-baylibre.

### DIFF
--- a/config/core/lab-configs.yaml
+++ b/config/core/lab-configs.yaml
@@ -9,7 +9,9 @@ labs:
       - passlist:
           plan:
             - baseline
-            - kselftest
+            - kselftest-filesystems
+            - kselftest-futex
+            - kselftest-lib
             - preempt-rt
 
   lab-broonie:
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -1804,6 +1804,7 @@ test_configs:
     test_plans:
       - baseline
       - baseline-nfs
+      - kselftest-lib
       - ltp-crypto
       - ltp-ipc
       - ltp-mm
@@ -1923,6 +1924,7 @@ test_configs:
     test_plans:
       - baseline
       - baseline-nfs
+      - kselftest-lib
 
   - device_type: meson-g12b-odroid-n2
     test_plans:
@@ -2035,6 +2037,7 @@ test_configs:
   - device_type: panda
     test_plans:
       - baseline
+      - kselftest-lib
 
   - device_type: panda-es
     test_plans:
@@ -2256,6 +2259,7 @@ test_configs:
   - device_type: sun50i-h6-pine-h64
     test_plans:
       - baseline
+      - kselftest-lib
       - ltp-crypto
       - ltp-ipc
       - ltp-mm
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -1960,6 +1960,9 @@ test_configs:
   - device_type: meson-gxl-s905x-libretech-cc
     test_plans:
       - baseline
+      - kselftest-filesystems
+      - kselftest-futex
+      - kselftest-lib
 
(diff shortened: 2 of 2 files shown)


gctucker:
Some tests are failing, it would be interesting to compare results between `lab-collabora` and `lab-baylibre`.  This is the set of device types that are online in both labs:
```
{'panda', 'r8a7796-m3ulcb', 'qemu', 'sun50i-h6-pine-h64', 'imx6q-sabrelite', 'meson-g12b-a311d-khadas-vim3'}
```
Could you please add a commit to this PR to enable `kselftest-lib` on all of them (except `qemu`) to gather some data?

gctucker:
Please rebase now that #629 has been merged.

config/core/lab-configs.yaml
```diff
@@ -9,7 +9,9 @@ labs:
       - passlist:
           plan:
             - baseline
-            - kselftest
```
gctucker, line 12: The commit subject should mention lab-baylibre, otherwise it looks like this is about the test plan definition.  For example: `lab-configs.yaml: enable kselftest collections in lab-baylibre`

config/core/test-configs.yaml
```diff
@@ -1974,6 +1974,9 @@ test_configs:
   - device_type: meson-gxl-s905x-libretech-cc
     test_plans:
       - baseline
+      - kselftest-filesystems
```
gctucker, line 1979: The subject of this commit is wrong too, it's about the device types and not the lab config.  So this would seem better: `test-configs.yaml: enable kselftest on some libretech devices`

You can then mention the fact that lab-baylibre has those devices in the commit message.

gctucker (changes requested):
LGTM, with just a couple of minor things to tweak in the commit subjects.

gctucker (approved):

gctucker:
Changes look good now, but let's get more results on staging.  It seems like they're all failing to reach a login prompt in lab-baylibre.

khilman:
At first glance, the failures in lab-baylibre are related to the LAVA worker for some of these devices not having enough resources to unpack/extract the NFSroot.  Looking into that.

Also, for this to work with NFSroot and DHCP, the `ip=` cmdline needs to be modified in the device-type so that only the real, wired ethernet is probed via DHCP.  If you use just `ip=dhcp` then all network interfaces are tried for DHCP, and the kselftest config fragment enables a bunch of network drivers, all of which get attempted with DHCP, causing the boot to take forever and eventually timeout.

Adding @montjoie

gctucker:
@montjoie @khilman There seems to be other kinds of issue in your lab:
http://lava.baylibre.com:10080/scheduler/job/315926

It says the kernel image can't be found, but it's there:
```
$ curl -I http://storage.staging.kernelci.org/kernelci/staging-mainline/staging-mainline-20210512.0/arm64/defconfig+kselftest/clang-10/kernel/Image exists
HTTP/1.1 200 OK
```
The error message is about a proxy connection error, so maybe a broken reverse-proxy (squid)?

config/core/test-configs.yaml
```diff
@@ -2158,6 +2164,7 @@ test_configs:
   - device_type: r8a7796-m3ulcb
     test_plans: &r8a7796-m3ulcb_test-plans
       - baseline
+      - kselftest-lib
```
gctucker, line 2167: Please drop this line, it's conflicting with #672.

gctucker (changes requested):

gctucker:
Skipping to avoid conflict with #672 until comments have been addressed.

config/core/test-configs.yaml
alexandrasp: done.

gctucker:
Skipping as this is conflicting with #696.

PR #647, opened 2021-04-26
build-configs.yaml: more stable trees

sashalevin:
Add stable/stable-rc/stable-queue trees up to 5.20.

Signed-off-by: Sasha Levin <sashal@kernel.org>


gctucker (approved):
Tested OK locally with `kci_build list_configs | grep stable`.

PR #648, opened 2021-04-27
Add config fragment for gemini_defconfig

montjoie:
I have two gemini boards but their bootloader are an old vendor one.
There are no way to pass CMDLINE nor tell where to find initrd easily.

The easiest way is to hack CONFIG_CMDLINE and set initrd address in it.
There are no need to do two build since CONFIG_CMDLINE is not a
sufficiant hack to break a build.

### DIFF
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -148,6 +148,12 @@ fragments:
   debug:
     path: "kernel/configs/debug.config"
 
+  gemini:
+    path: "kernel/configs/gemini.config"
+    configs:
+      - 'CONFIG_CMDLINE="console=ttyS0,19200n8 ip=dhcp initrdmem=0x800000,10M"'
+      - 'CONFIG_CMDLINE_FORCE=y'
+
   kselftest:
     path: "kernel/configs/kselftest.config"
 
@@ -841,6 +847,7 @@ build_configs:
               - 'multi_v7_defconfig+CONFIG_EFI=y+CONFIG_ARM_LPAE=y'
               - 'allnoconfig'
               - 'allmodconfig'
+              - 'gemini_defconfig+gemini'
 
       # clang-10 is the current minium clang version for -next
       clang-10:


gctucker:
FYI this is how I enabled it on staging:

```
$ git remote update origin
$ git checkout origin/staging.kernelci.org
$ ./kci_build list_configs | grep staging
kernelci_staging-mainline
kernelci_staging-next
kernelci_staging-stable
$ ./kci_build list_kernel_configs --build-config=kernelci_staging-next --kdir=~/src/linux | grep "arm\ "
arm multi_v7_defconfig clang-12
arm multi_v7_defconfig gcc-8
```
This shows the list of kernel build combinations for the staging-next build config.  Then I made this patch to enable `gemini_defconfig+gemini`:
```diff
diff --git a/config/core/build-configs.yaml b/config/core/build-configs.yaml
index f81faf42f175..1693b1cee516 100644
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -774,7 +774,11 @@ build_configs:
         fragments: [kselftest]
         architectures:
           arc: *arc_defconfig
-          arm: *arm_defconfig
+          arm:
+            <<: *arm_defconfig
+            fragments: [gemini]
+            filters: [passlist: {defconfig: ['multi_v7', 'gemini']}]
+            extra_configs: ['gemini_defconfig+gemini']
           arm64: *arm64_defconfig
           i386: *i386_defconfig
           mips: *mips_defconfig
```

And now:

```
$ ./kci_build generate_fragments --build-config=kernelci_staging-next --kdir=/home/gtucker/src/linux
kernel/configs/kselftest.config
kernel/configs/gemini.config
$ ./kci_build list_kernel_configs --build-config=kernelci_staging-next --kdir=~/src/linux | grep "arm\ "
arm multi_v7_defconfig clang-12
arm gemini_defconfig+gemini gcc-8
arm multi_v7_defconfig gcc-8
```
The patch above has been added to the list on staging:
https://github.com/kernelci/kernelci-deploy/blob/main/patches/kernelci-core/0007-STAGING-enable-defconfig_gemini-gemini.patch

So staging builds should be using that for now.  When this PR is done, we can remove the patch on staging.

@montjoie Please share any kernel patches you may have for the device tree, they can be applied too.

gctucker:
Here's a build: https://staging.kernelci.org/build/id/608adccccf76a6adeb23ad3b/

But there aren't any test results.  @montjoie It looks like we're missing the dtb file, do you have patches to add support for this board that we can try on staging?

montjoie:
I started sending DTS/dt-binding patch upstream for this platform.

gctucker:
Not in kernelci-core, but we can add patches in kernelci-deploy which is where the kernel branches used on staging are being created.  If you send me the patches in any format I can add them there for staging-next builds.

gctucker:
@montjoie Are your device tree patches available anywhere?  Also you said maybe the fragment won't be needed if the initrd address can be set in the device tree, is that right?

montjoie:
I have sent DTB for mainlining. So less backports/hack.
I have postponed a bit since the first device I wanted to do kernelCI on it is not CI friendly at all.
I close this request since I could probably pass all hacks via DTB.

PR #651, opened 2021-04-28
config/core: drop kselftest test-plan

alexandrasp:
It removes the generic kselftest test-plan. We are now focusing on collection approach before to start to run all kselftest in a job/device.


gctucker:
Merged the actual commit by hand to avoid including the changes from #643 that appear on this PR by mistake.

PR #654, opened 2021-04-30
core/test-configs.yaml: Drop few LTP jobs

Lakshmipathi:
To reduce the load on specific devices we dropped
specific ltp testruns from few devices.

From now on, beaglebone-black runs only ltp-ipc.
rk3288-rock2-square runs only ltp-mm and
bcm2836-rpi-2-b runs only ltp-crypto. veyron-jaq
runs only ltp-timers.

Signed-off-by: lakshmipathi ganapathi <lakshmipathi.ganapathi@collabora.com>

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -1666,7 +1666,6 @@ test_configs:
     test_plans:
       - baseline
       - ltp-crypto
-      - ltp-ipc
       - sleep
       - usb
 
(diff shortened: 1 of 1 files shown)


gctucker:
For some reason there's been no results on staging.kernelci.org for a couple of days, taking a look.  This will hopefully get merged in time for the update next week instead.

gctucker:
Staging is back on track but the beaglebone-black devices in lab-collabora have a very long queue, let's wait a bit until we get actual results.

gctucker:
Collabora's lab is rather congested today but this is partly because this PR needs to be merged to reduce the number of jobs being submitted.  So it should be OK to merge it and deploy it (next prod update is either tomorrow or Monday).

PR #656, opened 2021-05-04
T27947 config/k8s: add KCI_DB_CONFIG

gctucker:
Add a KCI_DB_CONFIG parameter to the Kubernetes job-build.jinja2
template to be able to specify the --db-config option to pass to
kci_data rather than having a hard-coded staging.kernelci.org.

Depends on https://github.com/kernelci/kernelci-jenkins/pull/54


gctucker:
Tested OK on staging.

Note: The docker container is now automatically recreated on staging whenever the env file changes (or when there are any other changes in kernelci-jenkins-data).

PR #658, opened 2021-05-05
T27948 kernelci.build: add default make target for MIPS

gctucker:
Add default uImage.gz make target for MIPS otherwise some builds may
not produce any usable kernel image, depending on the defconfig.

### DIFF
--- a/kernelci/build.py
+++ b/kernelci/build.py
@@ -42,6 +42,7 @@ MAKE_TARGETS = {
     'arc': 'uImage',
     'i386': 'bzImage',
     'x86_64': 'bzImage',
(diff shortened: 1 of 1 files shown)


gctucker:
Tested OK on staging: https://staging.kernelci.org/build/next/branch/master/kernel/next-20210505/

There are 2 build failures there but they are for other reasons:
```
../arch/mips/Makefile:284: *** CONFIG_CPU_DADDI_WORKAROUNDS unsupported without -msym32.  Stop.
```
Maybe this flag should be added to build-configs.yaml, still it's a separate issue.

PR #660, opened 2021-05-14
test-configs.yaml: update kseltest test plan to use new rootfs

alexandrasp:
It updates the test configurations to make kselftest test plan use the new buster-kselftest rootfs image.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -49,6 +49,12 @@ file_systems:
     type: debian
     ramdisk: 'buster-igt/20210503.0/{arch}/rootfs.cpio.gz'
 
+  debian_buster_kselftest:
+    type: debian
+    ramdisk: 'buster-kselftest/20210514.0/{arch}/rootfs.cpio.gz'
+    nfs: 'buster-kselftest/20210514.0/{arch}/full.rootfs.tar.xz'
+    root_type: nfs
+
   debian_buster-v4l2_ramdisk:
     type: debian
     ramdisk: 'buster-v4l2/20210503.0/{arch}/rootfs.cpio.gz'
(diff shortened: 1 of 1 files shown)


config/core/test-configs.yaml
```diff
@@ -49,6 +49,12 @@ file_systems:
     type: debian
     ramdisk: 'buster-igt/20210503.0/{arch}/rootfs.cpio.gz'
 
+  debian_buster_kselftest:
+    type: debian
+    ramdisk: 'buster-kselftest/20210512.0/{arch}/rootfs.cpio.gz'
+    nfs: 'buster-kselftest/20210512.0/{arch}/full.rootfs.tar.xz'
```
alexandrasp, line 55: @gctucker I am not sure about this date definition in the case of staging/production. I am trying to use in this PR rootfs generated by me on staging.

gctucker: I see, if you want to use your own staging build you need to add a temporary patch to use the staging image with the staging base URL:
```yaml
  debian-staging:
    url: 'http://storage.staging.kernelci.org/images/rootfs/debian'
```
Then once it has been tested on staging, the patch can be dropped and the PR can be merged.

But we can already build a buster-kselftest image in production, let's see if we can have this today in which case you'll be able to use it directly in this PR.

gctucker: OK here's some rootfs images built from the current `main` branch:
https://storage.kernelci.org/images/rootfs/debian/buster-kselftest/20210514.0/

Maybe try to use them in this PR if you don't need any extra changes?

alexandrasp: Nice, updated to use this rootfs generated by `main` branch.

Let's see if all goes well with the next jobs.

gctucker:
Let's take a look again on Monday or Tuesday once the Collabora lab is back on track with running jobs.

alexandrasp:
some new jobs using new rootfs image: 
https://staging.kernelci.org/test/job/kernelci/branch/staging-mainline/kernel/staging-mainline-20210518.0/plan/kselftest-lib/

gctucker:
@alexandrasp And it's failing to boot, do you know what caused the issue?  I'm pretty sure that platform used to boot fine with kselftest.

gctucker (approved):

PR #672, opened 2021-05-20
test-configs.yaml: remove usb from r8a7796-m3ulcb

alexandrasp:
Drop usb tests from r8a7796-m3ulcb until we can run usb tests only for mainline and linux-next.


gctucker:
Tested OK on staging.

PR #688, opened 2021-05-21
test-configs.yaml: remove ltp-ipc from odroid-xu3

alexandrasp:
Drop ltp-ipc until we add more devices in the lab, it has just one odroid-xu3 and queue is getting longer.


gctucker (approved):
Tested OK on staging, baseline and other test plans got scheduled but not ltp-ipc as expected.

PR #690, opened 2021-05-24
test-configs.yaml: update Debian rootfs URLs

gctucker:
Update all the Debian rootfs URLs to 20210520.0.

Signed-off-by: Guillaume Tucker <guillaume.tucker@collabora.com>

PR #692, opened 2021-05-25
Revert "test-configs.yaml: update kseltest test plan to use new rootfs"

gctucker:
This reverts commit f1fb532033fdc770235074fd835a96803ed7d66c.

The kselftest jobs appear to all be all failing with this rootfs, so
revert to using the plain Buster one instead for now.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -49,12 +49,6 @@ file_systems:
     type: debian
     ramdisk: 'buster-igt/20210503.0/{arch}/rootfs.cpio.gz'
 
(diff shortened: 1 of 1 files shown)


alexandrasp:
I don't think this PR is necessary, the problem in #690 is caused by a typo.

gctucker:
This PR is necessary to get kselftest jobs to run again on staging, but yes if the root cause of the issue is fixed we can then close this PR without merging it.  See my comment on #690 (the issue was introduced earlier when the buster-kselftest rootfs was defined, #690 just changes the date in the URL).

PR #707, opened 2021-05-28
Check skipped kselftests

hardboprobot:
This allows to use a different skipfile for each test set instead of
using a single skipfile for all kselftests. Finer granularity, better
control and better reporting.

Depends on:

- [ ] https://github.com/kernelci/test-definitions/pull/8

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -193,6 +193,7 @@ test_plans:
     params:
       job_timeout: '10'
       kselftest_collections: "futex"
+      kselftest_skipfile: "skipfile-futex.yaml"
 
   kselftest-lib:
     <<: *kselftest
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -193,7 +193,6 @@ test_plans:
     params:
       job_timeout: '10'
       kselftest_collections: "futex"
-      kselftest_skipfile: "skipfile-futex.yaml"
 
   kselftest-lib:
     <<: *kselftest
--- a/config/lava/kselftest/kselftest.jinja2
+++ b/config/lava/kselftest/kselftest.jinja2
@@ -21,6 +21,6 @@
       name: {{ plan }}
       parameters:
         TESTPROG_URL: {{ kselftests_url }}
-        SKIPFILE: {{ kselftest_skipfile }}
+        SKIPFILE: {{ kselftest_skipfile|default("skipfile-lkft.yaml", true) }}
         TST_CMDFILES: {{ kselftest_collections }}
         TST_CASENAME: {{ kselftest_tests }}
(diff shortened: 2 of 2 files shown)


config/lava/kselftest/kselftest.jinja2
```diff
@@ -21,6 +21,6 @@
       name: {{ plan }}
       parameters:
         TESTPROG_URL: {{ kselftests_url }}
-        SKIPFILE: skipfile-lkft.yaml
```
gctucker, line 24: I think we should have a default value defined somewhere, for the kselftest variants that don't define `kselftest_skipfile`.  Maybe this could be done in the template:
```yaml
SKIPFILE: {{ kselftest_skipfile|default("skipfile-lkft.yaml", true) }}
```

config/core/test-configs.yaml
```diff
@@ -193,6 +193,7 @@ test_plans:
     params:
       job_timeout: '10'
       kselftest_collections: "futex"
+      kselftest_skipfile: "skipfile-futex.yaml"
```
gctucker, line 196: Is this file actually defined anywhere?  It would most likely need a PR in the [test-definitions](https://github.com/kernelci/test-definitions) repository to add this file.

alexandrasp: @gctucker it's here: https://github.com/kernelci/test-definitions/pull/8

config/lava/kselftest/kselftest.jinja2
hardboprobot: What you suggested looks good, I'll do the change

config/lava/kselftest/kselftest.jinja2
```diff
@@ -21,6 +21,6 @@
       name: {{ plan }}
       parameters:
         TESTPROG_URL: {{ kselftests_url }}
-        SKIPFILE: {{ kselftest_skipfile }}
+        SKIPFILE: {{ kselftest_skipfile|default("skipfile-lkft.yaml", true) }}
```
hardboprobot, line 24: About this, I forgot to mention that it won't fix the problem with the current reporting, as it will still use the general skipfile if none is supplied. Another option would be to not use a default skipfile and to simply omit this parameter. That will require changes in https://github.com/kernelci/test-definitions/blob/main/automated/linux/kselftest/kselftest.sh but I'm having some trouble testing that.

alexandrasp:
@gctucker @hardboprobot maybe we could close this PR.

gctucker:
The PR looks strange, there are 4 commits including things like specifying a different skipfile for kselftest-futex but the diff only shows the template change with the default file name.

Adding skip labels to this PR and https://github.com/kernelci/test-definitions/pull/8 to see if the results on staging show any tests that should be skipped.  It's not clear whether we really need different skipfiles for each collection, or if this was just a way to avoid the bug with the skipped test cases being always added to the results.

gctucker:
There seems to be an issue with the diff in this PR, also @alexandrasp is now working on fixing test-configurations to produce a correct list of skipped test cases.  Closing for now, although the idea of having a template parameter to specify the skiplist file in each test plan is a valid one and we might implement that at some point in the future.

Issue #712, opened 2021-05-31
kci_build: Support defconfigs from Chromeos


ribalda:
ChromeOS defines its defconfigs as "flavours" in the kernel repository. (Eg: https://chromium.googlesource.com/chromiumos/third_party/kernel/+/refs/heads/chromeos-5.10/chromeos/config/ )

Allow kci_build to create defconfigs base on ChromeOS flavours to have configs like:

```yaml
 next:
    tree: next
    branch: 'master'
    variants:
      chromiumos-qualcomm:
        build_environment: clang-12
        arm64: *arm64_arch
        base_defconfig: 'cros://chromeos-5.10/config/arm64/chromiumos-qualcomm.flavour.config' 
```

gctucker:
> ```yaml
>         base_defconfig: 'cros://chromeos-5.10/config/arm64/chromiumos-qualcomm.flavour.config' 
> ```

Is the `cros://` protocol documented anywhere?

ribalda:
It is not a real network protocol, just a naming convention.

gctucker:
@ribalda We should be able to close this now that #713 has been merged, unless you want to wait for #729 to be merged too?

gctucker:
All merged now, closing this issue.

PR #713, opened 2021-05-31
kernelci: Add support for cros defconfigs

ribalda:
ChromeOS maintain its own defconfigs as splitconfig files in its
repository.

Create a helper that is capable of fetching that configuration and
convert it into a standard oldconfig.

Fixes: #712
Signed-off-by: Ricardo Ribalda <ribalda@chromium.org>

### DIFF
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -237,7 +237,7 @@ build_environments:
   clang-12:
     cc: clang
     cc_version: 12
-    arch_params:
+    arch_params: &clang_12_arch_params
       <<: *clang_arch_params
       riscv:
         <<: *riscv_params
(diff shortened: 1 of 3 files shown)


gctucker:
Thanks, looks good at first glance.  We'll be enabling automatic deployment and test for Chrome OS changes later today, then we'll see if it gets built as expected on chromeos.kernelci.org.

ribalda:
Thanks for taking a look to this. Seems like chromeos.kernelci.org only shows the mainline tree, and I have added my config under next. Will you modify  chromeos.kernelci.org  to support all the configs?

gctucker:
Yes, this will get enabled today.  We can decide which kernel revisions to build, I guess mainline and linux-next would be good ones to start with, using a small subset of build configurations.  But we can adjust that as we see fit.

gctucker:
We've been distracted with other things, here's the issue about enabling automated builds and tests for chromeos.kernelci.org: https://github.com/kernelci/kernelci-deploy/issues/4

gctucker:
Ah, this happened as a side-effect of creating a temporary branch called `chromeos.kernelci.org`.  Sounds like we'll need 2 branches then (one to deploy and one to merge, like we have `staging.kernelci.org` to deploy and PRs get merged in `main`).

PR #715, opened 2021-06-01
build-configs: Add ltp-crypto fragment

crazoes:
Almost all of the ltp-crypto tests are being skipped since the required
kernel config option for crypto algorithms are not enabled.
Hence add ltp_crypto fragment in build-configs.yaml file which would
enable the config options.

Signed-off-by: Shreeya Patel <shreeya.patel@collabora.com>

### DIFF
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -148,6 +148,60 @@ trees:
 
 fragments:
 
+  crypto:
+    path: "kernel/configs/crypto.config"
+    configs:
+      - 'CONFIG_CRYPTO_CHACHA20POLY1305=y'
+      - 'CONFIG_CRYPTO_CHACHA20=y'
+      - 'CONFIG_CRYPTO_POLY1305=y'
+      - 'CONFIG_CRYPTO_AEAD=y'
+      - 'CONFIG_CRYPTO_MANAGER=y'
+      - 'CONFIG_CRYPTO_LIB_CHACHA_GENERIC=y'
+      - 'CONFIG_CRYPTO_SKCIPHER=y'
+      - 'CONFIG_CRYPTO_HASH=y'
+      - 'CONFIG_CRYPTO_LIB_POLY1305_GENERIC=y'
+      - 'CONFIG_CRYPTO_AEAD2=y'
+      - 'CONFIG_CRYPTO_ALGAPI=y'
+      - 'CONFIG_CRYPTO_MANAGER2=y'
+      - 'CONFIG_CRYPTO_SKCIPHER2=y'
+      - 'CONFIG_CRYPTO_HASH2=y'
+      - 'CONFIG_CRYPTO_ALGAPI2=y'
+      - 'CONFIG_CRYPTO_NULL2=y'
+      - 'CONFIG_CRYPTO_RNG2=y'
+      - 'CONFIG_CRYPTO_AKCIPHER2=y'
+      - 'CONFIG_CRYPTO_KPP2=y'
+      - 'CONFIG_CRYPTO_ACOMP2=y'
+      - 'CONFIG_SGL_ALLOC=y'
+      - 'CONFIG_CRYPTO_VMAC=y'
+      - 'CONFIG_CRYPTO_PCRYPT=y'
+      - 'CONFIG_PADATA=y'
+      - 'CONFIG_CRYPTO_HMAC=y'
+      - 'CONFIG_CRYPTO_AES=y'
+      - 'CONFIG_CRYPTO_LIB_AES=y'
+      - 'CONFIG_CRYPTO_SHA256=y'
+      - 'CONFIG_CRYPTO_CBC=y'
+      - 'CONFIG_CRYPTO_MD5=y'
+      - 'CONFIG_CRYPTO_SHA1=y'
+      - 'CONFIG_CRYPTO_AUTHENC=y'
+      - 'CONFIG_CRYPTO_USER=y'
+      - 'CONFIG_CRYPTO_USER_API_HASH=y'
+      - 'CONFIG_CRYPTO_USER_API_SKCIPHER=y'
+      - 'CONFIG_CRYPTO_USER_API_RNG=y'
+      - 'CONFIG_CRYPTO_USER_API_AEAD=y'
+      - 'CONFIG_CRYPTO_SHA3=y'
+      - 'CONFIG_CRYPTO_LIB_CHACHA20POLY1305=y'
+      - 'CONFIG_CRYPTO_DEV_FSL_CAAM_CRYPTO_API=y'
+      - 'CONFIG_CRYPTO_DEV_FSL_CAAM_CRYPTO_API_DESC=y'
+      - 'CONFIG_CRYPTO_AEAD=y'
+      - 'CONFIG_CRYPTO_AUTHENC=y'
+      - 'CONFIG_CRYPTO_SKCIPHER=y'
+      - 'CONFIG_CRYPTO_LIB_DES=y'
+      - 'CONFIG_CRYPTO_XTS=y'
+      - 'CONFIG_CRYPTO_DEV_FSL_CAAM_CRYPTO_API_QI=y'
+      - 'CONFIG_CRYPTO_DEV_FSL_CAAM_CRYPTO_API_DESC=y'
+      - 'CONFIG_CRYPTO_DES=y'
+      - 'CONFIG_CRYPTO_ECB=y'
+
   debug:
     path: "kernel/configs/debug.config"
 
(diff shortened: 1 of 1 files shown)


montjoie (changes requested):
Why do you enable CONFIG_PADATA ? it is a non-user option
Same question for SGL_ALLOC.
Probably all CRYPTO_LIB_XXX should not also be enabled at hand.

You miss also to enable CRYPTO selftests
See https://github.com/kernelci/kernelci-core/pull/511/commits

crazoes:
> Why do you enable CONFIG_PADATA ? it is a non-user option
> Same question for SGL_ALLOC.
> Probably all CRYPTO_LIB_XXX should not also be enabled at hand.
> 
These config options are dependencies hence I enabled them as well.

> You miss also to enable CRYPTO selftests
> See https://github.com/kernelci/kernelci-core/pull/511/commits

Why do I need CRYPTO selftests? I dom't get your point here.
https://github.com/linux-test-project/ltp/tree/master/testcases/kernel/crypto these are the test cases that we are running
in kernelCI. I don't see any readme file describing which configs are needed for them so I checked these test cases and
and tried to enable whatever was needed. Ofcourse there is a possibility of something being missed but I can only say that after seeing logs in staging.

montjoie:
For example, PADATA is not a dependency, it is selected by CRYPTO_PCRYPT.
Same for SGL_ALLOC, nothing "depend on" it.
CRYPTO_CHACHA20 select CRYPTO_LIB_CHACHA_GENERIC and so on....
You need to add in fragment only final needed configs or thoses one where a "depend on" exists.

crazoes:
> For example, PADATA is not a dependency, it is selected by CRYPTO_PCRYPT.
> Same for SGL_ALLOC, nothing "depend on" it.
> CRYPTO_CHACHA20 select CRYPTO_LIB_CHACHA_GENERIC and so on....
> You need to add in fragment only final needed configs or thoses one where a "depend on" exists.

But we need CRYPTO_PCRYPT and whenever we enable it, it selects some other configs which is PADATA.
If we would select it through menuconfig GUI then it would automatically select PADATA as well right?

gctucker:
> > You miss also to enable CRYPTO selftests
> > See https://github.com/kernelci/kernelci-core/pull/511/commits
> 
> Why do I need CRYPTO selftests? I dom't get your point here.

That's right, please don't enable the kselftest config in this fragment.  There's a kselftest fragment for that.  If we need both kselftest and crypto to be enabled, we need to add a build config with kselftest+crypto, but that's for another PR.

gctucker:
I've changed the base branch to `main`, the `staging.kernelci.org` branch is automatically generated using open PRs for the `main` branch.

config/core/build-configs.yaml
```diff
@@ -190,6 +190,42 @@ fragments:
   x86_kvm_guest:
     path: "kernel/configs/kvm_guest.config"
 
+  ltp_crypto:
```
gctucker, line 193: Please rename to `crypto` as this is not necessarily for LTP.  It can be useful for any tests that require crypto algorithms to be enabled.

config/core/build-configs.yaml
```diff
@@ -190,6 +190,42 @@ fragments:
...
+    path: "kernel/config/ltp_crypto.config"
```
gctucker, line 194:
```suggestion
    path: "kernel/configs/ltp_crypto.config"
```

gctucker:
I've created a one-off staging branch with this patch:
https://github.com/kernelci/kernelci-core/commit/05ea97a468de9c182122fd025a64965658ed64b4

Once the comments in this PR have been addressed, we can enable the `crypto` fragment on staging for all regular builds.

gctucker:
I've added a temporary patch on staging to test this, see the results for `staging-mainline-20210602.2` when the build completes.

crazoes:
@gctucker for some reason I don't see all the configs enabled in `staging-mainline-20210602.2`
Also, I believe that arm has different output for test cases than arm64 and x86
So if possible can we run tests on arm and arm64 in staging.

I do not see any effect of my change even though the configs enabled seem right to me as per the test cases.
I will have to investigate more into it.

gctucker:
OK I've added this patch to enable it on arm, arm64 and x86_64 on staging: https://github.com/kernelci/kernelci-deploy/blob/main/patches/kernelci-core/0008-STAGING-enable-crypto-fragment.patch

gctucker:
(with gcc-8 on linux-next only for now)

montjoie:
> > For example, PADATA is not a dependency, it is selected by CRYPTO_PCRYPT.
> > Same for SGL_ALLOC, nothing "depend on" it.
> > CRYPTO_CHACHA20 select CRYPTO_LIB_CHACHA_GENERIC and so on....
> > You need to add in fragment only final needed configs or thoses one where a "depend on" exists.
> 
> @montjoie But we need CRYPTO_PCRYPT and whenever we enable it, it selects some other configs which is PADATA.
> If we would select it through menuconfig GUI then it would automatically select PADATA as well right?

All "select" are done by kconfig, you dont have to do it yourself.
The only CONFIG you need to handle are the "depends on" ones.

So you do not have to take care about SGL_ALLOC, PADATA and co.

crazoes:
@montjoie are you saying that all dependent configs are automatically selected?
I don't think this is true when we manually change the config file. Consider a config option as X which selects Y and then Y selects Z. If you manually try to change .config file in kernel to select X then it won't automatically select Y or Z. We have to do it since we are manually changing the file. That is why it is easier to use menuconfig GUI which would select Y and Z automatically.

montjoie:
> @montjoie are you saying that all dependent configs are automatically selected?
> I don't think this is true when we manually change the config file. Consider a config option as X which selects Y and then Y selects Z. If you manually try to change .config file in kernel to select X then it won't automatically select Y or Z. We have to do it since we are manually changing the file. That is why it is easier to use menuconfig GUI which would select Y and Z automatically.

I think you should test it.
Example on a clean linux:
make x86_64_defconfig
cp .config .config.orig
echo "CONFIG_CRYPTO_PCRYPT=y" >> arch/x86/configs/x86_64_defconfig
make x86_64_defconfig
diff -u /.config.orig .config
--- .config.orig	2021-06-03 13:41:24.909102350 +0200
+++ .config	2021-06-03 14:13:52.114431935 +0200
@@ -813,6 +813,7 @@
 # CONFIG_IOSCHED_BFQ is not set
 # end of IO Schedulers
 
+CONFIG_PADATA=y
 CONFIG_ASN1=y
 CONFIG_INLINE_SPIN_UNLOCK_IRQ=y
 CONFIG_INLINE_READ_UNLOCK=y
@@ -4247,7 +4248,7 @@
 CONFIG_CRYPTO_GF128MUL=y
 CONFIG_CRYPTO_NULL=y
 CONFIG_CRYPTO_NULL2=y
-# CONFIG_CRYPTO_PCRYPT is not set
+CONFIG_CRYPTO_PCRYPT=y
 # CONFIG_CRYPTO_CRYPTD is not set
 CONFIG_CRYPTO_AUTHENC=y
 # CONFIG_CRYPTO_TEST is not set

As you see PADATA was automatically added since it is selected.
If you manually change a .config, you should do some make oldconfig to handle all CONFIGs change.

gctucker:
@crazoes Please update the commit message, it's still referring to the `ltp-crypto` fragment.

gctucker:
Tested OK on staging:
https://staging.kernelci.org/build/kernelci/branch/staging-next/kernel/staging-next-20210603.1/

Builds:
* arm64 https://staging.kernelci.org/build/id/60b8a8026b1b76355b902fd4/
* arm https://staging.kernelci.org/build/id/60b8aafdc95f7ebe6b902fd3/
* x86_64 https://staging.kernelci.org/build/id/60b8ab691e6566b748902fcf/

`ltp-crypto` results:
* arm64 https://staging.kernelci.org/test/plan/id/60b8a90fbc04ff1294902fe0/ (10 pass, 6 skip)
* arm https://staging.kernelci.org/test/case/id/60b8b03dac263550bb902fd6/ (failed to reach a login prompt...)
* x86_64 No results because `ltp-crypto` is only enabled on x86 Chromebooks and they require the `x86-chromebook` fragment as well.  I'll update the staging patch to include a `x86_64_defconfig+x86-chromebook+crypto` build, in the same way that we have one for `kselftest`.

crazoes:
@gctucker

I think the current result has all of them passing except one (af_alg0) which was expected because it uses salsa20 algorithm which is not present in kernel anymore hence I cannot enable it's config.
Also for arm, I don't think login is failing because of the changes we made. It was also happening before and that might be a different problem.

@montjoie I understood your point. Thank you for detailed explanation. I will have to check how exactly KernelCI enables the config. As far as I know, it uses merge_script and it is not necessary that the way how KernelCI is enabling the config would lead to also enabling the dependent config options. So to be on the safer side I enabled the dependent config options as well. But I can dig in more to see if it enables them automatically then we can remove it from fragment. For now, the current config options seems to be correct and we are able to see almost all tests running except one which uses salsa20 algorithm.
https://staging.kernelci.org/test/plan/id/60b92f76c652ee4b96902fe4/

crazoes:
@gctucker just changed the commit message.

gctucker:
OK thanks, we can merge it as-is and improve how the fragment is managed or defined as a follow-up.

gctucker (approved):

PR #718, opened 2021-06-01
kernelci.build: make and install perf

hiwang123:
make linux perf in linux/tools/perf and install the binary as an
artifact.

Signed-off-by: Hsin-Yi Wang <hsinyi@chromium.org>

### DIFF
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -278,15 +278,19 @@ build_environments:
         cross_compile: 'arc-elf32-'
       arm: &arm_params
         cross_compile: 'arm-linux-gnueabihf-'
+        build_perf: true
       arm64: &arm64_params
         cross_compile: 'aarch64-linux-gnu-'
         cross_compile_compat: 'arm-linux-gnueabihf-'
+        build_perf: true
       i386: &i386_params
         name: 'x86'
+        build_perf: true
       mips: &mips_params
         cross_compile: 'mips-linux-gnu-'
       x86_64: &x86_64_params
         name: 'x86'
+        build_perf: true
       riscv: &riscv_params
         name: 'riscv64'
         cross_compile: 'riscv64-linux-gnu-'
@@ -297,9 +301,11 @@ build_environments:
     arch_params: &clang_arch_params
       arm:
         <<: *arm_params
+        build_perf: false
         name:
       arm64:
         <<: *arm64_params
+        build_perf: false
         name:
       i386:
         <<: *i386_params
--- a/config/k8s/job-build.jinja2
+++ b/config/k8s/job-build.jinja2
@@ -94,7 +94,8 @@ set -x; \
 ./kci_build make_kernel && \
 ./kci_build make_modules && \
 ./kci_build make_dtbs && \
-./kci_build make_kselftest; \
+./kci_build make_kselftest && \
+./kci_build make_perf; \
 export KERNEL_BUILD_RESULT=$?;
 \
 set +x; \
--- a/kernelci/build.py
+++ b/kernelci/build.py
@@ -540,7 +540,7 @@ class Metadata:
         """
         self._steps.append(data)
         total_duration = sum(s['duration'] for s in self._steps)
-        all_status = set(s['status'] for s in self._steps)
+        all_status = set(s['status'] for s in self._steps if s['critical'])
         self._bmeta['build'] = {
             'duration': total_duration,
             'status': 'PASS' if all_status == {'PASS'} else 'FAIL'
@@ -726,7 +726,7 @@ class Step:
                 res = False
         return res
 
-    def _add_run_step(self, status, jopt=None, action=''):
+    def _add_run_step(self, status, jopt=None, action='', critical=True):
         start_time = datetime.fromtimestamp(self._start_time).isoformat()
         run_data = {
             'name': ' '.join([self.name, action]) if action else self.name,
@@ -739,6 +739,7 @@ class Step:
             run_data['threads'] = str(jopt)
         if self._log_path and os.path.exists(self._log_path):
             run_data['log_file'] = self._log_file
+        run_data['critical'] = critical
         run_data['status'] = "PASS" if status is True else "FAIL"
         self._meta.add_step(run_data)
         self._meta.save(save_artifacts=False)
@@ -787,8 +788,6 @@ class Step:
     def _get_make_opts(self, opts, make_path):
         env = self._meta.get('bmeta', 'environment')
         make_opts = env['make_opts'].copy()
-        if opts:
-            make_opts.update(opts)
 
         arch = env['arch']
         make_opts['ARCH'] = arch
@@ -818,6 +817,9 @@ class Step:
         if self._output_path and (self._output_path != make_path):
             # due to kselftest Makefile issues, O= cannot be a relative path
             make_opts['O'] = format(os.path.abspath(self._output_path))
+
+        if opts:
+            make_opts.update(opts)
 
         return make_opts
 
@@ -871,7 +873,7 @@ class Step:
         """
         raise NotImplementedError("Step.run() needs to be implemented.")
 
-    def install(self, verbose=False, status=True):
+    def install(self, verbose=False, status=True, critical=True):
         """Base method to install the build artifacts.
 
         The default behaviour is to install bmeta.json and steps.json in the
@@ -881,7 +883,7 @@ class Step:
         *verbose* is whether to show what is being installed
         *status* is True if install commands succeeded, False otherwise
         """
-        self._add_run_step(status, action='install')
+        self._add_run_step(status, action='install', critical=critical)
         files = [
             (self._meta.bmeta_path, '', ''),
             (self._meta.steps_path, '', ''),
@@ -983,6 +985,7 @@ class EnvironmentData(Step):
             'platform': platform_data,
             'use_ccache': shell_cmd("which ccache > /dev/null", True),
             'make_opts': make_opts,
+            'build_perf': build_env.get_build_perf(arch)
         }
         return self._add_run_step(True)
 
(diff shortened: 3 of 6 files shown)


hiwang123:
This is tested with following kernelci.conf:
[DEFAULT]
verbose: true

[kci_build]
kdir: linux
build_env: gcc-8
arch: x86_64
install: true

./kci_build make_perftool

However I only tested one env+arch locally. Can I test this change to see if it can generate artifacts successfully on other env+arch? 

Currently it only generates artifact but didn't install into the image

kernelci/build.py
```diff
@@ -787,8 +787,6 @@ def _output_to_file(self, cmd, file_path, rel_path=None):
     def _get_make_opts(self, opts, make_path):
         env = self._meta.get('bmeta', 'environment')
         make_opts = env['make_opts'].copy()
-        if opts:
-            make_opts.update(opts)
```
eballetbo, line 791: Maybe I'm missing something, but why you're removing this?

eballetbo (commented):
Hey Hsin-Yi. Thank you for contributing at this. I have one question, maybe I miss something, though.

kernelci/build.py
hiwang123: It's moved to below (L820). So we can overwrite O= option.
The reason is, it seems that is we assign O=/some/path/other/than/make_path, the compile will fail. And set O to makefile path is fine. The error:
In file included from bench/../../arch/x86/lib/memcpy_64.S:8,
                 from bench/mem-memcpy-x86-64-asm.S:14:
/root/kernelci-core/linux/build/arch/x86/include/generated/asm/export.h:1:10: fatal error: asm-generic/export.h: No such file or directory
 #include <asm-generic/export.h>
          ^~~~~~~~~~~~~~~~~~~~~~
compilation terminated.

The full step:
./kci_build make_config --defconfig=defconfig+kselftest
./kci_build make_kernel
./kci_build make_perftool

config/k8s/job-build.jinja2
```diff
@@ -94,7 +94,8 @@ set -x; \
 ./kci_build make_kernel && \
 ./kci_build make_modules && \
 ./kci_build make_dtbs && \
-./kci_build make_kselftest; \
+./kci_build make_kselftest && \
+./kci_build make_perftool; \
```
gctucker, line 98: We'll have to see whether we want to build perf with all the kernel builds, or if we should have an option to only select some of them.  Part of that will depend on the available build resources and the extra storage required to store the artifacts.

kernelci/build.py
gctucker: OK this shouldn't be breaking existing use-cases where the `O=` option is not defined for specific builds, and it gives the possibility to override it when needed.  We'll see if all the builds work as expected on staging to confirm this.

config/k8s/job-build.jinja2
gctucker: I see `CONFIG_PERF_EVENTS` is enabled in `multi_v7_defconfig` but not in arm64 defconfig or x86_64_defconfig.  So the `is_enabled()` method will currently result in perf being built only for arm, unless we define a config fragment in `build-configs.yaml` to enable it.  Maybe we could have a fragment to force it to be disabled, and another one to enable it.  Then we could explicitly configure which builds have perf enabled and which ones don't.

hiwang123: Sounds good. I updated this commit to add a new fragment to include a few configs (though they are selected by default defconfig).  now is_enabled() will check if this fragment is presented.
I'm not sure which arch would like to use this, so I added this fragment into default_fragments as well.

gctucker: Actually, in `init/Kconfig`:
```
config PERF_EVENTS
	bool "Kernel performance events and counters"
	default y if PROFILING
	depends on HAVE_PERF_EVENTS
```
And `CONFIG_PROFILING` is enabled in all the arch defconfigs as far as I can tell.  So we'll just need a config fragment to disable it if we want to have builds without perf.  But first, let's see if this builds fine and how much extra resources it takes as maybe we can just have perf always enabled in all the builds.

gctucker: Slight correction: `multi_v7_defconfig` doesn't have `CONFIG_PROFILING=y` but it does have `CONFIG_PERF_EVENTS`...  Same conclusion though.

kernelci/build.py
gctucker: It took about 20s to build on my regular builder with 4 cores, and I believe this is the only artifact we need to store:
```
-rwxr-xr-x 1 root root 14M Jun  2 14:01 tools/perf/perf
```
And after running `strip` on it:
```
-rwxr-xr-x 1 root root 5.7M Jun  2 14:04 tools/perf/perf
```
So I think we can try to just enable it for all the main defconfigs, it's not that big.  Maybe we can just build it for the main defconfig of each architecture if we can configure this easily in YAML.

config/core/build-configs.yaml
```diff
@@ -154,6 +154,13 @@ fragments:
   kselftest:
     path: "kernel/configs/kselftest.config"
 
+  perf:
```
gctucker, line 157: This happens to be actually enabled by default in all configs, so we might need a fragment to disable it.  But for now, please drop this fragment and let's see how much extra load it adds to always build it.

kernelci/build.py
```diff
@@ -1385,6 +1386,60 @@ def install(self, verbose=False):
...
+    def is_enabled(self):
+        """Check whether the perf is enabled
+
+        Return True if perf config fragment is enabled in the build
+        meta-data and CONFIG_PERF_EVENTS is in kernel config, or False
+        otherwise.
+        """
+        return 'perf' in self._meta.get(
+            'bmeta', 'kernel', 'defconfig_extras'
+        ) and self._kernel_config_enabled('PERF_EVENTS')
```
gctucker, line 1404:
```suggestion
        return self._kernel_config_enabled('PERF_EVENTS')
```
Please don't rely on the defconfig_extras, I can see how that would work but then the logic is not consistent with `PERF_EVENTS` being enabled in the `.config`.  So I think it's better to just rely on the `.config` and make sure the `PERF_EVENTS` is there when we want to build `perf`.

kernelci/build.py
```diff
@@ -1385,6 +1386,60 @@ def install(self, verbose=False):
...
+        return self._add_run_step(res, jopt)
+
+    def install(self, verbose=False):
+        """Install the perf binary
+
+        Install the perf binary
+
+        *verbose* is whether the build output should be shown
+        """
+        perf = os.path.join(self._kdir, 'tools/perf/perf')
```
gctucker, line 1428: If the only artifact is this `perf` binary, then we don't need to create a tarball.

kernelci/build.py
```diff
@@ -1385,6 +1386,60 @@ def install(self, verbose=False):
         return super().install(verbose)
 
 
+class MakePerfTool(Step):
```
gctucker, line 1389: How about `MakePerf`, and have a `kci_build make_perf` command?  That's what I typed initially and was confused why it didn't work.  Does anyone usually refer to this as `perftool` rather than `perf`?

gctucker:
I've started a build on staging with a temporary patch to enable it, see the results for `staging-next-20210602.1` when the build completes.

hiwang123:
this currently only builds with:
x86: gcc, clang
arm(64): gcc (with this suggested fix: https://lore.kernel.org/patchwork/patch/1434412/)
arm(64) with clang seems will still enable __x86_64__ macro and builds fail:
https://storage.staging.kernelci.org/kernelci/staging-next/staging-next-20210602.1/arm64/defconfig/clang-12/logs/perf.log

gctucker:
It build OK, for example with gcc-8 on arm64:
https://staging.kernelci.org/build/id/60b799eac30b5c2988902fd0/
https://storage.staging.kernelci.org/kernelci/staging-next/staging-next-20210602.1/arm64/defconfig/gcc-8/logs/perf.log

It took 33s (cross-compiled):
https://storage.staging.kernelci.org/kernelci/staging-next/staging-next-20210602.1/arm64/defconfig/gcc-8/steps.json

gctucker:
So I think this is fine, we just need to include the binary artifacts now in the installation step and this PR should be ready to be merged, if the binary size is not an issue.

hiwang123:
this currently only builds with:
x86: gcc, clang
arm(64): gcc (with this suggested fix: https://lore.kernel.org/patchwork/patch/1434412/)
arm(64) with clang seems will enable __x86_64__ macro and builds fail:
https://storage.staging.kernelci.org/kernelci/staging-next/staging-next-20210602.1/arm64/defconfig/clang-12/logs/perf.log

other arcs will fail. Should we add a fragment that only enables in x86 and arm(64)? And I'm not sure how to deal with arm(64)+clang case.

gctucker:
OK, yes we can enable it only for the combinations of compilers and arches that work.  That's basically what I did on staging, with only the 3 main architectures and only gcc-8 for now.

@hiwang123 What is the reason for specifying `O=`?  Is it not possible to build perf with an alternative output path?

hiwang123:
If we build kernel first then build perf, we'll see this error:
In file included from bench/../../arch/x86/lib/memcpy_64.S:8,
from bench/mem-memcpy-x86-64-asm.S:14:
/root/kernelci-core/linux/build/arch/x86/include/generated/asm/export.h:1:10: fatal error: asm-generic/export.h: No such file or directory
#include <asm-generic/export.h>
^~~~~~~~~~~~~~~~~~~~~~
compilation terminated.

config/k8s/job-build.jinja2
gctucker, line 98:
```suggestion
./kci_build make_perf; \
```

gctucker (changes requested):
Small thing to tweak since the command has been renamed to `make_perf`.

kernelci/build.py
hiwang123: In the latest version, I add a build_perf parameter into build_environments>arch_params. So that we can enable and disable by different arch+cc combination

hiwang123:
In the latest version, I add a build_perf parameter into build_environments>arch_params. So that we can enable and disable by different arch+cc combination.
currently enabled:
{x86_64, i386}{gcc, clang}
{arm64}{gcc}

gctucker:
> In the latest version, I add a build_perf parameter into build_environments>arch_params. So that we can enable and disable by different arch+cc combination.

I guess that works, but it's going to be very difficult to maintain as we keep adding things.  For example, if we did that we would also need to add `build_kselftest`, `build_documentation`...  So maybe we should have a dictionary to tell which steps to build, using the step names and true / false as values instead.

hiwang123:
Using steps names sounds more generic.

gctucker:
Skipping this PR for now to avoid having "failed" build status just because of perf failures.

gctucker:
Dropping skip label to see if this is what caused the problem with results not loading on the web dashboard.

gctucker:
The issue is back with this PR, so adding skip label again.  Seems worth investigating further.

hiwang123:
Previously this didn't work on clang13, but there are a few changes on tools/perf recently. And now it can be built 
 with clang13 successfully. Also the arm with gcc works. 
currently enabled now is:
{x86_64, i386}{gcc, clang*}
{arm64, arm}{gcc}

Should I update the config and try again?

gctucker:
There is still the issue that if for any reason the `perf` build stage fails, the overall build status is also "fail".  So yes we can try again with clang-13, but to have it merged and used in production we need to improve how build stage results are being handled.

Ideally, building each step should be considered another result just like test cases.  Until we have that, we can try to put in place something special for `perf` to build it separately and report the status as a test result.

hiwang123:
This version makes the build perf as non critical. If only non critical steps failed, the overall build is still PASS.
The steps status are still in bmeta.json, and we can add a test to check them. Should the test be in config/lava/ as other kernel tests? Since it won't test anything on the device, simply just check the 'status' in steps

gctucker:
> > In the latest version, I add a build_perf parameter into build_environments>arch_params. So that we can enable and disable by different arch+cc combination.
> 
> I guess that works, but it's going to be very difficult to maintain as we keep adding things. For example, if we did that we would also need to add `build_kselftest`, `build_documentation`... So maybe we should have a dictionary to tell which steps to build, using the step names and true / false as values instead.

@hiwang123 Please see this earlier comment, it's also something to address before we can merge this PR.

gctucker:
> This version makes the build perf as non critical. If only non critical steps failed, the overall build is still PASS.
> The steps status still records in bmeta.json, and we can add a test to check them. Should the test be in config/lava/ as other kernel tests? Since it won't test anything on the device, simply just check the 'status' in steps

I guess this works, but `critical` is a very arbitrary definition.  And with this approach, the result of the perf build isn't actually tracked in the backend.

So instead, I think we should leave the status logic as-is, but build perf only after sending the build meta-data and binaries.  So when sending the build, the result doesn't include building perf.  Then the job builds perf, and sends the result with `kci_data` as a test result.  The main issue with KernelCI build entries is that they have to be plain kernel builds, so anything else needs to be sent as a test.

The whole database schema needs a redesign to have a common base class for each step, and express dependencies between steps.  Until then, let's try to avoid adding features that will then need to be removed such as a `critical` flag which only follows the current definition of a kernel build.

aliceinwire:
@hiwang123 any update on this?

nuclearcat:
As no answer received - closing PR.

PR #720, opened 2021-06-02
Add lc-compliance test

nfraprado:
Signed-off-by: Nícolas F. R. A. Prado <nfraprado@collabora.com>

PR #723, opened 2021-06-02
STAGING Add buster-libcamera

gctucker:


config/core/test-configs.yaml
```diff
@@ -55,6 +58,10 @@ file_systems:
     nfs: 'buster-kselftest/20210520.0/{arch}/full.rootfs.tar.xz'
     root_type: nfs
 
+  debian_buster-libcamera_ramdisk:
+    type: debian-staging
+    ramdisk: 'buster-libcamera/20210602.0/{arch}/rootfs.cpio.gz'
+
```
alexandrasp, line 64: @gctucker libcamera uses nfs, could you include it?

gctucker: No it's still configured to use a ramdisk: https://github.com/kernelci/kernelci-core/pull/720/files#
```yaml
  lc-compliance:
    rootfs: debian_buster-libcamera_ramdisk
    pattern: 'lc-compliance/{category}-{method}-{protocol}-{rootfs}-lc-compliance-template.jinja2'
  ```

gctucker:
Replaced with extra patch in #720.

Issue #710, opened 2021-05-31
New test plan ltp-ima


pevik:
Could you please create new test plan for [LTP IMA tests](https://github.com/linux-test-project/ltp/tree/master/testcases/kernel/security/integrity/ima/)?
They're maintained by me.
It'd be enough to run them on mainline, next tree and [Mimi Zohar's git tree on next-integrity-testing](https://git.kernel.org/pub/scm/linux/kernel/git/zohar/linux-integrity.git/log/?h=next-integrity-testing) (maintainer's branch).

gctucker:
If this test is already built in the buster-ltp rootfs image we're using then this should be trivial.  I've added it to the backlog for this sprint, let's see if someone has the time to take a look.

williamklin:
ltp-ima requires some specific kernel configs to work, do I need to build a new kernel for this, or just to choose whatever device types (in test_configs.yaml)  which already have these configs enabled to run?

gctucker:
Sounds like having a new config fragment would be the best option.  Then we can decide when to build it, and the test plan would depend on the fragment to be in the build (for example, see `v4l2-compliance-vivid` which relies on the `virtualvideo` fragment).

pevik:
Thanks a lot for working on this!
FYI all IMA required kernel options are in https://github.com/linux-test-project/ltp/tree/master/testcases/kernel/security/integrity/ima/. You can probably ignore EVM test just now, as it document unfixed problem on mainline.

PR #724, opened 2021-06-04
Add ltp-ima test plan (#710)

williamklin:
Added a 'ltp-ima' fragment and a 'ltp-ima' test plan.
The test-definition part seems work fine with the latest LTP version by passing 'ima' to tst_cmdfiles.

Don't know if it's the correct way to use 'filters/passlist' in the test_plan to force it using a kernel with ltp-ima fragment.

### DIFF
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -154,6 +154,13 @@ fragments:
   kselftest:
     path: "kernel/configs/kselftest.config"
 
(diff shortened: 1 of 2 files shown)


williamklin:
close because sanity check fails

gctucker:
@williamklin You can re-open this and just fix the issue, it's about alphabetical order in the YAML file.

PR #726, opened 2021-06-04
Add cros defconfigs

gctucker:
Recreate #713 now that we have automated builds on chromeos.kernelci.org and it was previously merged by accident outside of GitHub.


gctucker:
Replaced by #729

PR #727, opened 2021-06-04
build-configs.yaml: add more drivers to x86-chromebook

sonnydeez:
These were needed to boot coral onto a Chrome OS filesystem on emmc.
Some like nvme would be needed for other machines but weren't tested.
The dm_* drivers are need to get some of the filesystems mounted,
otherwise the Chrome OS scripts are unhappy and reboot.  Pstore is
useful for debugging kernel crashes, so I enabled that as well.


gctucker:
Please create a branch in a fork and submit a pull request from there.

Issue #719, opened 2021-06-02
Extend x86_64_defconfig to be able to support more drivers


eballetbo:
The defconfig used to build amd64 kernels in KernelCI is very basic (basically the x86_64_defconfig + small fragment), in fact, is a minimal set of kernel config options. While is quite easy to upstream new config options for other platforms (like arm64 and armhf) it looks hard doing this for x86_64_defconfig.

If we want to have more coverage we need to enable more and more drivers in the kernel configuration. The purpose of this task is add a mechanism to be able to enable more drivers in amd64 kernel configs. As a first step, we can try to enable all the Cros-EC related drivers for Chromebooks.

eballetbo:
From the cros-ec side, this fragment:
```
CONFIG_CHROME_PLATFORMS=y
CONFIG_CROS_EC=m
CONFIG_CROS_EC_LPC=m

CONFIG_IIO=m
CONFIG_IIO_CROS_EC_SENSORS_CORE=m
CONFIG_IIO_CROS_EC_SENSORS=m
```
Makes the cros-ec test results go from 0 tests

* https://lava.collabora.co.uk/results/3945366

to 7 tests passed.

* https://lava.collabora.co.uk/results/3946291

sonnydeez:
So will try to find some more configs that help with octopus -- what are the other targets in the lab? Grunt maybe?

eballetbo:
We have Grunt, Rammus, Sona and Coral too.

gctucker:
@eballetbo Sounds like adding a `cros-ec` fragment?

eballetbo:
It depends on how many fragments you want to deal with. That's one, but there are more to enable sound drivers, tpm, touchpad and touchscreen. Not sure if we should create a fragment for every feature or just a big fragment complementing the x86_64_defconfig.

eballetbo:
Note that for ARM all these configs are includes in the default defconfig in the kernel.

gctucker:
OK if this is intended to only be used on top of `x86_64_defconfig` then we may include it in the `x86-chromebook` fragment.

The only thing is that it's good to have a very minimal fragment to just boot Chromebooks, so we can more easily compare results with other x86 platforms that don't use any fragments.  So maybe we could keep `x86-chromebook` as small as possible, and just have one more e.g. `x86-chromebook-full` with all the options needed to fully enable everything.  That may actually be what #726 is really about though, and we could be building that on the chromeos.kernelci.org instance.

[...]

PR #728, opened 2021-06-04
build-configs.yaml: add more drivers to x86-chromebook

sonnydeez:
These were needed to boot coral onto a Chrome OS filesystem on emmc.
Some like nvme would be needed for other machines but weren't tested.
The dm_* drivers are need to get some of the filesystems mounted,
otherwise the Chrome OS scripts are unhappy and reboot.  Pstore is
useful for debugging kernel crashes, so I enabled that as well.

Fixes: #719 

### DIFF
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -190,6 +190,20 @@ fragments:
       - 'CONFIG_MMC_SDHCI=y'
       - 'CONFIG_MMC_SDHCI_PCI=y'
       - 'CONFIG_MMC_SDHCI_ACPI=y'
+      - 'CONFIG_MMC_BLOCK=y'
+      - 'CONFIG_CHROME_PLATFORMS=y'
+      - 'CONFIG_CHROMEOS_LAPTOP=y'
+      - 'CONFIG_CHROMEOS_PSTORE=y'
+      - 'CONFIG_CHROMEOS_TBMC=y'
+      - 'CONFIG_PSTORE=y'
+      - 'CONFIG_EFI_VARS_PSTORE=y'
+      - 'CONFIG_BLK_DEV_NVME=y'
+      - 'CONFIG_NVME_CORE=y'
+      - 'CONFIG_DM_BUFIO=y'
+      - 'CONFIG_DM_CRYPT=y'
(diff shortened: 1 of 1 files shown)


10ne1:
I can confirm the above changes are required to be able to boot to userspace the mainline kernel on the x86_64 EVE Chromebook, though more changes are required because in my case the Chrome UI crashes as well as attestationd.

@gctucker @eballetbo Can we please merge this?

gctucker:
Thanks for the review.  Let's wait for some staging results today to confirm the kernels built with this fragment don't introduce regressions.

The only potentially issue here is that the `x86-chromebook` fragment is meant to be as small as possible, to reduce the scope for differences with the main defconfig e.g. `x86_64_defconfig` and be able to compare results with other platforms that don't use the fragment.  If some of the configs listed here are only required to be able to boot a Chrome OS user-space, then they should be moved to the `chromeos` branch which is used with the chromeos.kernelci.org instance.

10ne1:
@gctucker Thanks for the context.

This config fragment change could definitely be made smaller, for example by running `make savedefconfig` on it, (eg `CONFIG_MMC_BLOCK=y` is the default and can be omitted).

Some of these changes are also just useful for debugging and should be moved to a `x86-chromebook-full` as you suggested in the issue.

Also these are required to boot to the ChromeOS userspace, so they should be moved to the `chromeos` branch.

Some more changes are required for which I'll post PR (eg for the i2c tpm chip on EVE) and I can pick these up as well if you want me to.

So as is, please do NOT merge this. I'll mark it as a draft.

10ne1:
FYI I created PR: #764 which incorporates the changes from this one. Can you please review and if you're ok we can move the discussion there? Thank you!

gctucker:
I've added the `staging-skip` label as this PR is conflicting with #764.  In fact, #764 is meant to replace this one so it's now going to get tested on staging.kernelci.org.

@sonnydeez Can you please close this PR if you're happy with #764 being merged instead?

gctucker:
Closing while wrapping up the hackfest project board, but please re-open and move to the hackfest 2 board if there is something that's not covered by #764.

PR #731, opened 2021-06-11
lc-compliance fixup

gctucker:
Replace ramdisk LAVA job templates with NFS ones.  Fixes #720.


gctucker:
Cherry-picked into #720

PR #729, opened 2021-06-04
Add cros defconfigs

ribalda:
Recreate #726 to fix the yaml errors.

PR #733, opened 2021-06-14
Replace characters introduced by cros://

gctucker:
Add the colon`:` to the list of characters to replace with dashes in the Kubernetes job names.

Also replace more than just slash `/` characters in the upload file path to avoid issues and keep it readable.

Depends on #729


gctucker:
Tested OK on staging with just the `:` k8s fix:
https://chromeos.kernelci.org/build/kernelci/branch/chromeos-mainline/kernel/chromeos-mainline-20210614.1/

PR #735, opened 2021-06-15
test-configs.yaml: Add seccomp and lkdtm kselftest collections

kees:
Add "seccomp" and "lkdtm" kselftests to the default test configs, since they both have architectural variations.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -210,6 +210,18 @@ test_plans:
       job_timeout: '10'
       kselftest_collections: "lib"
 
+  kselftest-lkdtm:
+    <<: *kselftest
+    params:
+      job_timeout: '10'
+      kselftest_collections: "lkdtm"
+
+kselftest-seccomp:
+    <<: *kselftest
+    params:
+      job_timeout: '10'
+      kselftest_collections: "seccomp"
+
   ltp: &ltp
     rootfs: debian_buster-ltp_nfs
     pattern: 'ltp/{category}-{method}-{protocol}-{rootfs}-ltp-template.jinja2'
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -216,7 +216,7 @@ test_plans:
       job_timeout: '10'
       kselftest_collections: "lkdtm"
 
-kselftest-seccomp:
+  kselftest-seccomp:
     <<: *kselftest
     params:
       job_timeout: '10'


config/core/test-configs.yaml
```diff
@@ -210,6 +210,18 @@ test_plans:
       job_timeout: '10'
       kselftest_collections: "lib"
 
+  kselftest-lkdtm:
+    <<: *kselftest
+    params:
+      job_timeout: '10'
+      kselftest_collections: "lkdtm"
+
+kselftest-seccomp:
```
gctucker, line 219:
```suggestion
  kselftest-seccomp:
```

gctucker (commented):
Just a small indentation glitch to fix.

The new kselftest collections also need to be enabled in test configs to specify which devices they should be run on.  So I've made #736 on top of this one to do just that in Collabora's lab based on current devices' availability.  That's on x86 Intel and AMD, as well as arm64 Rockchip.  We'll see on staging.kernelci.org if the tests run as expected.

config/core/test-configs.yaml
kees: Whoops! I'm not sure how that happened. Thanks for catching that.

gctucker:
s/intendation/indentation/ (in the commit message of my own fixup... I'll clean this up when merging it)

gctucker:
Tested `seccomp` OK on staging in combination with #736:
https://staging.kernelci.org/test/plan/id/60cb785c8d9134a74a24bfa2/

However the `lkdtm` tests all failed:
https://staging.kernelci.org/test/job/kernelci/branch/staging-next/kernel/staging-next-20210617.0/plan/kselftest-lkdtm/
https://storage.staging.kernelci.org/kernelci/staging-next/staging-next-20210617.0/x86_64/x86_64_defconfig+x86-chromebook+kselftest/clang-13/lab-collabora/kselftest-lkdtm-asus-C523NA-A20057-coral.html

@kees Do you know what caused all these failures?

kees:
> https://storage.staging.kernelci.org/kernelci/staging-next/staging-next-20210617.0/x86_64/x86_64_defconfig+x86-chromebook+kselftest/clang-13/lab-collabora/kselftest-lkdtm-asus-C523NA-A20057-coral.html
> 
> @kees Do you know what caused all these failures?

Uh... where did the dmesg output go? OH! Crap. I found it. $SHELL isn't defined. Well that's fun. I will send patches.

kees:
Okay, I've sent a patch to fix the lkdtm tests. As a work-around, if kernelci.org would export "SHELL=/bin/sh" before running tests, that would also solve it. :)

https://lore.kernel.org/lkml/20210617231027.3908585-1-keescook@chromium.org/T/#u

gctucker:
> Okay, I've sent a patch to fix the lkdtm tests. As a work-around, if kernelci.org would export "SHELL=/bin/sh" before running tests, that would also solve it. :)
> 
> https://lore.kernel.org/lkml/20210617231027.3908585-1-keescook@chromium.org/T/#u

Nice, I've added the patch to the staging-next branch so we'll see if it does fix the issue on staging.kernelci.org.

gctucker:
Ah, now `bc` is missing:
```
# ./stack-entropy.sh: 27: ./stack-entropy.sh: bc: not found
```
https://storage.staging.kernelci.org/kernelci/staging-next/staging-next-20210618.0/x86_64/x86_64_defconfig+x86-chromebook+kselftest/clang-13/lab-collabora/kselftest-lkdtm-asus-C523NA-A20057-coral.html#L4982
I'll be adding the package to the Debian rootfs image, it'll take a bit longer to get there but it's not a big issue.

gctucker:
I've re-run the job and added `apt install bc` to get over this issue and see if it would run, but now it's either skipping or failing all the test cases:
https://lava.collabora.co.uk/scheduler/job/4051364

@kees Can you please take a look and see if you know what's causing these failures?

I can still make a PR to add `bc` to the rootfs as that's likely to be needed for other tests too anyway.

kees:
If this contains the SHELL fix, I'm stumped. Given that I can see dmesg output before and after the other lkdtm tests, it looks very much like this line is failing:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/tools/testing/selftests/lkdtm/run.sh#n82

What's the best way for me to set up a way to duplicate the execution environment? Adding `set -x` to run.sh might tell us more, or dropping the `2>/dev/null`

gctucker:
@kees To answer your question, I guess you could download the rootfs image, the kselftest tarball and the kernel and run that locally with QEMU.

But in fact, I've found the root cause of the issue and replied to your email.  Essentially, we can and we should define `SHELL` to be `/bin/bash` in KernelCI kselftest jobs.

gctucker:
See #736 and #737 for the `/bin/bash` fix.

gctucker:
Tested OK on staging in combination with other pending patches:
https://staging.kernelci.org/test/plan/id/60d2dbdc3cfb88da0924bf41/ lkdtm
https://staging.kernelci.org/test/plan/id/60d2de49c460b226ec24bf31/ seccomp

PR #736, opened 2021-06-16
Extend kselftest coverage

gctucker:
Enable the newly added kselftest collections lkdtm and seccomp on x86 devices.

Depends on #735 and #737

PR #737, opened 2021-06-18
T28818 rootfs params

gctucker:
Add params to rootfs configs.


gctucker (commented):
Tested OK in combination with #736

PR #740, opened 2021-06-24
test-configs.yaml: add igt-gpu-i915 to grunt

alexandrasp:
Includes igt-gpu-i915 plan to the chromebook grunt.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -1828,6 +1828,7 @@ test_configs:
       - baseline
       - baseline-nfs
       - cros-ec
+      - igt-gpu-i915
       - kselftest-filesystems
       - kselftest-futex
       - kselftest-lib


alexandrasp:
It seems be ok, first kselftests are already enabled on chromebooks and just grunt was without igt-gpu-i915.

gctucker:
The `grunt` Chromebook has an AMD chip and therefore doesn't have an Intel i915 GPU.  So this won't produce any useful results.

However, we may consider running IGT on AMD for the `amdgpu` kernel driver but I believe this would require a bit more work.

gctucker:
@alexandrasp I've checked with some graphics experts, there's nothing needed other than the `amdgpu` kernel driver to run IGT on `grunt`.  However `CONFIG_DRM_AMDGPU` is not enabled in `x86_64_defconfig` so it would need to be added with a KernelCI config fragment.  I guess we would then have some new builds with a `x86_64_defconfig+x86-chromebook+amdgpu` config.  Once we have that, we can define the `igt-amdgpu` test plan with a list of test cases useful to run on this GPU driver.

gctucker:
And, the `stoney_*` binaries would need to be imported from https://git.kernel.org/pub/scm/linux/kernel/git/firmware/linux-firmware.git/tree/amdgpu (thanks @fooishbar for the tips).

gctucker:
> the `igt-amdgpu` test plan with a list of test cases useful to run on this GPU driver.

More tips, here's a list to get started:

amd_basic, amd_bypass, amd_color, amd_cs_nop, amd_info, and (if amdgpu is a module rather than builtin) amd_module_load

It would seem worth creating a roadmap item on the GitHub workboard for this.

fooishbar:
Thanks for the work on this! Super awesome to see more coverage :)

mupuf:
> I've checked with some graphics experts, there's nothing needed other than the amdgpu kernel driver to run IGT on grunt.

You'll need firmwares though. You may however do it like me: embed them in the kernel image using `CONFIG_EXTRA_FIRMWARE`.

How are you keeping track of failures here? Just as a diff on the status? If so, be aware that IGT has millions of failure modes and the most reliable way to do so is to check out stderr and dmesg :)

gctucker:
Thanks @mupuf for another round of expert tips :)

There's a Debian Buster rootfs image dedicated to IGT, so I think we can include the firmware binaries there.  But yes, built-in firmware is also something to keep in mind if it works better in practice in some cases.

About failures, yes that's how it's currently being done.  Each IGT test case reports a status (pass/fail/skip), and regressions are detected when one test case was passing on a previous revision of the same kernel branch and started failing.  The serial console output between each test case is also saved with the results, which helps with quick manual inspection.  So far this has worked pretty well but we're only running a small subset of IGT.  There's definitely scope for improvement there, especially when we start getting unstable results and we'll need to rely more on stats than binary pass/fail.

mupuf:
> Thanks @mupuf for another round of expert tips :)

If only... I'm still ascending the everest-like learning curve ;)

> 
> There's a Debian Buster rootfs image dedicated to IGT, so I think we can include the firmware binaries there. But yes, built-in firmware is also something to keep in mind if it works better in practice in some cases.

Yeah, in your case, keeping the firmwares in the rootfs might be the better choice, as it reduces the duplication of bytes in there, and should speed up deployment!

> 
> About failures, yes that's how it's currently being done. Each IGT test case reports a status (pass/fail/skip), and regressions are detected when one test case was passing on a previous revision of the same kernel branch and started failing. The serial console output between each test case is also saved with the results, which helps with quick manual inspection. So far this has worked pretty well but we're only running a small subset of IGT. There's definitely scope for improvement there, especially when we start getting unstable results and we'll need to rely more on stats than binary pass/fail.

I'm surprised you manage to fix regressions that fast, but I guess if you have only a couple of tests, it should be relatively easy indeed :) In any case, good luck to you all!

gctucker:
Replaced with #743.

PR #741, opened 2021-06-25
STAGING use buster-kselftest 20210618.0 with bc

gctucker:
To use the latest buster-kselftest with `bc` installed on staging until a rootfs image is in production.


gctucker:
The buster-kselftest rootfs images have now been enabled in production, so this can be dropped.

PR #745, opened 2021-06-29
Add cip config fragment

aliceinwire:

### DIFF
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -722,9 +722,12 @@ build_configs:
       gcc-8:
         build_environment: gcc-8
         architectures:
-          arm: *arm_arch
-          arm64: *arm64_arch
-          x86_64: *x86_64_arch
+          arm:
+            base_defconfig: 'cip://4.19.y-cip/arm/qemu_arm_defconfig'
+          arm64:
+            base_defconfig: 'cip://4.19.y-cip/arm64/qemu_arm64_defconfig'
+          x86_64:
+            base_defconfig: 'cip://4.19.y-cip/x86/cip_qemu_defconfig'
 
   cip_4.19-rt:
     tree: cip
@@ -733,9 +736,12 @@ build_configs:
       gcc-8:
         build_environment: gcc-8
         architectures:
-          arm: *arm_arch
-          arm64: *arm64_arch
-          x86_64: *x86_64_arch
+          arm:
+            base_defconfig: 'cip://4.19.y-cip-rt/arm/qemu_arm_defconfig'
+          arm64:
+            base_defconfig: 'cip://4.19.y-cip-rt/arm64/qemu_arm64_defconfig'
+          x86_64:
+            base_defconfig: 'cip://4.19.y-cip-rt/x86/cip_qemu_defconfig'
 
   clk:
     tree: clk
--- a/kernelci/build.py
+++ b/kernelci/build.py
@@ -1119,6 +1119,8 @@ scripts/kconfig/merge_config.sh -O {output} '{base}' '{frag}' {redir}
             env['name'],
         ])
 
+        publish_path = publish_path.replace(':', '-')
+
         bmeta['kernel'] = {
             'defconfig': target,
             'defconfig_full': defconfig,
--- a/kernelci/build.py
+++ b/kernelci/build.py
@@ -21,6 +21,7 @@ import itertools
 import json
 import os
 import platform
+import re
 import shutil
 import tarfile
 import time
@@ -34,6 +35,10 @@ from kernelci.storage import upload_files
 # This is used to get the mainline tags as a minimum for git describe
 TORVALDS_GIT_URL = \
     "git://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git"
+
+CIP_CONFIG_URL = \
+    "https://gitlab.com/cip-project/cip-kernel/cip-kernel-config/-\
+/raw/master/{branch}/{config}"
 
 # Hard-coded make targets for each CPU architecture
 MAKE_TARGETS = {
@@ -1065,6 +1070,13 @@ scripts/kconfig/merge_config.sh -O {output} '{base}' '{frag}' {redir}
             cmd = self._output_to_file(cmd, self._log_path, self._kdir)
         return shell_cmd(cmd, True)
 
+    def _create_cip_config(self, config):
+        [(branch, config)] = re.findall(r"cip://([\w\-.]+)/(.*)", config)
+        cip_config = os.path.join(self._output_path, ".config")
+        url = CIP_CONFIG_URL.format(branch=branch, config=config)
+        if not _download_file(url, cip_config):
+            raise FileNotFoundError("Error reading {}".format(url))
+
     def run(self, jopt=None, verbose=False, opts=None):
         """Make the kernel config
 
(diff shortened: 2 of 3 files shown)


config/core/build-configs.yaml
```diff
@@ -177,6 +177,35 @@ fragments:
       - 'CONFIG_VIDEO_VIVID=y'
       - 'CONFIG_VIDEO_VIVID_MAX_DEVS=64'
 
+  x86_4.19.y_cip:
```
gctucker, line 180: Would it be possible to have a config fragment that didn't depend on the kernel revision?  The `merge_config.sh` script will ignore config names that don't exist, so if you combine fragments for newer kernels with older ones it might work.

Alternatively, you could have one common fragment and then smaller ones specific to some particular kernel versions.

gctucker (commented):
This looks fine in principle, we should try to build it on staging to confirm.

Note: Once this is merged, we should have a follow-up PR to move all the CIP configuration into a separate YAML file I think.  But that would be part of an overall clean-up with a file for android, one for stable, one for CIP and maybe others too.

patersonc:
One thing to note is that CIP currently uses a lot of defconfigs, some different for different Kernel versions, a lot of different ones per arch:
https://gitlab.com/cip-project/cip-kernel/cip-kernel-config

I wonder if it would get a bit messy if we start trying to add yaml definitions for them all, especially if the current set up is done on a per-arch basis.

Perhaps we'd be better to add in support to pull the configs directly from gitlab/an external URL? This would make maintenance easier as well.

gctucker:
Yes, having fragments defined with a URL to a file rather than config options inline would be great.  Maybe we can start with this one inline, and consider generalising the approach initiated on the `chromeos` branch for other CIP fragments with git or HTTP protocols?

gctucker:
See this commit for the `cros://` type of fragments:
https://github.com/kernelci/kernelci-core/commit/744b87c058c3e4a7af9bd472610975940bf24444

gctucker:
@aliceinwire Hello, how would you like to resolve this?  How about pulling config fragments over HTTP and have them built eventually on the new cip.kernelci.org instance?  We could start with staging.kernelci.org to test the changes until the new instance is ready.

kernelci/build.py
```diff
@@ -1107,6 +1119,8 @@ def run(self, jopt=None, verbose=False, opts=None):
             env['name'],
         ])
 
+        publish_path = publish_path.replace(':', '-')
```
gctucker, line 1122: This is a bit hacky but let's see if it works and then we can refine the implementation.

kernelci/build.py
```diff
@@ -1115,6 +1129,10 @@ def run(self, jopt=None, verbose=False, opts=None):
             'publish_path': publish_path,
         }
 
+        if target.startswith("cip://"):
+            self._create_cip_config(target)
+            target = "olddefconfig"
+
         res = self._make(target, jopt, verbose, opts)
```
gctucker, line 1136: I think it's better to not change the value of `target` as it's coming from the input options.  Instead, we could do this:
```suggestion
        if target.startswith("cip://"):
            self._create_cip_config(target)
            res = self._make('olddefconfig', jopt, verbose, opts)
        else:
            res = self._make(target, jopt, verbose, opts)
```

gctucker (commented):
Looks fine to me, with maybe a few things to tweak.  Please check whether this is working as expected on staging.

kernelci/build.py
aliceinwire: ok, works for me

aliceinwire:
yes looks working with no errors https://staging.kernelci.org/build/cip/branch/linux-4.19.y-cip/kernel/v4.19.205-cip56/

kernelci/build.py
```diff
@@ -1115,7 +1129,12 @@ def run(self, jopt=None, verbose=False, opts=None):
             'publish_path': publish_path,
         }
 
-        res = self._make(target, jopt, verbose, opts)
+        if target.startswith("cip://"):
+            self._create_cip_config(target)
+            target = "olddefconfig"
```
gctucker, line 1134:
```suggestion
```

config/core/build-configs.yaml
```diff
@@ -722,9 +722,12 @@ build_configs:
       gcc-8:
         build_environment: gcc-8
         architectures:
-          arm: *arm_arch
-          arm64: *arm64_arch
-          x86_64: *x86_64_arch
+          arm:
+            base_defconfig: 'cip://4.19.y-cip/arm/qemu_arm_defconfig'
```
gctucker, line 726: Tested OK on staging: https://staging.kernelci.org/build/id/61371cfa24b1645bfe36a89f/

config/core/build-configs.yaml
```diff
@@ -722,9 +722,12 @@ build_configs:
...
+          arm64:
+            base_defconfig: 'cip://4.19.y-cip/arm64/qemu_arm64_defconfig'
```
gctucker, line 728: Tested OK on staging: https://staging.kernelci.org/build/id/6137240c428e1b1b5136a8f5/

config/core/build-configs.yaml
```diff
@@ -722,9 +722,12 @@ build_configs:
...
+          x86_64:
+            base_defconfig: 'cip://4.19.y-cip/x86/cip_qemu_defconfig'
```
gctucker, line 730: Tested OK on staging: https://staging.kernelci.org/build/id/61371c9444b490febd36a88b/

config/core/build-configs.yaml
```diff
@@ -733,9 +736,12 @@ build_configs:
...
         architectures:
-          arm: *arm_arch
-          arm64: *arm64_arch
-          x86_64: *x86_64_arch
+          arm:
+            base_defconfig: 'cip://4.19.y-cip-rt/arm/qemu_arm_defconfig'
+          arm64:
+            base_defconfig: 'cip://4.19.y-cip-rt/arm64/qemu_arm64_defconfig'
+          x86_64:
+            base_defconfig: 'cip://4.19.y-cip-rt/x86/cip_qemu_defconfig'
```
gctucker, line 744: These -rt configs don't seem to exist:
```
$ curl -sI https://gitlab.com/cip-project/cip-kernel/cip-kernel-config/-/raw/master/4.19.y-cip/x86/cip_qemu_defconfig | head -1
HTTP/2 200
$ curl -sI https://gitlab.com/cip-project/cip-kernel/cip-kernel-config/-/raw/master/4.19.y-cip-rt/x86/cip_qemu_defconfig | head -1
HTTP/2 404
```
There appears to be only some Siemens configs in the repository:
https://gitlab.com/cip-project/cip-kernel/cip-kernel-config/-/tree/master/4.19.y-cip-rt/x86

So maybe drop the -rt ones for now in this PR and make another one later when the config files are there?

gctucker (commented):
Found an issue with the -rt configs.  See also some suggested changes on https://github.com/aliceinwire/kernelci-core/pull/1

config/core/build-configs.yaml
gctucker: See this fixup PR: https://github.com/aliceinwire/kernelci-core/pull/2

Please feel free to merge it on top of this branch if you're OK with the changes.

gctucker:
OK merging this and then the fixup one https://github.com/aliceinwire/kernelci-core/pull/2 as discussed on Slack.

PR #747, opened 2021-07-02
config/rootfs/debos: add nosuspend overlay

gctucker:
Add "nosuspend" overlay for Debian Buster rootfs images to disable
suspend in systemd.  This is to work around spurious suspend requests
on Chromebook devices in particular.  Remove this file in the ramdisk
image as it's only an issue with nfsroot.

On a side note, the "sleep" and "usb" test plans use rtcwake which
bypasses systemd so this should not have any effect anyway.

### DIFF
--- a/config/core/rootfs-configs.yaml
+++ b/config/core/rootfs-configs.yaml
@@ -28,6 +28,7 @@ rootfs_configs:
     extra_files_remove: &extra_files_remove_buster
       - '*networkd*'
       - '*resolved*'
+      - nosuspend.conf
       - tar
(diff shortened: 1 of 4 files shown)


gctucker:
Tested OK on staging, now the `hana` tests with Debian don't suspend and all complete.  Here's some sample results:
https://staging.kernelci.org/test/plan/id/60e6514898f01bda2474029f/ baseline-nfs
https://staging.kernelci.org/test/plan/id/60e649c483f7259e0574027c/ kselftest-seccomp

PR #751, opened 2021-07-08
Add buildroot Dockerfile

mgalka:
Add Dockerfile that creates an image that contains all dependencies and
tools needed to build buildroot based rootfs images.

This docker image is to be used by Jenkins instead of a dedicated machine/node for buildroot builds.

### DIFF
--- a/config/docker/build-and-push.sh
+++ b/config/docker/build-and-push.sh
@@ -32,6 +32,7 @@
 set -e
 tag_px='kernelci/'
 
+
 options='npbdrikqQt:'
 while getopts $options option
 do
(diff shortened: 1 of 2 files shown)


mgalka:
Tested locally with `x86` and `arm64` images.

1. create docker image
```
docker image build --tag kernelci/buildroot:latest .
```

2. Build image
```
git clone https://github.com/kernelci/buildroot.git buildroot
cd buildroot

```
```

2. Run the container
```
git clone https://github.com/kernelci/buildroot.git buildroot
cd buildroot
docker container run -it -v $(pwd):/buildroot kernelci/buildroot /bin/bash
```

3. Inside the container
```
cd /buildroot
./configs/frags/build x86 baseline
# for arm64 put arm64 instead of x86

```

File: config/docker/buildroot/Dockerfile [hunk 1]
```diff
@@ -0,0 +1,31 @@
...
+    bzip2 \
+    cpio \
+    gcc \
+    g++ \
+    git \
+    make \
+    rsync \
+    patch \
+    python3 \
+    file \
```
gctucker, line 27: Alphabetical order tweak:
```suggestion
    cpio \
    file \
    gcc \
    g++ \
    git \
    make \
    rsync \
    patch \
    python3 \
```

gctucker (commented):
Tested OK locally.

@mgalka Please also add a patch to this PR to enable building this Docker image with the `build-and-push.sh` script.

config/docker/buildroot/Dockerfile
mgalka: fixed

mgalka: build-and-push.sh changes tested locally

config/docker/build-and-push.sh
```diff
@@ -100,6 +102,18 @@ then
   fi
 fi
 
+if [ "x${buildroot}" == "xtrue" ]
+then
+  tag=${tag_px}buildroot
+  echo_build $tag
+  docker build ${quiet} ${cache_args} debos -t $tag
```
gctucker, line 109: Erm, that's blatant copy-paste mistake.  Please try the script before adding a change to the PR.

gctucker (changes requested):

config/docker/build-and-push.sh
mgalka: You're right I made copy-pasted the part of the script for debos and changed the tag name, but haven't actually changed the Dockerfile to use.
I did have run the script before adding changes to the PR. I built the image and it was tagged properly, I haven't spotted though it's not the right image. My fault.

gctucker: I've added the `buildroot` images to be built automatically on staging, so if you update this PR it should get built later today and then you can pull the image `kernelci/staging-buildroot` and check it works.

mgalka: Refactored `build-and-push.sh` script. I got rid of the redundant code blocks and extracted them to functions. It should make it easier to add new images in future and improve readability of the script.

I tested it locally with the following calls, which are the same as in `kernelci-deploy/kernelci.org` script.

```
./build-and-push.sh -drn
./build-and-push.sh -bikQn
```
I couldn't spot any misbehaviour. 

**Pushing to docker hub has not been tested.**

build logs attached
[build-cmd_docker.log](https://github.com/kernelci/kernelci-core/files/6800620/build-cmd_docker.log)
[build-cmd_rootfs_docker.log](https://github.com/kernelci/kernelci-core/files/6800622/build-cmd_rootfs_docker.log)

gctucker:
Great, let's take a look later today when this has been pushed to https://hub.docker.com/repository/docker/kernelci/staging-buildroot

Note: The last images pushed there were actually `debos` due to the previous bug in the script.

mgalka:
> Great, let's take a look later today when this has been pushed to https://hub.docker.com/repository/docker/kernelci/staging-buildroot
> 
> Note: The last images pushed there were actually `debos` due to the previous bug in the script.

Yes, I noted that these images were `debos`. I got a bit mislead by the way script communicated `Building kernelci/buildroot`, but  actually the image to build was specified in another place. I hope that after refactoring it becomes more clear.

gctucker:
The image has now been built and pushed on staging: https://hub.docker.com/repository/docker/kernelci/staging-buildroot

Please check it looks correct and then this should be merged.

mgalka:
Staging docker images were build, tagged and pushed to docker hub with new `build-and-push.sh` script.

I was able to pull `kernelci/staging-buildroot` container from docker hub and build buildroot based rootfs image using steps from https://github.com/kernelci/kernelci-core/pull/751#issuecomment-876543753

gctucker (approved):

PR #758, opened 2021-07-14
test-configs.yaml: only run baseline on minnowboard-turbot

gctucker:
As this platform doesn't have enough capacity available in any lab,
drop baseline-nfs to reduce the coverage to the bare minimal with only
baseline.


gctucker:
Tested OK on staging: https://lava.collabora.co.uk/scheduler/queue?search=staging-next-20210719.0#table

PR #764, opened 2021-07-20
Extend x86_64_defconfig to support more drivers

10ne1:
This is a simplification of PR 728 [1] to remove values enabled
by default (via make savedefconfig) and adds a few more configs
to enable EC tests which are independent of ChromeOS (eg can work
on Debian).

Manually tested on Eve (Adrian) and Coral (Sonny).
Ran cros-ec-tests on KernelCI Octopus and Coral boards [2] [3]
where test passes went from 0 to 4-7 depending on the board.

[1] #728
[2] https://lava.collabora.co.uk/scheduler/job/4249724#results_158406973
[3] https://lava.collabora.co.uk/scheduler/job/4249260#results_158391223

Fixes: #728

### DIFF
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -251,6 +251,17 @@ fragments:
       - 'CONFIG_MMC_SDHCI=y'
       - 'CONFIG_MMC_SDHCI_PCI=y'
       - 'CONFIG_MMC_SDHCI_ACPI=y'
+      - 'CONFIG_CHROME_PLATFORMS=y'
+      - 'CONFIG_CHROMEOS_LAPTOP=y'
+      - 'CONFIG_CHROMEOS_TBMC=y'
+      - 'CONFIG_CROS_EC=y'
+      - 'CONFIG_CROS_EC_LPC=y'
+      - 'CONFIG_CROS_EC_I2C=y'
+      - 'CONFIG_IIO=m'
+      - 'CONFIG_IIO_CROS_EC_SENSORS_CORE=m'
+      - 'CONFIG_IIO_CROS_EC_SENSORS=m'
+      - 'CONFIG_MFD_CROS_EC_DEV=m'
+      - 'CONFIG_I2C_DESIGNWARE_PLATFORM=y'
 
   x86_kvm_guest:
     path: "kernel/configs/kvm_guest.config"


sonnydeez (approved):
LGTM on build-configs.yaml: add more minimal defconfigs to x86-chromebook

gctucker:
What's the difference between the "full" fragment and builds with actual Chrome OS configs?

For example, this one: https://chromeos.kernelci.org/build/id/60f71c035913db1b565a7072/

10ne1:
> What's the difference between the "full" fragment and builds with actual Chrome OS configs?
> 
> For example, this one: https://chromeos.kernelci.org/build/id/60f71c035913db1b565a7072/

The difference is that the `x86-chromebook-full` fragment contains all configs in one place to be applied on top of upstream x86_64_defconfig. It's main advantage is simplicity, but maybe this is only useful for development purposes, and not very useful for KernelCI?

From a developer perspective:

In a normal ChromeOS kernel build, config fragments are scattered all over the place and applied rather chaotically even in places outside the kernel tree, for example based on specific use-flags enabled in board overlays, [any of these fragments](https://chromium.googlesource.com/chromiumos/overlays/chromiumos-overlay/+/main/eclass/cros-kernel2.eclass#145) can be silently enabled, or defconfigs from the chromium overlay can silently "usurp / replace" x86_64_defconfig leading to surprises depending if specific build-time variables are set (eg `CHROMEOS_KERNEL_CONFIG`).

From a KernelCI perspective:

Maybe it would be more useful to not take the "maximal" approach to `x86-chromebook-full` and just add some very specific known configs which are required for tests to pass? Like enabling EC, TPM, PM and others which are not enabled in the minimalist `x86-chromebook`?

Or maybe renounce the idea of `x86-chromebook-full` entirely and just use the ChromeOS created configs which are downloaded as an archive from the internet? :smile:

eballetbo:
I think it would make sense have x86-chromebook-full but add these configs only when they are needed. So right now, we have a minimal fragment that makes a chromebook boot and have serial connection, but that is not enough to run some tests, so while we are enabling more and more tests EC, TPM and other stuff we might require to add config options to the fragment in order to run tast tests or other types of tests.

gctucker:
So, which tests do we want to run on the main linux.kernelci.org instance?

It seems to me that we just need to add a few config options to the `x86-chromebook` fragment used on the main instance to extend coverage there.  The advantage of having only one fragment is that we don't create more kernel builds just for Chromebooks.  The only issue is that every time we add something to the fragment, the build becomes less like a plain defconfig and results become harder to compare with other platforms.  But that's a trade-off to be made.

That fragment would also be available on the Chrome OS instance, where we would have extra kernel builds with the Chrome OS fragments on top of it to run even more tests with a full Chrome OS user-space.

10ne1:
> So, which tests do we want to run on the main linux.kernelci.org instance?

For now we want to enable the EC tests (cros-ec-tests) and also add some TPM tests because we spotted a regression upstream in the i2c cr50 tpm driver and it would be good to have some kernelci tests for at least something like does `/dev/tpm0` exist after booting, or see if communication with the TPM chip works.

I guess these two can be added directly to the main instance because they can be run on Debian as well.

In the future we'd need to decide on a case-by-case basis depending on the test to be added / enabled.

> It seems to me that we just need to add a few config options to the `x86-chromebook` fragment used on the main instance to extend coverage there. The advantage of having only one fragment is that we don't create more kernel builds just for Chromebooks. The only issue is that every time we add something to the fragment, the build becomes less like a plain defconfig and results become harder to compare with other platforms. But that's a trade-off to be made.

You are the maintainer so you get to decide on the trade-off. :smile: I agree with the above.

> That fragment would also be available on the Chrome OS instance, where we would have extra kernel builds with the Chrome OS fragments on top of it to run even more tests with a full Chrome OS user-space.

Sounds good. I'll update this PR to make it more minimal  (replace the massive x86-chromebook-full with just EC & TPM configs) and target the main branch. 

Thank you both @eballetbo and @gctucker for the feedback !

10ne1:
@gctucker @eballetbo  I've re-targeted the merge base to the main branch and updated the `x86-chmoebook-full` commit. Can you please review again? Thanks!

gctucker:
> The x86-chromebook fragment should be as minimal as possible and
remain as close to the mainline x86_64_defconfig as possible but
still allow booting a ChromeOS userspace image

Not quite, Chrome OS is only meant to be used on chromeos.kernelci.org at the moment.  So a fragment used in builds for linux.kernelci.org doesn't need configs that are only useful to Chrome OS.

config/core/build-configs.yaml
```diff
@@ -262,6 +262,21 @@ fragments:
...
       - 'CONFIG_BLK_DEV_NVME=y'
 
+  x86-chromebook-full:
+    path: "kernel/configs/x86-chromebook-full.config"
+    configs:
+      - 'CONFIG_CROS_EC=m'
+      - 'CONFIG_CROS_EC_LPC=m'
+      - 'CONFIG_IIO=m'
+      - 'CONFIG_IIO_CROS_EC_SENSORS_CORE=m'
+      - 'CONFIG_IIO_CROS_EC_SENSORS=m'
```
gctucker, line 272: Is there a good reason to build them as modules rather than having them built-in?

gctucker (commented):
I think there's a bit of confusion here.  This is what I believe would be optimal:

* A single `x86-chromebook` fragment built on linux.kernelci.org, to enable hardware features for tests run on linux.kernelci.org, as long as this remains small enough
* If a large fragment is needed to boot and run meaningful tests on linux.kernelci.org then we may create a separate one, but this doesn't seem to be the case with the config options listed in this PR
* Kernels built on chromeos.kernelci.org are meant to be run with Chrome OS user-space and the Tast framework, in fact there are already builds with Chrome OS kernel config fragment files from the chromiumos tree.  In principle, these fragments combined with the x86-chromebook one from the main branch should be enough to enable Chrome OS user-space to boot in LAVA.  If that's not the case, then extra configs may be added to a fragment built on chromeos.kernelci.org only.

config/core/build-configs.yaml
eballetbo: Usually having as module give you more chances to catch issues, I was solving in the past lots of issues and dependencies problems that were only noticed when the driver is built as module.

Said that, I'd follow the same rule that is followed in other defconfigs (like the arm64 defconfig), try to use always modules unless the option is really needed to boot (i.e some mmc controllers, etc.). In the case of cros-ec, some Chomebooks has problems booting unless is built-in, i.e when the cros-ec controls a regulator that is used for the emmc. I'd propose:

'CONFIG_CROS_EC=y'
'CONFIG_CROS_EC_LPC=y'
'CONFIG_IIO=m'
'CONFIG_IIO_CROS_EC_SENSORS_CORE=m'
'CONFIG_IIO_CROS_EC_SENSORS=m'

I'm missing, though, CONFIG_MFD_CROS_EC_DEV

Note that these options should be needed to increase coverage of the cros-ec-tests done (non chromeos userspace dependent). And that should be the argument to include these configs in this fragment.

@aratiu: Would be nice to see a couple of jobs running the cros-ec-tests with and without this fragment, to see  that there is a difference.

10ne1: I've pushed the new version and also added links to some runs on LAVA boards for cros-ec-tests.

config/core/build-configs.yaml
```diff
@@ -251,6 +251,29 @@ fragments:
       - 'CONFIG_MMC_SDHCI=y'
       - 'CONFIG_MMC_SDHCI_PCI=y'
       - 'CONFIG_MMC_SDHCI_ACPI=y'
+      - 'CONFIG_CHROME_PLATFORMS=y'
+      - 'CONFIG_CHROMEOS_LAPTOP=y'
+      - 'CONFIG_CHROMEOS_TBMC=y'
+      - 'CONFIG_DM_CRYPT=y'
+      - 'CONFIG_DM_THIN_PROVISIONING=y'
+      - 'CONFIG_DM_INIT=y'
+      - 'CONFIG_DM_VERITY=y'
```
eballetbo, line 260: Just to make it clear these DM options are only useful to boot chromeos userspace (so more suitable for the chromeos.kernelci.org) instance. Not sure if @gctucker wants these options here.

config/core/build-configs.yaml
```diff
@@ -251,6 +251,29 @@ fragments:
...
+      - 'CONFIG_CFS_BANDWIDTH=y'
+      - 'CONFIG_USER_NS=y'
+      - 'CONFIG_BLK_DEV_NVME=y'
```
eballetbo, line 263: These 3 option seems to not be needed also for the Chromebooks that currently are attached to the lab (grunt, coral, octopus)

config/core/build-configs.yaml
```diff
@@ -251,6 +251,29 @@ fragments:
...
+      - 'CONFIG_CROS_EC=y'
+      - 'CONFIG_CROS_EC_LPC=y'
+      - 'CONFIG_CROS_EC_I2C=y'
+      - 'CONFIG_IIO=m'
+      - 'CONFIG_IIO_CROS_EC_SENSORS_CORE=m'
+      - 'CONFIG_IIO_CROS_EC_SENSORS=m'
+      - 'CONFIG_MFD_CROS_EC_DEV=m'
```
eballetbo, line 270: These extend the cros-ec-tests coverage, from not testing the cros-ec to test basic functionalities of the cros-ec. I ran it on the three devices (grunt, coral, octopus)

Without this fragment:
* https://lava.collabora.co.uk/results/4247977
* https://lava.collabora.co.uk/results/4264092
* https://lava.collabora.co.uk/results/4267113

With the fragment, the results are:
* https://lava.collabora.co.uk/results/4276864
* https://lava.collabora.co.uk/results/4276865
* https://lava.collabora.co.uk/results/4276866

config/core/build-configs.yaml
```diff
@@ -251,6 +251,29 @@ fragments:
...
+      - 'CONFIG_TCG_TPM=y'
+      - 'CONFIG_TCG_TIS=y'
+      - 'CONFIG_TCG_TIS_SPI=y'
+      - 'CONFIG_TCG_TIS_SPI_CR50=y'
+      - 'CONFIG_TCG_TIS_I2C_CR50=y'
```
eballetbo, line 275: These might be required to enable some tpm tests, but right now, AFAIK we don't have a running test suite for this.

config/core/build-configs.yaml
```diff
@@ -251,6 +251,29 @@ fragments:
...
+      - 'CONFIG_I2C_DESIGNWARE_PLATFORM=y'
```
eballetbo, line 276: This config is probably required for both, the cros-ec-tests and tpm.

config/core/build-configs.yaml
```diff
@@ -251,6 +251,29 @@ fragments:
       - 'CONFIG_MMC_SDHCI=y'
       - 'CONFIG_MMC_SDHCI_PCI=y'
       - 'CONFIG_MMC_SDHCI_ACPI=y'
+      - 'CONFIG_CHROME_PLATFORMS=y'
+      - 'CONFIG_CHROMEOS_LAPTOP=y'
+      - 'CONFIG_CHROMEOS_TBMC=y'
```
eballetbo, line 256: These three are required for the cros-ec-tests

eballetbo (commented):
Up to @gtucker which fragments he really wants to pick, IMO at least the ones that extends the cros-ec-test coverage will be interesting to land. Then there are some that are only needed, for now, to boot the ChromeOS userspace. Still not sure how to deal with that, if needs to be a fragment that only applies to the chromeos.kernelci.org instance or should be applied always.

config/core/build-configs.yaml
10ne1: @eballetbo Do we need a separate test suite just for TPM? I'm adding some boot sanity checks to bootrr which I think should be enough for now to test these configs, eg check that dev node exists, check TPM chip communication, check TCG version etc.

gctucker:
@eballetbo Thanks for clarifying what each config was for.  I think we should start by adding the ones that are useful either to boot new Chromebook device types or enable cros-ec tests to the main `x86-chromebook` fragment as that will be used on linux.kernelci.org.

On top of that, we can enable a few more config options to boot Chrome OS on the `chromeos` branch, but I believe the Chrome OS fragments already have those configs turned on anyway.  These config options should be dropped from this PR anyway as booting Chrome OS is not useful in builds on linux.kernelci.org.

config/core/build-configs.yaml
gctucker: Great, so I think we should keep these in this PR then.

config/core/build-configs.yaml
gctucker: OK so let's drop them for now and consider adding them if/when we actually add a TPM test suite.

config/core/build-configs.yaml
gctucker: Sounds like we should keep it too if it's useful for cros-ec tests.

gctucker (changes requested):

config/core/build-configs.yaml
10ne1: Except for `CONFIG_BLK_DEV_NVME` they are needed to boot Eve (Google Pixelbook), but since we do not appear to have that in the pipeline I guess we can drop these and I'll build my test kernels with local changes for Eve.

I do not know which Chromebooks come with NVME drives and we don't appear to have any in the pipeline so we can drop  `CONFIG_BLK_DEV_NVME` as well.

@sonnydeez Can you tell us for which specific boards did you add this config initially?

config/core/build-configs.yaml
10ne1: Great, I'll drop these and add them via a separate review once we merge some TPM tests.

config/core/build-configs.yaml
10ne1: They are dropped now. marking as resolved.

gctucker:
Thanks, this fragment will be used in the next staging builds.  Let's check the results a bit later, the next one is scheduled to start in 5h.

gctucker (approved):

config/core/build-configs.yaml
10ne1: Since we do not have any boards with NVME I'm marking this as resolved. When we will add a device with NVME we can also add this config according to the principles discussed in this review. Thanks everyone! :)

gctucker:
Here's a build with `x86_64_defconfig+x86-chromebook`:
https://staging.kernelci.org/build/id/61034aee1957ef8dd3d74f62/

These are the configs that were added by KernelCI on top of the defconfig:
https://storage.staging.kernelci.org/kernelci/staging-mainline/staging-mainline-20210730.0/x86_64/x86_64_defconfig+x86-chromebook/gcc-8/config/kernelci.config

And the resulting kernel config:
https://storage.staging.kernelci.org/kernelci/staging-mainline/staging-mainline-20210730.0/x86_64/x86_64_defconfig+x86-chromebook/gcc-8/config/kernel.config

Test results are looking OK, but it's worth checking things in details.  In particular, check that the regressions are not new with the extra configs and check that cros-ec is running things as expected.

`coral` is failing to boot apparently now, not sure if it's related to this change:
https://staging.kernelci.org/test/plan/id/61034edf73bdfc5214d74f54/

There are still a lot of cros-ec skipped tests, is that expected?
https://staging.kernelci.org/test/plan/id/61034e7a3df854d466d74f6e/

10ne1:
@gctucker LGTM including the skipped tests. There might be some RESULT=unknown cros-ec-tests but I already sent a patch for them to the test suite maintainers.

Regarding the Coral failure, it appears to be unrelated, I used that config on top of the released Linux 5.14-rc3 commit ff1176468d36 and it booted fine:

https://lava.collabora.co.uk/scheduler/job/4285272

gctucker:
@10ne1 Well coral passes the baseline test plan with the same kernel build:
https://staging.kernelci.org/test/plan/id/61034b016165b3360ad74f5e/

And it was already failing to boot with the cros-ec rootfs before, here's some results from Sunday:
https://staging.kernelci.org/test/plan/id/60fd280b0771d3e9b1c28cb3/

So this PR is not introducing a regression on coral.  Still we need to investigate what's going on.

gctucker:
If you look at the log, it appears to be a bug in how the `login` test result is created:
https://storage.staging.kernelci.org/kernelci/staging-mainline/staging-mainline-20210725.0/x86_64/x86_64_defconfig+x86-chromebook/gcc-8/lab-collabora/cros-ec-asus-C523NA-A20057-coral.html

The coral did boot fine, then there was some collision on the serial console between kernel and user-space messages which caused a LAVA "signal" error.  Maybe this is what resulted in the `login` test case to be reported as failing, see the original LAVA job which has a status of "incomplete" although it kind of worked:
https://lava.collabora.co.uk/scheduler/job/4244850

I guess we should be using `/dev/kmsg` with cros-ec, as discussed earlier with @eballetbo.  That should hopefully clear this kind of errors.

gctucker:
Ah I see there's now #778 to use `/dev/kmsg` so let's just wait and see :)

gctucker:
Moving to the main project board as this is now out of scope for the hackfest.

10ne1:
All dependencies for this PR were merged (the /devkmesg and cros-ec-tests fixes).

Can we please test & merge it as well if you are ok with the current iteration?

10ne1:
@gctucker Any update on this? Thanks!

gctucker (approved):
Tested OK on staging, here's the KernelCI fragment showing all the config options:
https://storage.staging.kernelci.org/kernelci/staging-next/staging-next-20210901.2/x86_64/x86_64_defconfig+x86-chromebook/clang-13/config/kernelci.config
```
# fragment from: x86-chromebook
CONFIG_SERIAL_8250_DW=y
CONFIG_X86_AMD_PLATFORM_DEVICE=y
CONFIG_MFD_INTEL_LPSS_PCI=y
CONFIG_USB_GADGET=y
CONFIG_USB_ETH=y
CONFIG_USB_RTL8152=y
CONFIG_MMC=y
CONFIG_MMC_SDHCI=y
CONFIG_MMC_SDHCI_PCI=y
CONFIG_MMC_SDHCI_ACPI=y
CONFIG_CHROME_PLATFORMS=y
CONFIG_CHROMEOS_LAPTOP=y
CONFIG_CHROMEOS_TBMC=y
CONFIG_CROS_EC=y
CONFIG_CROS_EC_LPC=y
CONFIG_CROS_EC_I2C=y
CONFIG_IIO=m
CONFIG_IIO_CROS_EC_SENSORS_CORE=m
CONFIG_IIO_CROS_EC_SENSORS=m
CONFIG_MFD_CROS_EC_DEV=m
CONFIG_I2C_DESIGNWARE_PLATFORM=y
```
Full build details for x86_64:
https://staging.kernelci.org/build/id/612f45ea7b7aee76db36a933/

Including some cros-ec tests:
https://staging.kernelci.org/test/plan/id/612f4fc33ca83b644936a925/

gctucker:
This just missed today's production update, but it should be part of the next update on Monday.

PR #767, opened 2021-07-21
test-configs.yaml: adjust kselftest coverage to match documentation

gctucker:
Enable a few kselftest plans on arm platforms to match the
documentation on https://kernelci.org/docs/tests/.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -2394,6 +2394,8 @@ test_configs:
   - device_type: sun50i-h6-pine-h64
     test_plans:
       - baseline
+      - kselftest-filesystems
+      - kselftest-seccomp
       - ltp-crypto
       - ltp-ipc
       - ltp-mm


config/core/test-configs.yaml
```diff
@@ -2305,6 +2305,7 @@ test_configs:
     test_plans:
       - baseline
       - baseline-nfs
+      - kselftest-futex
```
gctucker, line 2308: This isn't working for some reason:
https://lava.collabora.co.uk/scheduler/job/4246916

Dropping this test plan from the PR for now.

config/core/test-configs.yaml
```diff
@@ -2401,6 +2402,8 @@ test_configs:
   - device_type: sun50i-h6-pine-h64
     test_plans:
       - baseline
+      - kselftest-filesystems
+      - kselftest-seccomp
```
gctucker, line 2406: Failing to DHCP in the kernel: https://lava.collabora.co.uk/scheduler/job/4246975

The network interface unique name needs to be defined in the device dictionary like for hana devices.

gctucker (commented):
Not merging yet as this isn't producing good results.

config/core/test-configs.yaml
gctucker: Well no, that didn't work either with specific eth name:
https://lava.collabora.co.uk/scheduler/job/4260493
So this PR can be closed, these devices will require more work until they can run kselftest.

PR #771, opened 2021-07-27
test-configs.yaml: Add xilinx-zcu102 device

aliceinwire:

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -1673,6 +1673,16 @@ device_types:
       - blocklist: *allmodconfig_filter
       - blocklist: {defconfig: ['multi_v7_defconfig+linaro-base+distribution']}
 
+  zynqmp-zcu102:
+    name: 'zynqmp-zcu102'
+    mach: zynqmp
+    arch: arm64
+    class: arm64-dtb
+    boot_method: uboot
+    dtb: 'xilinx/zynqmp-zcu102-rev1.0.dtb'
+    filters:
+      - passlist: {defconfig: ['defconfig']}
+
 
 test_configs:
 
@@ -2580,3 +2590,8 @@ test_configs:
   - device_type: zynq-zc702
     test_plans:
       - baseline
+
+  - device_type: zynqmp-zcu102
+    test_plans:
+      - baseline
+      - smc


config/core/test-configs.yaml
```diff
@@ -1640,6 +1640,14 @@ device_types:
             - 'tinyconfig'
             - 'kvm_guest.config'
 
+  xilinx-zcu102:
+    mach: zynq
+    class: arm64-dtb
+    boot_method: uboot
+    filters:
+      - blocklist: *allmodconfig_filter
+      - blocklist: {defconfig: ['multi_v7_defconfig+linaro-base+distribution']}
```
gctucker, line 1649: Where is `+linaro-base+distribution` coming from? As far as I know there aren't any fragment called like that in build-configs.yaml.

gctucker: Ah fair enough, I guess these fragments were added there as well by mistake.  Could you please drop it from `xilinx-zcu102` and maybe also add a patch to drop it from `zynq-zc702`?

config/core/test-configs.yaml
```diff
@@ -1640,6 +1640,11 @@ device_types:
             - 'tinyconfig'
             - 'kvm_guest.config'
 
+  xilinx-zcu102:
```
gctucker, line 1643: This device type doesn't appear to be in any lab.  However it is defined in LAVA upstream.

There is this instead in lab-cip:
https://lava.ciplatform.org/scheduler/device_type/zynqmp-zcu102

Is this the same device type but with a different name?

gctucker (commented):
There's no results on staging, we need to understand why.

config/core/test-configs.yaml
aliceinwire: yes is this one.

aliceinwire: fixed the name

gctucker:
This device type is already in upstream LAVA:
https://git.lavasoftware.org/lava/lava/-/blob/master/etc/dispatcher-config/device-types/xilinx-zcu102.jinja2

Waiting for staging results before merging.

gctucker:
@aliceinwire The reason why jobs aren't getting run on this device is because the device type name doesn't match any existing device tree name.  Here's the list of dtb files from an arm64 defconfig build:
```
                "xilinx/zynqmp-zcu102-rev1.0.dtb",
                "xilinx/zynqmp-zcu102-rev1.1.dtb",
                "xilinx/zynqmp-zcu102-revA.dtb",
                "xilinx/zynqmp-zcu102-revB.dtb",
```

Which revision is the device in the CIP lab?

Once we know this, the issue can be solved either by renaming the device type to match the device tree name (and that should be also fixed upstream in LAVA), or the `dtb` name could be explicitly defined in test-configs.yaml like in this example:

```yaml
  odroid-xu3:
    mach: exynos
    class: arm-dtb
    boot_method: uboot
    dtb: 'exynos5422-odroidxu3.dtb'
```

aliceinwire:
Tested and works https://staging.kernelci.org/soc/zynqmp/
we can close this issue

gctucker:
Yep, only the 16K pages flavour isn't booting:
https://staging.kernelci.org/soc/zynqmp/job/next/kernel/next-20210910/plan/baseline/

But that's not a blocker for this PR, other platforms are failing too.  It would be a bit neater to either add a filter to not run the 16K page build on this platform if it's known to not work, or report it so someone can fix it in the kernel.

If you don't think a filter is necessary then this can be merged now.

gctucker:
Also it may be worth adding something to the commit message, to explain what this device is and which tests are being enabled on it.

aliceinwire:
> Also it may be worth adding something to the commit message, to explain what this device is and which tests are being enabled on it.

done. this pull request canbe merged

gctucker (approved):

PR #773, opened 2021-07-27
test-configs.yaml: Enable kselftest and LTP on qemu platforms

broonie:
Enable kselftest and LTP on representitive qemu platforms for arm, arm64
and x86. These are low level testsuites with limited hardware interaction
so quite well matched to emulated platforms.

This is particularly useful for kselftest which is where tests for low
level architecture features tend to be added, these features are initially
only going to be available in emulation.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -2215,6 +2215,18 @@ test_configs:
     test_plans:
       - baseline_qemu
       - baseline-qemu-docker
+      - kselftest-filesystems
+      - kselftest-futex
+      - kselftest-lib
+      - kselftest-lkdtm
+      - kselftest-seccomp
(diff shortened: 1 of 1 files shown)


gctucker:
We need to do a bit of clean-up in the LAVA test templates before we can do this, and then have a template for running kselftest and LTP on QEMU as currently there's only U-Boot, GRUB and Depthcharge.

lubinszARM:
> gctucker

How is it going?

gctucker:
@broonie Support for running LTP on QEMU is now enabled, there's already `ltp_qemu`.  Feel free to make a new PR to enable the other ones.  Support for kselftest will require more work though.

Closing this one as it's now stale.

PR #776, opened 2021-07-28
build-configs.yaml: enable CONFIG_FB in virtualvideo fragment

gctucker:
As CONFIG_DRM_FBDEV_EMULATION now doesn't enable CONFIG_FD as per
f611b1e7624c ("drm: Avoid circular dependencies for CONFIG_FB"),
enable it in the virtualvideo fragment used by KernelCI.  This fixes
the issue where the vivid driver is not built and the
v4l2-compliance-vivid.device-presence test case fails.

### DIFF
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -229,6 +229,7 @@ fragments:
   virtualvideo:
     path: "kernel/configs/virtualvideo.config"
     configs:
+      - 'CONFIG_FB=y'
       - 'CONFIG_MEDIA_SUPPORT=y'
       - 'CONFIG_MEDIA_CAMERA_SUPPORT=y'
       - 'CONFIG_MEDIA_TEST_SUPPORT=y'


gctucker:
Fixes this regression: https://linux.kernelci.org/test/case/id/60ecaf5b508aeda41a1179ea/

gctucker:
Tested on staging with single job: https://bot.staging.kernelci.org/job/gtucker/job/single/155/console

gctucker:
Discussed on linux-media, this seems like the best approach:
https://lore.kernel.org/linux-media/02487fea-09ba-cda8-3418-d1db4529ee7c@xs4all.nl/T/#m4815bc7c4b54d38a3028d7ea6bf4e0dbe83db8b7

gctucker:
Waiting for this patch to be reviewed:
https://lore.kernel.org/linux-media/5a6270fd-57a5-e3f0-5485-1c514493d659@collabora.com/T/#m7015d37edbff2507c81e0e1a2a01da385432e34f

If it gets applied then the build configuration should be changed to leave the virtualvideo fragment as-is without `CONFIG_FB=y` since vivid could then be built without it, but we should instead add some `+virtualvideo+CONFIG_FB=y` extra builds (at least x86_64) for the media tree to cover both configurations.

gctucker (commented):
As the kernel patch may take a while to get applied, merge this for now.  We can update the build configs again if and when the vivid dependency on `CONFIG_FB` gets dropped.

PR #746, opened 2021-06-30
Add smc test

aliceinwire:
Signed-off-by: Alice Ferrazzi <alice.ferrazzi@miraclelinux.com>

PR #777, opened 2021-07-30
WIP use buildroot for smc test

gctucker:
Depends on #746


gctucker:
This didn't "fix" the issue, so closing now.  We can stick to Debian for the smc test plan, and the `hana` device type probably needs a firmware update.

PR #781, opened 2021-08-03
T29625 doc test configs

gctucker:
Improve the YAML config documentation section and add a page for test configurations.


gctucker:
Originally imported from https://github.com/kernelci/kernelci-doc/wiki/Test-configurations

PR #787, opened 2021-08-09
Update clang-13 repository

mgalka:
llvm-toolchain-buster leads now to clang-14 hence path to clang-13
packages needed an update.


mgalka:
Docker build tested locally

PR #789, opened 2021-08-13
build-configs: Enable builds of 16K pages on arm64

broonie:
Add coverage of 16K pages on arm64 in -next since while it's available on
a number of systems it is not widely covered by test systems so can turn
up issues from time to time.


gctucker:
Tested OK on staging: https://staging.kernelci.org/build/id/61289e156897ae872636a896/

gctucker (approved):

PR #783, opened 2021-08-04
lab-configs.yaml: Expand test coverage in lab-cip

patersonc:
Enable kselftest, ltp and preempt-rt tests.

At the moment kselftest and ltp only support limited platforms, so not many
(if any) tests will actually get run in lab-cip, but let's enable them
anyway so everything is ready to go when test support is expanded.

Signed-off-by: Chris Paterson <chris.paterson2@renesas.com>

PR #790, opened 2021-08-20
Add cip core

aliceinwire:


gctucker:
This is the kind of thing that would need to be done on a separate instance, as it's not using standard KernelCI rootfs images.

We may create buster-cip rootfs images as regular KernelCI images if they're basically a variant on Buster with just a few custom packages.  They would need to be built with kci_rootfs like the other images, and stored on storage.kernelci.org.  That would seem like a valid approach for the main linux.kernelci.org instance.

gctucker:
> This is the kind of thing that would need to be done on a separate instance, as it's not using standard KernelCI rootfs images.
> 
> We may create buster-cip rootfs images as regular KernelCI images if they're basically a variant on Buster with just a few custom packages. They would need to be built with kci_rootfs like the other images, and stored on storage.kernelci.org. That would seem like a valid approach for the main linux.kernelci.org instance.

After some more discussion and looking into the cip-core images, they are essentially Debian rootfs images using only packages found upstream and a few scripts and configuration files added on top of them.  So they are not for any particular product, they are similar to other rootfs variants created by KernelCI.  Also, reproducing these images isn't completely trivial outside of the cip-core build system.  So it makes more sense to have the images uploaded directly from the cip-core build system (Gitlab) to KernelCI storage.

The only real requirement is to have a unique URL for each of them (e.g. version or date) so the `test-configs.yaml` file can explicitly choose which ones to use and change atomically during production updates.

If for any reason the cip-core images become product-oriented, or if some changes are required for KernelCI but they can't be done in the main cip-core images, then they would need to be built using `kci_rootfs` like the other regular KernelCI rootfs images.  As long as we're not in any of these situations, we should be able to keep reusing the images directly from the cip-core build system.

config/core/test-configs.yaml
```diff
@@ -123,7 +143,7 @@ test_plans:
     pattern: 'baseline/generic-qemu-baseline-docker-template.jinja2'
     filters:
       - blocklist: *kselftest_defconfig_filter
-
+
```
gctucker, line 146: Please delete the whitespace.

gctucker (commented):
The `baseline-cip-nfs` test plan doesn't seem to be enabled on any devices in `lab-cip`, apart from beaglebone-black maybe.  Is that intentional?

Also, while I think this is useful to try on staging, if the `baseline-cip` tests don't add any value compared to the regular `baseline` test plan based on buildroot then I don't think we should be enabling it in production.  We should however try to enable CIP tests that need the cip-core rootfs images.

config/core/test-configs.yaml
```diff
@@ -14,6 +14,12 @@ file_system_types:
       x86:     [{arch: i386}, {arch: x86_64}]
       mipsel:  [{arch: mips}]
 
+  cip:
+    url: 'https://storage.staging.kernelci.org/images/rootfs/cip/20210910/'
```
gctucker, line 18: Note: We will need to upload the images on https://storage.kerneci.org before deploying this in production.

aliceinwire: yes, I need a production token for upload to production storage.

aliceinwire:
Created pull request for adding missing CIP devices to KernelCI https://github.com/kernelci/kernelci-core/pull/846

aliceinwire:
> Looks like it's not in the `cpio` archive format.

thanks for finding this

aliceinwire:
@gctucker This passed correctly https://lava.ciplatform.org/scheduler/job/507687

gctucker:
Ah I see we have a fresh set of results today, great:
https://staging.kernelci.org/test/job/cip/branch/linux-5.10.y-cip/kernel/v5.10.8-8301-g3d6168cb89f6/

config/core/test-configs.yaml
```diff
@@ -2186,6 +2204,7 @@ test_configs:
   - device_type: mt8173-elm-hana
     test_plans:
       - baseline
+      - baseline-cip-nfs
```
gctucker, line 2207: This one is failing but it's a known issue, the network is broken in mainline at the moment (well, missing something in defconfig):
https://staging.kernelci.org/test/plan/id/6185115274f9828eaaa6c483/

So it's not a problem for this PR.

config/core/test-configs.yaml
```diff
@@ -1843,6 +1858,7 @@ test_configs:
   - device_type: beaglebone-black
     test_plans:
       - baseline
+      - baseline-cip-nfs
```
gctucker, line 1861: OK: https://staging.kernelci.org/test/plan/id/618510a0210225a29aa6c454/

gctucker (commented):
I've found that this was conflicting with #783 and as a result it wasn't being deployed on staging.  Now we should start getting baseline-cip-nfs test results for other branches on staging too, which will help confirm it works with all the platforms enabled here.

In the meantime, I believe we should already be able to send the rootfs images to storage.kernelci.org and update the URL in this PR.

gctucker:
Latest results from staging only show results for the `hana` Chromebook:
https://staging.kernelci.org/test/job/kernelci/branch/staging-mainline/kernel/staging-mainline-20211108.0/plan/baseline-cip-nfs/

Investigating why the tests weren't run on the x86 Chromebooks.

gctucker:
@aliceinwire There's a problem with the x86 rootfs images apparently:
https://lava.collabora.co.uk/scheduler/job/4924712

```
$ curl -I https://storage.kernelci.org/images/rootfs/cip/20211105//qemu-x86_64/cip-core-image-kernelci-cip-core-buster-qemu-x86_64.tar.gz
HTTP/2 404 
```
One issue is that there are two slash `//` characters, but that's probably not what's causing the error.  The actual problem is that the architecture name is not x86_64 but amd64:
```
$ curl -I https://storage.kernelci.org/images/rootfs/cip/20211105/qemu-amd64/cip-core-image-kernelci-cip-core-buster-qemu-amd64.tar.gz
HTTP/2 200 
```

gctucker:
I've added a commit to this PR to fix the issue `test-configs.yaml: add missing arch map for cip images`.  Tested locally, now checking if this is working as expected on staging.

gctucker:
OK that's better: https://lava.collabora.co.uk/scheduler/job/4926571

config/core/test-configs.yaml
```diff
@@ -1919,6 +1937,7 @@ test_configs:
   - device_type: hp-11A-G6-EE-grunt
     test_plans:
       - baseline
+      - baseline-cip-nfs
```
gctucker, line 1940: OK: https://staging.kernelci.org/test/plan/id/618900b6bbb919bcbea6c431/

config/core/test-configs.yaml
gctucker: OK: https://staging.kernelci.org/test/plan/id/6188ddfdbc4b48580ea6c45f/

(staging-mainline has some extra patches to fix this platform, it won't work with current mainline)

config/core/test-configs.yaml
```diff
@@ -1938,6 +1957,7 @@ test_configs:
   - device_type: hp-x360-12b-n4000-octopus
     test_plans:
       - baseline
+      - baseline-cip-nfs
```
gctucker, line 1960: OK: https://staging.kernelci.org/test/plan/id/61890121f71c565aeea6c436/

config/core/test-configs.yaml
```diff
@@ -2335,14 +2356,17 @@ test_configs:
   - device_type: r8a774a1-hihope-rzg2m-ex
     test_plans:
       - baseline
+      - baseline-cip-nfs
```
gctucker, line 2359: There's a problem with this one: https://lava.ciplatform.org/scheduler/job/512994

config/core/test-configs.yaml
```diff
@@ -2335,14 +2356,17 @@ test_configs:
...
 
   - device_type: r8a774b1-hihope-rzg2n-ex
     test_plans:
       - baseline
+      - baseline-cip-nfs
 
   - device_type: r8a774c0-ek874
     test_plans:
       - baseline
+      - baseline-cip-nfs
```
gctucker, line 2369: OK: https://staging.kernelci.org/test/plan/id/6189050d44c46d52a1a6c425/

config/core/test-configs.yaml
```diff
@@ -2335,14 +2356,17 @@ test_configs:
   - device_type: r8a774a1-hihope-rzg2m-ex
     test_plans:
       - baseline
+      - baseline-cip-nfs
 
   - device_type: r8a774b1-hihope-rzg2n-ex
     test_plans:
       - baseline
+      - baseline-cip-nfs
```
gctucker, line 2364: Currently offline: https://lava.ciplatform.org/scheduler/device_type/r8a774b1-hihope-rzg2n-ex

config/core/test-configs.yaml
gctucker: OK with stable 5.4 as well: https://staging.kernelci.org/test/plan/id/61890e45786ed47d60a6c425/

config/core/test-configs.yaml
gctucker: Tested OK with stable 5.4: https://staging.kernelci.org/test/plan/id/61890f6e98d5806260a6c438/

So there might be an issue in linux-next with this platform.  It's not a problem for merging this PR though.

config/core/test-configs.yaml
gctucker: OK from a few days ago with CIP 5.10 kernel: https://staging.kernelci.org/test/plan/id/618514e7471f2400eda6c499/

gctucker: Tested OK again today with production rootfs image: https://lava.ciplatform.org/scheduler/job/513021

gctucker:
Also had to add a commit to enable `baseline-cip-nfs` in `lab-cip`.

gctucker (approved):
All good to merge now.  This should get rolled out in production today.

PR #792, opened 2021-08-31
test-configs.yaml: update Debian rootfs URLs

mgalka:
update Debian rootfs URLs


gctucker (approved):
Tested OK on staging
https://staging.kernelci.org/test/job/kernelci/branch/staging-next/kernel/staging-next-20210901.2/

PR #793, opened 2021-09-02
Detect better the endianess of a build

montjoie:
The current way to detect endianess is to check if a BIG ENDIAN defconfig frag was added.
But this is wrong for defconfig which is are nativly big endian like arm
ixp4xx_defconfig.

So the ultimate way to know the endianess is to download the kernel
config and grep it.

Signed-off-by: Corentin Labbe <clabbe@baylibre.com>

### DIFF
--- a/kernelci/build.py
+++ b/kernelci/build.py
@@ -1173,6 +1173,11 @@ class MakeKernel(Step):
         if target:
             kbmeta['image'] = target
 
+        if self._kernel_config_enabled('CPU_BIG_ENDIAN'):
+            kbmeta['endianness'] = 'big'
+        else:
+            kbmeta['endianness'] = 'little'
+
         res = self._make(target, jopt, verbose)
 
         if res:
--- a/kernelci/test.py
+++ b/kernelci/test.py
@@ -108,7 +108,6 @@ def get_params(meta, target, plan_config, storage):
     rootfs = plan_config.rootfs
     defconfig_full = kernel['defconfig_full']
     defconfig = ''.join(defconfig_full.split('+')[:1])
-
     endian = kernel["endianness"]
     describe = rev['describe']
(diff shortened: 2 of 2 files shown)


gctucker:
I think this should be done at build time, when the `.config` file has been generated.  Then it can be grepped, or we can use the existing method in the build steps to check if a config is enabled, and populate some meta-data flag with the result.  For example, a similar thing is done to see if modules or device tree is enabled.

kernelci/test.py
```diff
@@ -108,7 +108,12 @@ def get_params(meta, target, plan_config, storage):
     rootfs = plan_config.rootfs
     defconfig_full = kernel['defconfig_full']
     defconfig = ''.join(defconfig_full.split('+')[:1])
-    endian = 'big' if 'BIG_ENDIAN' in defconfig_full else 'little'
+
+    if "endian" in kernel:
+        endian = kernel["endian"]
+    else:
```
gctucker, line 114: I don't think we need to keep backward compatibility, the code is expecting the data to be valid so it should always have the `endian` attribute.

kernelci/build.py
```diff
@@ -1173,6 +1173,11 @@ def run(self, jopt=None, verbose=False, opts=None):
         if target:
             kbmeta['image'] = target
 
+        if self._kernel_config_enabled('CPU_BIG_ENDIAN'):
```
gctucker, line 1176: I think `endianness` is a bit more appropriate.

gctucker (commented):
LGTM, just a couple of minor things.

kernelci/test.py
montjoie: what happen if someone use kci_test against an old build ? I think we should keep compatibility for this case.

gctucker:
LGTM, let's check results on staging.

PR #795, opened 2021-09-04
Ixp4xx

montjoie:


gctucker:
I think this approach is going to lead to some problems because `armeb-xscale` is not a real achitecture name, and relying on the CPU name is very specific to this platform.

Instead, we can define `buildroot-xscale` as a rootfs type:
```yaml
rootfs_configs:
  buildroot-baseline: &baseline
    rootfs_type: buildroot
    arch_list:
      - arc
      - arm64
      - arm64be
      - armeb
      - armel
      - mipsel
      - riscv
      - x86
    frags:
      - baseline

  - buildroot-baseline-xscale:
    <<: *baseline
    arch_list:
      - armeb
    frags:
      - baseline
      - armeb-xscale
```

Then this `buildroot-baseline-xscale` rootfs type can be used in `test-configs.yaml` to create a `baseline_xscale` variant in the same way that there is a `baseline_qemu` one.  Platforms with the xscale architecture would then be running the `baseline_xscale` test plan instead of `baseline`.

aliceinwire:
@montjoie can you rebase this?

nuclearcat:
I guess if there is no plans anymore for this PR - maybe we can close it?

montjoie:
I still want to test ixp4xx, but the rootfs problem and the strangeness of this platform wont help.
I close this PR until I got the time to re-work on it.

PR #799, opened 2021-09-07
config: move stable build configs to build-configs-stable.yaml

gctucker:
Create a new YAML config file for the stable builds to make it easier
to maintain by reducing the main build-configs.yaml file and having
another one dedicated to stable.

Depends on #804


aliceinwire:
@gctucker Build configs broken order: 'khilman' before 'kernelci'

gctucker:
Yes, that's because the stable build configs are now added at the end of the ones from the main file.  The check should be done in each file rather than on the full list with data merged from several files.

mgalka (approved):
looks good to me

Issue #698, opened 2021-05-27
Include static list of kernel firmware files to add to Debian rootfs images


gctucker:
A first step to start using kernel firmware files is to define a static list with the ones to copy from the upstream repository into rootfs images, such as buster-v4l2.  This one is used as a ramdisk, so the size should be kept to only the firmware files that may be used on platforms currently running v4l2-compliance tests in KernelCI.

Typically, the [buster-v4l2.sh script](https://github.com/kernelci/kernelci-core/blob/main/config/rootfs/debos/scripts/buster-v4l2.sh) would add a step to check out the [linux-firmware](https://git.kernel.org/pub/scm/linux/kernel/git/firmware/linux-firmware.git) repository and copy files.

This could then be generalised with a separate script that can be used in any rootfs image, with a list of firmware files to copy defined in [rootfs-configs.yaml](https://github.com/kernelci/kernelci-core/blob/main/config/core/rootfs-configs.yaml) in the same way that Debian packages and files to remove are listed there.

eballetbo:
Some examples of missing firmware:

[baseline-mt8173-elm-hana-next-2021-05-27](https://storage.kernelci.org/next/master/next-20210527/arm64/defconfig/clang-12/lab-collabora/baseline-mt8173-elm-hana.html)
- [   14.823968] r8152 2-1.1:1.0: Direct firmware load for rtl_nic/rtl8153b-2.fw failed with error -2
- [   14.833068] r8152 2-1.1:1.0: Falling back to sysfs fallback for: rtl_nic/rtl8153b-2.fw

[asus-C433TA-AJ0005-rammus](https://lava.collabora.co.uk/scheduler/job/3483318)

- [    0.892283] i915 0000:00:02.0: Direct firmware load for i915/kbl_dmc_ver1_04.bin failed with error -2
- [    0.893101] i915 0000:00:02.0: [drm] Failed to load DMC firmware i915/kbl_dmc_ver1_04.bin. Disabling runtime power management.
- [    5.049971] r8152 2-3:1.0: Direct firmware load for rtl_nic/rtl8153b-2.fw failed with error -2
- [    5.059940] r8152 2-3:1.0: unable to load firmware patch rtl_nic/rtl8153b-2.fw (-2)
- [   10.906283] platform regulatory.0: Direct firmware load for regulatory.db failed with error -2

gctucker:
[...]

PR #801, opened 2021-09-08
kci_build: add make_firmware command

10ne1:
This adds a make_firmware command which in a similar manner
to make_modules creates some firmware tarballs based on the
configuration definitions in the new firmware-configs.yaml.

A new "push_firmware" command should be added in the future
so artifacts resulting from this command could be added to
storage, then logic for LAVA boards to download & use them.

Fixes: #698 

### DIFF
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -55,6 +55,9 @@ trees:
   evalenti:
     url: "https://git.kernel.org/pub/scm/linux/kernel/git/evalenti/linux-soc-thermal.git"
 
+  firmware:
+    url: "https://git.kernel.org/pub/scm/linux/kernel/git/firmware/linux-firmware.git"
+
   gtucker:
     url: 'https://gitlab.collabora.com/gtucker/linux.git'
 
@@ -854,6 +857,10 @@ build_configs:
   evalenti:
     tree: evalenti
     branch: 'for-kernelci'
+
+  firmware:
+    tree: firmware
+    branch: main
 
   gtucker_stable:
     tree: gtucker
--- a/config/core/firmware-configs.yaml
+++ b/config/core/firmware-configs.yaml
@@ -1,0 +1,17 @@
+firmware_archives:
+
+  # include all device firmware in a single archive
+  firmware-all.tar.xz:
+    - '*'
+
+  # include all i195 firmware in one archive
+  firmware-i915.tar.xz:
+    - i915/*
+
+  # only include octopus specific fw in this archive
+  firmware-hp-x360-12b-n4000-octopus.tar.xz:
+    - rtl_nic/rtl8153b-2.fw
+    - i915/kbl_dmc_ver1_04.bin
+    - i915/glk_dmc_ver1_04.bin
+
+  # add below more device-specific firmware archives as needed
--- a/kci_build
+++ b/kci_build
@@ -388,6 +388,15 @@ class cmd_make_modules(MakeCommand):
         return step.install(args.verbose, args.j)
 
 
+class cmd_make_firmware(MakeCommand):
+    help = "Package kernel firmware"
+    step_cls = kernelci.build.MakeFirmware
+
+    def __call__(self, configs, args):
+        self.step_cls.fw_config = configs['firmware_configs']
+        return super().__call__(configs, args)
+
+
 class cmd_make_dtbs(MakeCommand):
     help = "Build device trees"
     step_cls = kernelci.build.MakeDeviceTrees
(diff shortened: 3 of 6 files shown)


gctucker:
I thought the firmware files were meant to be part of the rootfs image builds, rather than the kernel builds as they don't change with each kernel revision.

@eballetbo What was the outcome of our previous discussions?  It would be good to confirm the plans here.

10ne1:
@gctucker That is a good point, that the firmware doesn't change with each kernel build,  so I can move the new command logic to the kci_rootfs builds. Thanks!

@eballetbo What do you think?

eballetbo:
> @gctucker That is a good point, that the firmware doesn't change with each kernel build, so I can move the new command logic to the kci_rootfs builds. Thanks!
> 
> @eballetbo What do you think?

We discussed a bit with Guillaume on Monday during the Open Hours and we were agree that would make more sense to have the firmware files part of the rootfs builds having and extra parameter in rootfs_config that can be added or not per rootfs definition. I think the idea was have something similar to i.e the extra_packages_remove but with installing the firmware.

gctucker:
@10ne1 So, I think you need to close this and start again :p

10ne1:
@eballetbo @gctucker 

Let's keep this discussion in written form so the design trade-offs decision is clearly documented.

I am not opposed to abandoning this route and just directly installing firmware during rootfs creation, but:

1. I think the separate firmware tarball + yaml config design proposed in this review is better than adding firmware directly in the rootfs (as I said we could move firmware tarball generation to kci_rootfs, or even create a separate kci_firmware). The implementation is a bit more complex, that is true, because we have a separate firmware build / push / pull step,  but is more efficient to minimize the root size and allows easily adding or changing the firmware during testing (just point to a new tarball).

2. My problem with the install in rootfs approach is how to decide which firmware to install for which board in which rootfs to avoid bloating the rootfs images unnecessarily?

3. Should we always install a subset of firmware we think is necessary in the rootfs for all boards? I assume creating board-specific rootfs'es is not an option.

Thanks again and sorry for being such a PITA with this.

10ne1:
Also @gctucker quoting you from issue #698 

> A longer-term solution would probably be to have the ability to provide a tarball with some firmware files determined just before scheduling each job, in the same way that we have a tarball for kernel modules. That way we could include files depending on the target device type, the kernel build, tests to run... and avoid making special rootfs images. It's a bit more complicated to implement though.

Shouldn't we implement the long-term solution directly? This is what I attempted here.

eballetbo:
> @eballetbo @gctucker
> 
> Let's keep this discussion in written form so the design trade-offs decision is clearly documented.
> 

Sounds good :-)

> I am not opposed to abandoning this route and just directly installing firmware during rootfs creation, but:
> 
> 1. I think the separate firmware tarball + yaml config design proposed in this review is better than adding firmware directly in the rootfs (as I said we could move firmware tarball generation to kci_rootfs, or even create a separate kci_firmware). The implementation is a bit more complex, that is true, because we have a separate firmware build / push / pull step,  but is more efficient to minimize the root size and allows easily adding or changing the firmware during testing (just point to a new tarball).

I think that (correct me @gtucker if I am wrong) this solution implies also some modification in the backend or lava in order to pick the firmware tarball like we do for kernel modules? So this solution is also a bit more complicated than just apply the above PR and ...

> 2. My problem with the install in rootfs approach is how to decide which firmware to install for which board in which rootfs to avoid bloating the rootfs images unnecessarily?

... I see the firmware files more tied to the rootfs than the kernel. Right now, we have different rootfs in function of the type of tests they run, i.e if the running rootfs is the igt one makes sense to only have gpu related firmware inside, if it's the v4l2 one makes sense to only install the media related firmware needed. And the same for baseline tests, only makes sense to install the firmware that makes some more tests to pass, the other is not really interesting/needed. What I had in my mind is a rootfs-configs.yaml with an extra parameter that says which firmware installs on every rootfs. I.e

buster-igt:
    ...
    add_firmware: 
         firmware-baseline-needed
         firmware-igt-related
   ...


> 3. Should we always install a subset of firmware we think is necessary in the rootfs for all boards? I assume creating board-specific rootfs'es is not an option.
> 

More than board-specific I'd say test specific.

> Edit:
> 
> 1. One downside of the separate firmware tarball + yaml approach is that firmware might be required for properly booting, eg to bring up the network hardware, but in that case we could just include it in the kernel config / image.
> 

Or just have a base firmware list that will be installed on every rootfs.

> Thanks again and sorry for being such a PITA with this.

Not a PITA, the discussions are for that. Nice you shared your thoughts :-)

gctucker:
Thanks for going through all the aspects of this.  I think we should put a summary on #698

PR #802, opened 2021-09-08
setup.py: install kernelci.data package

evdenis:
Otherwise, kci_data fails.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -29,6 +29,7 @@ setuptools.setup(
         "kernelci",
         "kernelci.config",
         "kernelci.lab",
+        "kernelci.data",
     ],
     package_data={
         '': ['../doc/*.md'],


setup.py
```diff
@@ -29,6 +29,7 @@
         "kernelci",
         "kernelci.config",
         "kernelci.lab",
+        "kernelci.data"
```
gctucker, line 32:
```suggestion
        "kernelci.data",
```
to avoid making changes to this line again when appending another line.

gctucker (commented):
A small tweak, and maybe also add this to the commit message:
```
Fixes: 220eb53ea127 ("setup.py: install kci_data script")
```

setup.py
evdenis: Fixed

gctucker (approved):
Thanks, tested OK manually after installing the package in `kernelci/build-base` Docker image and running kci_data list_configs from the installed package.

PR #804, opened 2021-09-08
Validate YAML files individually

gctucker:
Validate each YAML file individually as top-level entries can now be merged from various separate files.  When this happens, the entries are put together in the order they appear in separate files, so the alphabetical check is irrelevant.

### DIFF
--- a/kernelci/config/__init__.py
+++ b/kernelci/config/__init__.py
@@ -28,32 +28,30 @@ import kernelci.config.rootfs
 import kernelci.config.test
 
 
-def _find_yaml_files(config_path):
+def _iterate_yaml_files(config_path):
     if config_path.endswith('.yaml'):
         yaml_files = [config_path]
     else:
         yaml_files = glob.glob(os.path.join(config_path, "*.yaml"))
-    return yaml_files
-
-
-def validate_yaml(config_path, entries):
-    yaml_files = _find_yaml_files(config_path)
     for yaml_path in yaml_files:
         with open(yaml_path) as yaml_file:
             data = yaml.safe_load(yaml_file)
-            for name, value in data.items():
-                if name not in entries:
-                    continue
-                if hasattr(value, 'update'):
-                    keys = value.keys()
-                elif hasattr(value, 'extend'):
-                    keys = value
-                else:
-                    keys = []
-                err = kernelci.sort_check(keys)
-                if err:
-                    return "Broken order in {} {}: '{}' is before '{}'".format(
-                        yaml_path, name, err[0], err[1])
+            yield yaml_path, data
+
+
+def validate_yaml(config_path, entries):
+    for yaml_path, data in _iterate_yaml_files(config_path):
+        for name, value in ((k, v) for k, v in data.items() if k in entries):
+            if isinstance(value, dict):
+                keys = value.keys()
+            elif isinstance(value, list):
+                keys = value
+            else:
+                keys = []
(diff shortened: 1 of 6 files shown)


kernelci/config/__init__.py
```diff
@@ -28,32 +28,32 @@
...
-                    keys = value.keys()
-                elif hasattr(value, 'extend'):
-                    keys = value
-                else:
-                    keys = []
-                err = kernelci.sort_check(keys)
-                if err:
-                    return "Broken order in {} {}: '{}' is before '{}'".format(
-                        yaml_path, name, err[0], err[1])
+            yield(yaml_path, data)
```
mgalka, line 39: missing space?

kernelci/config/__init__.py
```diff
@@ -28,32 +28,32 @@
...
+
+
+def validate_yaml(config_path, entries):
+    for yaml_path, data in _iterate_yaml_files(config_path):
+        for name, value in data.items():
+            if name not in entries:
+                continue
+            if hasattr(value, 'update'):
```
mgalka, line 47: Wouldn't it be more clear to use `isinstance(value, dict)` instead?

kernelci/config/__init__.py
```diff
@@ -28,32 +28,32 @@
...
+                keys = value.keys()
+            elif hasattr(value, 'extend'):
```
mgalka, line 49: `isinstance(value, list)` wouldn't be better?

kernelci/config/__init__.py
```diff
@@ -27,7 +28,36 @@
...
+        yaml_files = glob.glob(os.path.join(config_path, "*.yaml"))
+    for yaml_path in yaml_files:
+        with open(yaml_path) as yaml_file:
+            data = yaml.safe_load(yaml_file)
+            yield(yaml_path, data)
+
+
+def validate_yaml(config_path, entries):
+    for yaml_path, data in _iterate_yaml_files(config_path):
+        for name, value in data.items():
```
mgalka, line 44: Maybe we could use a generator expression here instead of for loop with if..continue?
```python
for name, value in (n, v for n, v in data.items() if n in entries):
```

mgalka (changes requested):
Looks good, just a few small things worth considering.

kernelci/config/__init__.py
gctucker: Yes, that was copied from the code that loads the config where it makes more sense to check for the `update` method.  I'll change it here to check the class instead.

kernelci/config/__init__.py
gctucker: Ditto.

kernelci/config/__init__.py
gctucker: Fair enough, with a couple more brackets though: `(n, v) for n, v in data.items`

kernelci/config/__init__.py
gctucker: Fixed

kernelci/config/__init__.py
gctucker: Fixed

kernelci/config/__init__.py
gctucker: Fixed

kernelci/config/__init__.py
gctucker: Fixed

mgalka (approved):
LGTM

PR #809, opened 2021-09-09
Add gdb-minimal package to enable additional ltp-mm test case

pawiecz:
This PR enables following ltp-mm test case:
- vma05

Depends on #808


gctucker:
Pushed the changes directly to the `main` branch to resolve the conflicts.

PR #810, opened 2021-09-10
WIP: Update root file systems to Bullseye

alexandrasp:
Debian Bullseye was released as the new stable version. This commit updates rootfs images from Buster to Bullseye.


Lakshmipathi:
@alexandrasp I see few deleted `.sh` files. May be you need to push the renamed or new files?

Yesterday,  For `bullseye-git` rootfs found couple of issues with the `buster-igt.sh` script.  Following changes works for amd64: 
https://github.com/Lakshmipathi/kernelci-core/commit/75895e19dd702f178f9ec0d2881f822e384b941e

gctucker:
@Lakshmipathi is already working on this, so this PR is going to conflict with his work.  Adding the staging-skip label for now but I think it should be closed.

@alexandrasp There are still many things to fix for the Bullseye Docker images so I think you can focus on that for now.

PR #811, opened 2021-09-10
config/docker: respect prefix in Dockerfiles

evdenis:
It's required to automate containers creation for
$ ./build-and-push.sh -t kernelci/staging- ...

### DIFF
--- a/config/docker/build-and-push.sh
+++ b/config/docker/build-and-push.sh
@@ -75,7 +75,7 @@ docker_push() {
 docker_build_and_tag() {
   tag=${tag_px}$2
   echo_build $tag
-  docker build ${quiet} ${cache_args} $1 -t $tag
+  docker build --build-arg PREFIX=$tag_px ${quiet} ${cache_args} $1 -t $tag
   if [ "x${push}" == "xtrue" ]
   then
     docker_push $tag
--- a/config/docker/build-k8s/Dockerfile
+++ b/config/docker/build-k8s/Dockerfile
@@ -1,7 +1,8 @@
 #
 # Container to drive Kubernetes cluster from KernelCI
 #
-FROM kernelci/build-base
+ARG PREFIX=kernelci/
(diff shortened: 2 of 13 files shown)


config/docker/build-and-push.sh
```diff
@@ -75,7 +75,7 @@ docker_push() {
 docker_build_and_tag() {
   tag=${tag_px}$2
   echo_build $tag
-  docker build ${quiet} ${cache_args} $1 -t $tag
+  docker build --build-arg PREFIX=$tag ${quiet} ${cache_args} $1 -t $tag
```
gctucker, line 78:
```suggestion
  docker build --build-arg PREFIX=$tag_px ${quiet} ${cache_args} $1 -t $tag
```

gctucker (commented):
Just a small glitch, LGTM otherwise.  Tested OK locally with the script run on staging and the suggested fix applied.

config/docker/build-and-push.sh
evdenis: Sorry, I missed this.

evdenis:
fixed

gctucker:
There's a benign warning with some Docker images that aren't based on `build-base` such as the `debos` one:
```
[Warning] One or more build-args [PREFIX] were not consumed
```
But I think we can live with that.

evdenis:
Adding unused `ARG PREFIX` to build-base/Dockerfile suppresses this warning. But I don't think it's a good idea to add it.

gctucker:
> Adding unused `ARG PREFIX` to build-base/Dockerfile suppresses this warning. But I don't think it's a good idea to add it.

Exactly, so let's leave it as-is.

gctucker:
I'm doing a full rebuild of all the Docker images with cache disabled just to double check it's all working as expected, then this should be merged.

gctucker:
OK that just worked, thank you.

PR #813, opened 2021-09-13
test-configs.yaml: disable kselftest on mt8183-kukui-jacuzzi

gctucker:
Disable all kselftest on mt8183-kukui-jacuzzi-juniper-sku16 as the
hack to set a custom ip=::: kernel command line argument is breaking
other use-cases on this device type in Collabora's LAVA lab.  This is
a temporary measure until the issue is fixed properly in the LAVA
device type templates or elsewhere.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -2207,11 +2207,6 @@ test_configs:
     test_plans:
       - baseline
       - baseline-nfs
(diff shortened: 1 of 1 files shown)


gctucker:
Tested OK manually with `./kci_test list_jobs` using both a plain arm64 defconfig and one with kselftsest fragment.

PR #807, opened 2021-09-09
Move Docker images to Debian Bullseye

alexandrasp:
Debian Bullseye was released as the new stable version. Updates build-base, build-k8s, clang-11, clang-12, clang-13, gcc-9_arm, gcc-9_arm64 and gcc-9_x86 docker images using Buster to Bullseye.

- [x] build-base: built, pushed, tested with staging-next
- [x] build-k8s: built, pushed, tested with staging-next
- [x] clang-11: built, pushed, tested by hand (not used in build-configs.yaml so we might just drop the Docker image)
- [x] clang-12: built, pushed, tested with linux-next on staging
- [x] clang-13: built, pushed, tested with staging-next
- [x] gcc-9_arm: built, pushed, tested with staging-next
- [x] gcc-9_arm64: built, pushed, tested with staging-next
- [x] gcc-9_x86: built, pushed, tested with staging-next

PR #817, opened 2021-09-16
build-configs.yaml: upgrade from gcc-8 to gcc-9

gctucker:
Now that the Docker images are being upgraded from Debian Buster to
Bullseye, the standard GCC version needs to change from gcc-8 to
gcc-9.

Depends on #807


alexandrasp (approved):
LGTM

PR #818, opened 2021-09-17
Drop broken Bullseye Docker images for now

gctucker:
Temporarily drop the Docker images that are failing to build with Bullseye so we can keep building the other ones and test them on staging.

Depends on #807


gctucker:
This is now all being fixed as part of #824

PR #819, opened 2021-09-21
Add qemu-utils

mgalka:
Add qemu-utils package which is necessary to convert raw images to qcow2

### DIFF
--- a/config/docker/qemu/Dockerfile
+++ b/config/docker/qemu/Dockerfile
@@ -9,5 +9,6 @@ RUN apt-get update && \
 	qemu-system-ppc \
 	qemu-system-s390x \
 	qemu-system-sparc \
-	qemu-system-x86 && \
+	qemu-system-x86 \
+	qemu-utils && \
 	apt-get clean


config/docker/qemu/Dockerfile
```diff
@@ -10,4 +10,5 @@ RUN apt-get update && \
 	qemu-system-s390x \
 	qemu-system-sparc \
 	qemu-system-x86 && \
+	qemu-utils \
```
pawiecz, line 13:
```suggestion
	qemu-system-x86 \
	qemu-utils && \
```

mgalka: Didn't push all changes, thanks for spotting. Corrected.

gctucker:
> Add qemu-utils package which is necessary to convert raw images to qcow2

This Docker image was initially intended to be used while running QEMU jobs in LAVA.  Are you using it to do other things, such as converting images outside of LAVA?

mgalka:
> > Add qemu-utils package which is necessary to convert raw images to qcow2
> 
> This Docker image was initially intended to be used while running QEMU jobs in LAVA. Are you using it to do other things, such as converting images outside of LAVA?

Not exactly. As the images are usually quite large and downloading them during tests may be time consuming it may be beneficial to generate a qcow2 image backed up by a raw one, which is fast and doesn't take much space. In such scenario the latter image may be re-used by the tests because we can be sure it wasn't modified.

gctucker:
My understanding based on the [lab docs merge request](https://gitlab.collabora.com/lava/docs/-/merge_requests/28) is that `qemu-utils` is only useful for converting images to qcow2, which isn't something that LAVA jobs would typically do.  So it's a slight deviation from what this Docker image was originally intended for, which is to run QEMU in LAVA jobs.  I think it's OK since the `qemu-utils` package is rather small and shouldn't cause any undesired side-effects for the primary purpose of this Docker image, but we shouldn't keep adding more utilities to this image or it would become harder to maintain it for multiple use-cases.

PR #820, opened 2021-09-21
doc: kci_rootfs: point docker to non-staging images

10ne1:
Staging images are not as stable and can break more often
so it's best to point instructions to non-staging images.


gctucker (approved):
Deployed OK on staging: https://static.staging.kernelci.org/docs/core/kci_rootfs/

alexandrasp (approved):

PR #824, opened 2021-09-27
Bullseye Docker images

gctucker:
This is to merge the `bullseye-docker` topic branch once all the Docker images have been converted to Bullseye.

### DIFF
--- a/config/core/build-configs-stable.yaml
+++ b/config/core/build-configs-stable.yaml
@@ -1,7 +1,7 @@
 # Build fewer kernel configs with stable branches
 stable_variants: &stable_variants
-  gcc-8:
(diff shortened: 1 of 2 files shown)


gctucker:
I've disabled automatic rebuild of Docker images on staging to be sure they don't get replaced with Buster ones while the Bullseye ones are being built and pushed manually.

gctucker:
Staging jobs were broken due to some patch conflicts, it's running again now.  This should help confirm the build-base and all the toolchain images are working as expected.

gctucker:
I've added the `staging-skip` label, updating the straging branch now so we should be able to progress with #834

PR #825, opened 2021-09-29
config/lava: add earlycon to kernel command line

gctucker:
Add "earlycon" to the default kernel command line as it should just
work on platforms with stdout-path set in their device tree.

Suggested-by: Rob Herring <robh@kernel.org>


gctucker:
Tested OK on staging, verified that it has no apparent side effects:
https://lava.collabora.co.uk/scheduler/job/4603304#L759

PR #826, opened 2021-10-05
Draft: Add support for gz rootfs

aliceinwire:
draft for adding support to gz rootfs in KernelCI

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -43,6 +43,7 @@ file_systems:
     type: debian
     ramdisk: 'buster/20210913.0/{arch}/initrd.cpio.gz'
     nfs: 'buster/20210913.0/{arch}/full.rootfs.tar.xz'
+    compression: gz
     root_type: nfs
 
   debian_buster-cros-ec_ramdisk:
--- a/config/lava/base/kernel-ci-base-tftp-deploy.jinja2
+++ b/config/lava/base/kernel-ci-base-tftp-deploy.jinja2
@@ -19,7 +19,11 @@ actions:
 {%- if nfsrootfs_url %}
     nfsrootfs:
       url: {{ nfsrootfs_url }}
-      compression: xz
+      {% if compression %}
+        compression: {{ compression }}
+      {% else %}
+        compression: xz
+      {% endif  %}
 {%- endif %}
 {%- if initrd_url %}
     ramdisk:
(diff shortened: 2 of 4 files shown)


config/core/test-configs.yaml
```diff
@@ -43,6 +43,7 @@ file_systems:
     type: debian
     ramdisk: 'buster/20210913.0/{arch}/initrd.cpio.gz'
     nfs: 'buster/20210913.0/{arch}/full.rootfs.tar.xz'
+    compression: gz
```
gctucker, line 46: The problem with this is there is a different compression for each rootfs image (e.g. nfs and ramdisk).  So I think relying on the file name extension works better in practice.

To support a `compression` attribute, we would need to have a configuration object for each rootfs image.  At the moment we just have a string (URI).

aliceinwire: actually we don't if we set a default compression attribute.

gctucker: And we can't do it at the moment, unless we add a configuration object (e.g. YAML dictionary) for each rootfs image as they may all have different compression like in this particular example (ramdisk is gz, nfsroot is xz).

aliceinwire: My Idea was to have a default compression attribute xz if no attribute is specified. So that the compression attribute can be added only for compression usage different from xz and not changing the current test-configs.yaml template

aliceinwire:
> Adding `staging-skip` label as this is causing some Python exceptions:
> 
> ```
> Traceback (most recent call last):
>   File "/data/workspace/bot.staging.kernelci.org/test-runner/workspace/test-runner/kernelci-core/./kci_test", line 277, in <module>
>     status = opts.command(configs, opts)
>   File "/data/workspace/bot.staging.kernelci.org/test-runner/workspace/test-runner/kernelci-core/./kci_test", line 231, in __call__
>     job = api.generate(params, target, plan, callback_opts)
>   File "/data/workspace/bot.staging.kernelci.org/test-runner/workspace/test-runner/kernelci-core/kernelci/lab/lava.py", line 95, in generate
>     short_template_file = plan.get_template_path(target.boot_method)
>   File "/data/workspace/bot.staging.kernelci.org/test-runner/workspace/test-runner/kernelci-core/kernelci/config/test.py", line 386, in get_template_path
>     compression=self.rootfs.compression,
>   File "/data/workspace/bot.staging.kernelci.org/test-runner/workspace/test-runner/kernelci-core/kernelci/config/test.py", line 286, in compression
>     print(compression)
> NameError: name 'compression' is not defined
> ```
thanks for checking this

config/core/test-configs.yaml
gctucker: The compression attribute is defined in the template:
https://github.com/kernelci/kernelci-core/blob/main/config/lava/base/kernel-ci-base-tftp-deploy.jinja2#L24

aliceinwire: ok. so what I have to do ?

PR #836, opened 2021-10-12
kernelci.test: derive compression type from URLs

gctucker:
Rather than hard-coding the compression type in the LAVA base
template, derive it from the rootfs and modules URLs and pass it as
template arguments.  This allows more flexibility by relying on the
convention with usual file extensions i.e. .gz, .bz2 and .xz.  If the
end of the URL doesn't match any of these (say, .tar), the compression
attribute is left empty and not defined in the LAVA template.

### DIFF
--- a/kernelci/test.py
+++ b/kernelci/test.py
@@ -16,7 +16,10 @@
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
 
 import os
+import re
 import urllib.parse
+
+COMPRESSION_RE = re.compile(r'.*\.(gz|bz2|xz)$')
 
 
 def match_configs(configs, meta, lab):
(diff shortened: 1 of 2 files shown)


gctucker:
Alternative implementation for #836 which relies on file naming convention rather than YAML configuration.

kernelci/test.py
```diff
@@ -16,8 +16,11 @@
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
 
 import os
+import re
```
aliceinwire, line 19: What about using a more Pythonic style instead of regex, as we are already importing os ?
For reference, this is how I did in the previous pull request https://github.com/kernelci/kernelci-core/blob/a338de1c8f1fa0703bc56e6f78719af9e37fe3db/kernelci/config/test.py#L265

gctucker: Sure, but I think it would mean doing something like this:
```python
import urllib.parse
import os

COMPRESSION_FORMATS = ['gz', 'bz2', xz']

def get_compression(url):
    path = urllib.parse.spliturl(url).path
    ext = os.path.splitext(path)
    fmt = ext.rpartition('.')[2]
    return fmt if fmt in COMPRESSION_FORMATS else ''
```
So it's a bit more complicated than the regex implementation.

Which one do you prefer?  Or, do you have another approach in mind?

aliceinwire: ok works for me

aliceinwire (approved):
Looks good to me

Issue #845, opened 2021-10-19
Improve Thumb2 coverage for ARM build and boot tests


ardbiesheuvel:
None of the ARM configs select CONFIG_THUMB2_KERNEL (with the exception of milbeaut_m10v_defconfig, which can only boot on that specific platform), and so we have very little build coverage and no boot test coverage at all for ARM v7 kernels built in Thumb2 mode.

Given that Thumb2 kernels use completely different instruction sequences for early boot, resume from suspend, exception entry, etc., it seems appropriate to make sure that Thumb2 kernels are covered at least by the minimal, baseline boot tests. In general, any CPU that can run v7 ARM code should be able to run v7 Thumb2 code, although there may be bootloader/firmware limitations in individual cases causing some platforms to be unsuitable to take part in these tests.

PR #851, opened 2021-10-21
build-configs.yaml: drop the ardb_arm-kaslr-latest config

gctucker:
Drop the arm-kaslr-latest branch from ardb's tree as it is now stale
and not used any more.

Suggested-by: Ard Biesheuvel <ardb@kernel.org>


gctucker:
As discussed on #845

PR #861, opened 2021-11-05
config/k8s: Merge gen.py and wait.py into a single script to run obs

broonie:
Currently we have two scripts for driving Kubernetes, one to generate a
YAML file with a job description in it and another one to monitor the
status of such jobs after they have been submitted via kubectl. Let's cut
out the middleman and mush the two scripts together into one which directly
submits the generated job and then monitors the status. The structure is
taken from wait.py with generate.py moved to a function job_create() with
a small addition to actually submit the job to Kubernetes, returning the
resulting job name which is then used by the rest of the code instead of
having it provided on the command line.

An option is provided allowing the generated YAML to be written out to disk
for diagnostic purposes.

This is not a great way of doing things, it's an in place refactoring to
merge the two scripts with minimal changes prior to further work to take
more advantage of having things in a single script such as avoiding polling
by name or using notification facilities provided by the Kubernetes APIs.

### DIFF
--- a/config/k8s/run.py
+++ b/config/k8s/run.py
@@ -1,0 +1,252 @@
+#!/usr/bin/env python3
+
+import os
+import sys
+import time
+import argparse
+import yaml
+from kubernetes import client, config, utils
+from pprint import pprint
+from jinja2 import Environment, FileSystemLoader
+import re
+
+def env_override(value, key):
+    return os.getenv(key, value)
+
+def job_create(yaml_output, namespace):
+    d = os.getenv('DEFCONFIG').split('+')
+    defconfig = d[0]
+    frag = None
+    if len(d) > 1:
+        frag = d[1]
+
+    job_name = 'build-j{}-{}-{}-{}'.format(
+        os.getenv('BUILD_ID'),
+        os.getenv('ARCH'),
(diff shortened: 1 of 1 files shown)


broonie:
This isn't particularly useful in and of itself - I'm partly publishing it now to move to a different machine but figured some review was useful even at this stage. This won't do anything without corresponding updates in the Jenkins configuration.

@khilman

gctucker:
Adding the `staging-skip` label as I guess this will break the build jobs on staging without the aforementioned missing Jenkins changes.

broonie:
It won't break anything on staging, it's adding a completely new script so it'll literally do nothing without a Jenkins update.

gctucker:
@khilman @broonie Are you happy with this script as it currently stands?

Unless someone finds a problem with it, I think we should be able to merge it now.  Some follow-up work would be to turn it into a runtime implementation like LAVA and shell scripts to run arbitrary jobs in Kubernetes, in the coming months probably.

broonie:
I think it's fine in so far as it is - it'll get discarded once the proper rework is here, this is just moving us to an event driven rather than polling monitoring of existing jobs.

PR #868, opened 2021-11-10
README.md: Add KernelCI project logo

patersonc:
Add the KernelCI project logo to the top level readme.

This was suggested by LF legal to help with our trademark application.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,7 @@
+<img src="https://kernelci.org/image/kernelci-horizontal-color.png"
+     alt="KernelCI project logo"
+     width="40%" />
+
 # Welcome to KernelCI
 
 The KernelCI project is dedicated to testing the upstream Linux kernel.  Its


gctucker:
I wonder why it should be in this README and not in kernelci-api, kernelci-admin, kernelci-deploy, kcidb (not to mention the ones that are being deprecated e.g. kernelci-frontend, kernelci-backend, kernelci-jenkins...).  Is it not enough to have it on kernelci.org?  Maybe we could just remove the content of this README and have a link to the actual documentation: https://kernelci.org/docs/core/.

patersonc:
> I wonder why it should be in this README and not in kernelci-api, kernelci-admin, kernelci-deploy, kcidb (not to mention the ones that are being deprecated e.g. kernelci-frontend, kernelci-backend, kernelci-jenkins...). Is it not enough to have it on kernelci.org? Maybe we could just remove the content of this README and have a link to the actual documentation: https://kernelci.org/docs/core/.

Good question. I think it's mainly because a link to this repo was part of the trademark application, not the others.

We could add the logo to the other repos as well of course. Up to you/maintainers.

gctucker:
I see, it just seemed a bit arbitrary and superficial to have it here only.

It probably would make sense to add it to the kernelci-api, kernelci-project and kcidb repos, but maybe we should be using a URL from kernelci.org rather than copy the image file everywhere.  We could add the image file to kernelci-project where other images are already stored.  Then the URL would be https://kernelci.org/image/kernelci-horizontal-color.png (or .jpeg actually, to load quicker on low bandwidth).

patersonc:
> I see, it just seemed a bit arbitrary and superficial to have it here only.
> 
> It probably would make sense to add it to the kernelci-api, kernelci-project and kcidb repos, but maybe we should be using a URL from kernelci.org rather than copy the image file everywhere. We could add the image file to kernelci-project where other images are already stored. Then the URL would be https://kernelci.org/image/kernelci-horizontal-color.png (or .jpeg actually, to load quicker on low bandwidth).

Agreed, I'll try and get some MRs up later.

gctucker:
See https://github.com/kernelci/kernelci-project/pull/73

If that works for you we can merge and deploy it now so the definitive URL on kernelci.org can be used directly in README files.

gctucker:
@patersonc OK it's there https://kernelci.org/image/kernelci-horizontal-color.png

gctucker (approved):

PR #870, opened 2021-11-12
K8s build

broonie:


gctucker:
How about this, to fix pycodestyle and maybe simplify the syntax a bit?
```diff
diff --git a/kernelci/k8s.py b/kernelci/k8s.py
index 0b048c8c2d2e..637e0f323c97 100644
--- a/kernelci/k8s.py
+++ b/kernelci/k8s.py
@@ -159,14 +159,13 @@ class k8s:
         else:
             fetch_template = "build-fetch-tar.jinja2"
 
-        if not in job_dict['spec']['template']['spec'] 'initContainers':
-            job_dict['spec']['template']['spec']['initContainers'] = []
+        init_containers = job_dict['spec']['template']['spec'].setdefault(
+            'initContainers', []
+        )
         template = env.get_template(fetch_template)
         fetch_yaml = template.render(params)
         fetch_dict = yaml.safe_load(fetch_yaml)
-        job_dict['spec']['template']['spec']['initContainers'] =
-            + job_dict['spec']['template']['spec']['initContainers']
-            + fetch_dict
+        init_containers.extend(fetch_dict)
 
         # Everything ready, dump a copy if requested and submit
         if (outdir):
```

gctucker:
Actually, is `fetch_dict` really a dict and `init_containers` a list?  You can't add those two things together.  If it's a dict, then setdefault() can be passed an empty dict `{}` and then you can do `init_containers.update(fetch_dict)` to merge them together.

gctucker:
@broonie Please rebase now that #873 has been merged.

gctucker:
Adding `staging-skip` since this is causing some exceptions on staging:
```
+ ./kci_build check_new_commit --build-config=stable-5.14 --storage=http://storage.staging.kernelci.org
  File "/data/workspace/bot.staging.kernelci.org/build-trigger/workspace/kernel-build-trigger/kernelci-core/./kci_build", line 26, in <module>
    import kernelci.k8s
  File "/data/workspace/bot.staging.kernelci.org/build-trigger/workspace/kernel-build-trigger/kernelci-core/kernelci/k8s.py", line 18, in <module>
    import pygit2
ModuleNotFoundError: No module named 'pygit2'
```
It looks like the k8s Docker image needs to be rebuilt with the updated requirements.txt.

broonie:
initContainers is a list of dicts, we're trying to add a new list element which is a dict to the end of the ordered list of initContainers.

gctucker:
Right, so to add just one element you need to use `.append()`.

gctucker:
I've deleted the k8s Docker image on staging to ensure it gets rebuilt from scratch in the next job.  I guess the change in requirements.txt wasn't seen by Docker.  So we should be able to remove the skip label now.

gctucker:
There's still a problem with missing `pygit2`, I think it needs to be added to `setup.py`.  It's an issue that the dependencies are listed in two places, however I'm not sure what the best solution for that is.  Loading requirements.txt in setup.py isn't the intended way to use it, maybe we should just drop requirements.txt and rely solely on setup.py.  The current situation isn't ideal either as having to maintain 2 lists of dependencies is bound to cause issues again in the future.

See the related docs about that: https://packaging.python.org/discussions/install-requires-vs-requirements/#requirements-files

aliceinwire:
@broonie any update on this?

gctucker:
I believe this will be deprecated before being merged with the new API & pipeline, but maybe there are some ideas we could reuse.

mgalka:
New API is going to have this feature implemented.

PR #875, opened 2021-11-13
rootfs-configs.yaml: Add firmware rootfs

nfraprado:
Add a rootfs containing the firmware to boot devices and the overlay
script that will be used to check which firmwares are missing.

For now, only the firmwares used by rk3399-gru-kevin are present.

### DIFF
--- a/config/core/rootfs-configs.yaml
+++ b/config/core/rootfs-configs.yaml
@@ -51,6 +51,25 @@ rootfs_configs:
     script: "scripts/install-bootrr.sh"
     test_overlay: "overlays/baseline"
 
+  bullseye-firmware:
+    rootfs_type: debos
+    debian_release: bullseye
+    arch_list:
+      - amd64
+      - arm64
+      - armhf
+    extra_packages:
+      - wireless-regdb
+    extra_packages_remove:
+      - bash
+      - e2fslibs
+      - e2fsprogs
+      - libext2fs2
+    extra_firmware:
+      - mrvl/pcieusb8997_combo_v4.bin
+      - rockchip/dptx.bin
+    test_overlay: "overlays/firmware"
+
   bullseye-igt:
     rootfs_type: debos
     debian_release: bullseye
--- a/config/core/rootfs-configs.yaml
+++ b/config/core/rootfs-configs.yaml
@@ -66,8 +66,14 @@ rootfs_configs:
       - e2fsprogs
       - libext2fs2
     extra_firmware:
+      - i915/bxt_dmc_ver1_07.bin
+      - i915/glk_dmc_ver1_04.bin
+      - i915/kbl_dmc_ver1_04.bin
+      - imx/sdma/sdma-imx6q.bin
       - mrvl/pcieusb8997_combo_v4.bin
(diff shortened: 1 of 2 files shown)


nfraprado:
An example run of this can be seen at https://lava.collabora.co.uk/scheduler/job/4975847

nfraprado:
@mgalka Can this be run on staging now that it was rebased and the conflicts resolved?

gctucker:
How is this going to scale with all the devices being tested on linux.kernelci.org?  We could potentially include all the firmware files following this trend, and that would be too much to fit in a single rootfs image.

I don't think we can keep adding hard-coded lists of firmware files in a general purpose rootfs image.  This worked well as a starting point to get something in place for a specific purpose, for example to be able to run particular tests that require some drivers to be probed on some platforms (such as cros-ec and igt tests).

What difference does it actually make to have the firmware files listed here, other than avoid some warnings and a few drivers not completely probing?

nfraprado:
Agreed that this isn't very scalable, though I don't think there's a better alternative with the current mechanisms in place. The ideal approach would probably be to have a list of firmwares used on a per device basis, which would be applied independently of the rootfs and to all tests of that device.

But I think what needs to be discussed first is the purpose and usefulness of this test.

I've re-read the discussion on #801 to get some more background and thought a bit more, and I think the issue here is that while KernelCI tests focus on asserting that the kernel behaves correctly, this test doesn't. Instead, this test acts more as a "sanity check", it just checks whether we have all the firmware that the kernel is trying to load. Failing to do so could *potentially* impact the kernel behavior in meaningful ways, but we don't check for the behavior here.

So what's the usefulness of this test? In my opinion, it could make it easier to pinpoint the failure reason for other behavior tests. For example, if an igt test fails, and this firmware test also fails, it would then be more obvious that there is a firmware missing for the igt test, which could be causing the issue. In fact I think the presence of a test like this would make it part of the process of adding a new device to KernelCI to also add all the firmware required by it.

But then again, these scenarios assume that the same firmware would be installed for a device independent of the test, which is not the case currently. So perhaps the firmware installation mechanism in KernelCI would need to be improved before this test can really prove to be useful.

nfraprado:
@gctucker so, what are your thoughts on this?

Is this still something planned for the future? (quoted from #698)
> A longer-term solution would probably be to have the ability to provide a tarball with some firmware files determined just before scheduling each job, in the same way that we have a tarball for kernel modules. That way we could include files depending on the target device type, the kernel build, tests to run... and avoid making special rootfs images. It's a bit more complicated to implement though.

And if it is, should we hold this test for missing firmware until after that's in place, so we can really enjoy the benefits of this test without needing to bloat a single image with all the firmwares?

gctucker:
> So perhaps the firmware installation mechanism in KernelCI would need to be improved before this test can really prove to be useful.

I think that is true.  At the moment, we can only install firmware files for specific and limited use-cases.  Having the ability to load any firmware is beyond what we can currently cope with using hard-coded lists of files in fixed rootfs images.

nuclearcat:
As there is no update on PR, i will open issue and try to add missing firmware files if they are critical - in next firmware PR

PR #877, opened 2021-11-15
Enable kselftest on stable 5.14

gctucker:
Enable kselftest on stable kernels based on linux-5.14.y as it should be working well enough.  Future stable branches may be enabled if this appears to be producing valid results.

Also move the stable trees definitions to the stable builds config file to keep it more self-contained.


gctucker:
Tested OK on staging: https://staging.kernelci.org/build/id/61927dcb8b7c023850a6c43b/

gctucker:
x86 Chromebooks: https://staging.kernelci.org/build/id/619279002f1bf0d9aca6c42a/

PR #881, opened 2021-11-17
drop old lava_rest lab type

gctucker:
Drop the lava_rest lab type as it has now been replaced with
lava.lava_rest.


mgalka (approved):
lgtm

PR #884, opened 2021-11-18
rootfs-configs.yaml: bullseye-igt: Add firmware for amdgpu

nfraprado:
Add firmware used by the amdgpu driver in the bullseye-igt rootfs. This
will allow amdgpu igt tests to be run later.

Fixes #853. Since the `bullseye-igt` rootfs has been added in the meantime, I added the firmware to that rootfs instead of the `buster-igt` originally mentioned in the issue. 

Signed-off-by: Nícolas F. R. A. Prado <nfraprado@collabora.com>

PR #890, opened 2021-11-19
Add igt amdgpu test plan and enable it on grunt

### LINKED ISSUE #883: Add igt-gpu-amd test plan
Once the `amdgpu` kernel driver is being built with at least one config, the next step is to add a `igt-gpu-amd` test plan in `test-configs.yaml` with a list of relevant IGT test cases to run on a platform with an AMD GPU (probably grunt).

- [ ] List of test cases determined see this https://github.com/kernelci/kernelci-core/issues/743#issuecomment-870406385 as a starting point
- [ ] PR created to add the new test plan and enable it on `hp-11A-G6-EE-grunt` (or any other suitable platform)

nfraprado:
Add a test plan to run IGT tests, both Core and AMD specific, on the AMDGPU driver, and enable it to run on hp-11A-G6-EE-grunt (aka grunt). This relies on the igt rootfs containing the firmware required by AMDGPU, which will get added by #884.

Fixes #883.

At the moment some tests still fail:

```
amd_bypass.8bpc-bypass-mode
amd_color.crtc-lut-accuracy
amd_info.query-timestamp
amd_info.query-timestamp-while-idle
```

Some more in-depth debugging will be needed to try and find out why, since just enabling the other amdgpu config symbols didn't solve it. If it's not possible to get all tests working then we could disable them.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -148,6 +148,25 @@ test_plans:
     rootfs: debian_buster-igt_ramdisk
     pattern: 'igt/{category}-{method}-{protocol}-{rootfs}-igt.jinja2'
 
+  igt-gpu-amd:
+    <<: *igt
+    params:
+      drm_driver: "amdgpu"
+      tests: >
+        core_auth
+        core_getclient
+        core_getstats
+        core_getversion
+        core_setmaster_vs_auth
+        amd_basic
+        amd_bypass
+        amd_color
+        amd_cs_nop
+        amd_info
+        amd_link_settings
+    filters:
+      - passlist: {defconfig: ['amdgpu']}
+
   igt-gpu-i915:
     <<: *igt
     params:
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -1965,6 +1965,7 @@ test_configs:
       - baseline-cip-nfs
       - baseline-nfs
       - cros-ec
+      - igt-gpu-amd
       - kselftest-filesystems
       - kselftest-futex
       - kselftest-lib


nfraprado:
Tried the other missing amd igt tests.

`amd_assr`, `amd_hotplug`, `amd_mem_leak` and `amd_vrr_range` suspend the device as part of their tests, and since grunt doesn't have suspend/resume working it never comes back. So I'll not add these tests to the list.

`amd_abm`, `amd_max_bpc` and `amd_prime` just skip when running on grunt, so I'm also keeping these tests out of the list for now. (could be added later if other devices have these features enabled)

`amd_module_load` passes although the driver is compiled built-in in the amdgpu fragment, so there's no reason to run it for now.

Finally, `amd_link_settings` does run and pass on grunt, so I'm adding it to the test plan.

Additionally, reviewing the test results I noticed that `gem_create` was being skipped on grunt. On closer inspection it looks like this test is specific to i915 (intel), so I'm removing it from the test plan.

gctucker:
Here's some test results from staging:
https://staging.kernelci.org/test/plan/id/619d094491436532b81760b7/
with the matching LAVA job:
https://lava.collabora.co.uk/scheduler/job/5056131

All the test cases are being skipped, it looks like a firmware file is missing:
```
<4>[    5.341333] amdgpu 0000:00:01.0: Direct firmware load for amdgpu/stoney_pfp.bin failed with error -2
<3>[    5.350736] amdgpu 0000:00:01.0: amdgpu: gfx8: Failed to load firmware \"amdgpu/stoney_pfp.bin\"
<3>[    5.359598] [drm:gfx_v8_0_sw_init.cold] *ERROR* Failed to load gfx firmware!
<3>[    5.366903] [drm:amdgpu_device_init.cold] *ERROR* sw_init of IP block <gfx_v8_0> failed -2
```

The rootfs image is this one:
http://storage.kernelci.org/images/rootfs/debian/buster-igt/20211118.0/amd64/rootfs.cpio.gz

Are the firmware files included in the ramdisk?  Are they not only in the NFS tarball?  Maybe this test plan should be switched to NFS root if that's the case.

nfraprado:
Oh, right. That's because IGT is still using the buster rootfs:
```
  igt: &igt
    rootfs: debian_buster-igt_ramdisk
```

You mentioned in #884 that @Lakshmipathi was still wiring up `bullseye-igt`. So, what is the plan with `bullseye-igt`, will it replace `buster-igt`, in which case we should just wait for that to happen and this test will work, or will it run alongside buster, in which case I can add the firmware files for amdgpu to `buster-igt` as well so we can get this test going right now?

gctucker:
The plan is to migrate all the Buster rootfs images to Bullseye, so trying to get amdgpu working on Buster doesn't seem like a good use of anyone's time.

We should instead get Bullseye deployed, and as far as I can tell there's nothing blocking progress on that front.

gctucker (approved):
Tested OK on staging: https://staging.kernelci.org/test/plan/id/619e0af6055aa919bf1760b5/

gctucker:
Tested OK on staging again:
https://staging.kernelci.org/test/plan/id/619e1388b1c39d282d1760cb/ (amdgpu)
https://staging.kernelci.org/test/plan/id/619e1c4e5ad04340d51760b5/ (i915)

PR #891, opened 2021-11-20
kernelci.build: Make each build step overwrite its meta-data when re-run

nfraprado:
Currently, each build step always appends its meta-data to steps.json.
This works well for automated building, where all steps are always run
in the same sequence. However when kci_build is used manually to build
an image, multiple runs of the same step may happen as part of the
iterative development process. Since the overall build status is the
aggregation of the individual steps' status, a previous failed build can
keep the overall status as FAIL in bmeta.json, preventing the usage of
kci_test.

This could be worked around by the user always running init_bmeta before
starting each build, but a better alternative is to make each step
overwrite the meta-data from a previous run of the same step. This
improves the usability from manual usage while not affecting the
automated usecase.

### DIFF
--- a/kernelci/build.py
+++ b/kernelci/build.py
@@ -538,12 +538,21 @@ class Metadata:
                 break
         return value
 
-    def add_step(self, data):
-        """Add meta-data for a build step
+    def update_step(self, data):
+        """Update meta-data for a build step
+
+        Update meta-data for a build step so that if there is already a step
+        with the same name, the new data overwrites the old one, otherwise the
+        data is appended at the end.
 
         *data* is the data for the step, following the schema
         """
-        self._steps.append(data)
+        for i, item in enumerate(self._steps):
+            if item['name'] == data['name']:
+                self._steps[i] = data
+                break
+        else:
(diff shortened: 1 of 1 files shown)


kernelci/build.py
```diff
@@ -538,12 +538,21 @@ def get(self, *keys):
...
+        with the same name, the new data overwrites the old one, otherwise the
+        data is appended at the end.
 
         *data* is the data for the step, following the schema
         """
-        self._steps.append(data)
+        for i, item in enumerate(self._steps):
+            if item['name'] == data['name']:
+                self._steps[i] = data
+                break
```
gctucker, line 553: The for-loop provides references to the dictionaries in the list, so it's also possible to replace their content directly rather than using indexes:
```suggestion
        for step in self._steps:
            if step['name'] == data['name']:
                step.clear()
                step.update(data)
                break
```
I find it a bit cleaner to solely rely on the iterator, but I guess using the indexes also works.

gctucker (commented):
A small comment.

Tested OK on staging, there are no visible side-effects.

PR #892, opened 2021-11-22
kernelci.lab.LabAPI: pass job_path to .submit()

gctucker:
Rather than passing a job definition object to LabAPI.submit(), pass
the path to a job definition file.  Then it's up to the implementation
code to open it and submit it.  All the current use-cases are based on
actual definition files; if being able to pass an object instead
becomes required at some point then the class can be adjusted
accoringly then.

Update kernelci.lab.lava and kci_test accordingly.

### DIFF
--- a/kci_test
+++ b/kci_test
@@ -239,14 +239,12 @@ class cmd_submit(Command):
         for path in job_paths:
             if not os.path.isfile(path):
                 continue
-            with open(path, 'r') as job_file:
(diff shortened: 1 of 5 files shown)


gctucker (commented):
Tested OK on staging:
https://staging.kernelci.org/test/job/kernelci/branch/staging-next/kernel/staging-next-20211125.0/

PR #894, opened 2021-11-22
Drop: ar9331-dpt-module device

Lakshmipathi:
mips big-endian architecture is not supported in bullseye, so drop
ar9331-dpt-module device.

Below debian mailing list announcement:
https://lists.debian.org/debian-devel-announce/2021/01/msg00002.html
said "it is probably time to drop this architecture from bullseye and sid.
(i.e. the same we had for buster minus mips)"

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -390,13 +390,6 @@ device_types:
             - 'allnoconfig'
             - 'defconfig+CONFIG_KASAN=y'
 
-  ar9331-dpt-module:
-    mach: ar9331
-    class: mips-dtb
-    boot_method: barebox
-    filters:
-      - blocklist: *allmodconfig_filter
-
   armada-370-db:
     mach: mvebu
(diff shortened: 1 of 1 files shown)


gctucker:
Please fix the commit subject, this is a change in test-configs.yaml

gctucker:
Confirmed with @montjoie and @khilman that there aren't any MIPS devices left in lab-baylibre.  It's not clear where the device in this PR came from but it's not available anywhere either so it looks like we can drop it.

gctucker:
@mgrzeschik Hello, any thoughts about this device?  Is it still online in the Pengutronix lab?

gctucker:
Well I see it's offline according to the LAVA web UI, but maybe you guys have it hooked up in LabGrid or something?
https://hekla.openlab.pengutronix.de/scheduler/device_type/ar9331-dpt-module

gctucker (approved):

PR #898, opened 2021-11-23
Enable kselftest on CIP kernels

aliceinwire:
Enable kselftest on CIP kernels that support kselftest and are producing valid results. 
Future CIP kernel may be enabled if they appear to be producing valid results.

Also move the CIP trees definitions to the CIP builds config file to keep it more self-contained.

### DIFF
--- a/config/core/build-configs-cip.yaml
+++ b/config/core/build-configs-cip.yaml
@@ -68,7 +68,7 @@ build_configs:
   cip_4.4:
     tree: cip
     branch: 'linux-4.4.y-cip'
-    variants: *cip_variants
+    variants: *cip_variants_kselftest
 
   cip_4.4-rt:
     tree: cip
@@ -83,15 +83,15 @@ build_configs:
         build_environment: gcc-10
         architectures:
           arm:
-            <<: *cip_variants
+            <<: *cip_variants_kselftest
             extra_configs:
               - 'cip://4.19.y-cip/arm/qemu_arm_defconfig'
           arm64:
-            <<: *cip_variants
+            <<: *cip_variants_kselftest
             extra_configs:
               - 'cip://4.19.y-cip/arm64/qemu_arm64_defconfig'
           x86_64:
-            <<: *cip_variants
+            <<: *cip_variants_kselftest
             extra_configs:
               - 'cip://4.19.y-cip/x86/cip_qemu_defconfig'
 
--- a/config/core/build-configs-cip.yaml
+++ b/config/core/build-configs-cip.yaml
@@ -45,6 +45,25 @@ cip_variants: &cip_variants
         fragments: [x86-chromebook]
 
 
+cip_variants_kselftest: &cip_variants_kselftest
+  gcc-10:
+    <<: *cip_gcc_10
+    architectures:
+      <<: *cip_architectures
+      arm:
+        base_defconfig: 'multi_v7_defconfig'
+        extra_configs: ['allnoconfig']
+        fragments: [kselftest]
+      arm64:
+        extra_configs: ['allnoconfig']
+        fragments: [kselftest]
+      x86_64:
+        base_defconfig: 'x86_64_defconfig'
+        extra_configs:
+          - 'allnoconfig'
+          - 'x86_64_defconfig+x86-chromebook+kselftest'
+        fragments: [x86-chromebook, kselftest]
+
 build_configs:
   cip_4.4:
     tree: cip
@@ -84,4 +103,4 @@ build_configs:
   cip_5.10:
     tree: cip
     branch: 'linux-5.10.y-cip'
-    variants: *cip_variants
+    variants: *cip_variants_kselftest
--- a/config/core/build-configs-cip.yaml
+++ b/config/core/build-configs-cip.yaml
@@ -1,0 +1,87 @@
+trees:
+  cip:
+    url: "https://git.kernel.org/pub/scm/linux/kernel/git/cip/linux-cip.git"
+
+
+# Build fewer kernel configs with cip branches
+cip_variants: &cip_variants
+  gcc-10: &cip_gcc_10
+    build_environment: gcc-10
+    fragments: [tinyconfig]
+    architectures: &cip_architectures
+      arc:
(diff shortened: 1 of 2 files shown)


gctucker:
The commit only seems to be moving the configs, not enabling kselftest.  Is this PR missing a 2nd commit for kselftest maybe?

gctucker:
Adding `staging-skip` label until the YAML configuration is fixed (aliases don't work across files so the missing bits need to be copied to the new file).

aliceinwire:
> Adding `staging-skip` label until the YAML configuration is fixed (aliases don't work across files so the missing bits need to be copied to the new file).

yaml configuration test passed correctly. so removed the skip label.

>The commit only seems to be moving the configs, not enabling kselftest. Is this PR missing a 2nd commit for kselftest maybe?

yes, I was only checking the divided configuration before sending the kselftest commits.

config/core/build-configs-cip.yaml
```diff
@@ -0,0 +1,125 @@
+trees:
+  cip:
+    url: "https://git.kernel.org/pub/scm/linux/kernel/git/cip/linux-cip.git"
+
+# Default config with full build coverage
+build_configs_defaults:
```
gctucker, line 6: Please don't redefine build_configs_defaults, this will override the one already defined in `build-configs.yaml`.  You can call it anything else instead, such as `cip_build_config_defaults`.

Note: Ideally `build_configs_defaults` should be dropped from the code as it is a special case, but that's another problem.

config/core/build-configs-cip.yaml
```diff
@@ -0,0 +1,125 @@
...
+          arm: *arm_arch
+          arm64: *arm64_arch
+          x86_64: *x86_64_arch
+
+  cip_5.10:
+    tree: cip
+    branch: 'linux-5.10.y-cip'
+    variants:
+      gcc-10:
+        build_environment: gcc-10
```
gctucker, line 112: The `kselftest` fragment can also be added at the top level, so it will apply to all architectures.

config/core/build-configs-cip.yaml
gctucker: Also, maybe the configs used for stable would be more appropriate for CIP?
https://github.com/kernelci/kernelci-core/blob/main/config/core/build-configs-stable.yaml#L16

config/core/build-configs-cip.yaml
aliceinwire: followed stable

config/core/build-configs-cip.yaml
aliceinwire: Followed stable

aliceinwire:
Found this on the logs:

```
+ ./kci_build init_bmeta --build-config=cip_5.10 --commit=3d6168cb89f6537a81a72269305d4437df856891 --describe=v5.10.8-8301-g3d6168cb89f6 --describe-verbose=v5.10.8-8301-g3d6168cb89f6
Initialising build meta-data in /tmp/kci/linux/build
Install directory: /tmp/kci/linux/build/_install_
+ ./kci_build make_config --defconfig=x86_64_defconfig+kselftest
  File "/scratch/kernelci-core/./kci_build", line 427, in <module>
  File "/scratch/kernelci-core/./kci_build", line 342, in __call__
    res = step.run(args.j, args.verbose, opts)
  File "/scratch/kernelci-core/kernelci/build.py", line 1109, in run
    self._gen_kci_frag(configs, fragments, kci_frag_name)
  File "/scratch/kernelci-core/kernelci/build.py", line 1040, in _gen_kci_frag
    with open(path) as frag:
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/kci/linux/kernel/configs/kselftest.config'
```

gctucker (commented):
I've run the `kci_build generate_fragments` command locally and it's working fine, the kselftest fragment is being generated as expected.  So I'm not sure what's going on, I've started a full build of `cip_5.10` with the staging branch:
https://bot.staging.kernelci.org/job/kernel-build-trigger/1397/

config/core/build-configs-cip.yaml
```diff
@@ -0,0 +1,106 @@
...
+    url: "https://git.kernel.org/pub/scm/linux/kernel/git/cip/linux-cip.git"
+
+
+# Build fewer kernel configs with cip branches
+cip_variants: &cip_variants
+  gcc-10: &cip_gcc_10
+    build_environment: gcc-10
+    fragments: [tinyconfig]
+    architectures: &cip_architectures
+      arc:
```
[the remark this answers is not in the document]
gctucker: Ah sorry it is included in the build configs, was confused with the part that overrides the arm, am64 and x86_64 architectures.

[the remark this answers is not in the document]
aliceinwire: right

gctucker:
Tested OK on staging, except the x86_64_defconfig+x86-chromebook-kselftest build appears to be missing:
https://staging.kernelci.org/build/kernelci/branch/staging-cip/kernel/staging-cip-20211124.1/

It was triggered by Jenkins but then timed out, so maybe something went wrong in Kubernetes.  I'm running it again with a different cluster and will try to get the log to see what's going on.

gctucker:
Note: To get the log in real-time from Kubernetes:
```
kubectl --context gke_android-kernelci-external_us-east4-c_kci-big-us-east4 logs -f build-j39383-x86-64-gcc-10-x86-64-defconfig-x86-chromebook959mf
```
for this build job: https://bot.staging.kernelci.org/job/kernel-build/39383/console

gctucker:
OK that worked: https://staging.kernelci.org/build/id/619ebb98d80d9c3e731760ab/

So I think there was a glitch in Kubernetes with the previous job, which is not a problem caused by this PR.

Now getting some tests scheduled with this build...

gctucker:
Tested OK on staging with CIP 5.10 kernel:
https://staging.kernelci.org/test/job/kernelci/branch/staging-cip/kernel/staging-cip-20211124.1/

gctucker (approved):

PR #901, opened 2021-11-23
igt-parser.sh: Add amdgpu binaries directory to path

nfraprado:
The amdgpu igt binaries are installed in a subdirectory of the usual
path where other igt binaries are installed. Add this subdirectory to
the PATH variable as well so we can execute the amdgpu igt tests.


gctucker (approved):
Tested OK in LAVA:
https://lava.collabora.co.uk/results/5059744/0_igt-gpu-amd?search=&length=50#table

gctucker:
Results for i915 are also fine: https://lava.collabora.co.uk/results/5059743

PR #902, opened 2021-11-23
Debian Bullseye 20211118.0

gctucker:
Replace the plain Debian Buster rootfs images with Bullseye.


gctucker:
Tested OK on staging: https://staging.kernelci.org/test/job/kernelci/branch/staging-mainline/kernel/staging-mainline-20211123.1/

baseline-nfs, sleep and usb tests are showing expected results

PR #909, opened 2021-11-25
test-configs.yaml: add lenovo-hr330a-7x33cto1ww-emag

gctucker:
Add the lenovo-hr330a-7x33cto1ww-emag arm64 server device type and
enable only baseline-nfs on it as it takes a very long time to boot
with the current setup.  It relies entirely on UEFI and doesn't use
any dtb file.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -926,6 +926,11 @@ device_types:
     dtb: 'freescale/fsl-ls1028a-kontron-sl28-var3-ads2.dtb'
     boot_method: uboot
 
+  lenovo-hr330a-7x33cto1ww-emag:
+    mach: arm64
+    arch: arm64
+    boot_method: grub
+
   meson-g12a-sei510:
     mach: amlogic
     class: arm64-dtb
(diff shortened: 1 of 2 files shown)


gctucker:
Tested OK with a job in LAVA:
https://staging.lava.collabora.dev/scheduler/job/75419

However this took 8min to complete, for an actual LTP test case that took under 15s to run.  The firmware seems to be taking about 4min before starting a kernel, which is most likely going to be a very big bottleneck.

gctucker:
Some changes to the device type need to be made in upstream LAVA, to drop unnecessary kernel command line argument and to separate it from the `d03` template which is for an unrelated device type.

Also the firmware is taking about 4min before booting the kernel, which makes this device impractical for the tests we're currently running.

alexandrasp (commented):
There is a typo in the commit message "tempalte"

gctucker (commented):
Tested OK on staging:
https://staging.kernelci.org/test/plan/id/61c9eeacf2cb8cbae563555c/

The `sleep` template for GRUB isn't actually used but it doesn't hurt to add it for completeness.  Other devices using GRUB are likely to start running these tests.

PR #910, opened 2021-11-26
kernelci.build: replace more characters in publish path

gctucker:
Rather than simply replacing slash `/` characters, also replace
back-slash and colon characters to make the path is less likely to
cause issues with escape charaters and also more readable.  This is in
particular useful for kernels using a Chrome OS config.

### DIFF
--- a/kernelci/build.py
+++ b/kernelci/build.py
@@ -1110,16 +1110,16 @@ scripts/kconfig/merge_config.sh -O {output} '{base}' '{frag}' {redir}
 
         bmeta = self._meta.get('bmeta')
         rev, env = (bmeta[cat] for cat in ('revision', 'environment'))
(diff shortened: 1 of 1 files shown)


gctucker:
Note: This patch was originally added to the `chromeos` branch for the chromeos.kernelci.org instance.  It appears to be a useful thing to merge into the main branch to reduce "downstream" patches for Chrome OS support.

gctucker:
@aliceinwire Hello, would you like to review this PR and check there is no impact on the CIP builds?

gctucker:
Tested OK on staging with CIP config as well:
https://storage.staging.kernelci.org/cip/linux-4.19.y-cip/v4.19.220-cip63/x86_64/cip---4.19.y-cip-x86-cip_qemu_defconfig/gcc-10/

PR #912, opened 2021-11-30
CHROMEOS config/docker: add chromeos-tast image

gctucker:
Add a kernelci/chromeos-tast Docker image to run the host tools for
Chrome OS Tast tests.  This is intended to be used on LAVA dispatchers
to run tests on Chromebook devices running Chrome OS user-space.

### DIFF
--- a/config/docker/chromeos-tast/Dockerfile
+++ b/config/docker/chromeos-tast/Dockerfile
@@ -1,0 +1,43 @@
+FROM debian:bullseye-slim
+MAINTAINER "KernelCI TSC" <kernelci-tsc@groups.io>
+
+ARG DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update && apt-get install --no-install-recommends -y \
+    apt-transport-https \
+    bzip2 \
+    ca-certificates strace \
+    ssh
+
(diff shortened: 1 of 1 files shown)


gctucker:
Tested OK on staging, the image was pushed to `kernelci/chromeos-tast`.

config/docker/chromeos-tast/Dockerfile
```diff
@@ -0,0 +1,40 @@
+FROM debian:bullseye
```
mgalka, line 1: Do we need full `bullseye` image or could we do with `bullseye-slim` one?

alexandrasp (approved):
In general looks good.

config/docker/chromeos-tast/Dockerfile
gctucker: Good point, I'll give it a try to find out.

gctucker: Yep that worked, updated.

gctucker:
Tested OK with a LAVA job that just opens an SSH connection to a Chromebook running Chrome OS:
https://staging.lava.collabora.dev/scheduler/job/78777

PR #914, opened 2021-11-30
build-configs.yaml: remove clang-10

alexandrasp:
Now that it was raised the minimum supported version of LLVM to 11.0.0 we can drop clang-10 from Kernel CI.


config/core/build-configs.yaml
```diff
@@ -685,9 +685,9 @@ build_configs:
               - 'allnoconfig'
               - 'allmodconfig'
 
-      # clang-10 is the current minium clang version for -next
-      clang-10:
-        build_environment: clang-10
+      # clang-11 is the current minium clang version for -next
```
gctucker, line 688: Actually this comment is out-of-date, as clang-10 (and maybe clang-11 soon) is the minimal version for mainline.

gctucker:
On 02/12/2021 08:26, Masahiro Yamada wrote:
> Only positive feedback so far.
> 
> All applied to linux-kbuild.

This should land in linux-next soon.  I've updated the patches on staging to use clang-11 so we can start testing it ahead of time.

gctucker:
OK the changes are now in linux-next.  I think we can merge this but a few things will need to be fixed as a follow-up, since we still need to test with clang-10 until the next mainline merge window when the change will get merged.

PR #919, opened 2021-12-01
rootfs-configs.yaml: Add bullseye-cros-ec config

mgalka:
Add bullseye based image configuration for cros-ec tests.

### DIFF
--- a/config/rootfs/debos/scripts/strip.sh
+++ b/config/rootfs/debos/scripts/strip.sh
@@ -10,8 +10,7 @@ rm -rf /etc/localtime
 cp /usr/share/zoneinfo/Etc/UTC /etc/localtime
 
 
-UNNEEDED_PACKAGES=" libfdisk1"\
-" tzdata"\
+UNNEEDED_PACKAGES=" libfdisk1"
 
 export DEBIAN_FRONTEND=noninteractive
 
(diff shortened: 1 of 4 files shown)


gctucker:
Has it been built on staging?  I think that would be the next step.

mgalka:
> Has it been built on staging? I think that would be the next step.

Yes, it was on Friday and I also ran a test job with the new rootfs for x86_64 on a device. The `unittest` package seems still to be uninstalled. I investigated it and it seems that `strip.sh` script ran by `kci_rootfs` removes the package `libpython3.9-stdlib` before it gets marked as "manual". I've just pushed a workaround for it. I'll give it a testing round.

gctucker:
Have you tested your workaround on staging or with a LAVA manual resubmit?

config/rootfs/debos/scripts/strip.sh
```diff
@@ -10,8 +10,7 @@ rm -rf /etc/localtime
 cp /usr/share/zoneinfo/Etc/UTC /etc/localtime
 
 
-UNNEEDED_PACKAGES=" libfdisk1"\
-" tzdata"\
+UNNEEDED_PACKAGES=" libfdisk1"
```
gctucker, line 13: What is the difference in ramdisk size?  Ideally this should be a parameter in rootfs-configs.yaml to remove it when it's not neeed.  Maybe we could run cros-ec tests with NFS instead in this case, as it has the full rootfs image before being stripped.

mgalka:
Closing this PR as the fix has been split into:
https://github.com/kernelci/kernelci-core/pull/995
https://github.com/kernelci/kernelci-core/pull/996

PR #769, opened 2021-07-26
Allow LTP and kselftest to run in my lab

broonie:

PR #920, opened 2021-12-01
test-configs: Enable MTE for virtual arm64 platforms

broonie:
The arm64 Memory Tagging Extension (FEAT_MTE) is supported by qemu but is
not enabled, even with cpu=max. Get some test coverage for it by enabling
it for our virtual arm64 platforms when the built kernel supports it. There
is some specific coverage for it in kselftest too which will be exercised
when this is enabled.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -1257,7 +1257,7 @@ device_types:
       arch: arm64
       cpu: 'max'
       guestfs_interface: 'virtio'
-      machine: 'virt,gic-version=2,accel=kvm:tcg'
+      machine: 'virt,gic-version=2,mte=on,accel=kvm:tcg'
     filters:
       - regex: { defconfig: 'defconfig' }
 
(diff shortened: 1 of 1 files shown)


gctucker:
This conflicts with the earlier commit https://github.com/kernelci/kernelci-core/commit/e35518652ef0c08f5a686c45cbf96ca36c6a918a.  Please rebase

gctucker:
@broonie OK I've now merged #769 so please rebase again.  That's the one that was conflicting as it's enabling KVM in QEMU.

gctucker:
All `arm64` QEMU jobs are failing `baseline` since today in `lab-broonie` and `lab-baylibre`, I wonder if it's related to this change.  Could you please take a look?

If it's due to the QEMU version being used, we probably should start transitioning all the QEMU jobs to Docker in LAVA so they'll all be running the exact same binaries regardless of the lab.

gctucker:
@broonie Hello, could you please take a look at the QEMU errors as per the comment above?

broonie:
It looks like both labs are running too old a version of qemu to have MTE support. As discussed in the meeting moving things over to scheduling jobs with our own dockerized qemu is probably the way forward here.

gctucker:
OK I think we'll merge this one and then rework things a bit to require Docker for QEMU jobs as a follow-up.

gctucker (approved):
Tested OK on labs with a recent enough version of QEMU:
https://staging.kernelci.org/test/plan/id/61f88b64a699d87bc138de3d/

Issue #887, opened 2021-11-19
buster-libcamera failing to build on arm64


gctucker:
The `buster-libcamera` rootfs image is failing to build today with this error:
```
2021/11/18 23:47:20 buster-libcamera.sh | qemu: uncaught target signal 11 (Segmentation fault) - core dumped

2021/11/18 23:47:20 buster-libcamera.sh | /tmp/script/buster-libcamera.sh: line 50:  9907 Segmentation fault      meson build --prefix=/usr -Ddocumentation=disabled -Dgstreamer=disabled -Dpipelines=uvcvideo,rkisp1

2021/11/18 23:47:20 Action `Build testsuite` failed at stage Run, error: exit status 139
Powering off.
```
It built fine last week, the binaries can be found here:
https://storage.kernelci.org/images/rootfs/debian/buster-libcamera/20211112.0/arm64/

None of the other arm64 builds are failing, apart from buster-ltp but that's a known issue.  So it's most likely due to some changes in libcamera or its dependencies.

We may try with QEMU 6 from Debian backports as it's now available for Bullseye (the `kernelci/debos` Docker image is based on Bullseye).  We may also try to upgrade with `bullseye-libcamera` as we'll need to do that soon anyway, and the plain `bullseye` rootfs images are nearly ready for production.

nfraprado:
A couple days ago, the way lc-compliance is built changed in libcamera: https://git.linuxtv.org/libcamera.git/commit/?id=0d50a04cc918e5122d1b22201cc949c2cfd337a5

Trying to compile lc-compliance locally using `kci_rootfs` in the `kernelci/debos` container reproduces the issue:
```
2021/11/19 14:24:06 buster-libcamera.sh | Downloading gtest source from https://github.com/google/googletest/archive/release-1.11.0.zip
2021/11/19 14:24:07 buster-libcamera.sh | Downloading file of unknown size.
2021/11/19 14:24:07 buster-libcamera.sh | Downloading gtest patch from https://wrapdb.mesonbuild.com/v2/gtest_1.11.0-1/get_patch
2021/11/19 14:24:08 buster-libcamera.sh | qemu: uncaught target signal 11 (Segmentation fault) - core dumped
[...]

PR #924, opened 2021-12-06
Revert "config/docker: upgrade debos to Bullseye"

alexandrasp:
This reverts commit 69cff1b4df3e6cecac4b9778c93807aedf7c7dab.

### DIFF
--- a/config/docker/debos/Dockerfile
+++ b/config/docker/debos/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:bullseye-slim
+FROM debian:buster-slim
 
 ARG DEBIAN_FRONTEND=noninteractive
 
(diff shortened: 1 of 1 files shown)


gctucker:
What is this about?

aliceinwire:
What's happen with Bullseye? Any problem to link ?

alexandrasp:
@gctucker This is about libcamera bug, I have tried with qemu 6 from Bullseye backports and got the same error. I put a comment on issue #887, I also tried using a Debian testing, same problem. So, you have suggested to revert that commit as a workaround until we've come up with a better way of building things from source, ok? 

@aliceinwire The problem is not with Bullseye, but with the qemu package used by Bullseye, it is breaking the build for libcamera rootfs on arm64.

gctucker:
@alexandrasp No, I was referring to the commit in libcamera that @nfraprado mentioned.

OK so we don't need to revert to Buster, closing this now.

PR #925, opened 2021-12-06
T31651 detached process

gctucker:
Run scripts in `kernelci.lab.Shell` as detached processes and pass the YAML configuration for the remote API to the template so they can send the results directly.


gctucker:
Tested OK locally with https://github.com/kernelci/kernelci-pipeline

PR #928, opened 2021-12-10
test configs: Allow ltp-timers to be tested in qemu device

lubinszARM:
Signed-off-by: Robin Lu <robin.lubin@shopee.com>

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -72,6 +72,10 @@ file_systems:
     nfs: 'bullseye-ltp/20211221.0/{arch}/full.rootfs.tar.xz'
     root_type: nfs
 
+  debian_bullseye-ltp_ramdisk:
+    type: debian
+    ramdisk: 'bullseye-ltp/20211221.0/{arch}/rootfs.cpio.gz'
+
   debian_bullseye-rt_ramdisk:
     type: debian
     ramdisk: 'bullseye-rt/20211216.0/{arch}/rootfs.cpio.gz'
@@ -357,6 +361,16 @@ test_plans:
   ltp-timers:
     <<: *ltp
     pattern: 'ltp/{category}-{method}-{protocol}-{rootfs}-ltp-open-posix-template.jinja2'
+    params:
+      <<: *ltp-params
+      grp_test: "TMR"
+      job_timeout: '30'
+
+  ltp-timers_qemu:
+    <<: *ltp
+    base_name: ltp-timers
+    rootfs: debian_bullseye-ltp_ramdisk
+    pattern: 'ltp/{category}-{method}-{rootfs}-ltp-open-posix-template.jinja2'
     params:
       <<: *ltp-params
       grp_test: "TMR"
@@ -2322,6 +2336,7 @@ test_configs:
     test_plans:
       - baseline_qemu
       - baseline-qemu-docker
+      - ltp-timers_qemu
       - smc
       - v4l2-compliance-vivid
 
--- a/config/lava/ltp/generic-qemu-ramdisk-ltp-open-posix-template.jinja2
+++ b/config/lava/ltp/generic-qemu-ramdisk-ltp-open-posix-template.jinja2
@@ -1,0 +1,11 @@
+{% extends 'boot/generic-qemu-boot-template.jinja2' %}
+{% block actions %}
+{{ super () }}
+
+{% include 'ltp/ltp-open-posix.jinja2' %}
+
+{% endblock %}
+
+{%- block image_arg %}
+        image_arg: '-kernel {kernel} -append "console={{ console_dev }},115200 root=/dev/ram0 debug verbose {{ extra_kernel_args }}"'
+{%- endblock %}


lubinszARM:
Hi @gctucker 
I have deployed qemu+ltp/kselftest related tests in my local kernelci+lava environment and it has been running stably for a period of time.
Also, please see the log of qemu+ltp as reference.
[lava_qemu_ltp_timers.log](https://github.com/kernelci/kernelci-core/files/7691823/lava_qemu_ltp_timers.log)

lubinszARM:
cc @broonie

config/core/test-configs.yaml
```diff
@@ -84,7 +84,7 @@ file_systems:
 
   debian_buster-ltp_nfs:
     type: debian
-    ramdisk: 'buster-ltp/20211203.0/{arch}/initrd.cpio.gz'
+    ramdisk: 'buster-ltp/20211203.0/{arch}/rootfs.cpio.gz'
```
gctucker, line 87: The `initrd.cpio.gz` is created especially to then be able to switch the root file system to NFS.  So this change can't be done, the `rootfs.cpio.gz` is to boot with just a ramdisk which has a full root file system.

We either need to fix NFS root on QEMU or make a variant for LTP tests to run on QEMU with a ramdisk.

config/core/test-configs.yaml
```diff
@@ -1342,6 +1342,7 @@ device_types:
     context:
       arch: x86_64
       cpu: 'qemu64'
+      memory: 1024
```
gctucker, line 1345: Why was this needed?

config/core/test-configs.yaml
lubinszARM: I see. I will try to add a new variant for this case.

config/core/test-configs.yaml
lubinszARM: The default 512M seems too small. When running ltp/kselftest+ qemu related cases, there will be random OOM crash problems. Especially in ltp-mm, kselftest-lib ...

config/core/test-configs.yaml
lubinszARM: Hi @gctucker 
I modified a new version according to your suggestion.
Please help to check it.
Thanks.

config/core/test-configs.yaml
gctucker: OK thanks, let's see if we can have ltp-timers working with the default memory size on QEMU and then if it's needed for other tests we can potentially increase it.

One issue is that some labs may be configured for a certain maximum amount of memory per QEMU instance, so changing the memory attribute may not be enough if there isn't enough actual physical RAM on the machines where the jobs are run.  We might need to add an attribute in the test plan definitions about the amount of memory required so labs can enable only those that they will be able to run.  But that would be a discussion for a follow-up PR I think.

config/core/test-configs.yaml
```diff
@@ -336,6 +339,15 @@ test_plans:
       grp_test: "TMR"
       job_timeout: '30'
 
+  ltp-timers-qemu:
+    <<: *ltp
+    rootfs: debian_buster-ltp_qemu
```
gctucker, line 344: To group the results under ltp-timers, with the tests run on non-QEMU devices:
```suggestion
  ltp-timers_qemu:
    <<: *ltp
    basename: ltp-timers
    rootfs: debian_buster-ltp_qemu
```

gctucker (commented):
Tested OK on staging, except that it didn't work at all in lab-broonie for some reason:
https://staging.kernelci.org/test/job/kernelci/branch/staging-mainline/kernel/staging-mainline-20211214.0/plan/ltp-timers-qemu/

@broonie Do you know why these failed only in your lab?

It worked in the Collabora and LKFT labs although there were a few test case failures in all the jobs.  It would be good to see if these can be fixed (missing config?) or if they should be skipped.

config/core/test-configs.yaml
```diff
@@ -339,8 +339,9 @@ test_plans:
       grp_test: "TMR"
       job_timeout: '30'
 
-  ltp-timers-qemu:
+  ltp-timers_qemu:
```
lubinszARM, line 342: In the test plan is still ltp-timers-qemu, we also need to change it.

gctucker: Sure.  Actually I think this fix can be squashed with the first commit.

gctucker:
Adding `staging-skip` label as there's a YAML validation error as anticipated in a comment:
```
KeyError: 'ltp-timers-qemu'
```

config/core/test-configs.yaml
lubinszARM: Done.

lubinszARM:
@gctucker 
Any progress? Can we merge it?

gctucker:
I believe the change in the rootfs images hasn't been tested yet, with the symlink stage moved to a different place.

lubinszARM:
> symlink

Oh. I see.
But https://github.com/kernelci/kernelci-core/pull/929 has nothing to do with this PR.
full.rootfs.cpio.gz is only required by kselftest@qemu related cases.
For ltp@qemu, rootfs.cpio.gz can already meet the requirements.

config/core/test-configs.yaml
```diff
@@ -88,6 +88,9 @@ file_systems:
     nfs: 'buster-ltp/20211203.0/{arch}/full.rootfs.tar.xz'
     root_type: nfs
 
+  debian_buster-ltp_qemu:
```
gctucker, line 91:
```suggestion
  debian_buster-ltp_ramdisk:
```

config/core/test-configs.yaml
```diff
@@ -336,6 +339,16 @@ test_plans:
       grp_test: "TMR"
       job_timeout: '30'
 
+  ltp-timers_qemu:
+    <<: *ltp
+    basename: ltp-timers
+    rootfs: debian_buster-ltp_qemu
```
gctucker, line 345:
```suggestion
    rootfs: debian_buster-ltp_ramdisk
```

config/core/test-configs.yaml
```diff
@@ -336,6 +339,16 @@ test_plans:
       grp_test: "TMR"
       job_timeout: '30'
 
+  ltp-timers_qemu:
+    <<: *ltp
+    basename: ltp-timers
```
gctucker, line 344:
```suggestion
    base_name: ltp-timers
```

gctucker (commented):
The tests were run fine:
https://staging.kernelci.org/test/job/kernelci/branch/staging-next/kernel/staging-next-20211220.0/plan/ltp-timers_qemu/

A couple of small tweaks are still needed then this should be ready to be merged.

lubinszARM:
@gctucker 
Sorry for delay due to the holiday. Please check it again. :)

config/core/test-configs.yaml
```diff
@@ -88,7 +88,9 @@ file_systems:
     type: debian
     ramdisk: 'buster-v4l2/20211224.3/{arch}/rootfs.cpio.gz'
 
-
+  debian_buster-ltp_ramdisk:
```
gctucker, line 91: We've moved on to Bullseye now for this rootfs, so please use `debian_bullseye-ltp_ramdisk` and move it after the `-nfs` variant.

config/core/test-configs.yaml
```diff
@@ -356,6 +358,16 @@ test_plans:
       grp_test: "TMR"
       job_timeout: '30'
 
+  ltp-timers_qemu:
+    <<: *ltp
+    base_name: ltp-timers
+    rootfs: debian_buster-ltp_ramdisk
```
gctucker, line 364:
```suggestion
    rootfs: debian_bullseye-ltp_ramdisk
```

config/core/test-configs.yaml
lubinszARM: Done.

config/core/test-configs.yaml
lubinszARM: Done.

gctucker (approved):
This one is failing to boot in @broonie's lab:
https://staging.kernelci.org/test/plan/id/61d4f37a573563a138635574/

But otherwise results seem valid in the other labs:
https://staging.kernelci.org/test/plan/id/61d4e8c5b30a9bbea1635589/ (gcc-10)
https://staging.kernelci.org/test/plan/id/61d500d188931996d763555c/ (clang-11)

So I think this PR can be merged but we should take a look at what's causing the failure.

gctucker:
The commits also need to be squashed or reorganised.

gctucker:
@lubinszARM Please rebase and squash fixups

lubinszARM:
> @lubinszARM Please rebase and squash fixups

Hi
Done.

lubinszARM:
Any comments?

gctucker (approved):
Tested OK on staging:
https://staging.kernelci.org/test/plan/id/61f17f2f5753e3db1338decf/

It's failing to boot in @broonie's lab but that's a separate issue I believe:
https://staging.kernelci.org/test/plan/id/61f1794cc1032fb06d38de22/

PR #938, opened 2021-12-14
config/docker: Factor out common elements of clang containers

broonie:
The clang docker containers contain a lot of duplicated content installing
utlilities used in the clang build, runtime dependencies for the kernel
build system and removing GCC. This means that any updates to those shared
elements need to touch multiple Dockerfiles. Simplify this by creating a
clang-base container which has everything except the actual clang packages
and basing the version specific clang packages on that.

Since clang-10 is now deprecated and the containers for it are based on
an older Debian release which isn't quite the same as the rest it is not
updated.

### DIFF
--- a/config/docker/build-and-push.sh
+++ b/config/docker/build-and-push.sh
@@ -87,6 +87,7 @@ docker_build_and_tag() {
 
 for target in $targets; do
     if [ "$target" = "compilers" ]; then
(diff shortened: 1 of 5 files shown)


gctucker (approved):
LGTM.  Tested OK on staging.

PR #949, opened 2021-12-16
test-configs.yaml: add arm64-chromebook to default filter

gctucker:
Add defconfig+arm64-chromebook to the default test plan filter so test
plans will be enabled by default on Chromebooks that rely on this
config fragment (mt8173-elm-hana at the moment).

Fixes: aeab1ef0cdc1 ("build-configs.yaml: add arm64-chromebook fragment")

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -92,6 +92,7 @@ test_plan_default_filters:
       values:
         - ['arm', 'multi_v7_defconfig']
         - ['arm64', 'defconfig']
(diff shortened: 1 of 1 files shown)


gctucker:
Tested OK on staging, for example the preempt-rt results are now here:
https://staging.kernelci.org/test/job/kernelci/branch/staging-mainline/kernel/staging-mainline-20211217.0/plan/preempt-rt/

(clang-11 ones are missing but that was due to a Jenkins/Kubernetes sync issue)

PR #951, opened 2021-12-17
Enable cpufreq kselftest on mt8173-elm-hana

nfraprado:
This adds a new kselftest test plan for the cpufreq collection, and enables it to run on mt8173-elm-hana.

The test output log can be seen here: https://lava.collabora.co.uk/scheduler/job/5242522

To run this test, `CONFIG_ARM_MEDIATEK_CPUFREQ` needs to be enabled as built-in. [A patch has been sent to the mailing list adding it to the defconfig](https://lore.kernel.org/linux-mediatek/20211217154452.868419-1-nfraprado@collabora.com/T/#u).

After that patch gets merged, we will still either need to block this test on older kernel branches tested by KernelCI or have a config fragment anyway for those, right?

### DIFF
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -140,6 +140,7 @@ fragments:
     path: "kernel/configs/arm64-chromebook.config"
     configs:
       - 'CONFIG_REGULATOR_DA9211=y'
+      - 'CONFIG_ARM_MEDIATEK_CPUFREQ=y'
 
   crypto:
     path: "kernel/configs/crypto.config"
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -227,6 +227,12 @@ test_plans:
     filters:
       - passlist: {defconfig: ['kselftest']}
 
+  kselftest-cpufreq:
+    <<: *kselftest
+    params:
+      job_timeout: '10'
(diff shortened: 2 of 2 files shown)


gctucker:
> After that patch gets merged, we will still either need to block this test on older kernel branches tested by KernelCI or have a config fragment anyway for those, right?

Yes, but in fact there's already the `arm64-chromebook` fragment which is needed for `mt8173-elm-hana`.  So we could add that extra config to it.  Or we can add a filter to this test plan to block some kernel revisions.  But let's see if your patch is going to get merged as other arm64 defconfig changes have been waiting to be applied for a long time (which is why we now have this fragment).

gctucker:
Alright I've added a commit to the PR to enable this config option, we can drop it later if the patch gets merged upstream.

gctucker (approved):
Tested OK on staging:
https://staging.kernelci.org/build/id/61c5c1a7d868697e086355c1/
https://storage.staging.kernelci.org/kernelci/staging-next/staging-next-20211224.0/arm64/defconfig+arm64-chromebook/gcc-10/config/kernelci.config
https://staging.kernelci.org/build/id/61c5bf7aedebb8daab63559c/
https://staging.kernelci.org/test/plan/id/61c5c0a400bace1a0363555d/ (kselftest-cpufreq)

PR #953, opened 2021-12-20
kernelci.config: default filters and clean up

gctucker:
Group default filters under `default_filters` and clean up YAML configuration.


gctucker (commented):
Tested OK on staging.

PR #957, opened 2021-12-20
T32412 kernelci.lab.LabAPI: add .save_file()

gctucker:
Add LabAPI.save_file() method to let each lab implementation the
possibility to save job definition files in a particular way.  Move
the implementation from kci_test to the LabAPI base class to provide
the same generic default behaviour.


mgalka (approved):
looks good to me

PR #963, opened 2021-12-22
T32412 Add Shell lab type

gctucker:
Add `kernelci.lab.shell` to run tests in a local process. Update `kci_test` accordingly to be able to use it.

Note: The script templates are not part of this PR as they are instead being added to [`kernelci-pipeline`](https://github.com/kernelci/kernelci-pipeline).

### DIFF
--- a/kernelci/lab/shell.py
+++ b/kernelci/lab/shell.py
@@ -1,0 +1,29 @@
+import os
+import subprocess
+from jinja2 import Environment, FileSystemLoader
+from kernelci.lab import LabAPI
+
+
+class Shell(LabAPI):
+
+    def generate(self, params, device_config, plan_config,
+                 callback_opts=None, templates_path='config/scripts'):
+        jinja2_env = Environment(loader=FileSystemLoader(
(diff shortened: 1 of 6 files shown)


kernelci/lab/shell.py
```diff
@@ -0,0 +1,33 @@
...
+            params['db_config_yaml'] = db_config.to_yaml()
+        return template.render(params)
+
+    def save_file(self, *args, **kwargs):
+        output_file = super().save_file(*args, **kwargs)
+        os.chmod(output_file, 0o775)
+        return output_file
+
+    def submit(self, job_path, get_process=False):
+        process = subprocess.Popen(job_path, shell=True)
```
mgalka, line 27: Is it needed here to set `shell=True`.  If it's not necessary to run the command in a shell we should consider setting it to False.

gctucker: Yes that's right, it's not required.  I've removed it now.

gctucker (commented):
Tested OK on staging and locally with kernelci-pipeline.

PR #965, opened 2021-12-22
WIP: Create a new lava job template to be used by Chromebook

alexandrasp:
config/core/test.yaml: add test_plan, file_system and buildroot hypothetical entries to be used in Chromebook devices and their lava job template.

lava/chrome: include basic jinja2 template for Chromebook devices.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -14,6 +14,11 @@ file_system_types:
       x86:     [{arch: i386}, {arch: x86_64}]
       mipsel:  [{arch: mips}]
 
+  chrome-os:
+    url: 'to be defined'
+    arch_map:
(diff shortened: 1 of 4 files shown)


alexandrasp:
@gctucker I opened this PR as draft because we have some points in progress in the chrome-os integration in Kernel CI, not sure yet if it is the right option use boot-nfs as base, so..we can start discussing and improving step-by-step until we complete the whole pipeline.

gctucker:
Running Tast tests needs to be done with the Chrome OS flashed on the device, so it's not NFS root but using the on-board memory.  The boot method is `emmc`.  See this sample job:
https://lava.collabora.co.uk/scheduler/job/5252267

PR #971, opened 2021-12-24
build-configs.yaml: add INTERCONNECT_QCOM_SC7180 to arm64-chromebook

gctucker:
Add CONFIG_INTERCONNECT_QCOM_SC7180=y to the arm64-chromebook fragment which is needed for sc7180-trogdor-lazor-limozeen.

### DIFF
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -145,6 +145,9 @@ fragments:
       - 'CONFIG_REGULATOR_DA9211=y'
       - 'CONFIG_ARM_MEDIATEK_CPUFREQ=y'
       - 'CONFIG_RTC_DRV_MT6397=y'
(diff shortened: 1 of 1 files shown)


nuclearcat:
According to email: https://groups.io/g/kernelci/message/1245?p=%2C%2C%2C20%2C0%2C0%2C0%3A%3Arecentpostdate%2Fsticky%2C%2CEnable+SC7180+interconnect+driver%2C20%2C2%2C0%2C86672319
kernel was hanging before.
It seems in some tests this chromebook passing tests successfully: https://lava.collabora.dev/scheduler/job/7206151

gctucker (commented):
Thanks for checking, yes it looks like it's working now.  Also I've added a comment about the missing config upstream, that might be dropped in the future if it eventually gets merged.

PR #972, opened 2021-12-27
build-configs-android: enable clang-14 android builds

musamaanjum:
Enable clang-14 for android builds. Also enable the GKI builds for
android with only clang-14.


gctucker:
Please use a branch from your forked repository when creating pull requests.  I think it means recreating one in this case as PRs can't have their incoming branch edited, only the base target branch can be edited.

Issue #489, opened 2020-09-02
build Android common kernel trees with Clang


nickdesaulniers:
Clang and LLD are **required** for building Android kernels, and there are VTS tests that enforce that.  We should be testing Android trees with Clang (GCC doesn't hurt, but we should also be using the configs that Android uses such as the GKI configs, which may not build with GCC.  I'll file a separate bug about GKI though).  cc @adelva1984 

- [x] #933
- [x] #934
- [x] #935
- [x] #936
- [x] #988

musamaanjum:
I'm starting working on https://github.com/kernelci/kernelci-core/issues/935. We should be aware that build on android tree will fail because of the following two issues in some configurations:
1. https://github.com/ClangBuiltLinux/linux/issues/1473. It has been fixed. The fix is present in linux-next (2e705706), but not in v5.15-rc6 and hence not in Andriod kernels.
2. https://github.com/ClangBuiltLinux/linux/issues/1497. Fixes of these bugs haven't reached to linux-next, mainline and hence not in Andriod kernels.

musamaanjum:
From Nathan on IRC:
> The first patch is in -next and it will be merged into 5.17 when the merge window opens. The second patch is not in -next (but should be) and I was hoping it would be merged in 5.16 but it does not seem like it will be. Once the second patch is at least in -next, clang-14 can be turned on for that tree.

gctucker:
OK, I think we could apply these patches on the staging-next branch for builds run on staging.kernelci.org.  Could you please try to identify these pending patches?

musamaanjum:
1. Fix is in linux-next: 2e70570656ad ("drm/i915: Avoid bitwise vs logical OR warning in snb_wm_latency_quirk()")
2. In [power supply tree](https://git.kernel.org/pub/scm/linux/kernel/git/sre/linux-power-supply.git) which doesn't flow to linux-next: 644106cdb898 ("power: reset: ltc2952: Fix use of floating point literals")

gctucker:
OK thanks, I've added `644106cdb898 ("power: reset: ltc2952: Fix use of floating point literals")` as a patch file for the staging-next kernel branch used on staging.kernelci.org.  It's just copied to a directory on the server where a script is run to push that branch, you'll find it in GitHub when it's run.  We'll see if this fixes some clang-14 builds.

[...]

PR #974, opened 2021-12-27
build-configs-android: enable clang-14 android builds

musamaanjum:
Enable clang-14 builds on 5.10 and 5.15 android branches. Also enable
the GKI configs.

### DIFF
--- a/config/core/build-configs-android.yaml
+++ b/config/core/build-configs-android.yaml
@@ -41,6 +41,27 @@ android_variants: &android_variants
             - 'allmodconfig'
             - 'allnoconfig'
 
+clang_android_variants: &clang_android_variants
+  clang-14:
+    build_environment: clang-14
+    architectures:
+        arm: *arm_arch
+        arm64:
+          <<: *arm64_arch
+          extra_configs:
+            - 'allmodconfig'
+            - 'allnoconfig'
+            - 'defconfig+CONFIG_CPU_BIG_ENDIAN=y'
+            - 'defconfig+CONFIG_RANDOMIZE_BASE=y'
+            - 'gki_config'
+        i386: *i386_arch
+        riscv: *riscv_arch
+        x86_64:
+          <<: *x86_64_arch
+          extra_configs:
+            - 'allmodconfig'
+            - 'allnoconfig'
+            - 'gki_config'
 
 build_configs:
 
@@ -142,19 +163,27 @@ build_configs:
   android12-5.10:
     tree: android
     branch: 'android12-5.10'
-    variants: *android_variants
+    variants:
+      <<: *android_variants
+      <<: *clang_android_variants
 
   android12-5.10-lts:
     tree: android
     branch: 'android12-5.10-lts'
-    variants: *android_variants
+    variants:
+      <<: *android_variants
+      <<: *clang_android_variants
 
   android13-5.10:
     tree: android
     branch: 'android13-5.10'
-    variants: *android_variants
+    variants:
+      <<: *android_variants
+      <<: *clang_android_variants
 
   android13-5.15:
     tree: android
     branch: 'android13-5.15'
-    variants: *android_variants
+    variants:
+      <<: *android_variants
+      <<: *clang_android_variants


config/core/build-configs-android.yaml
```diff
@@ -41,6 +41,25 @@ android_variants: &android_variants
...
+            - 'defconfig+CONFIG_RANDOMIZE_BASE=y'
+            - 'gki_config'
+        i386: *i386_arch
+        riscv: *riscv_arch
+        x86_64:
+          base_defconfig: 'x86_64_defconfig'
+          extra_configs:
+            - 'allmodconfig'
+            - 'allnoconfig'
+            - 'gki_config'
```
gctucker, line 62: It's a bit better to inherit from the definition in `gcc-10` and override fields that change, for example:
```suggestion
        x86_64:
          <<: * x86_64_arch
          extra_configs:
            - 'allmodconfig'
            - 'allnoconfig'
            - 'gki_config'
```
The same can be done for `arm64` even though there's nothing to inherit.  It makes the config easier to maintain as future changes in the `gcc-10` section that are architecture-specific would be automatically propagated to the other sections such as `clang-14`.

gctucker (commented):
LGTM, let's see if we can get some results on staging.

If some builds aren't working at all, we may have to drop them from this PR and deal with them separately (if additional compiler options are needed etc.).  But that will depend on the results.

gctucker:
I wonder if all the kernel branches should be built with `clang-14` or even with Clang at all, the list goes back to 4.4.  It might be worth asking on IRC, and if we need to only enable recent kernels with `clang-14` then we could create an `android_variants_clang` definition to use with those branches only.

gctucker:
Same thing for `gki_config` - in which kernel revision were they introduced?

musamaanjum:
I'd asked on IRC yesterday. I'm waiting for reply:
> In kernelCI, we have android kernels which date back to `4.4`. GKI configs for android are present in android-4.19-stable, android11-5.4 and so on. We should enable the android builds with clang-14 only on recent kernels. Which kernels seem good candidates to be enabled on clang-14? kernels 5.4 onward seem good candidates. Android's currently enabled branches can be found here: config/core/build-configs-android.yaml

gctucker:
OK it's probably also possible to look for the presence of the `gki_config` files in each branch.

gctucker:
Staging builds are here:
https://staging.kernelci.org/build/android/branch/android13-5.15/kernel/ASB-2021-12-05_13-5.15-461-gf4bb4932c9ec/

In particular:
FAIL https://staging.kernelci.org/build/id/61c9be82a2f2f74e6363557b/ `x86_64_defconfig`
FAIL https://staging.kernelci.org/build/id/61c9b2fca6ca40fe9663557c/ `multi_v7_defconfig`

The `defconfig` builds for `riscv` and `arm64` passed.  We should check whether these build errors are expected (i.e. known problems in the kernel tree) or if they are due to KernelCI not building correctly.

musamaanjum:
All the failing builds [here](https://staging.kernelci.org/build/android/branch/android13-5.15/kernel/ASB-2021-12-05_13-5.15-461-gf4bb4932c9ec/) will be solved by the already identified two patches except the following builds:
[ixp4xx_defconfig ‐ arm ‐ clang-14](https://storage.staging.kernelci.org/android/android13-5.15/asb-2021-12-05_13-5.15-461-gf4bb4932c9ec/arm/ixp4xx_defconfig/clang-14/logs/build-errors.log)
[rpc_defconfig ‐ arm ‐ clang-14](https://storage.staging.kernelci.org/android/android13-5.15/asb-2021-12-05_13-5.15-461-gf4bb4932c9ec/arm/rpc_defconfig/clang-14/logs/build-errors.log)
[pxa_defconfig ‐ arm ‐ clang-14](https://storage.staging.kernelci.org/android/android13-5.15/asb-2021-12-05_13-5.15-461-gf4bb4932c9ec/arm/pxa_defconfig/clang-14/logs/build-errors.log)
These are failing with different build errors.

gctucker:
> All the failing builds [here](https://staging.kernelci.org/build/android/branch/android13-5.15/kernel/ASB-2021-12-05_13-5.15-461-gf4bb4932c9ec/) will be solved by the already identified two patches except the following builds: [ixp4xx_defconfig ‐ arm ‐ clang-14](https://storage.staging.kernelci.org/android/android13-5.15/asb-2021-12-05_13-5.15-461-gf4bb4932c9ec/arm/ixp4xx_defconfig/clang-14/logs/build-errors.log) [rpc_defconfig ‐ arm ‐ clang-14](https://storage.staging.kernelci.org/android/android13-5.15/asb-2021-12-05_13-5.15-461-gf4bb4932c9ec/arm/rpc_defconfig/clang-14/logs/build-errors.log) [pxa_defconfig ‐ arm ‐ clang-14](https://storage.staging.kernelci.org/android/android13-5.15/asb-2021-12-05_13-5.15-461-gf4bb4932c9ec/arm/pxa_defconfig/clang-14/logs/build-errors.log) These are failing with different build errors.

OK, thanks for checking.  I don't think this will be a blocking issue but we should probably review the list of defconfigs being built with Android branches anyway.  Maybe we can stop building all the `arm` ones specific to a particular platform, it's something to confirm with Google.  It would save resources and make the results land quicker.

gctucker:
Confirmed on staging that all clang-14 builds are passing on linux-next with the extra fix applied:
https://staging.kernelci.org/build/kernelci/branch/staging-next/kernel/staging-next-20211228.1/

musamaanjum:
After discussing with @gctucker, I've updated the PR to enable only 5.10 and 5.15 andorid branches with clang 14. GKI config is also present on these branches. We can always include more android branches after consulting Google.

gctucker (approved):
LGTM, starting a build job on staging to check it's producing valid results.

musamaanjum:
Checked the results: https://staging.kernelci.org/build/android/branch/android13-5.15/kernel/ASB-2022-01-05_13-5.15-55-g9e280ea43eaa/

Other than builds already mentioned [failing builds above](https://github.com/kernelci/kernelci-core/pull/974#issuecomment-1002433487), everything should be good after we add the [previously identified 2 patches](https://github.com/kernelci/kernelci-core/issues/489#issuecomment-1001963194) to android stagging as well.

gctucker (approved):
Tested OK on staging.

PR #975, opened 2021-12-27
Fix GitHub actions running checks

JenySadadia:

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -19,21 +19,23 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        python-version: ['3.7', '3.10']
 
     steps:
 
-      # Checks out the repository under $GITHUB_WORKSPACE
-      - uses: actions/checkout@v2
+      - name: Check out source code
+        uses: actions/checkout@v2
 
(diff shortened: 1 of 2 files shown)


JenySadadia:
Tested on 'fix-python-version' branch. Github actions check completed successfully.
https://github.com/JenySadadia/kernelci-core/actions/runs/1626830604

.github/workflows/main.yml
```diff
@@ -19,21 +19,23 @@ jobs:
     runs-on: ubuntu-20.04
     strategy:
       matrix:
-        python-version: [3.7, 3.8, 3.9]
+        python-version: ['3.10'] # Only supported one at the moment
```
gctucker, line 22: Actually that was true for kernelci-api, but I believe kernelci-core might work with earlier versions.  So it's worth trying with 3.7 and 3.10 to see what happens.

gctucker (commented):
LGTM - just maybe one thing to tweak to see if earlier Python versions work as well.

.github/workflows/main.yml
JenySadadia: I have checked with python version 3.7, 3.8, 3.9, and 3.10
https://github.com/kernelci/kernelci-core/actions/runs/1626891867
It is working with all the versions.

gctucker (approved):

PR #979, opened 2022-01-04
Add the implementation in kernelci.data.kernelci_api.submit() to use PUT

JenySadadia:
Fixes: https://github.com/kernelci/kernelci-api/issues/66

kernelci/data/kernelci_api.py: Add put handler
kernelci/data/kernelci_api.py: Add logic in submit() to use PUT
in case the object already exists

### DIFF
--- a/kernelci/data/kernelci_api.py
+++ b/kernelci/data/kernelci_api.py
@@ -79,7 +79,10 @@ class KernelCI_API(Database):
         obj_list = []
         for path, item in data.items():
             try:
-                resp = self._post(path, json.dumps(item))
+                if 'id' in item.keys():
+                    resp = self._put(path, json.dumps(item))
+                else:
+                    resp = self._post(path, json.dumps(item))
             except requests.exceptions.HTTPError as ex:
                 self._print_http_error(ex, verbose)
                 raise(ex)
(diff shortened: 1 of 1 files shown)


kernelci/data/kernelci_api.py
```diff
@@ -69,6 +69,12 @@ def get_node(self, node_id):
     def get_node_from_event(self, event):
         return self.get_node(event.data['id'])
 
+    def _put(self, path, data=None):
```
gctucker, line 72: Maybe move it next to `_get()` and `_post()` to keep the primitive HTTP wrappers together

JenySadadia: Okay.

gctucker (approved):
LGTM.  I'll try to use it with the related API changes to manually check it's working as expected, then it should be ready to be merged.

gctucker (approved):

JenySadadia:
@gctucker We need to merge changes from this PR to 'api-pipeline' branch to fix https://github.com/kernelci/kernelci-pipeline/issues/7 issue.
The reason is we are using that branch for kernelci-pipeline.
```
# Install kernelci-core from api-pipeline topic branch
RUN pip install git+https://github.com/kernelci/kernelci-core.git@api-pipeline
```

gctucker:
Well we need to stop using the `api-pipeline` branch now as I believe all the required changes have been merged in `main`.

PR #980, opened 2022-01-11
test-configs.yaml: Enable kselftest-rtc test plan on rk3399-gru-kevin

nfraprado:
The rk3399-gru-kevin device uses the rtc-cros-ec driver for RTC. This PR adds the kselftest-rtc test on kevin to exercise this driver. The config needed to enable the driver is already part of the arm64 defconfig.

LAVA run: https://lava.collabora.co.uk/scheduler/job/5409783

As can be seen from the LAVA output, the `alarm_alm_set_minute` test, which sets an alarm, is failing with timeout. From a first look it seems like a bug in the `rtc-cros-ec` driver.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -2411,6 +2411,7 @@ test_configs:
       - baseline
       - baseline-nfs
       - cros-ec
+      - kselftest-rtc
       - ltp-fcntl-locktests
       - ltp-pty
       - ltp-timers


gctucker:
@nfraprado Could you please take a look at the `sleep` test results and compare?  They use `rtcwake`.

nfraprado:
Since the whole rtc selftest was timing out, I increased its timeout setting to let all tests run, and with this I was able to see that even though the `alarm_alm_set_minute` test is failing with timeout, `alarm_wkalm_set_minute` is working fine. This means that the alarm functionality is working at least with the `wkalm` variant. So the failure on the `alarm_alm_set_minute` is most likely a driver bug, so shouldn't hold back this test from being added.

Lava log with increased timeout: https://lava.collabora.co.uk/scheduler/job/5415176

A patch has been sent upstream changing the timeout setting to allow all tests to run: https://lore.kernel.org/all/20220112165359.3206047-1-nfraprado@collabora.com/

nfraprado:
@gctucker sure, I'll do that next. I'll also double-check why this timeout increase is needed, since the response on the ML is that it shouldn't be.

gctucker:
OK thanks.  It would be good to also have a GitHub issue in kernelci-project to track this on the [kernel bugs](https://github.com/orgs/kernelci/projects/6) board.

nfraprado:
Opened the kernel bug issue at https://github.com/kernelci/kernelci-project/issues/86. I don't have permission to add it to the project/board though.

nfraprado:
Regarding the comparison between the `sleep` test and this test:

Both tests use the RTC, but their purpose is different. The `sleep` test uses RTC alarms and then suspends the device, so that the alarm will wake it up. It does this for different sleep levels. On the other hand, the rtc kselftest (added here) tries a few different functions of the RTC, including setting an alarm.

So what both have in common is working with RTC alarms through use of the `RTC_WKALM_RD` and `RTC_WKALM_SET` IOCTLs. Both of these IOCTLs seem to be working fine on this device and driver, since the `alarm_wkalm_set_minute` test passes (and at least some of the sleep test cases pass).

The failing test in the rtc kselftest, `alarm_alm_set_minute`, uses the `RTC_ALM_SET` IOCTL, which has slightly different semantics, and is not used by `rtcwake` in the `sleep` test.

nfraprado:
@gctucker I'd say this is ready to be tested on staging and get merged.

The failing test case, `alarm_alm_set_minute`, really looks like a bug in the rtc driver and is being tracked at https://github.com/kernelci/kernelci-project/issues/86.

The fix for the selftest timeout bug has been sent upstream and accepted by the maintainer: https://lore.kernel.org/all/f286a86d-6371-d8bd-6ca9-5f622468d1a4@linuxfoundation.org/. When the fix reaches mainline it will allow the last test case, `alarm_wkalm_set_minute`, to run (and pass).

gctucker (approved):

PR #986, opened 2022-01-14
kernelci.build: don't convert publish path to lower-case

gctucker:
Stop converting all the elements in the publish path to lower case as
the config string with e.g. +CONFIG_SOMETHING=y needs to remain in
capital letters.  This was most likely introduced by mistake as
there's no apparent reason for using lower-case, even with Chrome OS
config names which the breaking change was originally required for.

Fixes: 2eccce7ab993 ("kernelci.build: replace more characters in publish path")


gctucker:
Ah, staging was jammed so there's no result since this PR was created.  It's back on track now, will check again tomorrow.

gctucker:
Tested OK on staging:
https://storage.staging.kernelci.org/next/master/next-20220115/arm64/?C=M&O=D
Now the capital case is preserved, for example: `defconfig+CONFIG_ARM64_16K_PAGES=y`

PR #952, opened 2021-12-17
Enable rtc kselftest on mt8173-elm-hana

nfraprado:
This adds a new kselftest test plan for the rtc collection, and enables it to run on mt8173-elm-hana.

The test output log can be seen here: https://lava.collabora.co.uk/scheduler/job/5227298

As can be seen, 2 out of the 7 tests fail. Namely, `rtc.alarm_alm_set` and `rtc.alarm_alm_set_minute`. The reason behind this is that the MT6397 RTC driver doesn't implement the `alarm_irq_enable()` operation, which is required for the `RTC_AIE_ON/OFF` IOCTLs used in those tests.

That said, that doesn't seem to be a problem since the `wkalm` variants of those two tests pass, which use the `RTC_WKALM_*` IOCTLs instead, which are preferred over the `ALM` ones anyway. (According to `Documentation/ABI/testing/rtc-cdev` and `drivers/rtc/dev.c`)

To run this test, `CONFIG_RTC_DRV_MT6397` needs to be enabled (either as module or built-in). [A patch has been sent to the mailing list adding it to the defconfig](https://lore.kernel.org/all/20211217162445.876034-1-nfraprado@collabora.com/).

PR #989, opened 2022-01-18
Add kselftest-rtc on two x86 chromebooks

nfraprado:
The x86 chromebook devices chosen to run the test were hp-11A-G6-EE-grunt and asus-C436FA-Flip-hatch, due to having a lot of them available at the LAVA lab, and them being from different platforms (Stoney Ridge/AMD vs Cometlake-U/Intel).

Lava log for grunt: https://lava.collabora.co.uk/scheduler/job/5453615
Lava log for hatch: https://lava.collabora.co.uk/scheduler/job/5453691

Both devices use the same RTC driver, so their logs and results are similar. The intent in adding both devices is to have some redundancy and achieving the goal of having two devices for each architecture running each kselftest.

All 7 tests pass, although given the rtc kselftest timeout bug, the whole rtc kselftest suite might fail sometimes due to the timeout. That bug should be fixed in the next kernel release (5.17): https://lore.kernel.org/all/20220112194142.3276385-1-nfraprado@collabora.com/. This might cause KernelCI to detect it as a regression, in which case might make more sense to merge this PR after 5.17.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -1910,6 +1910,7 @@ test_configs:
       - kselftest-lib
       - kselftest-livepatch
       - kselftest-lkdtm
+      - kselftest-rtc
       - kselftest-seccomp
       - ltp-crypto
       - ltp-fcntl-locktests
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -1763,6 +1763,7 @@ test_configs:
       - baseline-nfs
       - kselftest-filesystems
       - kselftest-futex
+      - kselftest-rtc
       - ltp-ipc
       - ltp-mm
       - ltp-timers


gctucker:
> All 7 tests pass, although given the rtc kselftest timeout bug, the whole rtc kselftest suite might fail sometimes due to the timeout. That bug should be fixed in the next kernel release (5.17): https://lore.kernel.org/all/20220112194142.3276385-1-nfraprado@collabora.com/. This might cause KernelCI to detect it as a regression, in which case might make more sense to merge this PR after 5.17.

Tests first need to pass in KernelCI before they can be detected as a regression.  Adding tests that fail will just be reported as failure.

config/core/test-configs.yaml
```diff
@@ -1763,6 +1763,7 @@ test_configs:
       - baseline-nfs
       - kselftest-filesystems
       - kselftest-futex
+      - kselftest-rtc
```
gctucker, line 1766: OK: https://staging.kernelci.org/test/plan/id/61e84bce6e7616b16d217804/
Interestingly, there are failures with the Clang build:
https://staging.kernelci.org/test/plan/id/61e848fa5d62af4ad321782f/

Maybe they are just intermittent issues, more tests would probably be needed to determine whether it's related to the compiler difference.

config/core/test-configs.yaml
```diff
@@ -1910,6 +1911,7 @@ test_configs:
       - kselftest-lib
       - kselftest-livepatch
       - kselftest-lkdtm
+      - kselftest-rtc
```
gctucker, line 1914: OK with one failure: https://staging.kernelci.org/test/plan/id/61e84d3088d83e941f2178dd/

and 2 failures with Clang:
https://staging.kernelci.org/test/plan/id/61e849bacf88d41e5b217807/

gctucker (approved):
LGTM, and it's nice to have the context explained in the git commits.

The fact that it's the same RTC hardware but failures are different tends to show that there are intermittent problems, unless they're due to how the hardware is set up.  At least it's the same kernel driver so that's providing interesting comparisons.

config/core/test-configs.yaml
nfraprado: Hmm the Clang log looks like it never even reached the login prompt, let alone run the test. Seems like an infrastructure issue (on either KernelCI or LAVA).

config/core/test-configs.yaml
nfraprado: Those are the same failure. Must be a timing issue whether LAVA decides to report both (the test case it was running as well as the test suite that contains it) or just one as failed.

nfraprado:
@gctucker Yeah, so, to clarify, the timeout issue I mentioned is intermittent. Basically depending on the current time when the test starts, it might be able to run through completion or not, due to the timeout expiring (if the time has a high value in the second, like above 30, when the `alarm_alm_set_minute` test case starts, it will be able to complete in time).

So until kernel 5.17 lands, it might fail sometimes. After 5.17 (5.17-rc2 most likely, actually), this intermittent issue will be fixed, and whichever tests fail are truly failing.

That's why I mentioned that it might cause KernelCI to accuse a regression. Come to think of it, the same is true for the test running in kevin, merged in #980. However on that device there's a test that is truly failing, so until 5.17, it might sometimes fail 1 test, and sometimes fail 2, which might cause trouble for KernelCI's regression detection (not sure)...

The test running on hana, #952, runs faster due to its driver's handling, so this intermittent issue is not present there.

TL;DR: We might need to revert #980 and hold this PR until 5.17 is released to avoid triggering regressions on KernelCI.

gctucker:
How about linux-next then?  Does it not already have the changes that will land in 5.17?  If there are any pending patches, we could apply them for staging jobs in the meantime.

nfraprado:
As per https://lore.kernel.org/all/f286a86d-6371-d8bd-6ca9-5f622468d1a4@linuxfoundation.org/ , should be applied to Shuah's subtree after merge window closes, which should be next week. I'll track it and and notify here when we get a tree that could run with the fix applied.

gctucker:
Thanks for the context. I think it's OK to merge it, having sporadic regressions in KernelCI is normal and it's not a problem with this PR. If a test failure is intermittent then the bisection job will detect that and not land on a wrong commit.

gctucker:
We should probably still add these pending patches to the staging jobs, to check all the tests are passing with them.

gctucker:
OK let's just wait for this to land in linux-next next week.

nfraprado:
The fix is now in the linux-kselftest subtree: https://git.kernel.org/pub/scm/linux/kernel/git/shuah/linux-kselftest.git/log/?h=fixes

It should reach mainline either next or the following week, for 5.17-rc2 or rc3, at which point it should automatically be tested by KernelCI I believe.

gctucker:
Yeah, we probably should consider testing linux-kselftest as well.  Thanks for the update.

PR #990, opened 2022-01-19
build-configs-android: Remove gki_config

musamaanjum:
All the *_defconfig are built by default. This is why even if gki_config
was a typo (instead of correct gki_defconfig) gki config is being built.
Remove it and simplify the file.
    
Fixes: af4646952a07 ("build-configs-android: enable clang-14 android builds")

### DIFF
--- a/config/core/build-configs-android.yaml
+++ b/config/core/build-configs-android.yaml
@@ -7,7 +7,7 @@ trees:
 android_variants: &android_variants
   gcc-10:
     build_environment: gcc-10
-    architectures:
+    architectures: &architectures
         arm: &arm_arch
           base_defconfig: 'multi_v7_defconfig'
           extra_configs:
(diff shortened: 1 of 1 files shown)


musamaanjum:
@gctucker it seems like gki wasn't enabled with the correct name and we couldn't find any error that it was wrong. Not sure how gki_defconfig config was used in a build on [staging](https://staging.kernelci.org/build/id/61dca65efed2653d3a63556c/).

gctucker:
I think it was built anyway because by default all the `*_defconfig` files found in the kernel source tree get built.  We're not defining a filter with a pass list to reduce that (yet), so adding `gki_defconfig` in the `extra_configs` is actually redundant at the moment.  All the other ones there are special ones are `make` targets rather than config files, or additional `+CONFIG_*` variants.

So I guess we may just drop `gki_config` in this PR, and we'll add it again if we narrow down the number of configs built for Android kernel branches.

musamaanjum:
I'm unable to find any gki builds on staging now. @gctucker Any idea about it?

gctucker:
OK this looks like a kernel issue based on https://github.com/ClangBuiltLinux/linux/issues/1497 so this PR is ready to be merged now.  We'll see when the issue gets fixed with production builds.

gctucker (approved):
Tested OK on staging.

PR #997, opened 2022-01-25
Update documentation links

nuclearcat:
Links to wiki deprecated, updated to new documentation, and some git links updated.


README.md
```diff
@@ -84,18 +84,18 @@ this works.
 ## YAML Configuration files
```
gctucker, line 84: I think this section could have been replaced with a link to https://kernelci.org/docs/core/config/

In fact, maybe this README should be entirely rewritten in a much smaller version now that we have more documentation available.

gctucker (approved):
Thanks for spotting these issues and fixing them.

PR #999, opened 2022-01-26
test-configs.yaml: use latest bullseye-ltp for QEMU

gctucker:
Update the URL of the newly defined bullseye-ltp ramdisk which is used
on QEMU.


gctucker (commented):
Tested OK on staging
https://staging.kernelci.org/test/job/kernelci/branch/staging-mainline/kernel/staging-mainline-20220128.0/

PR #1000, opened 2022-01-26
Fix how templates_path is used in kernelci.lab

gctucker:
Fix a couple of places where the `templates_path` was not being used correctly.


gctucker (commented):
Tested OK on staging and with kernelci-pipeline.

PR #1001, opened 2022-01-26
kernelci.config.base: fix default filters

gctucker:
Fix default filters when they're not defined in YAML using an empty
dictionary.

Fixes: 6374126064e3 ("kernelci.config: group default filters under default_filters")


gctucker (commented):
Tested OK on staging.

PR #1004, opened 2022-01-27
test-configs.yaml: Add CIP device x86-simatic-ipc227e

aliceinwire:

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -1672,6 +1672,19 @@ device_types:
     arch: i386
     boot_method: grub
 
+  x86-simatic-ipc227e:
+    mach: x86
+    arch: i386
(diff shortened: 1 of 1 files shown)


gctucker:
This didn't work with GRUB:
https://lava.ciplatform.org/scheduler/job/612881

Looking at the health check jobs, it seems to only support IPXE:
https://lava.ciplatform.org/scheduler/job/612849/definition#defline40

So either support for GRUB boot method should be added to the device type, or a KernelCI job template should be created to use IPXE.

Other test jobs are using IPXE so it looks like this would be the way forward:
https://lava.ciplatform.org/scheduler/job/612796/definition

gctucker:
Closing for now, please re-open when the platform is known to be working.

PR #1011, opened 2022-01-28
kernelci/data/kernelci_api: add KernelCI_API.get_nodes_by_commit_hash 

JenySadadia:
Add method to request kernelci-api 'nodes' endpoint
to get a list of node objects matching with the commit SHA-1.

### DIFF
--- a/kernelci/data/kernelci_api.py
+++ b/kernelci/data/kernelci_api.py
@@ -72,6 +72,14 @@ class KernelCI_API(Database):
         resp = self._get('/'.join(['node', node_id]))
         return json.loads(resp.text)
 
+    def get_nodes_by_commit_hash(self, commit_hash):
+        """
+        Get list of node objects matching the commit SHA-1
+        provided to 'nodes' endpoint
+        """
+        resp = self._get('?'.join(['nodes', 'revision.commit='+commit_hash]))
+        return resp.json()
+
     def get_node_from_event(self, event):
         return self.get_node(event.data['id'])
 


kernelci/data/kernelci_api.py
```diff
@@ -94,3 +94,12 @@ def submit(self, data, verbose=False):
 def get_db(config, token):
     """Get a KernelCI API database object"""
     return KernelCI_API(config, token)
+
+
+def get_node_by_commit_id(db_url, commit_id):
```
gctucker, line 99: Please make this a method of the `KernelCI_API` class.

kernelci/data/kernelci_api.py
```diff
@@ -72,6 +72,14 @@ def get_node(self, node_id):
         resp = self._get('/'.join(['node', node_id]))
         return json.loads(resp.text)
 
+    def get_node_by_commit_id(self, commit_id):
+        """
+        Get node object by sending commit ID to 'nodes' endpoint
+        """
+        get_node_path = '?'.join(['nodes', 'revision.commit='+commit_id])
```
gctucker, line 79: This could be using `self._get()` like the `get_node()` method

kernelci/data/kernelci_api.py
```diff
@@ -72,6 +72,14 @@ def get_node(self, node_id):
         resp = self._get('/'.join(['node', node_id]))
         return json.loads(resp.text)
 
+    def get_node_by_commit_id(self, commit_id):
+        """
+        Get node object by sending commit ID to 'nodes' endpoint
```
gctucker, line 77: It's actually the git commit SHA1, using "id" might be a bit ambiguous here as it might be confused with a database id.

JenySadadia: Renamed the function.

kernelci/data/kernelci_api.py
```diff
@@ -72,6 +72,13 @@ def get_node(self, node_id):
         resp = self._get('/'.join(['node', node_id]))
         return json.loads(resp.text)
 
+    def get_node_by_commit_hash(self, commit_hash):
```
gctucker, line 75: This will actually return a list of nodes, with all the ones matching the git SHA1. So we probably should call it `get_nodes_by_commit_hash` and mention it in the docstring.

gctucker (approved):
LGTM

PR #1012, opened 2022-01-31
test-configs.yaml: update rootfs URLs

mgalka:
Update rootfs URLs to 20220128.0. Buildroot and bullseye-ltp rootfs
images failed to build and are not updated.


gctucker:
Merging now to verify they're all OK with the pipe cleaner job in production.

PR #1015, opened 2022-02-01
Run QEMU LAVA tests only in Docker containers

### LINKED ISSUE #1014: Only run QEMU jobs in Docker
As shown by https://github.com/kernelci/kernelci-core/pull/920 and other tests in the past, having disparities between labs in terms of QEMU versions means we're not comparing like-to-like. Labs running older versions show failures that aren't related to kernel issues but lack of some feature support in QEMU. To avoid this issue, all the KernelCI QEMU jobs should require to be run in the Docker image provided by KernelCI. Most if not all the online LAVA labs are recent enough to support this, and if some don't then they should have QEMU jobs blocked from running on them to avoid false positives.

gctucker:
Convert configuration and LAVA job templates to always use Docker when running QEMU tests.

Fixes #1014

### DIFF
--- a/config/core/lab-configs.yaml
+++ b/config/core/lab-configs.yaml
@@ -85,7 +85,6 @@ labs:
     filters: &collabora-filters
       - blocklist:
           tree: [android]
-          plan: [baseline-qemu-docker]
 
   lab-collabora-staging:
     lab_type: lava.lava_rest
@@ -133,7 +132,6 @@ labs:
       - passlist:
           plan:
             - baseline
-      - blocklist: {plan: [baseline-qemu-docker]}
 
   lab-pengutronix:
     lab_type: lava.lava_rest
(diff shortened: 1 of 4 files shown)


gctucker:
@patersonc The CIP lab is also failing to use Docker:
https://lava.ciplatform.org/scheduler/job/627469#L162

Leaving out of the passlist filter for now.

gctucker:
Otherwise tested OK on staging: 
https://staging.kernelci.org/test/job/kernelci/branch/staging-cip/kernel/staging-cip-20220210.0/plan/baseline/

gctucker:
OK merging this now. In principle, `lab-linaro-lkft` will get Docker support fixed soon and `lab-collabora` should be upgraded to a working version in March. Until then, we can already make the move with the labs that already support it.

PR #1016, opened 2022-02-01
test-config: distribute core kernel kselftests

musamaanjum:
Distribute core kernel kselftests on all available devices per
architecture. One kselftest should run on at least two different
devices of same architecture when there is no capacity issue. Add or
remove kselftests to different devices to make this normalization.
These changes have been done after doing analysis that how many devices
devices are available of certain architecture.

---
The latest proposal, "Proposal - 2" can be seen in the [spreadsheet](https://docs.google.com/spreadsheets/d/1ffrb_zP1WqKNaLx3W6eOdp39CbXzvpUKBwiWtgtxfwM/edit#gid=0).

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -2025,6 +2025,12 @@ test_configs:
   - device_type: imx6q-sabrelite
     test_plans:
       - baseline
+      - kselftest-filesystems
+      - kselftest-futex
+      - kselftest-lib
+      - kselftest-livepatch
+      - kselftest-lkdtm
+      - kselftest-seccomp
 
   - device_type: imx6q-sabrelite
     test_plans:
@@ -2146,6 +2152,9 @@ test_configs:
     test_plans:
       - baseline
       - baseline-nfs
+      - kselftest-filesystems
+      - kselftest-futex
+      - kselftest-lib
 
   - device_type: meson-g12b-odroid-n2
     test_plans:
@@ -2227,6 +2236,7 @@ test_configs:
       - kselftest-filesystems
       - kselftest-futex
       - kselftest-lib
+      - kselftest-livepatch
       - kselftest-lkdtm
       - kselftest-rtc
       - kselftest-seccomp
@@ -2245,6 +2255,8 @@ test_configs:
     test_plans:
       - baseline
       - baseline-nfs
+      - kselftest-lkdtm
+      - kselftest-seccomp
 
   - device_type: mustang
     test_plans:
@@ -2443,7 +2455,6 @@ test_configs:
       - cros-ec
       - igt-gpu-panfrost
       - igt-kms-rockchip
-      - kselftest-lib
       - ltp-timers
       - sleep
       - usb


config/core/test-configs.yaml
```diff
@@ -2016,6 +2018,12 @@ test_configs:
   - device_type: imx6q-sabrelite
     test_plans:
       - baseline
+      - kselftest-filesystems
+      - kselftest-futex
+      - kselftest-lib
+      - kselftest-livepatch
+      - kselftest-lkdtm
+      - kselftest-seccomp
```
gctucker, line 2026: I'm not sure there are enough imx6q-sabrelite available to cover all this.  See this earlier commit: 3578f05e6cb9be3a8de9b25bb7e8578a227111a7

```
commit 3578f05e6cb9be3a8de9b25bb7e8578a227111a7
Author: Guillaume Tucker <guillaume.tucker@collabora.com>
Date:   Wed May 26 16:35:21 2021 +0100

    test-configs.yaml: only run baseline on imx6q-sabrelite
    
    There aren't enough imx6q-sabrelite boards available in the Collabora
    lab to run all the test plans that were assigned to it, so reduce to
    only baseline until this is resolved with more hardware.  Enable
    baseline-nfs to keep running only in lab-baylibre in the meantime.
    
```

So it's probably better to only keep running baseline on it.

config/core/test-configs.yaml
```diff
@@ -1777,8 +1777,6 @@ test_configs:
     test_plans:
       - baseline
       - baseline-nfs
-      - kselftest-lkdtm
-      - kselftest-seccomp
```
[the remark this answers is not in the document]
gctucker: I would rather add up to 3 extra kselftest plans to rammus and hatch before adding any more to coral.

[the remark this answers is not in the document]
musamaanjum: Got it. I'll distribute these tests on rammus, hatch and coral.

config/core/test-configs.yaml
musamaanjum: I've found 9 imx6q-sabrelite in Collabora's Lava: https://lava.collabora.co.uk/scheduler/alldevices/active?search=&length=100&page=1#table Maybe there have been added recently?

musamaanjum:
I'm working on a new proposal which can be found [here](https://docs.google.com/spreadsheets/d/1ffrb_zP1WqKNaLx3W6eOdp39CbXzvpUKBwiWtgtxfwM/edit#gid=0).

config/core/test-configs.yaml
gctucker: Ah yes, good point. Let's check the results and the queue size with the staging jobs to get an idea.

gctucker:
Staging jobs had stopped today, they're running again now so let's check the results tomorrow.

config/core/test-configs.yaml
```diff
@@ -2099,6 +2102,7 @@ test_configs:
     test_plans:
       - baseline
       - baseline-nfs
+      - kselftest-seccomp
```
gctucker, line 2105: The `jetson-tk1` is very old and there's only one left online, so I would rather not add anything to it. So it's probably best to drop this.

config/core/test-configs.yaml
```diff
@@ -2146,6 +2150,9 @@ test_configs:
     test_plans:
       - baseline
       - baseline-nfs
+      - kselftest-filesystems
+      - kselftest-futex
+      - kselftest-lib
```
gctucker, line 2155: There are 7 vim3 in Collabora's lab and I believe a few more in some other labs, so this looks fine. We might even be able to enable another 2 or 3 kselftest plans on this platform, but let's see. Other projects are using them too (Mesa CI).

config/core/test-configs.yaml
```diff
@@ -2227,6 +2234,7 @@ test_configs:
       - kselftest-filesystems
       - kselftest-futex
       - kselftest-lib
+      - kselftest-livepatch
```
gctucker, line 2237: This should be fine, there are 10 `hana` Chromebooks online in Collabora's lab and they're not used by any other project. We could even try to run every possible test plan on them, until we start reaching a capacity limit.

config/core/test-configs.yaml
```diff
@@ -2245,6 +2253,8 @@ test_configs:
     test_plans:
       - baseline
       - baseline-nfs
+      - kselftest-lkdtm
+      - kselftest-seccomp
```
gctucker, line 2257: Looks good to me, this is a new device type with 5 instances online so running several kselftest plans should be fine.

config/core/test-configs.yaml
```diff
@@ -2276,6 +2286,7 @@ test_configs:
   - device_type: panda
     test_plans:
       - baseline
+      - kselftest-lkdtm
```
gctucker, line 2289: I think the `panda` board is not really well supported any more, there's only one in Collabora's lab and it's offline. So it's probably best to drop this.

config/core/test-configs.yaml
```diff
@@ -2432,6 +2443,8 @@ test_configs:
     test_plans:
       - baseline
       - baseline-nfs
+      - kselftest-futex
+      - kselftest-livepatch
```
gctucker, line 2447: We don't have that many `rock2` board online, so I'm not quite sure about this. Only one online in Collabora's lab at the moment. It's probably best to drop this now or we'll just be filling the queue with jobs that can't be run.

config/core/test-configs.yaml
```diff
@@ -2443,6 +2456,7 @@ test_configs:
       - cros-ec
       - igt-gpu-panfrost
       - igt-kms-rockchip
+      - kselftest-filesystems
```
gctucker, line 2459: I don't think it's a good idea to add tests to `veyron-jaq` as it's an old platform with only 2 online. Maybe just adding this one will be fine, we could do that and see if they can deal with the load. But even if this works, I wouldn't be surprised if we were actually pretty close to the capacity limit.

gctucker (commented):
Looks good overall, with probably a few things to tweak to avoid hitting some capacity limits.

The staging jobs can give us an idea of what production will look like, but it's hard to really tell. So it's best to add things incrementally and check after a production update how much load is being added to each device type. Eventually we should have a more dynamic orchestration mechanism which would be querying the labs for their capacity in real-time for each device type, but in the meantime we just have to make baby steps.

config/core/test-configs.yaml
```diff
@@ -1809,11 +1809,7 @@ test_configs:
       - baseline-nfs
       - cros-ec
       - igt-gpu-i915
-      - kselftest-filesystems
-      - kselftest-futex
       - kselftest-lib
-      - kselftest-lkdtm
-      - kselftest-seccomp
```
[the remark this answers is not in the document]
musamaanjum: We have a lot of x86 devices: 
10 [hp-11A-G6-EE-grunt](https://lava.collabora.co.uk/scheduler/device_type/hp-11A-G6-EE-grunt): filesystems, futex, lib, livepatch, lkdtm, seccomp
5 [asus-C433TA-AJ0005-rammus](https://lava.collabora.co.uk/scheduler/device_type/asus-C433TA-AJ0005-rammus): lkdtm, seccomp
5 [asus-C436FA-Flip-hatch](https://lava.collabora.co.uk/scheduler/device_type/asus-C436FA-Flip-hatch): filesystems, futex
5 [asus-C523NA-A20057-coral](https://lava.collabora.co.uk/scheduler/device_type/asus-C523NA-A20057-coral): lib and lkdtm
As discussed, we want to keep running one test on 2 devices at max of one architecture. So I'm keeping all 6 selftests enabled on grunt and distributing the remaining tests on the remaining three.

We can add more tests later to coral instead of running same tests on more than 2 devies. What do you think?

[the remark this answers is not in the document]
gctucker: I think it's more about running tests on 2 devices at least, not max. Some tests will provide different results on different platforms so it's also useful to cover them, such as different Intel CPU families. Having at least 2 just means we can see if the tests fail at the same time on both or not and to provide redundancy.  It depends on the kinds of tests, some are more likely to show differences than others.

Things like `lib` and `lkdtm` are probably unlikely to be hardware-specific, whereas `futex` and `seccomp` are more likely to differ at least across CPU architectures but also potentially within chip families. This has to be balanced with running more kselftests and other test suites. I think we'll have to keep adjusting things incrementally, probably on a weekly basis while keeping an eye on the load.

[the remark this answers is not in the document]
musamaanjum: Thank you for your explanation. I've updated the PR by removing changes for this board. We can keep iterating and adding more tests.

config/core/test-configs.yaml
```diff
@@ -2025,6 +2025,13 @@ test_configs:
   - device_type: imx6q-sabrelite
     test_plans:
       - baseline
+      - kselftest-filesystems
```
gctucker, line 2028: Tested OK:
https://staging.kernelci.org/test/plan/id/623377f7aac9882c9e4b3165/

config/core/test-configs.yaml
```diff
@@ -2025,6 +2025,13 @@ test_configs:
...
+      - kselftest-futex
```
gctucker, line 2029: Failed: https://staging.kernelci.org/test/plan/id/62337904137c97d1e14b3155/

config/core/test-configs.yaml
```diff
@@ -2025,6 +2025,13 @@ test_configs:
...
+      - kselftest-lib
```
gctucker, line 2030: Tested OK: https://staging.kernelci.org/test/plan/id/623377cbedc070ee604b314a/
Many skip results are showing up but that's an artifact of how the skiplist works.

config/core/test-configs.yaml
```diff
@@ -2025,6 +2025,13 @@ test_configs:
...
+      - kselftest-livepatch
```
gctucker, line 2031: Tested OK:
https://staging.kernelci.org/test/plan/id/623378250a513edc394b312e/
However, most of the livepatch test cases were skipped.

config/core/test-configs.yaml
```diff
@@ -2025,6 +2025,13 @@ test_configs:
...
+      - kselftest-lkdtm
```
gctucker, line 2032: Failed: https://staging.kernelci.org/test/plan/id/623379d4df4661d6d44b31a0/

config/core/test-configs.yaml
```diff
@@ -2025,6 +2025,13 @@ test_configs:
...
+      - kselftest-rtc
```
gctucker, line 2033: The test job was run but all rtc test cases failed:
https://staging.kernelci.org/test/plan/id/623378e36d56c7a3c34b3130/

gctucker (commented):
Checked the results on sabrelite in combination with #1080. Some tests are OK, some are failing and some are skipped. It would be good to understand the failures and determine whether they're expected due to some kernel issues or if it's because of some KernelCI or infrastructure problems.

The skiplist might need to be tweaked too in order to run more tests, in test-definitions.

gctucker:
It's proving difficult to merge this in one go, so I think we should split it into smaller chunks. We could start with the sabrelite.

config/core/test-configs.yaml
musamaanjum: Timeout occurred in this case. Maybe some testcase failed. In the last couple of weeks, this test hasn't run on this target again. So we aren't sure if this is a problem or not.

config/core/test-configs.yaml
musamaanjum: Logs on this link aren't found. In the last couple of weeks, this test hasn't run on this target again. So we aren't sure if this is a problem or not.

config/core/test-configs.yaml
musamaanjum: Remove rtc from here. It was mistakenly added. This PR is not related to platform tests.

gctucker:
@musamaanjum OK I'll go through all this one more time and probably split things out into different PRs so we can already merge what's working and decide what to do with the things that aren't.

config/core/test-configs.yaml
gctucker: It looks like a genuine kernel problem: https://lava.collabora.co.uk/scheduler/job/6019351
So I think it's OK to enable it, and the problem should be fixed in the kernel.

config/core/test-configs.yaml
```diff
@@ -2025,6 +2025,13 @@ test_configs:
...
+      - kselftest-seccomp
```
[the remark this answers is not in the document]
gctucker: OK: https://staging.kernelci.org/test/plan/id/624b3191b1869a732205f4d1/

gctucker (approved):
Tested OK with valid kernel failures to be reported.  This worked in combination with #1080

musamaanjum:
@gctucker thank you so much for helping out.

PR #1017, opened 2022-02-02
rootfs-configs.yaml: drop buster-v4l2

gctucker:
Drop the buster-v4l2 rootfs config as bullseye-v4l2 is now available.
Also drop the related buster-v4l2.sh build script.


Lakshmipathi:
LGTM

PR #1018, opened 2022-02-02
Add ChromiumOS image building functionality as rootfs option

nuclearcat:
With just:

docker run -itd -v $(pwd):/kernelci-core --device /dev/kvm -v /dev:/dev --privileged kernelci/chromiumos
./kci_rootfs build --rootfs-config chromiumos --data-path config/rootfs/chromiumos --arch amd64-generic
(And several hours of build time)
I can get qemu bootable image:
config/rootfs/chromiumos/amd64-generic/chromiumos_test_image.bin
(Must be used with USB storage disk, if we use libvirt)

As ChromiumOS dont have strictly speaking "architecture", in our case it means ChromiumOS board name.
So it can be amd64-generic, hatch, etc. 

### DIFF
--- a/config/core/rootfs-configs.yaml
+++ b/config/core/rootfs-configs.yaml
@@ -255,6 +255,13 @@ rootfs_configs:
       - wget
       - xz-utils
 
+  chromiumos-amd64-generic:
+    rootfs_type: chromiumos
+    board: amd64-generic
+    branch: release-R99-14469.B
+    arch_list:
+      - amd64
+
   sid:
     rootfs_type: debos
     debian_release: sid
--- a/config/rootfs/chromiumos/scripts/build_board.sh
+++ b/config/rootfs/chromiumos/scripts/build_board.sh
@@ -1,0 +1,41 @@
+#!/bin/sh
+set -e
+BOARD=$1
+BRANCH=$2
+DATA_DIR=$(pwd)
+
+# Building SDK itself
+if [ ! -d chromiumos-sdk ]; then
+  sudo mkdir chromiumos-sdk
+  sudo chown user chromiumos-sdk
+  cd chromiumos-sdk
+  git config --global user.email "bot@kernelci.org"
+  git config --global user.name "KernelCI Bot"
+  git config --global color.ui false
+  repo init --repo-url https://chromium.googlesource.com/external/repo --manifest-url https://chromium.googlesource.com/chromiumos/manifest --manifest-name default.xml --manifest-branch ${BRANCH} --depth=1
+  repo sync -j$(nproc)
+  cros_sdk --create
+else
+  cd chromiumos-sdk
+  # We might have different branch
+  repo init --repo-url https://chromium.googlesource.com/external/repo --manifest-url https://chromium.googlesource.com/chromiumos/manifest --manifest-name default.xml --manifest-branch ${BRANCH} --depth=1
+  repo sync -j$(nproc)
+  cros_sdk --create
+fi
+
+# Compiling ChromiumOS image
+# Future possible option --profile=x, for example kernel-5_15, profiles are at /mnt/host/source/src/overlays/overlay-${BOARD}/profiles/
+cros_sdk setup_board --board=${BOARD} --force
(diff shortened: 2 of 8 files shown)


config/rootfs/chromiumos/scripts/build_board.sh
```diff
@@ -0,0 +1,36 @@
...
+
+# Building SDK itself
+if [ ! -d chromiumos-sdk ]; then
+  sudo mkdir chromiumos-sdk
+  sudo chown user chromiumos-sdk
+  cd chromiumos-sdk
+  git config --global user.email "bot@kernelci.org"
+  git config --global user.name "KernelCI Bot"
+  git config --global color.ui false
+  repo init --repo-url https://chromium.googlesource.com/external/repo --manifest-url https://chromium.googlesource.com/chromiumos/manifest --manifest-name default.xml --manifest-branch ${BRANCH}
```
gctucker, line 15: How about `--depth=1` for shallow clones? Do we need full clones?

config/rootfs/chromiumos/scripts/build_board.sh
```diff
@@ -0,0 +1,36 @@
...
+  repo sync -j32
```
gctucker, line 16: Maybe `-j$(nproc)` to scale depending on the host

gctucker (commented):
Looks like a great start! We should also add this Docker image to the script that builds and pushes the images to the docker hub in `config/docker/build-and-push.sh`.

config/rootfs/chromiumos/scripts/build_board.sh
nuclearcat: Done!

config/rootfs/chromiumos/scripts/build_board.sh
nuclearcat: Done! (and tested)

config/core/rootfs-configs.yaml
```diff
@@ -220,6 +220,14 @@ rootfs_configs:
     script: "scripts/buster-v4l2.sh"
     test_overlay: "overlays/v4l2"
 
+  chromiumos:
```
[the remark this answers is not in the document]
nuclearcat: I think we can add for now just single entry arch_list and in future we might have to define boards better way,if we will compile more than just one generic qemu image.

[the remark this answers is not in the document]
gctucker: OK sounds good

kci_rootfs
```diff
@@ -103,7 +103,10 @@ class cmd_build(Command):
         data_path = (
             args.data_path or 'config/rootfs/{}'.format(config.rootfs_type)
         )
-        return kernelci.rootfs.build(config_name, config, data_path, args.arch)
+        if (config.rootfs_type == "chromiumos" and args.board is None):
+            # Set default board
+            args.board = "amd64-generic"
+        return kernelci.rootfs.build(config_name, config, data_path, args)
```
gctucker, line 109: How about setting an `board` attribute in the YAML configuration instead? That's what I meant by having separate config entries for each board.

gctucker: Another attribute in the YAML config is probably going to be the chromium branch name, as each board will be using a different one. For the generic ones we could keep using the latest stable release and update it when needed.

config/rootfs/chromiumos/scripts/build_board.sh
gctucker: I don't see it in the diff though

nuclearcat: I will recheck and retest that.

kci_rootfs
nuclearcat: Done as well

gctucker:
@nuclearcat I've rebased the `chromeos` branch on top of latest `main`, please rebase this PR.

gctucker (commented):
A couple of minor things about comments, so no functional changes.

gctucker (approved):

PR #1020, opened 2022-02-02
rootfs-configs.yaml: Remove buster-cros-ec

mgalka:
Remove buster-cros-ec rootfs definition, bullseye-cros-ec should be used
instead.


config/core/rootfs-configs.yaml
```diff
@@ -189,31 +189,6 @@ rootfs_configs:
...
-      - e2fslibs
-      - e2fsprogs
-      - libext2fs2
-    extra_firmware:
-      - mrvl/pcieusb8997_combo_v4.bin
-      - mrvl/sd8897_uapsta.bin
-      - qca/rampatch_00440302.bin
-      - rtl_nic/rtl8153b-2.fw
-      - rockchip/dptx.bin
-    script: "scripts/buster-cros-ec-tests.sh"
```
gctucker, line 214: This needs to be removed too

gctucker: I mean, the script mentioned here

mgalka: Removed

PR #1024, opened 2022-02-04
Update kselftest from buster to bullseye

nuclearcat:
Add bullseye-kselftest rootfs image configuration.
Only required change was updating perl and perl modules version.

Signed-off-by: Denys Fedoryshchenko <denys.f@collabora.com>

PR #1025, opened 2022-02-04
Run kselftest-tpm2 on grunt

nfraprado:
This PR gets kselftest-tpm2 running on grunt.

Lava log: https://lava.collabora.co.uk/scheduler/job/5608585

Basically the results are:
```
# selftests: tpm2: test_smoke.sh
# test_read_partial_overwrite (tpm2_tests.SmokeTest) ... ok
# test_read_partial_resp (tpm2_tests.SmokeTest) ... ok
# test_seal_with_auth (tpm2_tests.SmokeTest) ... ERROR
# test_seal_with_policy (tpm2_tests.SmokeTest) ... ERROR
# test_seal_with_too_long_auth (tpm2_tests.SmokeTest) ... FAIL
# test_send_two_cmds (tpm2_tests.SmokeTest) ... ok
# test_too_short_cmd (tpm2_tests.SmokeTest) ... ok
# test_unseal_with_wrong_auth (tpm2_tests.SmokeTest) ... ERROR
# test_unseal_with_wrong_policy (tpm2_tests.SmokeTest) ... ERROR

# test_async (tpm2_tests.AsyncTest) ... ok

# selftests: tpm2: test_space.sh
# test_flush_context (tpm2_tests.SpaceTest) ... /usr/lib/python3.7/unittest/case.py:615: ResourceWarning: unclosed file <_io.FileIO name='/dev/tpmrm0' mode='rb+' closefd=True>
#   testMethod()
# ResourceWarning: Enable tracemalloc to get the object allocation traceback
# ok
# test_get_handles (tpm2_tests.SpaceTest) ... ok
# test_invalid_cc (tpm2_tests.SpaceTest) ... ok
# test_make_two_spaces (tpm2_tests.SpaceTest) ... ok
```

The errors are caused by `tpm2.ProtocolError: TPM_RC_LOCKOUT: cc=0x00000153, rc=0x00000921`. There was [a commit](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=a9920d3bad40201ee8ab1da36bee4674f7e50d69) in the past fixing those, but [it was later reverted](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=5be206eaac9a68992fc3b06fb5dd5634e323de86). However the patch might make sense in KernelCI's usecase, where devices are there just for testing. That said, that patch requires the `tpm2_clear` tool, which is only available in bullseye.

I ran the tests using the `bullseye-kselftest` rootfs from #1024, with the addition of the `tpm2-tools` package, and that patch applied (although modified to clear before rather than after the tests are run), and the errors were gone:
https://lava.collabora.co.uk/scheduler/job/5616044. A single FAIL remained from one of the tests.

So one option would be to, after #1024 is merged, add the `tpm2-tools`  package and carry a patch in KernelCI for kselftest. Ideally the change would be merged upstream, but given that it was reverted already, a different approach would be needed for upstream, and I'm not sure what that would be.

### DIFF
--- a/config/core/rootfs-configs.yaml
+++ b/config/core/rootfs-configs.yaml
@@ -159,6 +159,8 @@ rootfs_configs:
       - perl-modules-5.32
       - procps
       - publicsuffix
+      - python3-minimal
+      - python3-unittest2
       - wget
       - xz-utils
 
--- a/config/core/rootfs-configs.yaml
+++ b/config/core/rootfs-configs.yaml
@@ -161,6 +161,7 @@ rootfs_configs:
       - publicsuffix
       - python3-minimal
       - python3-unittest2
+      - tpm2-tools
       - wget
       - xz-utils
 
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -1975,6 +1975,7 @@ test_configs:
       - kselftest-lkdtm
       - kselftest-rtc
       - kselftest-seccomp
+      - kselftest-tpm2
       - kselftest-vm
       - ltp-crypto
       - ltp-fcntl-locktests
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -293,6 +293,12 @@ test_plans:
       job_timeout: '10'
       kselftest_collections: "seccomp"
 
+  kselftest-tpm2:
+    <<: *kselftest
+    params:
+      job_timeout: '10'
+      kselftest_collections: "tpm2"
+
   kselftest-vm:
     <<: *kselftest
     params:
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -268,6 +268,7 @@ fragments:
       - 'CONFIG_TCG_TIS_I2C_CR50=y'
       - 'CONFIG_SPI=y'
       - 'CONFIG_SPI_PXA2XX=y'
+      - 'CONFIG_PINCTRL_AMD=y'
 
   x86_kvm_guest:
     path: "kernel/configs/kvm_guest.config"


config/core/rootfs-configs.yaml
```diff
@@ -219,6 +219,8 @@ rootfs_configs:
       - perl-modules-5.28
       - procps
       - publicsuffix
+      - python3-minimal
+      - python3-unittest2
```
gctucker, line 223: You'll have to re-apply these change when bullseye-kselftest has been merged, maybe today or Monday

config/core/test-configs.yaml
```diff
@@ -1939,6 +1945,7 @@ test_configs:
       - kselftest-lkdtm
       - kselftest-rtc
       - kselftest-seccomp
+      - kselftest-tpm2
```
gctucker, line 1948: How about running it on other devices? Does it require particular hardware only available on grunt?

config/core/rootfs-configs.yaml
nfraprado: Sure, no problem.

config/core/test-configs.yaml
nfraprado: Yes, I intend to. I was having a few issues getting the test running on a few other devices, and grunt was the first one I got it working on. But there should be a few other devices in KernelCI where we can run this test on. By the end of next week I should probably be able to get at least one other device running this test.

config/core/rootfs-configs.yaml
gctucker: Actually I just merged it, so please rebase anytime :)

nfraprado: Hm, now that I think about it, since this PR both changes the rootfs and adds the test, shouldn't I wait for the kselftest test definition to be updated to use the new `buster-kselftest` rootfs before I rebase? Or I could also split this PR between the rootfs and the test part if you prefer.

nfraprado: OK, the bullseye rootfs seems to be in place now, so I've rebased and force-pushed.

gctucker:
This is conflicting with #823 which is blocked on a new buildroot image to be produced.

nfraprado:
Oops, I totally missed that PR and ended up duplicating some of that work. In any case, that particular PR doesn't look blocked in any way (it's just adding configs). So I'll rebase this PR on top of that one, and then we can start by merging that one. I'll also comment on it to answer some questions and get it unblocked.

gctucker:
@nfraprado Please rebase one more time, I think this is conflicting with the `kselftest-vm` changes that have just been merged.

gctucker (commented):
Here's the resulting job, manually submitted with the build and rootfs image mentioned above:
https://lava.collabora.co.uk/scheduler/job/5772426
https://staging.kernelci.org/test/plan/id/6218b696bb88e681ca4b318a/
which still showed the same errors as the kernel build didn't have the required patch.

Then the same job resubmitted with @nfraprado's kernel binary:
https://lava.collabora.co.uk/scheduler/job/5772538
which also shows the errors.

Could you please take a look and confirm which kernel changes are required to make this pass? We could then apply them on top of linux-next on a staging build to check the changes in this PR work with that kernel fix in place.

nfraprado:
@gctucker To get rid of the ERRORs, the commit I just pushed is needed in order to have tpm2-tools in the rootfs, and also the attached patch (renamed to a .txt extension since GitHub doesn't like .patch) needs to be applied to the kernel. Note that this patch is in kselftest, so it's the kselftest archive that will be affected rather than the kernel image.

[0001-DONOTUPSTREAM-selftest-tpm-Clear-TPM-before-running-.txt](https://github.com/kernelci/kernelci-core/files/8142876/0001-DONOTUPSTREAM-selftest-tpm-Clear-TPM-before-running-.txt)

But there's one other thing, it seems like some of the grunt devices have trouble running this test. The very same test [you ran here on cbg-7](https://lava.collabora.co.uk/scheduler/job/5772538) I re-ran [explicitly on cbg-4](https://lava.collabora.co.uk/scheduler/job/5773469), and as you can see, the errors I get are the `TPM_RC_LOCKOUT` ones, which can be worked around with the measures on the previous paragraph, while the errors you got were `Device or resource busy: '/dev/tpm0'` and `TPM_RC_BAD_AUTH`. So far it seems like `cbg-0` and `cbg-7` have this issue, while `cbg-1` and `cbg-4` don't. I still need to test on others.

gctucker:
Thanks for the update. We've just identified the issue with different grunt variants in the lab, they should be split as different device types.

Meanwhile I'll get your patch applied on staging linux-next to confirm the rest is working with these changes in place.  The kernel builds include the kselftest binaries, it's just part of the artifacts so that will all get tested together.  I might also re-run some jobs on different grunt devices to compare the results.

gctucker:
Also, starting another rootfs build with the extra patckage.

gctucker:
Starting another amd64 bullseye-kselftest rootfs build on staging now as other things have moved since then, and we'll be able to test that on staging.

gctucker:
Rootfs build: https://bot.staging.kernelci.org/job/rootfs-builder/345/console

gctucker:
Previously:
https://lava.collabora.co.uk/scheduler/job/5857148
```
kselftest: Running tests in tpm2
TAP version 13
1..2
# selftests: tpm2: test_smoke.sh
# ./test_smoke.sh: 9: python3: not found
# ./test_smoke.sh: 10: python3: not found
not ok 1 selftests: tpm2: test_smoke.sh # exit=127
# selftests: tpm2: test_space.sh
# ./test_space.sh: 9: python3: not found
```

Now, with the updated rootfs image:
https://lava.collabora.co.uk/scheduler/job/5857151
```
[   31.995459] kselftest: Running tests in tpm2
TAP version 13
1..2
# selftests: tpm2: test_smoke.sh
# test_read_partial_overwrite (tpm2_tests.SmokeTest) ... ok
# test_read_partial_resp (tpm2_tests.SmokeTest) ... ok
# test_seal_with_auth (tpm2_tests.SmokeTest) ... ok
# test_seal_with_policy (tpm2_tests.SmokeTest) ... ok
# test_seal_with_too_long_auth (tpm2_tests.SmokeTest) ... FAIL
# test_send_two_cmds (tpm2_tests.SmokeTest) ... ok
# test_too_short_cmd (tpm2_tests.SmokeTest) ... ok
# test_unseal_with_wrong_auth (tpm2_tests.SmokeTest) ... ok
# test_unseal_with_wrong_policy (tpm2_tests.SmokeTest) ... ok
# 
# ======================================================================
# FAIL: test_seal_with_too_long_auth (tpm2_tests.SmokeTest)
```
So the test results reported to KernelCI are still the same in terms of pass/fail/skip but at least the tests are actually being run now. I guess there could be test sets defined to include the details of each individual test case, then we would have `kselftest-tpm2.test_smoke.test_read_partial_overwrite` etc. But that's for a follow-up potential improvement in the test-definitions repository.

Strangely, in the job you ran the errors occurred but the results reported to LAVA were all "ok":
https://lava.collabora.co.uk/scheduler/job/5608585

Out of interest, do you know what is causing the error here with `test_seal_with_too_long_auth`?

Also, do you reckon the skiplist is masking too many tests and some should be run on grunt and maybe other platforms?

gctucker (approved):
Tested OK on staging with a few cases failing, but they can be investigated like any kernel failures.

nfraprado:
Sorry for the delay in the feedback. It is indeed strange that my first run passed even with the tests failing, but re-running the tests, I couldn't reproduce that anymore.

And no, I haven't looked further into the `test_seal_with_too_long_auth` error but I think that would need getting deeper into the TPM 2.0 spec to understand what's going wrong there.

Also, I saw some timeouts happening during some more runs of the tests, but the run times were fluctuating so much (the smoke would take anything between 34s to 69s to run) that I think it might have been just some issue with the boards on the lab. With the test being merged now, we'll get much more runs, and if the timeouts keep happening then I can send a patch upstream setting the timeout configuration for the tpm selftest to a higher value (it's currently at the default 45s).

By the way, is there a way to see which patches are currently being applied by KernelCI to run the tests? Even though I created an issue, https://github.com/kernelci/kernelci-project/issues/101, to track the TPM_RC_LOCKOUT issue that we needed a patch to workaround, it would be even better if it was very clear which patch is being used in KernelCI to workaround the issue. I'm just worried we might forget about this patch even. And since it's not upstreamable, we should have it at hand when looking into a proper solution upstream at a later point.

gctucker:
OK, if the test takes between 30 and 60s to run then the timeout should probably be like 120s anyway.

The kernel patches applied on staging can be see on the git branches, here's staging-next for example:
https://github.com/kernelci/linux/commits/staging-next

PR #1036, opened 2022-02-11
kernelci.rootfs: add RootfsBuilder class

gctucker:
Add a RootfsBuilder class as a base class to build rootfs images with
an abstract .build() method and refactor the implementations for debos
and buildroot to use it.


gctucker (commented):
Tested OK on staging

PR #1043, opened 2022-02-17
config/docker: add libhugetlbfs-dev for clang kselftest

gctucker:
Add missing libhugetlbfs-dev package in clang-base to fix some kselftest builds such as the vm ones.


gctucker:
Tested OK on staging:
https://storage.staging.kernelci.org/stable/linux-5.10.y/v5.10.101/x86_64/x86_64_defconfig%2Bx86-chromebook%2Bkselftest/clang-14/logs/kselftest.log
```
Emit Tests for vm
```
which shows that the tests were built correctly, which is not the case when the library is missing.

alexandrasp (approved):
LGTM

PR #1044, opened 2022-02-17
test-configs.yaml: add asus-cx9400-volteer

gctucker:
Add the asus-cx9400-volteer x86 Intel Tiger Lake Chromebook with a
small number of initial test plans to run.

Signed-off-by: Guillaume Tucker <guillaume.tucker@collabora.com>

PR #1045, opened 2022-02-17
test-configs.yaml: add kselftest-vm

gctucker:
Add a definition for the kselftest-vm test plan and enable it on a few devices initially.

Depends on #1044
Depends on #1046

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -293,6 +293,12 @@ test_plans:
       job_timeout: '10'
       kselftest_collections: "seccomp"
 
+  kselftest-vm:
+    <<: *kselftest
+    params:
+      job_timeout: '10'
+      kselftest_collections: "vm"
+
   lc-compliance:
     rootfs: debian_bullseye-libcamera_nfs
     pattern: 'lc-compliance/{category}-{method}-{protocol}-{rootfs}-lc-compliance-template.jinja2'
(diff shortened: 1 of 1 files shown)


alexandrasp (commented):
Initially sounds good, let's wait for some staging results.

gctucker:
I know there's a problem as the vm tests are in the default skiplist so they don't get run. I'm making a fix for that in test-definitions as well, will put an update when it's ready.

gctucker:
> I know there's a problem as the vm tests are in the default skiplist so they don't get run. I'm making a fix for that in test-definitions as well, will put an update when it's ready.

Actually this is now fixed with #1046. The skipfile logic doesn't work as expected when passing `-b ''`, so always providing the board name fixes it.  Maybe a better fix would be in the test-definitions logic but changing that is more likely to break other existing use-cases. It's fine to always pass the board name (and git tree name) anyway.

config/core/test-configs.yaml
```diff
@@ -1837,6 +1843,7 @@ test_configs:
       - baseline-nfs
       - cros-ec
       - kselftest-lib
+      - kselftest-vm
```
gctucker, line 1846: OK https://staging.kernelci.org/test/plan/id/621368f25669cbd30d4b3139/

config/core/test-configs.yaml
```diff
@@ -1962,6 +1969,7 @@ test_configs:
       - kselftest-lkdtm
       - kselftest-rtc
       - kselftest-seccomp
+      - kselftest-vm
```
gctucker, line 1972: OK: https://staging.kernelci.org/test/plan/id/62136a5c925f6bb55c4b3176/

It was run but results were not always parsed correctly. That seems to be more of an issue with the selftest to fix in the kernel tree (worked fine with volteer).

PR #1046, opened 2022-02-17
config/lava/kselftest: add missing parameters to template

gctucker:
Add the BOARD and BRANCH parameters to the kselftest.jinja2 template
as they will then be passed to the skiplist and used to correctly
compute the skipfile.

### DIFF
--- a/config/lava/kselftest/kselftest.jinja2
+++ b/config/lava/kselftest/kselftest.jinja2
@@ -24,3 +24,5 @@
         SKIPFILE: skipfile-lkft.yaml
         TST_CMDFILES: {{ kselftest_collections }}
         TST_CASENAME: {{ kselftest_tests }}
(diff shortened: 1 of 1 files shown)


gctucker:
Tested OK with a resubmitted LAVA job:
https://lava.collabora.co.uk/scheduler/job/5718524

Original one without the extra parameters:
https://lava.collabora.co.uk/scheduler/job/5718522

gctucker:
@alexandrasp By the way, this seems to have fixed the issue with skipped tests from unrelated collections:
https://staging.kernelci.org/test/plan/id/620f602a3ad2bf112a4b3160/

I still think there's a bug in test-definitions, as when no board is provided it should produce the same skipfile as when there is a board that doesn't match any of the skipped ones.  But adding the board parameter seems fine in any case, except it kind of hides the issue.

PR #1047, opened 2022-02-18
Pub/Sub filter implementation

JenySadadia:
Fixes https://github.com/kernelci/kernelci-api/issues/28 

### DIFF
--- a/kernelci/data/kernelci_api.py
+++ b/kernelci/data/kernelci_api.py
@@ -34,6 +34,7 @@ class KernelCI_API(Database):
             'Authorization': f'Bearer {self._token}',
             'Content-Type': 'application/json',
         }
+        self._filters = {}
 
     def _make_url(self, path):
         return urllib.parse.urljoin(self.config.url, path)
@@ -60,7 +61,15 @@ class KernelCI_API(Database):
         resp = self._post(f'subscribe/{channel}')
         return json.loads(resp.text)['id']
 
+    def subscribe_node_channel(self, filters=None):
+        resp = self._post(f'subscribe/node')
+        sub_id = json.loads(resp.text)['id']
+        self._filters[sub_id] = filters
+        return sub_id
+
     def unsubscribe(self, sub_id):
+        if sub_id in self._filters:
+            self._filters.pop(sub_id)
         self._post(f'unsubscribe/{sub_id}')
 
     def get_event(self, sub_id):
--- a/kernelci/data/kernelci_api.py
+++ b/kernelci/data/kernelci_api.py
@@ -103,6 +103,46 @@ class KernelCI_API(Database):
         resp = self._get('/'.join(['get_root_node', node_id]))
         return json.loads(resp.text)
 
+    def pubsub_event_filter(self, sub_id, event):
+        """Filter Pub/Sub events
+
+        Filter received Pub/Sub event using provided filter dictionary.
+        Return True if client has not provided any filter dictionary.
+        If filters are provided, return True if the event data matches with
+        the filter parameters, otherwise False.
+        """
+        filters = self._filters.get(sub_id)
+        if not filters:
+            return True
+        for key, value in filters.items():
+            if key not in event.keys():
+                continue
+            if isinstance(value, dict):
+                for sub_key, sub_value in value.items():
+                    if sub_key not in event.get(key):
+                        continue
+                    if sub_value != event.get(key).get(sub_key):
+                        return False
+            elif value != event[key]:
+                return False
+
+        return True
+
+    def receive_node(self, sub_id):
+        """
+        Listen to all the events on 'node' channel and apply filter on it.
+        Return node if event matches with the filter.
+        """
+        path = '/'.join(['listen', str(sub_id)])
+        while True:
+            resp = self._get(path)
+            event = from_json(resp.json().get('data'))
+            node = self.get_node_from_event(event)
+            if all(self.pubsub_event_filter(sub_id, obj) for obj in [
+                                                                node,
+                                                                event.data]):
+                return node
+
     def submit(self, data, verbose=False):
         obj_list = []
         for path, item in data.items():
--- a/kernelci/data/kernelci_api.py
+++ b/kernelci/data/kernelci_api.py
@@ -133,10 +133,8 @@ class KernelCI_API(Database):
         Listen to all the events on 'node' channel and apply filter on it.
         Return node if event matches with the filter.
         """
-        path = '/'.join(['listen', str(sub_id)])
         while True:
-            resp = self._get(path)
-            event = from_json(resp.json().get('data'))
+            event = self.get_event(sub_id)
             node = self.get_node_from_event(event)
             if all(self.pubsub_event_filter(sub_id, obj) for obj in [
                                                                 node,


JenySadadia:
The last commit can be removed after testing.

JenySadadia:
Removed the code with debug prints and code.

kernelci/data/kernelci_api.py
```diff
@@ -83,6 +95,47 @@ def get_nodes_by_commit_hash(self, commit_hash):
...
+        path = '/'.join(['listen', str(sub_id)])
+        resp = self._get(path)
+        event = from_json(resp.json().get('data'))
+        node = self.get_node_from_event(event)
+        node['op'] = event.data['op']
+        event_filter_status = self.pubsub_event_filter(sub_id, node)
+
+        if event_filter_status is False:
+            return
+        return node
```
gctucker, line 137: I believe this should be a `while` loop to only return a Node object when it matches the filter.  From a caller point of view, it basically means `db.receive_node()` will always return a node matching the filters set when subscribing.

kernelci/data/kernelci_api.py
```diff
@@ -83,6 +95,47 @@ def get_nodes_by_commit_hash(self, commit_hash):
...
+        Return True if the event matches with the filter, otherwise False.
+        """
+        event_filter_status = True
+        filters = self._filters.get(sub_id)
+        if not filters:
+            return event_filter_status
+        for key, value in filters.items():
+            if key not in event.keys():
+                continue
+            if isinstance(value, dict):
```
gctucker, line 111: This only looks for one sub-level in the object, which is fine with the current data schema we have. I think we can merge this code, but ideally a follow-up would be to make it truly recursive and try to match filers with any levels of nested dictionary values.

kernelci/data/kernelci_api.py
```diff
@@ -60,9 +61,20 @@ def subscribe(self, channel):
...
+        self._filters[sub_id] = filters
+        print("Subscribed with filters", self._filters)
+        return sub_id
+
     def unsubscribe(self, sub_id):
         self._post(f'unsubscribe/{sub_id}')
 
+    def unsubscribe_node_channel(self, sub_id):
+        self._filters.pop(sub_id)
+        self._post(f'unsubscribe/{sub_id}')
```
gctucker, line 76: I don't think we need a new method for this. It only needs `sub_id`, so we can do something like `self._filters.pop(sub_id) if sub_id in self._filters`

The advantage being that we would have only one method for unsubscribing, so if something needs to always be done to avoid resource leaks we would just have to implement it in that method.

kernelci/data/kernelci_api.py
```diff
@@ -60,9 +61,20 @@ def subscribe(self, channel):
         resp = self._post(f'subscribe/{channel}')
         return json.loads(resp.text)['id']
 
+    def subscribe_node_channel(self, filters=None):
+        resp = self._post(f'subscribe/node')
+        sub_id = json.loads(resp.text)['id']
+        self._filters[sub_id] = filters
+        print("Subscribed with filters", self._filters)
```
gctucker, line 68: This seems more like some debug print, it probably should be removed.

kernelci/data/kernelci_api.py
JenySadadia: Agreed.

kernelci/data/kernelci_api.py
JenySadadia: Okay.

kernelci/data/kernelci_api.py
JenySadadia: We need to run `db.receive_node` inside `while` loop. It will return node object if filter matches, otherwise None.

```
while True:
            node = db.receive_node(sub_id)
            if node:
                print("Node received:", node)
```

gctucker: But what is the value of returning None to the caller when the method could just keep waiting until a matching node is received? We could just remove the `if node` part on the caller side if the method always returns a matching node. Errors would be raised as exceptions.

JenySadadia: Updated the PR.

kernelci/data/kernelci_api.py
```diff
@@ -83,6 +92,49 @@ def get_nodes_by_commit_hash(self, commit_hash):
...
+                resp = self._get(path)
+                event = from_json(resp.json().get('data'))
+                node = self.get_node_from_event(event)
+                node['op'] = event.data['op']
+                if self.pubsub_event_filter(sub_id, node):
+                    return node
+        except Exception as e:
+            print(f"Error: {e}")
+        except KeyboardInterrupt as e:
+            print(f"Stopping.")
```
gctucker, line 136: Please remove this or we won't be able to abort the applications (e.g. trigger.py etc.) which relies on catching it. In fact I think we can drop all exception handling here, they can be caught by the code calling this method.

kernelci/data/kernelci_api.py
gctucker: > Errors would be raised as exceptions.

What I meant is that we don't need to rely on the method to return `None` to propagate errors, as that can be one using exceptions.  So the `receive_node()` method will always return a valid node object, or throw an exception if an error occured.

kernelci/data/kernelci_api.py
JenySadadia: Yes, I missed that scenario. Thanks for pointing it out.

kernelci/data/kernelci_api.py
```diff
@@ -83,6 +92,44 @@ def get_nodes_by_commit_hash(self, commit_hash):
...
+        """
+        Listen to all the events on 'node' channel and apply filter on it.
+        Return node if event matches with the filter.
+        """
+        path = '/'.join(['listen', str(sub_id)])
+        while True:
+            resp = self._get(path)
+            event = from_json(resp.json().get('data'))
+            node = self.get_node_from_event(event)
+            node['op'] = event.data['op']
```
gctucker, line 129: I think it's not great practice to change the data in the node dictionary as it's not a node any more. Instead, the node and event objects could be passed to the filter function which would then try to match the filters with both of them. Or you could call the function twice:
```python
if all(self.pubsub_event_filter(sub_id, obj) for obj in [node, event]):
    return node
````
Then the `node` object being returned will be what was received from the API with any modifications (i.e. without adding `op` to it).

**Comment by JenySadadia** [comment]
> Adding skip to test and merge #1052 first since they conflict with each other.

I rebased the PR.

File: kernelci/data/kernelci_api.py [hunk 7]
```diff
@@ -94,6 +103,45 @@ def get_root_node(self, node_id):
...
+            return event_filter_status
+        for key, value in filters.items():
+            if key not in event.keys():
+                continue
+            if isinstance(value, dict):
+                for sub_key, sub_value in value.items():
+                    if sub_key not in event.get(key):
+                        continue
+                    if sub_value != event.get(key).get(sub_key):
+                        return
```
gctucker, line 124: The docstring says the returned value should be a boolean, but here's returning `None`. This will evaluate to false with a basic `if` test but not if the returned value is compared with `False`. So it might work with some luck on how the returned value is used.
```suggestion
                        return False
```

kernelci/data/kernelci_api.py
```diff
@@ -94,6 +103,45 @@ def get_root_node(self, node_id):
         resp = self._get('/'.join(['get_root_node', node_id]))
         return json.loads(resp.text)
 
+    def pubsub_event_filter(self, sub_id, event):
+        """Filter Pub/Sub events
+
+        Filter received Pub/Sub event using provided filter dictionary.
+        Return True if the event matches with the filter, otherwise False.
+        """
+        event_filter_status = True
```
gctucker, line 112: This variable isn't actually used, because the method just returns every time it finds a filter that doesn't match the data. So I think we can drop it.
```suggestion
```

kernelci/data/kernelci_api.py
```diff
@@ -94,6 +103,45 @@ def get_root_node(self, node_id):
...
+        filters = self._filters.get(sub_id)
+        if not filters:
+            return event_filter_status
```
gctucker, line 115: Following the same idea:
```suggestion
            return True
```

kernelci/data/kernelci_api.py
```diff
@@ -94,6 +103,45 @@ def get_root_node(self, node_id):
...
+            if key not in event.keys():
+                continue
+            if isinstance(value, dict):
+                for sub_key, sub_value in value.items():
+                    if sub_key not in event.get(key):
+                        continue
+                    if sub_value != event.get(key).get(sub_key):
+                        return
+            elif value != event[key]:
+                return
```
gctucker, line 126: Ditto
```suggestion
                return False
```

kernelci/data/kernelci_api.py
```diff
@@ -94,6 +103,45 @@ def get_root_node(self, node_id):
...
+
+        return event
```
gctucker, line 128: This is probably because an earlier implementation was going to return either the event or `None`, but now it's returning a boolean. So, to be consistent:
```suggestion
        return True
```

kernelci/data/kernelci_api.py
```diff
@@ -94,6 +103,45 @@ def get_root_node(self, node_id):
         resp = self._get('/'.join(['get_root_node', node_id]))
         return json.loads(resp.text)
 
+    def pubsub_event_filter(self, sub_id, event):
+        """Filter Pub/Sub events
+
+        Filter received Pub/Sub event using provided filter dictionary.
+        Return True if the event matches with the filter, otherwise False.
```
gctucker, line 110: The wording might need to be tweaked a bit, because "matching the filter" could be interpreted as "the filter parameters are blocking the event". Say, if the filter parameters match the event data, then the result should be `False` even though it's "matching". Did I get this right?

If that's the case, i.e. `True` means the event should be kept and `False` means the event should be discarded, then the wording should reflect that more clearly.

gctucker (commented):
A couple of minor things to tweak

kernelci/data/kernelci_api.py
JenySadadia: > The wording might need to be tweaked a bit, because "matching the filter" could be interpreted as "the filter parameters are blocking the event". Say, if the filter parameters match the event data, then the result should be `False` even though it's "matching". Did I get this right?
> 
> If that's the case, i.e. `True` means the event should be kept and `False` means the event should be discarded, then the wording should reflect that more clearly.

Actually, the method returns True if no filter dict is provided or event data matches with the filter parameters. Returns False otherwise.

gctucker: I guess "no filter" or an empty dictionary is just a special case where the event is never filtered. So it's consistent with the idea that the method only returns `False` if there's a filter with at least one key that is present in the event data but with a value that doesn't match what is in the event data.

gctucker (approved):

gctucker (approved):
Added a fix related to the keep-alive events.

Tested OK on staging:
```
today at 8:10:07 AMListening for events... 
today at 8:10:07 AMPress Ctrl-C to stop.
today at 8:10:09 AMTime                        Commit        Status    Name
today at 8:10:09 AM2022-03-08 08:10:08.906433  abba0b6c99db  Pending   checkout
today at 8:11:00 AM2022-03-08 08:11:00.099377  abba0b6c99db  Pass      checkout
today at 8:11:00 AM2022-03-08 08:11:00.254930  abba0b6c99db  Pending   check-describe
today at 8:12:03 AM2022-03-08 08:12:03.275342  abba0b6c99db  Fail      check-describe
```

PR #1048, opened 2022-02-18
config/docker: add 32-bit support to gcc-10 x86

gctucker:
Add support for building 32-bit binaries to the gcc-10 x86 Docker
image.  This is useful for some kselftests (x86 and vm) which build
both 64-bit and 32-bit variants of some test programs.

### DIFF
--- a/config/docker/gcc-10_x86/Dockerfile
+++ b/config/docker/gcc-10_x86/Dockerfile
@@ -20,3 +20,9 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
    libpopt-dev \
    libasound2-dev \
    pkg-config
+
+# 32-bit support for kselftest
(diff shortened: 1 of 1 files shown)


gctucker (commented):
Tested OK locally with kselftest x86 build, for example this test binary was compiled both for 32-bit and 64-bit:
```
ccache gcc -Wall -I ../../../../usr/include  -no-pie -m32  protection_keys.c -lrt -lrt -ldl -lm -o /root/kernelci-core/linux/build/kselftest/vm/protection_keys_32
ccache gcc -Wall -I ../../../../usr/include  -no-pie -m64  protection_keys.c -lrt -lrt -ldl -o /root/kernelci-core/linux/build/kselftest/vm/protection_keys_64
```
It requires a fix in the Makefile though, so the logs on staging aren't showing this yet.

PR #1050, opened 2022-02-21
kci_rootfs: Add --output option

nuclearcat:
At current moment kci_rootfs generate artifacts in random, option specific directories,
which increase complexity of build configuration.

This patch adds mandatory --output option for directory, where artifacts generated.

Also it adds git clone option for buildroot, otherwise without it
buildroot rootfs option is incomplete and doesnt work "standalone".

Additionally it will have predictable structure for output:
{output_directory}/{config_name}/{arch}

### DIFF
--- a/kernelci/rootfs.py
+++ b/kernelci/rootfs.py
@@ -19,6 +19,7 @@
 from kernelci import shell_cmd
 from kernelci.storage import upload_files
 import os
+import shutil
 
 
 class RootfsBuilder:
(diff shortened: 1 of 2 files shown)


kernelci/rootfs.py
```diff
@@ -58,9 +73,10 @@ def build(self, config, data_path, arch):
                 for key, value in debos_params.items()
             )
         )
-        cmd = "cd {path} && debos --memory=4G {opts} rootfs.yaml".format(
-            path=data_path, opts=debos_opts
-        )
+        cmd = f"debos --memory=4G {debos_opts}"\
+            f" --artifactdir={absoutput_dir} "\
+            f"{self.kci_path}/config/rootfs/debos/rootfs.yaml"
```
gctucker, line 78: One issue is that we can't rely on the config files to always be in that sub-directory. This works when running `kci_rootfs` from a plain git checkout of kernelci-core, but if the package is installed then `kci_rootfs` may be in `/usr/local/bin` and the configuration files in `/etc/` or `/usr/local/share`. We don't have this problem with how Jenkins jobs are currently run, but if we lose the ability to specify where the config files are and where the build output should be then we'll most likely have to add it again at some point. So it's better if we can keep support for that without complicating the code. If it's an issue then we can simplify things and only make it work with the Jenkins use-case for now.

gctucker: Also, `os.path.join()` is the recommended way of constructing file paths.

nuclearcat: Ok, will put back this option

gctucker (approved):
Tested OK on staging:
https://bot.staging.kernelci.org/job/rootfs-builder/337/console

kernelci/rootfs.py
```diff
@@ -70,12 +88,34 @@ def __init__(self, *args, **kwargs):
...
+        artifact_dir = os.path.join(absoutput_dir, '_install_',
+                                    self.name, arch)
+        temp_dir = os.path.join(absoutput_dir, 'temp')
+
+        # Create directories if missing
+        if not os.path.isdir(artifact_dir):
+            os.makedirs(artifact_dir, exist_ok=True)
+        if not os.path.isdir(temp_dir):
+            os.makedirs(temp_dir, exist_ok=True)
+        os.chdir(temp_dir)
```
gctucker, line 102: This works, but relying on the current directory adds some implicit state to the logic so it's more likely to break with future changes. Ideally, each command should be run within a full relative path from the top-level directory.  See the `kernelci.build` code, there's `cd {kdir}` at the beginning of each command. Maybe this could be done here as a follow-up PR.

gctucker:
Tested OK on staging with a sample rootfs image:
https://lava.collabora.co.uk/scheduler/job/5772538

(The test cases are failing but that's a kselftest / kernel problem)

PR #1052, opened 2022-02-24
kernelci_api.py: Add KernelCI_API.get_root_node function

JenySadadia:
Add function to call 'get_root_node/' endpoint of
kernelci-api.


gctucker:
I think this is conflicting with #1047 so it's not currently being deployed on staging.  I'll add skip labels to the other PR relying on it in kernelci-pipeline.

gctucker (approved):
Tested OK on staging

PR #1055, opened 2022-03-01
WIP kernelci.data.kernelci_api: add 30s timeout to listen requests

gctucker:


gctucker:
This is causing some issues on the server side:
```
RuntimeError: read() called while another coroutine is already waiting for incoming data
```
Closing for now. Another approach would be to have events sent periodically to each listening client as a keep-alive feature.

PR #1056, opened 2022-03-01
kernelci.data.kernelci_api: handle keep-alive events

gctucker:
Handle keep-alive events received on the pub/sub interface.  These are
only there to ensure the TCP connection stays active.


gctucker:
Depends on https://github.com/kernelci/kernelci-api/pull/125

PR #1057, opened 2022-03-01
lab-configs: Enable more testsuites for my lab

broonie:
Should be academic until more machines come online (but I'm working on
that).

### DIFF
--- a/config/core/lab-configs.yaml
+++ b/config/core/lab-configs.yaml
@@ -22,6 +22,9 @@ labs:
             - baseline
             - kselftest
             - ltp
+            - preempt-rt
+            - sleep
+            - smc
 
   lab-cip:
     lab_type: lava.lava_xmlrpc


alexandrasp:
are these results available on KernelCI staging page? I can not found your lab info on https://staging.kernelci.org ..

gctucker:
> are these results available on KernelCI staging page? I can not found your lab info on https://staging.kernelci.org ..

No, because they're not online yet as per the PR description.  Guess this PR was created a bit too early, but that's benign.

@broonie Please put a comment when your devices become available in your lab and have staging KernelCI tests run on them.

broonie:
There's now a juno-r0-acpi device there. Not sure I see the blocker in merging even without that though given that the boards need individually listing too? My goal here is to get the lab permission granted and not have to worry about that when adding a board.

gctucker:
Sure, well there's probably no real blocker but it's adding something that's not used. So, why merge it now?

broonie:
Mainly to reduce the diff I'm carrying (so I don't have to remember to add things later), possibly also to help people add any new testsuites here. It will also allow things to be run on qemu if someone wants to do that (we should for at least some of kselftst). 

In any case there's a pull request for one board there now.

gctucker:
OK fine let's merge it now.

But please also add your lab to the QEMU passlist filter when Docker support is working, as the moment there won't be any QEMU jobs run at all (only lab-baylibre is really running QEMU jobs in production right now, collabora and lkft should be added soon).

gctucker (approved):

gctucker:
OK, it was failing last month:
https://lava.sirena.org.uk/scheduler/job/8547

I can make a PR to enable it again so we can check results on staging.

gctucker:
Done: #1088 Let's now wait for the next staging results to confirm.

PR #1058, opened 2022-03-01
CHROMEOS: Restore configuration for chromeos rootfs build and add support for --output option

nuclearcat:
ChromeOS rootfs builds should support --output option too


nuclearcat:
I will try to run builds on my branch first

gctucker:
Out of interest, did this get lost while resolving conflicts during a rebase or is it because a patch got dropped somehow?

PR #1059, opened 2022-03-02
Run IGT on mt8183

nfraprado:
This PR is based on https://github.com/kernelci/kernelci-core/pull/1352.

This adds an IGT test plan using the mediatek driver and runs the test on mt8183-jacuzzi-juniper.

Tested with linux-next-20220725.

This patch is needed to fix a display regression affecting all MTK platforms: https://lore.kernel.org/all/20220721172727.14624-1-angelogioacchino.delregno@collabora.com/
This patch is not required to get the tests running, but gets rid of the spam of kernel warnings during the run: https://lore.kernel.org/all/20220804194325.1596554-1-nfraprado@collabora.com/

LAVA run with only first patch applied: https://lava.collabora.dev/scheduler/job/7009915
LAVA run with both patches applied: https://lava.collabora.dev/scheduler/job/7009979

(Note: I added an initial delay on the run to make sure the DRM driver probes before the tests start. The test might not pass consistently as it currently stands, in which case we should find a way to check all modules have probed before starting tests in KernelCI.)


nfraprado:
Turns out that the wrong bus-type value was being used for the DSI/DPI operation of the anx7625 driver, so the related bindings are being reverted for 5.17: https://lore.kernel.org/linux-mediatek/20220308094911.2680291-1-robert.foss@linaro.org/.

So I'll mark this PR as a draft for now, because the driver, binding and devicetree now need new patches to use the correct bus-types before we can run IGT here.

gctucker:
The "convert to draft" link worked for me.

gctucker:
Please feel free to add links to patches that would fix the issue and we can have them deployed on staging to confirm tests are then passing with them.

gctucker:
@nfraprado Please rebase and provide an update regarding pending kernel patches.

nuclearcat:
ping :)

nfraprado:
Hi sorry for the delay, I'll try to work out the setup for this test again in the next few days. Lately I've been concentrating on improving the baseline tests, as they make it clearer if the basic support is good for all other tests, including this. So maybe take a look at https://github.com/kernelci/bootrr/pull/22 and https://github.com/kernelci/kernelci-core/pull/1352 in the meantime? Thanks :).

nfraprado:
PR description updated with required patches, and rebased on top of #1352 since that PR already provides all configs needed.

nuclearcat:
Might be expected some progress on this PR after ~6.3 kernel

gctucker:
How about `linux-next` if all the patches required have landed on maintainers' branches but are not in mainline yet?

gctucker:
@nfraprado Please rebase if this PR is still relevant (see earlier comments about depending on some upstream kernel patches).

Note: We might also apply some missing kernel patches on the branches used by staging.kernelci.org to help show this will then work.

nfraprado:
@gctucker Actually just enabled this as part of #1784 a few days ago. You can see results here: https://linux.kernelci.org/test/plan/id/6411519702d15a096f8c8648/

So closing this PR.

PR #1061, opened 2022-03-03
v4l2-compliance: Add support for capture and output format arguments

nfraprado:
This PR changes the v4l2-parser.sh script and v4l2-compliance.jinja2 template in order to allow for two new parameters to be passed in v4l2-compliance test plans: `v4l2_capture_format` and `v4l2_output_format`. These allow a video node to be matched also based on its capture and/or output format. This is useful since there are some platforms (like mt8173) that have more than one codec device using the same driver, and the only difference between them is the capture/output format. So these arguments allow each of the codec on platforms like this to be tested independently.

The matching of the device node based on the format needs to be done in the v4l-get-device utility, so [a Merge Request](https://gitlab.collabora.com/gtucker/v4l2-get-device/-/merge_requests/1) was created adding this functionality to it. Hence, these two new arguments added in the third commit of this PR can only be used after that MR in v4l2-get-device is merged.

Example LAVA run: https://lava.collabora.co.uk/scheduler/job/5810811
It uses the updated v4l2-get-device and changed argument syntax in v4l2-parser to match `/dev/video2`, which uses the `mtk-vcodec-enc` driver and has the `VP80` capture format, through the command `/usr/bin/v4l2-parser.sh -d mtk-vcodec-enc -f VP80`. There's another `mtk-vcodec-enc` device at `/dev/video1` which is (successfully) not matched by these arguments.

### DIFF
--- a/config/lava/v4l2-compliance/v4l2-compliance.jinja2
+++ b/config/lava/v4l2-compliance/v4l2-compliance.jinja2
@@ -1,6 +1,12 @@
 {% if v4l2_driver %}
     {% set v4l2_driver_arg = "-d " + v4l2_driver %}
 {% endif %}
+{% if v4l2_capture_format %}
+    {% set v4l2_capture_format_arg = "-f " + v4l2_capture_format %}
+{% endif %}
+{% if v4l2_output_format %}
+    {% set v4l2_output_format_arg = "-F " + v4l2_output_format %}
+{% endif %}
 - test:
     timeout:
       minutes: 10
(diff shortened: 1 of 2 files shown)


gctucker:
Also added a review on https://gitlab.collabora.com/gtucker/v4l2-get-device/-/merge_requests/1

The rootfs image build mentioned above doesn't have the MR changes in v4l2-get-device.

gctucker:
Please take a look at my review comments here:
https://gitlab.collabora.com/gtucker/v4l2-get-device/-/merge_requests/1

Then when that's merged we can have a new rootfs image built and test the whole thing on staging.

gctucker:
OK the MR mentioned above has been merged, so we should now be able to check this is all working as expected on staging.

gctucker:
Building a bullseye-v4l2 image for arm64 now on staging:
https://bot.staging.kernelci.org/job/rootfs-build-trigger/127/

Then the tests can be resubmitted with it to check it's all working as expected.

gctucker:
OK that worked: https://lava.collabora.co.uk/scheduler/job/6582980

PR #1062, opened 2022-03-04
CHROMEOS: Adjust uid for jenkins configuration

nuclearcat:
Jenkins run docker as "docker run -t -d -u 996:1001", so we are forced to use uid 996 to keep sudo and other scripts working.

### DIFF
--- a/config/docker/chromiumos/Dockerfile
+++ b/config/docker/chromiumos/Dockerfile
@@ -18,7 +18,7 @@ RUN apt-get update \
     && apt-get autoremove \
     && rm -rf /var/lib/apt/lists/*
 
-RUN useradd -u 1000 -ms /bin/sh user && adduser user sudo && echo '%sudo ALL=(ALL) NOPASSWD:ALL' >> /etc/sudoers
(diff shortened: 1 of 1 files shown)


gctucker (commented):
Jenkins is so idiosyncratic...

Please share a link to a job if you have tested this on staging to check it's working as expected.

nuclearcat:
https://bot.staging.kernelci.org/job/denysf/job/rootfs-builder-denysf/8/consoleFull

You can see here "docker run -t -d -u 996:1001 --privileged --device /dev/kvm -v /dev:/dev"
And inside those debug commands, i had more debug in this job than in commit, to make sure jenkins doesnt modify passwd or etc, because i was not sure why sudo fails in middle of script and why PATH variable was wrong.

gctucker:
Thanks, I see the job worked but the upload failed. The files have been copied manually here:
http://storage.chromeos.kernelci.org/images/rootfs/chromeos/chromiumos-amd64-generic/crostest003/

gctucker (approved):

PR #1075, opened 2022-03-12
test-configs.yaml: update rootfs URLs

mgalka:
update rootfs URLs


gctucker:
I've closed the conflicting PRs which were to test staging rootfs images with some changes. This is now deployed on staging, we can keep these rootfs images for Monday's production update even if it's a week old.

gctucker (approved):
Tested OK on staging

PR #1076, opened 2022-03-14
CHROMEOS: Add Chrome OS flashing template

mgalka:
Add LAVA test job template and configuration necessary for flashing
Chrome OS rootfs onto eMMC. The current configuration and scripts
support octopus only.


config/core/test-configs.yaml
```diff
@@ -155,6 +155,17 @@ test_plans:
       - passlist: &qemu_labs_filter
           lab: ['lab-baylibre', 'lab-collabora-staging']
 
+  chromeos-flash:
+    pattern: 'chromeos/chromeos-flash-template.jinja2'
+    params:
+      job_timeout: '15'
+      docker_image: 'kernelci/chromeos-flash'
+      chromeos_image: 'chromiumos_test_image.bin'
```
gctucker, line 163: As this is an `octopus` image, we probably should include `octopus` in the file name or in a sub-directory, ideally with the Chrome OS release number as well, to better differentiate with future images.  The `chromiumos_test_image.bin` name is the standard one produced by the Chrome OS SDK build system.

config/lava/chromeos/chromeos-flash-template.jinja2
```diff
@@ -0,0 +1,83 @@
...
+            - debian
+          scope:
+            - functional
+          environment:
+            - lava-test-shell
+        run:
+          steps:
+            - lsblk
+            - ls -l /root
+            - IMAGE_NAME=chromiumos_test_image.bin  /bin/bash /usr/bin/chromeos-flash-modules.sh flash_chromeos enable_rw fixup_tmpfiles
```
gctucker, line 77: This is still slightly some work-in-progress but I think the script should be renamed or refactored, to have a clearer distinction between flashing and installing modules.  Also this probably should be installed in `/usr/local` or in `/opt` rather than `/usr/bin`.  We could have `/opt/chromeos/flash`, `/opt/chromeos/install-modules` etc.

config/lava/chromeos/chromeos-flash-template.jinja2
```diff
@@ -0,0 +1,83 @@
...
+
+- test:
+    timeout:
+      minutes: 60
+    definitions:
+    - repository:
+        metadata:
+          format: Lava-Test Test Definition 1.0
+          name: hack
+          description: "hack"
```
gctucker, line 66: It's not a hack any more ;)

config/lava/chromeos/chromeos-flash-template.jinja2
```diff
@@ -0,0 +1,83 @@
...
+          steps:
+            - lsblk
+            - ls -l /root
+            - IMAGE_NAME=chromiumos_test_image.bin  /bin/bash /usr/bin/chromeos-flash-modules.sh flash_chromeos enable_rw fixup_tmpfiles
+      lava-signal: kmsg
+      from: inline
+      name: hack
+      path: inline/hack.yaml
+      
+{% endblock %}
```
gctucker, line 83: Missing line return

config/lava/chromeos/chromeos-flash-template.jinja2
```diff
@@ -0,0 +1,83 @@
...
+      wait:
+        device: true
+    definitions:
+    - repository:
+        metadata:
+          format: Lava-Test Test Definition 1.0
+          name: docker-network
+        run:
+          steps:
+            - lava-test-case uname --shell "uname -a"
```
gctucker, line 50: I think this was a debug thing early on, is it still useful?

config/lava/chromeos/chromeos-flash-template.jinja2
```diff
@@ -0,0 +1,83 @@
...
+            - ping -c 1 172.17.0.1
+            - wget http://172.17.0.1/chromeos/{{ chromeos_image }}
```
gctucker, line 52: We could also treat the `wget` command as a test case to show in the results whether it worked or not, like the `scp` command.

config/lava/chromeos/chromeos-flash-template.jinja2
gctucker: Also the `scp` command is sending the file using the template variable {{ chromeos_image }} but here the value of `$IMAGE_NAME` is hard-coded.

gctucker (commented):
I think this is generic enough and should work on any device, not just `octopus`. The only thing is that the kernel and rootfs images used for flashing are x86, so we may have to reflect that in the name.  The Chrome OS image name should also show that it's for `octopus`, but the template seems entirely generic.

config/lava/chromeos/chromeos-flash-template.jinja2
mgalka: It is, while it is a draft request and I use it to test the job

config/core/test-configs.yaml
```diff
@@ -155,6 +155,17 @@ test_plans:
...
+      flashing_kernel: https://people.collabora.com/~mgalka/stable/linux-5.10.y/v5.10.95/x86_64/x86_64_defconfig+x86-chromebook/gcc-10/kernel/bzImage
+      flashing_modules: https://people.collabora.com/~mgalka/stable/linux-5.10.y/v5.10.95/x86_64/x86_64_defconfig+x86-chromebook/gcc-10/modules.tar.xz
+      flashing_rootfs: https://storage.kernelci.org/images/rootfs/debian/bullseye-chromeos-flash/20220311.0/amd64/full.rootfs.tar.xz
+      flashing_ramdisk: http://storage.kernelci.org/images/rootfs/debian/buster/20210503.0/amd64/initrd.cpio.gz
```
gctucker, line 167: I think these images should eventually be stored on storage.chromeos.kernelci.org

It's probably not a blocker for this PR, but some tidying-up we can do once this is working.

config/core/test-configs.yaml
```diff
@@ -2000,6 +2011,7 @@ test_configs:
       - baseline
       - baseline-cip-nfs
       - baseline-nfs
+      - chromeos-flash
```
gctucker, line 2014: Adding it here means the pipeline logic in Jenkins will be automatically scheduling jobs, which is not what we want.  By specifying the `--plan` and `--target` command line arguments to `kci_test`, it's possible to generate a test definition that's not listed in the YAML.

config/lava/chromeos/chromeos-flash-template.jinja2
```diff
@@ -0,0 +1,85 @@
...
+          steps:
+            - lsblk
+            - ls -l /root
+            - IMAGE_NAME={{ chromeos_image }}  /bin/bash /usr/bin/chromeos-flash-modules.sh flash_chromeos enable_rw fixup_tmpfiles
+      lava-signal: kmsg
+      from: inline
+      name: chromeos-flash
+      path: inline/chromeos-flash.yaml
+      
+{% endblock %}
```
gctucker, line 85: Looks like it's still missing a line return.

gctucker:
@mgalka I've now run a job with an added `fixup_root` command instead of the one to install systemd temporary files and that's working fine:
https://lava.collabora.co.uk/scheduler/job/5879160

The next job to run Chrome OS worked without any manual fixup:
https://lava.collabora.co.uk/scheduler/job/5879162
(some JSON error but at least Chrome OS started completely)

So maybe it's time to respin the rootfs image with an updated version of the script and run the same commands with the `root_fixup` stuff.  See the job definition for the URL of the script I've been using.

Also, maybe this isn't a draft any more? What needs to be done before it can be merged aside from resolving the review comments and confirming it can flash a working image on a Chromebook?

config/core/test-configs.yaml
gctucker: I see you added the Chrome OS release number to the file name but not `octopus`. Is that on purpose? I believe the image was built specifically for `octopus` but we could double check with Brian.

gctucker:
@mgalka Please rebase as the `chromeos` branch has been updated.

mgalka:
PR rebased with current kernelci/chromeos branch.

Flashing template tested:
https://lava.collabora.co.uk/scheduler/job/5958715

config/rootfs/chromiumos/scripts/build_board.sh
```diff
@@ -50,3 +50,41 @@ gzip -1 "${DATA_DIR}/${BOARD}/chromiumos_test_image.bin"
 
 # Probably redundant, but better safe than sorry
 cleanup
+=======
```
gctucker, line 53: Seems like a leftover from a git conflict

gctucker (changes requested):
There are unexpected changes in `build_board.sh` which probably came from the rebase.

config/rootfs/chromiumos/scripts/build_board.sh
mgalka: Yes, it's fixed now

gctucker (approved):

PR #1077, opened 2022-03-14
config/docker: Fix dt-validation Docker image

mgalka:
Add dependencies needed to build and install dt-schema Python package.


gctucker (approved):
Built OK on staging

PR #1079, opened 2022-03-17
CHROMEOS Clean workspace on error or exit, compress image

nuclearcat:
Use trap function, so it will clean not only on exit, but
also if script quit early due error.
This resolve problems with jenkins.
Also compress image, as it shrink image from 6GB to approx 2GB.


nuclearcat:
Example of successful job:
https://bot.staging.kernelci.org/job/denysf/job/rootfs-builder-denysf/23/console

PR #1082, opened 2022-03-21
Update buildroot URLs with latest bootrr

gctucker:


gctucker:
This contains the changes from https://github.com/kernelci/bootrr/pull/14 and https://github.com/kernelci/bootrr/pull/15

gctucker:
Actually this build didn't include the changes from pending PRs since the buildroot job has changed and it was using production branch. Preparing a fix for that, will then build another set of images for staging.

gctucker:
Tested OK on staging. Closing now as this will then be picked up with a production build.

PR #1087, opened 2022-03-23
Add support for queue timeout

montjoie:
Since 2021.08, LAVA supports queue timeout.
Let's adds support for it in kernelCI.
Since my lab is recent enough, enable queue_timout in my lab.

Closes: 1086

### DIFF
--- a/config/lava/base/kernel-ci-base.jinja2
+++ b/config/lava/base/kernel-ci-base.jinja2
@@ -64,6 +64,15 @@ timeouts:
   actions:
     power-off:
       seconds: 30
+{%- if queue_timeout %}
+  queue:
+    {%- if queue_timeout.hours %}
+    hours: {{ queue_timeout.hours }}
+    {% endif %}
+    {%- if queue_timeout.days %}
+    days: {{ queue_timeout.days }}
+    {% endif %}
+{% endif %}
 priority: {{ priority }}
 visibility: public
 {% endblock %}
--- a/config/core/lab-configs.yaml
+++ b/config/core/lab-configs.yaml
@@ -4,6 +4,8 @@ labs:
   lab-baylibre:
     lab_type: lava.lava_xmlrpc
     url: 'https://lava.baylibre.com/RPC2/'
+    queue_timeout:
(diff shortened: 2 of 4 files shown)


gctucker:
LGTM, let's wait for some jobs on staging to confirm it's working as expected.

The commit message should have a prefix like `kernelci.config.lab: ` to be consistent with other changes. Keeping the YAML changes in a separate commit would also be good, but that's not a big issue.

config/lava/base/kernel-ci-base.jinja2
```diff
@@ -64,6 +64,10 @@ timeouts:
   actions:
     power-off:
       seconds: 30
+{%- if queue_timeout %}
+  queue:
+    {{ queue_timeout }}
+{% endif %}
```
gctucker, line 70: If you want to control how the YAML is going to be generated then I believe you have to list each possible field from the dictionary, like:
```
{%- if queue_timeout.hours %}
        hours: {{ queue_timeout.hours }}
{% endif %}
```
which isn't great, but at least it explicitly makes it clear how this should be in YAML.  The current implementation just works because the Python string representation of a dictionary is compatible with YAML, but that's mostly luck.  Maybe it could be done by iterating over the key/value pairs of the `queue_timeout` variable to avoid repeating the hours / minutes etc.

montjoie: Doing this could be a way to validate content of queue_timeout (we want basicly only days or hours)

gctucker: Right, well the validation could be done separately from generating a definition. There's already `kci_test validate` for that, basically the code that loads the config should be validating it.

Ideally we should be using something like `jsonschema` to describe the YAML config schema but there are other things to fix first.

gctucker:
It's a shame that `lab-clabbe` and `lab-baylibre` don't have a public web UI as I can't check the job definitions that have been submitted there.

I'll add a PR on top of that to enable timeouts in `lab-collabora` to check it's working as expected.

PR #1054, opened 2022-02-24
test-configs.yaml: enable QEMU Docker in lab-collabora

gctucker:
Enable QEMU Docker jobs in lab-collabora now that is has been upgraded
to LAVA 2021.01.3.

Depends on #1245 which in turn depends on some extra LAVA changes to be merged and deployed in the lab.

PR #1088, opened 2022-03-23
test-configs.yaml: add lab-broonie to QEMU passlist

gctucker:
Add lab-broonie to the passlist for QEMU tests which now requires
Docker support.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -153,7 +153,10 @@ test_plans:
     filters:
       - blocklist: *kselftest_defconfig_filter
       - passlist: &qemu_labs_filter
-          lab: ['lab-baylibre', 'lab-collabora-staging']
+          lab:
+            - 'lab-baylibre'
+            - 'lab-broonie'
+            - 'lab-collabora-staging'
 
   cros-ec:
     rootfs: debian_bullseye-cros-ec_ramdisk


gctucker:
Rebased on top of #1054 to avoid conflicts.

gctucker (commented):
Tested OK on staging:
https://staging.kernelci.org/test/job/kernelci/branch/staging-mainline/kernel/staging-mainline-20220323.0/plan/baseline/

All the baseline tests passed, but this LTP test failed:
https://lava.sirena.org.uk/scheduler/job/36500

It looks like a kernel problem, but there's no other lab running this at the moment so it's hard to tell. The BayLibre lab doesn't run LTP and the Collabora lab has an issue with the network config for Docker so it's not enabled yet.

While it would be good to fix this job failure, I don't think it's a blocker for this PR.

gctucker:
@broonie Does this look like a genuine kernel issue to you?
https://lava.sirena.org.uk/scheduler/job/36500

broonie:
On Wed, Mar 23, 2022 at 11:22:51AM -0700, Guillaume Tucker wrote:
> @broonie Does this look like a genuine kernel issue to you?
> https://lava.sirena.org.uk/scheduler/job/36500

It's a genuine issue - unclear to me without further investigation if
the kernel or the test is at fault (we've run out of memory, but it's
not clear to me if that's just the kernel being configured to need more
memory than the system has, about half the memory is reserved by the
time we print the Memory: line during boot which seems very high).

gctucker:
I see. This may be due to the fact that the LTP image is much larger than the baseline one.

So I think it should be fine to merge all this to enable QEMU tests in your lab, and if there's an issue with LTP jobs that can't be fixed easily (e.g. not enough RAM for all QEMU instances on the dispatcher...) then we could add some filter to not run it.

broonie:
Looking at some other systems with low memory I'm pretty confident it's just the combination of config and rootfs not fitting on the machine - I'm seeing a very similar OOM killer issue once we start running the initramfs.

PR #1089, opened 2022-03-23
test-configs: Enable arm64 kselftests

broonie:
Provide a basic hookup for the arm64 kselftests and enable them for a bunch
of platforms. Other than qemu these platforms are picked mostly at random,
just to ensure that there's a selection of physical implementations. The
one it's most important to cover here is qemu since the arm64 kselftests
exercise bleeding edge architecture features which are generally supported
in emulation well before they are available in physical implementations.
The tests will detect missing features at runtime and cleanly skip tests
that don't apply to the current platform.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -2012,6 +2012,8 @@ test_configs:
   - device_type: bcm2711-rpi-4-b
     test_plans:
       - baseline
+      - baseline-nfs
+      - kselftest-arm64
 
   - device_type: bcm2836-rpi-2-b
     test_plans:
@@ -2366,6 +2368,8 @@ test_configs:
   - device_type: meson-g12b-odroid-n2
     test_plans:
       - baseline
+      - baseline-nfs
+      - kselftest-arm64
       - kselftest-lib
 
   - device_type: meson-gxbb-nanopi-k2
@@ -2401,6 +2405,7 @@ test_configs:
       - baseline
       - baseline-nfs
       - kselftest-alsa
+      - kselftest-arm64
       - kselftest-cpufreq
       - kselftest-futex
       - kselftest-lib
@@ -2761,6 +2766,7 @@ test_configs:
     test_plans:
       - baseline
       - baseline-nfs
+      - kselftest-arm64
       - kselftest-cpufreq
       - kselftest-filesystems
       - kselftest-futex


config/core/test-configs.yaml
```diff
@@ -1863,6 +1869,7 @@ test_configs:
   - device_type: bcm2711-rpi-4-b
     test_plans:
       - baseline
+      - kselftest-arm64
```
gctucker, line 1872: Please also enable baseline-nfs as kselftest jobs use NFS root.

config/core/test-configs.yaml
```diff
@@ -2375,6 +2383,7 @@ test_configs:
   - device_type: qemu_arm64-virt-gicv3
     test_plans:
       - baseline_qemu
+      - kselftest-arm64
```
gctucker, line 2386: There's no support yet for running kselftest on QEMU in KernelCI, so please drop this for now.

config/core/test-configs.yaml
```diff
@@ -2481,6 +2490,7 @@ test_configs:
   - device_type: rk3328-rock64
     test_plans:
       - baseline
+      - kselftest-arm64
```
gctucker, line 2493: @montjoie Please confirm it's OK to enable this and baseline-nfs in BayLibre's lab

config/core/test-configs.yaml
```diff
@@ -2180,6 +2187,7 @@ test_configs:
   - device_type: meson-g12b-odroid-n2
     test_plans:
       - baseline
+      - kselftest-arm64
```
gctucker, line 2190: @montjoie Please confirm it's OK to enable this and baseline-nfs in BayLibre's lab.

config/core/test-configs.yaml
broonie: Shouldn't that be caught by lab filters if there's any issue?

config/core/test-configs.yaml
broonie: Same here...

config/core/test-configs.yaml
gctucker: Deciding which test plans to enable on which platform currently is a grey area between test coverage and lab capacity, so we just have to keep being pragmatic with the current system.

I think a better way forward would be to have some logic to dynamically schedule tests depending on each lab's availability, for example by limiting the number of jobs submitted to a particular device type or by querying the lab to see how long the job queue is. Then it would really become lab configuration, with test configuration only expressing what should ideally be tested.

broonie: Sure, my thought is more that if it's a question of "can we schedule this test in this lab?" that should be being handled by the lab config. An "is it worth running this test on this board vs another board/test" question is a different thing.

gctucker: Yes, the grey area we have at the moment comes from the fact that some boards are only available in a single lab. So adding a test plan to such board can be misleading as it looks like there's some coverage for it but if the tests can't actually be run in that lab then choosing another device type would be more appropriate (if applicable).

The config schema we have now worked very well when we just had basic tests that could (should) just be enabled on all the boards. It's something we have to deal with until we come up with something more appropriate with the growing variety of tests we're adding.

config/core/test-configs.yaml
```diff
@@ -2180,6 +2188,8 @@ test_configs:
   - device_type: meson-g12b-odroid-n2
     test_plans:
       - baseline
+      - baseline-nfs
```
gctucker, line 2191: OK: https://staging.kernelci.org/test/plan/id/623d1494834c683ef252098d/

config/core/test-configs.yaml
```diff
@@ -2481,6 +2491,8 @@ test_configs:
   - device_type: rk3328-rock64
     test_plans:
       - baseline
+      - baseline-nfs
```
gctucker, line 2494: OK: https://staging.kernelci.org/test/plan/id/623d14d00d2f18781752094c/

config/core/test-configs.yaml
gctucker: The job succeeded but there are no arm64 results: https://staging.kernelci.org/test/plan/id/623d178d0e60c006e0520902/

So it looks like there's a problem with running these tests, I haven't investigated any further.

gctucker (commented):
Some results are in, but it's not looking to good.  Adding staging-skip label so we can get the kselftest-alsa tests a run.

config/core/test-configs.yaml
broonie: Yes, looks like a general platform config issue.

config/core/test-configs.yaml
broonie: | No such collection 'arm64'

It looks like the image just doesn't contain the arm64 selftests.

config/core/test-configs.yaml
broonie: Looks like some potentially transient infra thing...

config/core/test-configs.yaml
broonie: It looks like the kselftest build is failing, and silently discarding those tests which fail to build in a non-obvious fashion not helped by our lack of UI around kselftest build failures. There's some issues with the docker containers lagging but the main issue for several of the tests is that the kselftest makefiles don't really cope properly with O= which breaks things (the main issue is with linking multi-object executables).

gctucker: Right, so we probably should drop this PR until the arm64 kselftest build is fixed.  The kselftest make command will only return an error if all the kselftest targets failed, so it's kind of expected that these error go under the radar. We would need to parse the logs or build the targes one by one... Ideally they should all get fixed and then the logic would be to return an error if any of them fails, I think we're getting closer to this with all the fixes currently being added.

gctucker:
@broonie Do you want to fix the arm64 kselftest build issue now or should we close this PR and reopen it when it's fixed?  I'm trying to reduce the number of stale PRs as they're a distraction when reviewing the backlog of things to resolve.

gctucker:
> @broonie Do you want to fix the arm64 kselftest build issue now or should we close this PR and reopen it when it's fixed? I'm trying to reduce the number of stale PRs as they're a distraction when reviewing the backlog of things to resolve.

@broonie Is someone working on a fix for this build error?

broonie:
The next -next build should have fixes for out of tree builds of the arm64 kselftests.

config/core/test-configs.yaml
```diff
@@ -2381,6 +2385,7 @@ test_configs:
       - baseline
       - baseline-nfs
       - kselftest-alsa
+      - kselftest-arm64
```
gctucker, line 2388: OK https://staging.kernelci.org/test/plan/id/627dbb86f9dd35fe39e12bdb/

broonie:
clang kselftest builds are just broken for cross compilation - it doesn't actually cross compile. That's a separate issue that needs to be fixed anyway.

broonie:
The bcm2711-rpi-4-b empty results you were pointing to are from a clang build which is why they're not showing anything.

gctucker:
> The bcm2711-rpi-4-b empty results you were pointing to are from a clang build which is why they're not showing anything.

The GCC builds aren't booting though, probably an issue with the kernel image size and load addresses:
https://lkft.validation.linaro.org/scheduler/job/5037080

There's one in the Collabora lab, I could try to tweak things to make it work.  I think it's fine now to merge this anyway with 2 known issues (kselftest clang build issue and boot problem on rpi-4-b) and try to fix them later on.  If the failing jobs are taking too much time in the lab we could add a PR to disable them.  It would be worth creating GitHub issues to keep track of this.

gctucker:
> Other than qemu these platforms are picked mostly at random,

So we could easily drop rpi-4-b and pick another one instead as a follow-up.  The `mt8173-elm-hana` Chromebook should be able to run all kselftest plans as there's 10 of them online and it's fast enough.

PR #1090, opened 2022-03-23
test-configs: Enable ALSA kselftests

broonie:
Hook up the ALSA kselftests. These do not require any physical setup of the
system so can be run on any device with the relevant hardware, they should
be very quick to run.

Since these tests are system specific I've enabled them for quite a wide
range of systems rather than picking a few per architectures, they will
gracefully handle systems with no sound cards so we could in theory just
enable them everywhere but there's the overhead from booting when run as
an independent testsuite. We may want to pare this back in future.

### DIFF
--- a/config/core/rootfs-configs.yaml
+++ b/config/core/rootfs-configs.yaml
@@ -143,6 +143,7 @@ rootfs_configs:
       - ca-certificates
       - iproute2
       - jdim
+      - libasound2
       - libatm1
       - libcap2-bin
       - libelf1
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -248,6 +248,12 @@ test_plans:
     filters:
       - passlist: {defconfig: ['kselftest']}
 
+  kselftest-alsa:
+    <<: *kselftest
+    params:
+      job_timeout: '10'
+      kselftest_collections: "alsa"
+
   kselftest-cpufreq:
     <<: *kselftest
     params:
@@ -1813,6 +1819,7 @@ test_configs:
     test_plans:
       - baseline
       - baseline-nfs
+      - kselftest-alsa
       - kselftest-lkdtm
       - kselftest-seccomp
       - ltp-fcntl-locktests
@@ -1823,6 +1830,7 @@ test_configs:
     test_plans:
       - baseline
       - baseline-nfs
+      - kselftest-alsa
       - kselftest-filesystems
       - kselftest-futex
       - kselftest-rtc
@@ -1836,6 +1844,7 @@ test_configs:
       - baseline-nfs
       - cros-ec
       - igt-gpu-i915
+      - kselftest-alsa
       - kselftest-filesystems
       - kselftest-futex
       - kselftest-lib
@@ -1857,6 +1866,7 @@ test_configs:
       - baseline
       - baseline-nfs
       - cros-ec
+      - kselftest-alsa
       - kselftest-lib
       - kselftest-vm
       - ltp-ipc
@@ -1981,6 +1991,7 @@ test_configs:
       - baseline-nfs
       - cros-ec
       - igt-gpu-amd
+      - kselftest-alsa
       - kselftest-filesystems
       - kselftest-futex
       - kselftest-lib
@@ -2005,6 +2016,7 @@ test_configs:
       - baseline
       - baseline-nfs
       - cros-ec
+      - kselftest-alsa
       - kselftest-cpufreq
 
   - device_type: hp-x360-12b-ca0500na-n4000-octopus
@@ -2272,6 +2284,7 @@ test_configs:
     test_plans:
       - baseline
       - baseline-nfs
+      # kselftest-alsa - Board capacity in a lab
 
   - device_type: minnowboard-turbot-E3826
     test_plans:
@@ -2323,6 +2336,7 @@ test_configs:
       - baseline
       - baseline-nfs
       - igt-kms-exynos
+      - kselftest-alsa
       - sleep
       - usb
 
@@ -2480,6 +2494,7 @@ test_configs:
     test_plans: &r8a7795-salvator-x_test-plans
       - baseline
       - baseline-nfs
+      # kselftest-alsa - Lab networking issue
 
   - device_type: r8a77950-salvator-x  # same as r8a7795-salvator-x
     test_plans: *r8a7795-salvator-x_test-plans
@@ -2506,6 +2521,7 @@ test_configs:
       - cros-ec
       - igt-gpu-panfrost
       - igt-kms-rockchip
+      # kselftest-alsa - Basic boot issue preventing testing
       - ltp-timers
       - sleep
       - usb
@@ -2513,12 +2529,15 @@ test_configs:
   - device_type: rk3328-rock64
     test_plans:
       - baseline
+      - baseline-nfs
+      # kselftest-alsa - lab stability/availability issue
 
   - device_type: rk3399-gru-kevin
     test_plans:
       - baseline
       - baseline-nfs
       - cros-ec
+      - kselftest-alsa
       - kselftest-rtc
       - ltp-fcntl-locktests
       - ltp-pty
@@ -2546,7 +2565,9 @@ test_configs:
   - device_type: sc7180-trogdor-lazor-limozeen
     test_plans:
       - baseline
+      # baseline-nfs - Current configs lack networking
       - cros-ec
+      # kselftest-alsa - Current configs lack networking
 
   - device_type: snow
     test_plans:


config/core/test-configs.yaml
```diff
@@ -2129,6 +2141,7 @@ test_configs:
     test_plans:
       - baseline
       - baseline-nfs
+      - kselftest-alsa
```
gctucker, line 2144: This platform has only one instance online so there's not enough capacity for running anything else than baseline. Please drop this line.

config/core/test-configs.yaml
```diff
@@ -2242,10 +2255,12 @@ test_configs:
...
 
   - device_type: minnowboard-turbot-E3826
     test_plans:
       - baseline
+      - kselftest-alsa
```
gctucker, line 2263: There isn't enough capacity in the Collabora lab for running anything else than baseline on this platform, partly because other projects run a lot of tests on them. So please drop this line too.

config/core/test-configs.yaml
```diff
@@ -2515,6 +2535,7 @@ test_configs:
     test_plans:
       - baseline
       - cros-ec
+      - kselftest-alsa
```
gctucker, line 2538: There are issues with mainline support for this platform, mainly some bits missing in the arm64 defconfig. So we could leave this here but it might not work for a little while until that's resolved. Most likely, we'll have to add things to the `arm64-chromebook` fragment in KernelCI if the changes aren't accepted upstream.

config/core/test-configs.yaml
```diff
@@ -2658,6 +2679,7 @@ test_configs:
       - baseline
       - cros-ec
       - igt-kms-tegra
+      - kselftest-alsa
```
gctucker, line 2682: Please drop this line, this platform doesn't have the capacity to run anything more.

config/core/test-configs.yaml
```diff
@@ -2481,12 +2499,14 @@ test_configs:
   - device_type: rk3328-rock64
     test_plans:
       - baseline
+      - kselftest-alsa
```
gctucker, line 2502: Please also add baseline-nfs since kselftest uses NFS. This helps with identifying issues that are related to booting with NFS as opposed to kselftest.

@montjoie This platform is only available in BayLibre's lab so please confirm it's OK to enable these tests on it.

config/core/test-configs.yaml
```diff
@@ -2473,6 +2490,7 @@ test_configs:
       - cros-ec
       - igt-gpu-panfrost
       - igt-kms-rockchip
+      - kselftest-alsa
```
gctucker, line 2493: We'll see how this goes, if it adds too much load in production we might have to drop it.

config/core/test-configs.yaml
```diff
@@ -2447,6 +2463,7 @@ test_configs:
     test_plans: &r8a7795-salvator-x_test-plans
       - baseline
       - baseline-nfs
+      - kselftest-alsa
```
gctucker, line 2466: @montjoie This device type is only available in BayLibre's lab, please confirm it's OK to enable these tests on it.

config/core/test-configs.yaml
```diff
@@ -2290,6 +2305,7 @@ test_configs:
       - baseline
       - baseline-nfs
       - igt-kms-exynos
+      - kselftest-alsa
```
gctucker, line 2308: I don't think there are enough XU3 boards online to run any more tests. It's probably better to drop it and rely on other platforms instead.

gctucker (commented):
This is currently conflicting with #1089 (arm64 kselftest) so I think we should try to get that one merged first and then rebase.

config/core/test-configs.yaml
broonie: Shouldn't this be restricted in the lab configuration? It does feel like we've got the lab and board configurations too tightly wound here...

config/core/test-configs.yaml
broonie: Note that as covered in the commit message the tests are device specific so we can't "rely" on other platforms - there's some shared stuff but a test running on one platform is very much not a direct substitute for a test running on another.

gctucker: Is that tied to the SoC or some audio chip on the XU3?

If this platform is really needed then maybe you could enable some in your lab, or see if another lab wants to add some more.  There might be a couple of spares in Collabora's stock but I can't guarantee they could be added to the lab.

gctucker:
Actually this PR is also conflicting with earlier PRs that are adding a bunch of kselftests on Chromebooks. So we should try to resolve the earlier changes first to unblock this situation.

gctucker:
Looking to resolve #1016 first, then this should get unblocked.

config/core/test-configs.yaml
broonie: The audio subsystem is a combination of both - the current tests mainly cover the off SoC components. It's not *needed* exactly any more than any other testing is needed but it would be good to have the coverage.

I really don't think the current approach is working well for these less generic tests, as someone describing a test for KernelCI this munging of the "what can this board do" and "what do we want to schedule at the current time" into one configuration is confusing to think about (especially when I also had to manually enumerate the tests for my lab) and makes it more complicated to remember what can be enabled where. It feels like we should be making more use of lab filters (especially for the ones where you're asking labs about their capacity), have a separate admin filter or have some way to note that a test is valid for a board even if it's not currently being run.

gctucker: Yes, this is all true. I'm not saying this config schema is ideal, far from it. What we need is people with time and skills to improve the software, so if you want to suggest a better way of doing it or even implement it then please do.  In the meantime, we can only use what is available and it does provide a way to run the things we want to run.

We could easily enough add some blocklist filters in lab-configs to not run some tests on some boards while enabling those tests in test-configs. But then we can't easily tell how test coverage is distributed across different boards. The more dynamic approach I mentioned in another comment is what I would suggest instead, a bit like moving from procedural to functional description of what needs to be tested.

broonie: We could at least be leaving stuff present but commented out rather than removing lines, that's at least retaining the information about the boards for later.

gctucker:
@broonie Please rebase now that older kselftest PRs have been merged.

gctucker (commented):
Jobs are being scheduled but no tests are run because:
```
No such collection 'alsa'
```
https://storage.staging.kernelci.org/kernelci/staging-next/staging-next-20220414.0/arm64/defconfig+kselftest/gcc-10/lab-baylibre/kselftest-alsa-r8a77950-salvator-x.html#L1798

From the build log:
```
/usr/lib/gcc-cross/aarch64-linux-gnu/10/../../../../aarch64-linux-gnu/bin/ld: mixer-test.c:(.text+0x6b0): undefined reference to `snd_config_top'
```
https://storage.staging.kernelci.org/kernelci/staging-next/staging-next-20220414.0/arm64/defconfig+kselftest/gcc-10/logs/kselftest.log

So I guess the Docker image is missing some -dev package, even though `libasound2-dev` is already included.

broonie:
The problem is that whatever pkg-config is installed in the container appears to be broken:

| /usr/bin/pkg-config: 1: Syntax error: "(" unexpected
| /usr/bin/pkg-config: 1: Syntax error: "(" unexpected

For some reason our docker containers install the architecture version of pkg-config but they've not set up the machinery to allow foreign executables to be executed.

broonie:
https://github.com/kernelci/kernelci-core/pull/1123

gctucker:
> #1123

Yes thanks, that fixed the build:
```
make[1]: Entering directory '/tmp/kci/linux/tools/testing/selftests/alsa'
ccache aarch64-linux-gnu-gcc     mixer-test.c -lasound -o /tmp/kci/linux/build/kselftest/alsa/mixer-test
make[1]: Leaving directory '/tmp/kci/linux/tools/testing/selftests/alsa'
```
https://storage.staging.kernelci.org/kernelci/staging-next/staging-next-20220414.0/arm64/defconfig+kselftest/gcc-10/logs/kselftest.log

Now I wonder how we never noticed this problem before. Anyway, maybe other things were also failing to build and got fixed as a side-effect.

config/core/rootfs-configs.yaml
```diff
@@ -143,6 +143,7 @@ rootfs_configs:
       - ca-certificates
       - iproute2
       - jdim
+      - libasound2
```
gctucker, line 146: Building a set of rootfs images on staging with this included:
https://bot.staging.kernelci.org/job/rootfs-build-trigger/115/

gctucker: They will be used on staging with this temporary PR:
https://github.com/kernelci/kernelci-core/pull/1128

broonie:
We seem to have a general problem where we don't notice build errors in kselftest (see also the arm64 issue that's waiting for a patch upstream) that has registered for me to look at.

gctucker:
> We seem to have a general problem where we don't notice build errors in kselftest (see also the arm64 issue that's waiting for a patch upstream) that has registered for me to look at.

Yes that's how kselftest builds work. If at least one subset managed to build then the exit code is 0. To detect errors, we need to add some logic by parsing the logs or having a list of subsets to expect.

config/core/test-configs.yaml
```diff
@@ -1813,6 +1819,7 @@ test_configs:
     test_plans:
       - baseline
       - baseline-nfs
+      - kselftest-alsa
```
gctucker, line 1822: OK: https://staging.kernelci.org/test/plan/id/6259565a70f1dd2db505f4de/

config/core/test-configs.yaml
```diff
@@ -1823,6 +1830,7 @@ test_configs:
...
```
gctucker, line 1833: OK: https://staging.kernelci.org/test/plan/id/625948e331b735880805f4c7/

config/core/test-configs.yaml
```diff
@@ -1836,6 +1844,7 @@ test_configs:
       - baseline-nfs
       - cros-ec
       - igt-gpu-i915
+      - kselftest-alsa
```
gctucker, line 1847: OK: https://staging.kernelci.org/test/plan/id/62598ff18f6776f03e05f4c4/

config/core/test-configs.yaml
```diff
@@ -1857,6 +1866,7 @@ test_configs:
       - baseline
       - baseline-nfs
       - cros-ec
+      - kselftest-alsa
```
gctucker, line 1869: OK: https://staging.kernelci.org/test/plan/id/625944ea9b6d8f722205f4c4/

config/core/test-configs.yaml
```diff
@@ -1981,6 +1991,7 @@ test_configs:
       - baseline-nfs
       - cros-ec
       - igt-gpu-amd
+      - kselftest-alsa
```
gctucker, line 1994: OK: https://staging.kernelci.org/test/plan/id/62595905830adffeb005f4e5/

config/core/test-configs.yaml
```diff
@@ -2005,6 +2016,7 @@ test_configs:
       - baseline
       - baseline-nfs
       - cros-ec
+      - kselftest-alsa
```
gctucker, line 2019: OK: https://staging.kernelci.org/test/plan/id/62593bc9edef2115f505f4d9/

config/core/test-configs.yaml
```diff
@@ -2272,6 +2284,7 @@ test_configs:
     test_plans:
       - baseline
       - baseline-nfs
+      - kselftest-alsa
```
gctucker, line 2287: Please drop this one, there isn't enough board capacity to run anything else than baseline.

gctucker: It did pass the tests though: https://staging.kernelci.org/test/plan/id/6259359ece93c2099905f4fd/

config/core/test-configs.yaml
gctucker: OK: https://staging.kernelci.org/test/plan/id/6259ace1cc7f4bcbf905f4c4/

I suggest we keep this one enabled for now and see next week in production whether it causes any congestion.

config/core/test-configs.yaml
gctucker: @montjoie @broonie No news and it's failing to boot now:
https://staging.kernelci.org/test/plan/id/62594298fa241475d805f521/

So I suggest we drop it for now.

config/core/test-configs.yaml
gctucker: Failed to boot: https://staging.kernelci.org/test/plan/id/6259450bd7075bb35505f4e6/

I suggest we drop this one too.

config/core/test-configs.yaml
gctucker: @montjoie @broonie No news and no results:
https://staging.kernelci.org/test/job/kernelci/branch/staging-mainline/kernel/staging-mainline-20220415.0/plan/kselftest-alsa/

So I suggest we drop this now.

config/core/test-configs.yaml
```diff
@@ -2506,19 +2521,23 @@ test_configs:
...
       - baseline
+      - baseline-nfs
+      - kselftest-alsa
 
   - device_type: rk3399-gru-kevin
     test_plans:
       - baseline
       - baseline-nfs
       - cros-ec
+      - kselftest-alsa
```
gctucker, line 2540: OK with some tests failing: https://staging.kernelci.org/test/plan/id/62594dea2ab208060405f4c4/

config/core/test-configs.yaml
gctucker: OK: https://staging.kernelci.org/test/plan/id/625944fba216fa032205f531/

config/core/test-configs.yaml
```diff
@@ -2546,7 +2565,9 @@ test_configs:
   - device_type: sc7180-trogdor-lazor-limozeen
     test_plans:
       - baseline
+      - baseline-nfs
```
gctucker, line 2568: This is failing for now as some config options are missing in the arm64-chromebook fragment (see #971):
https://staging.kernelci.org/test/plan/id/62593d43139f8f30b905f4c8/

So we could either drop it from this PR or wait until the missing configs have been added and the whole thing has been tested, as I believe there were other issues too in mainline with this platform.

gctucker (commented):
Some are producing useful results, others are not able to run the tests for various reasons. I think we could drop the ones I've suggested to get this merged quickly and create follow-up PRs for the ones with issues that can be resolved.

gctucker (approved):
OK this should be good to go now.

We might want to create GitHub issues to keep track of resolving the commented-out entries.  This should get reviewed and improved in the coming months anyway.

PR #1091, opened 2022-03-23
test-configs: Add at91sam9g20ek

broonie:
This board (a very old out of production AT91 reference system) is in my
lab.

The board only has 64M of RAM so is very restricted in terms of what it can
run from ramdisk but it has ethernet and is quite capable with NFS and is
a useful example of it's generation of audio hardware. With our current
setup it struggles to boot a multi_v5_defconfig in an initramfs due to the
size of the modules causing us to run out of memory, we may need to
explicitly block that.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -610,7 +610,6 @@ device_types:
 
   at91sam9g20ek:
     mach: at91
-    arch: armel
     class: arm-dtb
     boot_method: uboot
 
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -608,6 +608,12 @@ device_types:
     class: arm-dtb
     boot_method: uboot
 
+  at91sam9g20ek:
+    mach: at91
+    arch: armel
+    class: arm-dtb
+    boot_method: uboot
+
   bcm2711-rpi-4-b:
     mach: broadcom
     class: arm64-dtb
@@ -1857,6 +1863,10 @@ test_configs:
       - baseline
 
   - device_type: at91-sama5d4_xplained
+    test_plans:
+      - baseline
+
+  - device_type: at91sam9g20ek
     test_plans:
       - baseline
 


config/core/test-configs.yaml
```diff
@@ -608,6 +608,12 @@ device_types:
     class: arm-dtb
     boot_method: uboot
 
+  at91sam9g20ek:
+    mach: at91
+    class: arm-dtb
+    boot_method: uboot
+    dtb: "at91sam9g20ek.dtb"
```
gctucker, line 615: If the dtb name matches the device type name `at91sam9g20ek` then it shouldn't be necessary to specify it here.

config/core/test-configs.yaml
```diff
@@ -1860,6 +1866,13 @@ test_configs:
     test_plans:
       - baseline
 
+  - device_type: at91sam9g20ek
+    test_plans:
+      - baseline
+      - baseline-nfs
+    filters:
+      - passlist: {defconfig: ['at91_dt_defconfig']}
```
gctucker, line 1874: Building this extra defconfig on staging now to then get a set of tests scheduled:
https://bot.staging.kernelci.org/job/kernel-build/57083/consoleText

gctucker:
The device type doesn't appear to be in upstream LAVA:
https://git.lavasoftware.org/lava/lava/-/tree/master/etc/dispatcher-config/device-types

config/core/test-configs.yaml
```diff
@@ -1860,6 +1866,13 @@ test_configs:
     test_plans:
       - baseline
 
+  - device_type: at91sam9g20ek
+    test_plans:
+      - baseline
+      - baseline-nfs
```
gctucker, line 1872: It failed to boot:
https://lava.sirena.org.uk/scheduler/job/36823

config/core/test-configs.yaml
```diff
@@ -1860,6 +1866,13 @@ test_configs:
     test_plans:
       - baseline
 
+  - device_type: at91sam9g20ek
+    test_plans:
+      - baseline
```
gctucker, line 1871: Tested OK:
https://lava.sirena.org.uk/scheduler/job/36824
https://staging.kernelci.org/test/plan/id/623c40306b1bb198125208fb/

config/core/test-configs.yaml
broonie: Testsuite bug - I suspect the NFS root filesystem is a v7 one but this is an older machine that doesn't support that ISA.

gctucker:
> The device type doesn't appear to be in upstream LAVA: https://git.lavasoftware.org/lava/lava/-/tree/master/etc/dispatcher-config/device-types

There is one now: https://git.lavasoftware.org/lava/lava/-/merge_requests/1709

gctucker:
> > The device type doesn't appear to be in upstream LAVA: https://git.lavasoftware.org/lava/lava/-/tree/master/etc/dispatcher-config/device-types
> 
> There is one now: https://git.lavasoftware.org/lava/lava/-/merge_requests/1709

Merged now

gctucker:
The `at91_dt_defconfig` build is there but no test results:
https://staging.kernelci.org/build/id/62402075cd37f48937520924/

This PR is deployed on staging though, so I don't know why that is.  Maybe the build job took too long to complete in Kubernetes and Jenkins timed out waiting for it. Let's look at the next staging results.

gctucker:
Tested OK with a job definition generated by hand after dropping the `armel` attribute from the device config:
https://lava.sirena.org.uk/scheduler/job/46366

Will now wait for the next round of staging jobs to confirm this is working as expected.  Then we can try to switch to `multi_v5_defconfig` and see if it also works.  Using it might be a little bit better as it's more generic.

config/core/test-configs.yaml
gctucker: Tested OK (again, with the fix to drop `arch: armel`):
https://lava.sirena.org.uk/scheduler/job/47254
https://staging.kernelci.org/test/plan/id/6246b66fa57b0ef47b05f4c6/

gctucker (approved):
This can now be merged as baseline is using the buildroot armel images.  Support for Debian armel can be added as a follow up to enable kselftest (might require some mapping between architecture names in YAML).

PR #1093, opened 2022-03-24
test-configs.yaml: enable kselftest-arm64 on hana Chromebook

gctucker:
Enable kselftest-arm64 on mt8173-elm-hana.

Depends on #1089.


gctucker (commented):
Tested OK by hand in LAVA.

PR #1097, opened 2022-03-25
rootfs-configs: Enable armel kselftest builds

broonie:
With the pending addition of AT91SAM9G20-EK which is ARMv5 based we need
an armel build of kselftest in order to allow us to run kselftests on
that platform.

### DIFF
--- a/config/core/rootfs-configs.yaml
+++ b/config/core/rootfs-configs.yaml
@@ -163,6 +163,7 @@ rootfs_configs:
     arch_list:
       - amd64
       - arm64
+      - armel
       - armhf
     extra_packages:
       - bc


Lakshmipathi:
LGTM. If possible, please add Jenkins build rootfs job url.

gctucker:
rootfs image built:
https://storage.staging.kernelci.org/images/rootfs/debian/bullseye-kselftest/20220331.1/armel/

Next step is to run a kselftest job on the AT91SAM9G20-EK with it to confirm it's working as expected.  This should be done by hand as the YAML config is currently tied to the kernel architecture names only (i.e. only `arm`).

gctucker:
@broonie Finally got this running, but then it hit OOM pretty quickly:
https://lava.sirena.org.uk/scheduler/job/50665

Feel free to resubmit this job by hand if you want to investigate.  It won't be run automatically from staging yet because of the armel arch issue.  Also this is using a one-off `at91_dt_defconfig+kselftest` kernel built I triggered by hand in Jenkins as that's not included in the regular builds either.

broonie:
I guess that's the kselftest fragment pushing things over the edge - wouldn't 100% surprise me. Do we already prune the kselftest fragments to those for the kselftests we actually have setup to run? The first thought would be to run the kselftests with no fragments or fragments already in the base config against that base config - that'd get us some coverage without bloating the number of builds and is possibly good practice anyway while we're running each kselftest target in a separate job.

broonie:
BTW this does all seem a bit beyond the scope of the original pull request to just get the binaries built and enable such experiments.

gctucker:
This was to see whether the armel rootfs image actually worked with this platform, as otherwise there's no point building it.

The kselftest builds include all the kselftest fragments, so we have one build for all kselftest jobs.  This works on all other platforms and I don't think it's going to be practical to do some tailored builds just for this one.

broonie:
Well, there's a recursive thing where we can't look at enabling things if there's no images. There are other v5 platforms (versatilepb in qemu if nothing else).

For the additional configurations my first suggestion was to run the kselftests with the existing plain defconfig builds - that wouldn't mean any extra builds, just using the existing builds.

broonie:
Oh, and the other thing with using at91_dt_defconfig as the base is that it doesn't have all the bits needed for systemd - multi_v5_defconfig does though. That doesn't seem to be the issue here but it might cause trouble further down the line.

broonie:
That rootfs seems to run with a multi_v5_defconfig:

https://lava.sirena.org.uk/scheduler/job/51427

although it looks like the test wrapper stuff needs some education about older CPUs:

| + ./kselftest.sh -c rtc -T '' -t kselftest_armhf.tar.gz -s false -u http://storage.staging.kernelci.org/kernelci/staging-mainline/staging-mainline-20220404.0/arm/at91_dt_defconfig+kselftest/gcc-10/kselftest.tar.xz -L '' -S skipfile-lkft.yaml -b at91sam9g20ek -g kernelci -e '' -p /opt/kselftests/mainline/
| INFO: Generating a skipfile based on /lava-51437/0/tests/1_kselftest-rtc/automated/linux/kselftest/skipfile-lkft.yaml
| ERROR: Unsupported architecture: armv5tejl

Of the tests we support already we should be able to run at the very least the RTC tests with this defconfig, the board has an RTC which is supported here. The ALSA tests will run too once they're enabled.

gctucker:
So it sounds like the next step would be to check whether `kselftest-rtc` and `kselftest-alsa` work on at least one `armel` platform (I believe there is only `at91sam9g20ek` in KernelCI at the moment).

broonie:
We also need https://github.com/kernelci/kernelci-core/pull/1113 for the rootfs before we can test on platforms. The at91sam9g20ek has ALSA and RTC, and this is also suitable for the bcm2835-rpi I've got another pull request open for, it can't run the armhf rootfs so needs armel.

broonie:
Oh, and we'll also need to fix the kselftest build so the pre-v7 builds are done with an armel toolchain - at the minute we just use the armhf toolchain so the kselftest binaries won't actually run on older systems.

gctucker:
Building a new armel bullseye-kselftest image since the one from March has been garbage-collected:
https://bot.staging.kernelci.org/job/rootfs-builder/415/console

gctucker:
> Oh, and we'll also need to fix the kselftest build so the pre-v7 builds are done with an armel toolchain - at the minute we just use the armhf toolchain so the kselftest binaries won't actually run on older systems.

Right, so we need a new Dockerfile for `gcc-10_armel`. Could you please make a PR for that?

gctucker:
We'll also need some logic to pick that image for early arm kernel defconfigs but that can be dealt with separately I think. It's somewhat related to #1113.

broonie:
Too late... https://github.com/kernelci/kernelci-core/pull/1273

broonie:
Renamed. It seems to work, and obviously Debian have tested the actual binaries much more extensively.

gctucker:
@broonie I've built an armv5 kernel including kselftests, it's booting with the armel rootfs image but then I get this "illegal instruction" error:
https://lava.sirena.org.uk/scheduler/job/247565#L949

The kselftest files seem to be for the correct architecture:
```
kernelci@kernelci2:/var/www/html/gtucker/armv5/kselftest$ file timers/leap-a-day
timers/leap-a-day: ELF 32-bit LSB shared object, ARM, EABI5 version 1 (SYSV), dynamically linked, interpreter /lib/ld-linux.so.3, BuildID[sha1]=ecb96b37ca55efc9e5a1b4e079d449d56d6a1757, for GNU/Linux 3.2.0, not stripped
```

I don't think the test-definitions repository has any arch-specific code (binaries...) but I'll double check that.

Otherwise, can you think of anything that can be causing this? Debugging it in LAVA might be trick as the error message seems pretty vague.

gctucker:
Might be this line from test-deifinitions:
```
../../bin/${abi}/skipgen ${SKIPGEN_ARGS} "${SKIPFILE_YAML}" > "${SKIPFILE_PATH}"
```
Trying to find out where `skipgen` is coming from.

gctucker:
This looks fine, but I guess there must be something slightly wrong somewhere:
```
$ file automated/bin/armeabi/skipgen 
automated/bin/armeabi/skipgen: ELF 32-bit LSB executable, ARM, EABI5 version 1 (GNU/Linux), statically linked, no section header
```

broonie:
test-definitions for some reason downloads the skipgen binary from https://github.com/Linaro/skipgen/releases which has no armv5 binaries, it does have armv6 but that's not v5.

gctucker:
Doh, `skipgen` is in Go...  I hope there's ARMv5 support for that.  It's pretty trivial though, could be rewritten easily in Python or C I guess.

gctucker:
@broonie Hello, how would you like to resolve this?  Get a Go version with ARMv5 support? Fix test-definitions to reimplement skipgen not in Go? Try to work around this by getting a hard-coded list of skips for ARMv5?

broonie:
The simplest thing is to just not run a skiplist at all on armv5, the kselftest skiplist is essentially empty anyway and nothing I'm trying to schedule needs one.

gctucker:
@broonie OK, is this something you can take care of?

broonie:
Probably eventually - there's some other dependency that was stuck in limbo for ever IIRC.

gctucker:
How about closing this PR for now while all the other dependencies get resolved?  Having stale PRs is what's causing congestion in the reviews and makes it hard to merge others.

broonie:
I think it's near enough and there's nothing wrong with this PR itself, any changes will be orthogonal.

I have limited bandwidth to try to push through all the stop energy on the rootfs stuff and there's nothing particularly wrong with this PR as is. The main issue right now is that any time I want to look at it needs rootfs images building which means it gets stuck in the enormous logjam that is trying to get rootfs images built and the bodges applied to run tests with them.

gctucker:
The issue is that this PR has been open for about one year now, so while it might not be causing problems to have it here on the other hand it's contributing to the congestion we have with reviewing PRs.  So unless something is going to unblock this in the next couple of weeks, closing it now and re-opening it when it can actually be merged, or creating a new one then would be helpful for the review process imho.

gctucker:
Yes this needs to be improved, and anyone is welcome to help, but I don't think that makes any difference to the point about keeping stale PRs for months.  Closing stale PRs means more time for maintainers who then don't have to come back and look at them again, and more time to improve things such as automated rootfs builds for staging.

broonie:
AFAICT we already have enough results above to be confident that the rootfs itself is fine (and I've tested my own builds of it outside of kernelci quite happily), it's hard to tell what we'd gain from additional testing. We've demonstrated that the filesystem builds and boots, it's not like we cover every single kselftest for every change, the chances of some architecture specific breakage seem minimal and it's unlikely to be disruptive to anything else if there is any.

Meanwhile closing this PR or arguing about closing it both just consume additional time.

gctucker:
Please tell me which bit I'm getting wrong:

This PR enables the kselftest rootfs builds for `armel`.  If we try to run kselftest on `armel`, it fails with the error mentioned above in `skipgen`.  So we can't use this rootfs image until that's resolved.  So until then, there's no point in building the rootfs image and therefore no point in merging this PR.

broonie:
Right, we can't use the image without resolving the runtime issue with skipgen. The purpose in building the image would be to both cut down on the backlog of PRs (which you're actively pushing for, prompting this discussion) and reduce the hassle with working on skipgen caused by having to do the manual builds and bodge PRs at the expense of running an extra rootfs build every week (which seems doable). Personally I suspect that'd be a good tradeoff, though just leaving the PR open until everything else is sorted would also be fine.

broonie:
I guess - it'd certainly help in those 4 weeks (well, 2 or 3 by the time the image materialises with the latency).

gctucker:
Merging now after opening #1808 to keep track of this issue around `skipgen`.

PR #1103, opened 2022-03-31
CHROMEOS build-configs-chromeos.yaml: enable arm64-chromebook fragment

gctucker:
Add the arm64-chromebook fragment to all Chrome OS configs in the same
way that the x86-chromebook fragment was previously enabled.  This is
required to boot some Chromebook devices with serial console and other
tweaks that aren't in the upstream arm64 defconfig.


mgalka (approved):
LGTM

gctucker:
Tested OK: https://chromeos.kernelci.org/build/next/branch/master/kernel/next-20220331/

PR #1111, opened 2022-04-06
rootfs: parameterize the linux-firmware repo version

hardboprobot:
The config/rootfs/debos/scripts/install-firmware.sh script hardcodes the
commit to fetch from the linux-firmare repo. This changeset implements a
"linux_fw_version" key for rootfs configs that define
"extra_firmware". If defined (as a commit hash, tag or branch), it
overrides the commit originally defined in the script.

### DIFF
--- a/config/rootfs/debos/rootfs.yaml
+++ b/config/rootfs/debos/rootfs.yaml
@@ -60,11 +60,7 @@ actions:
   - action: run
     description: add firmware files
     chroot: false
-    {{ if $linux_fw_version }}
-    script: scripts/install-firmware.sh -v {{ $linux_fw_version }} -f {{ $extra_firmware }}
-    {{ else}}
-    script: scripts/install-firmware.sh -f {{ $extra_firmware }}
-    {{ end }}
+    script: scripts/install-firmware.sh -f {{ $extra_firmware }} {{ if $linux_fw_version }} -v {{ $linux_fw_version }} {{ end }}
 {{ end }}
 
   - action: run
--- a/config/rootfs/debos/rootfs.yaml
+++ b/config/rootfs/debos/rootfs.yaml
@@ -2,6 +2,7 @@
 {{- $basename := or .basename "." -}}
 {{- $extra_packages := or .extra_packages "" -}}
 {{- $extra_firmware := or .extra_firmware "" -}}
+{{- $linux_fw_version := or .linux_fw_version "" -}}
 {{- $suite := or .suite "buster" -}}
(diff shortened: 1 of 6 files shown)


config/rootfs/debos/rootfs.yaml
```diff
@@ -59,7 +60,11 @@ actions:
   - action: run
     description: add firmware files
     chroot: false
-    script: scripts/install-firmware.sh {{ $extra_firmware }}
+    {{ if $linux_fw_version }}
+    script: scripts/install-firmware.sh -v {{ $linux_fw_version }} -f {{ $extra_firmware }}
+    {{ else}}
+    script: scripts/install-firmware.sh -f {{ $extra_firmware }}
```
gctucker, line 66: Maybe move `{{ if $linux_fw_version }}` to the command line itself to avoid repeating it? Something like:
```
script: scripts/install-firmware.sh -f {{ $extra_firmware }}{{ if $linux_fw_version }} -v {{ $linux_fw_version }} {{ end }}
```

gctucker (commented):
How about adding a patch to use this in `rootfs-configs.yaml`?  See also my comment about using the `master` branch by default in the script.

gctucker:
> Which use case do you mean?

Yes I mean for the rootfs images built for KernelCI, as having a command line argument with a default value is only really useful if the script is used for another purpose outside of KernelCI.

gctucker:
Building `bullseye-igt` on staging to check it's working as expected:
https://bot.staging.kernelci.org/job/rootfs-build-trigger/117/

gctucker:
And also bullseye-cros-ec: https://bot.staging.kernelci.org/job/rootfs-builder/390/

gctucker:
Temporary draft PR to use the staging rootfs URLs: https://github.com/kernelci/kernelci-core/pull/1174

gctucker:
Tested OK with bullseye-igt:
https://staging.kernelci.org/test/plan/id/627a0bf002e0f47524e12c37/

The arm64 rootfs image was missing so I'll come back later to double check it with bullseye-cros-ec.

gctucker:
Tested OK with bullseye-cros-ec on arm64:
https://staging.kernelci.org/test/plan/id/627a2022b79de4fccce12bce/

Some test results are missing but that's due to some serial console noise:
https://lava.collabora.co.uk/scheduler/job/6320223

Using `/dev/kmsg` would help but that's not related to this PR.

gctucker (approved):

PR #1112, opened 2022-04-06
Lab priority ranges

broonie:
LAVA supports numerical priorities for jobs from 0 to 100. Currently we only allow a single priority to be configured for all KernelCI jobs in a lab, in order to allow labs to give us more flexibility in how we schedule jobs allow a minimum and maximum priority to be configured within which we can schedule jobs.

We then map this range onto a percentage priority configurable in each test plan, defaulting to 0 so all jobs are assigned the lowest priority allowed by the lab.

For robustness if only one of minimum or maximum is specified we set both minimum and maximum to a single value. For simplicity and robustness with people updating their configurations if the existing priority field is configured then we still retain it, if no range is specified it will be the minimum and maximum. The string based priorities low, medium and high map onto 0, 50 and 100 respectively as they do in LAVA.

It is likely to be desirable to extend this in future to allow more prioritisation based on the tree being tested as well, and possibly to make more dynamic use of the priority range (eg, favoring tests that have recently been unstable to make sure they get covered). Hopefully other lab types will also be able to map the percentage based prioritisation onto any similar features they have.

### DIFF
--- a/kernelci/config/lab.py
+++ b/kernelci/config/lab.py
@@ -67,14 +67,20 @@ class LabAPI(Lab):
 
 class Lab_LAVA(LabAPI):
 
-    def __init__(self, priority='medium', queue_timeout=None, *args, **kwargs):
+    def __init__(self, priority_min=50, priority_max=50,
+                 queue_timeout=None, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self._priority = priority
+        self._priority_min = priority_min
+        self._priority_max = priority_max
         self._queue_timeout = queue_timeout
 
     @property
-    def priority(self):
-        return self._priority
+    def priority_min(self):
+        return self._priority_min
+
+    @property
+    def priority_max(self):
+        return self._priority_max
 
     @property
     def queue_timeout(self):
@@ -84,7 +90,27 @@ class Lab_LAVA(LabAPI):
     def from_yaml(cls, lab, kw):
         priority = lab.get('priority')
         if priority:
-            kw['priority'] = priority
+            if priority == 'low':
+                priority = 0
+            elif priority == 'medium':
+                priority = 50
+            elif priority == 'high':
+                priority = 100
+            else:
+                priority = int(priority)
+
+            # If min/max are specified these will be overridden
+            kw['priority_min'] = priority
+            kw['priority_max'] = priority
+
+        # 0 is a valid value so explicitly check for None
+        priority_min = lab.get('priority_min')
+        if priority_min is not None:
+            kw['priority_min'] = int(priority_min)
+        priority_max = lab.get('priority_max')
+        if priority_max is not None:
+            kw['priority_max'] = int(priority_max)
+
         queue_timeout = lab.get('queue_timeout')
         if queue_timeout:
             kw['queue_timeout'] = queue_timeout
--- a/kernelci/lab/lava/__init__.py
+++ b/kernelci/lab/lava/__init__.py
@@ -35,12 +35,26 @@ class LavaAPI(LabAPI):
             print("Template not found: {}".format(template_file))
             return None
         base_name = params['base_device_type']
+
+        # Scale the job priority (from 0-100) within the available levels
+        # for the lab, or use the lowest by default.
+        if 'priority' in plan_config.params:
+            priority = plan_config.params['priority']
+            if priority > 100:
+                priority = 100
+        else:
+            priority = 0
+
+        prio_range = self.config._priority_max - self.config._priority_min
+        priority = int(((priority * prio_range) / 100) +
+                       self.config._priority_min)
+
         params.update({
             'template_file': template_file,
-            'priority': self.config.priority,
             'queue_timeout': self.config.queue_timeout,
             'lab_name': self.config.name,
             'base_device_type': self._alias_device_type(base_name),
+            'priority': priority,
         })
         if callback_opts:
             self._add_callback_params(params, callback_opts)
--- a/config/core/lab-configs.yaml
+++ b/config/core/lab-configs.yaml
@@ -17,7 +17,8 @@ labs:
   lab-broonie:
     lab_type: lava.lava_xmlrpc
     url: 'https://lava.sirena.org.uk/RPC2/'
-    priority: low
+    priority_min: 0
+    priority_max: 40
     filters:
       - passlist:
           plan:
--- a/config/core/lab-configs.yaml
+++ b/config/core/lab-configs.yaml
@@ -115,7 +115,8 @@ labs:
   lab-linaro-lkft:
     lab_type: lava.lava_rest
     url: 'https://lkft.validation.linaro.org/'
-    priority: low
+    priority_min: 0
+    priority_max: 25
     queue_timeout:
       days: 1
     filters:
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -125,6 +125,8 @@ test_plans:
 
   baseline:
     rootfs: buildroot-baseline_ramdisk
+    params:
+      priority: 90
     filters:
       - blocklist: *kselftest_defconfig_filter
 
--- a/kernelci/lab/__init__.py
+++ b/kernelci/lab/__init__.py
@@ -76,7 +76,8 @@ class LabAPI:
         return self.config.match(filter_data)
 
     def generate(self, params, device_config, plan_config,
-                 callback_opts=None, templates_path=None, db_config=None):
+                 callback_opts=None, templates_path=None, db_config=None,
+                 lab_config=None):
         """Generate a test job definition.
 
         *params* is a dictionary with the test parameters which can be used
@@ -95,6 +96,9 @@ class LabAPI:
 
         *db_config* is a Database configuration object for the database or API
             where the results should be sent
+
+        *lab_config* is a configuration object for the API
+            where the tests should be run
         """
         raise NotImplementedError("Lab.generate() is required")
 


kernelci/config/lab.py
```diff
@@ -84,7 +90,25 @@ def queue_timeout(self):
     def from_yaml(cls, lab, kw):
         priority = lab.get('priority')
         if priority:
-            kw['priority'] = priority
+            if priority == 'low':
+                priority = 0
+            elif priority == 'medium':
+                priority = 50
+            elif priority == 'high':
+                priority = 100
```
gctucker, line 98: We could probably just drop the old-style low/medium/high keywords and only rely on numeric values in the YAML config to simplify this.

kernelci/test.py
```diff
@@ -175,4 +176,19 @@ def _get_compression(url):
...
+            priority = 100
+    else:
+        priority = 0
+
+    prio_range = lab_config.priority_max - lab_config.priority_min
+    priority = int(((priority * prio_range) / 100) + lab_config.priority_min)
+    params.update({
+        'priority': priority,
+    })
+
```
gctucker, line 193: Until now, priorities were treated as a LAVA-specific properties. If the lab config is not a LAVA one it won't have `priority_max` / `priority_min` defined, and other types of "labs" (local shell, k8s...) may have a very different way of managing priorities or none at all. So I think it's best to keep this in `LavaAPI.generate()` rather than here.  Or we may start having lab-specific implementations of `.get_params()`, to call the base one and then update the params dictionary, but that would mean more changes to the code.

kernelci/lab/lava/__init__.py
```diff
@@ -37,7 +37,6 @@ def generate(self, params, device_config, plan_config, callback_opts=None,
         base_name = params['base_device_type']
         params.update({
             'template_file': template_file,
-            'priority': self.config.priority,
```
gctucker, line 40: See my other comment below - I believe the logic to work out the priority value belongs here as it's specific to LAVA labs.

gctucker (commented):
Sounds good, just a few stylistic things to consider.

It would also be good to update the documentation once this is all done, probably in `doc/config/test-configs.md`.

I guess this would also be a first step in adding criteria for dynamically scheduling tests: the plan's priority could be used to determine whether to run it at all, or on how many platforms to run it.

kernelci/config/lab.py
broonie: We could, but I didn't want to needlessly make configuration files for other deployments instabuggy on upgrade (especially with the sometimes shaky error reporting we have) and it seems helpful for talking to lab owners given that numerical priorities aren't that widely used.

The other use with dynamic scheduling is that if we use baseline test results (or other smoke test type things) to decide what else to schedule then it could be beneficial to get them in quickly so that anything we key off them gets scheduled more promptly.

kernelci/lab/lava/__init__.py
```diff
@@ -27,20 +27,35 @@
...
+        # Scale the job priority (from 0-100) within the available levels
+        # for the lab, or use the lowest by default.
+        if 'priority' in plan_config.params:
+            priority = plan_config.params['priority']
+            if priority > 100:
+                priority = 100
+        else:
+            priority = 0
+
+        prio_range = lab_config.priority_max - lab_config.priority_min
```
gctucker, line 49:
```
  File "/data/workspace/bot.staging.kernelci.org/test-runner/workspace/test-runner__2/kernelci-core/./kci_test", line 301, in <module>
  File "/data/workspace/bot.staging.kernelci.org/test-runner/workspace/test-runner__2/kernelci-core/./kci_test", line 258, in __call__
    job = lab_api.generate(params, device_config, plan_config,
  File "/data/workspace/bot.staging.kernelci.org/test-runner/workspace/test-runner__2/kernelci-core/kernelci/lab/lava/__init__.py", line 49, in generate
    prio_range = lab_config.priority_max - lab_config.priority_min
TypeError: unsupported operand type(s) for -: 'str' and 'str'
```
https://bot.staging.kernelci.org/job/test-runner/28034/console

gctucker (changes requested):
This is only working with `lab-broonie`, I guess there's a problem when the min/max range is not defined.
https://staging.kernelci.org/test/job/kernelci/branch/staging-next/kernel/staging-next-20220413.0/

broonie:
Bah, I tested that config but it was broken by the refactoring you asked for.

gctucker:
The problem is still there: https://bot.staging.kernelci.org/job/test-runner/28202/console

kernelci/config/lab.py
```diff
@@ -84,7 +90,25 @@ def queue_timeout(self):
...
+
+            # If min/max are specified these will be overridden
+            kw['priority_min'] = priority
+            kw['priority_max'] = priority
+
+        # 0 is a valid value so explicitly check for None
+        priority_min = lab.get('priority_min')
+        if priority_min is not None:
+            kw['priority_min'] = priority_min
```
gctucker, line 107: I haven't tested this but it looks like the YAML values are always parsed as strings even if they're only digits, in which case they would need to be converted to integer here:
```suggestion
            kw['priority_min'] = int(priority_min)
```

kernelci/config/lab.py
```diff
@@ -84,7 +90,25 @@ def queue_timeout(self):
...
+        priority_max = lab.get('priority_max')
+        if priority_max is not None:
+            kw['priority_max'] = priority_max
```
gctucker, line 110:
```suggestion
            kw['priority_max'] = int(priority_max)
```

broonie:
No idea if that fixes anything, I can't reproduce the issue you're seeing at all - I tested both before and after with empty priorities and it did the right thing.

broonie:
(I did fix one extra case there which *might* help but I couldn't trigger a failure with it either, I'm wondering if there's some Python version inconsistency or something.)

gctucker:
OK I've got to the bottom of this now.

Here's how to reproduce it, first by getting meta-data from a plain mainline build:
```
$ mkdir data/x86
$ wget -P data/x86 https://storage.staging.kernelci.org/kernelci/staging-mainline/staging-mainline-20220418.0/x86_64/x86_64_defconfig/gcc-10/{artifacts,bmeta,steps}.json
```

Using the the last staging tag that had the issue, it works with `lab-broonie` but fails with `lab-collabora` which doesn't have min/max priorities set in YAML:
```
$ git checkout staging-20220413.0
$ mkdir data/jobs
$ ./kci_test generate --lab-config=lab-broonie --install-path=data/x86 --output=data/jobs
data/jobs/kernelci-staging-mainline-staging-mainline-20220418.0-x86_64-x86_64_defconfig-gcc-10-no-dtb-qemu_x86_64-uefi-baseline_qemu.yaml
data/jobs/kernelci-staging-mainline-staging-mainline-20220418.0-x86_64-x86_64_defconfig-gcc-10-no-dtb-qemu_x86_64-ltp-timers_qemu.yaml
data/jobs/kernelci-staging-mainline-staging-mainline-20220418.0-x86_64-x86_64_defconfig-gcc-10-no-dtb-qemu_x86_64-baseline_qemu.yaml
data/jobs/kernelci-staging-mainline-staging-mainline-20220418.0-x86_64-x86_64_defconfig-gcc-10-no-dtb-qemu_x86_64-smc_qemu.yaml
data/jobs/kernelci-staging-mainline-staging-mainline-20220418.0-x86_64-x86_64_defconfig-gcc-10-no-dtb-qemu_x86_64-uefi-mixed-baseline_qemu.yaml
$ ./kci_test generate --lab-config=lab-collabora --install-path=data/x86 --output=data/jobs
  File "/home/gtucker/project/kernelci/kernelci-core/./kci_test", line 301, in <module>
  File "/home/gtucker/project/kernelci/kernelci-core/./kci_test", line 258, in __call__
    job = lab_api.generate(params, device_config, plan_config,
  File "/home/gtucker/project/kernelci/kernelci-core/kernelci/lab/lava/__init__.py", line 49, in generate
    prio_range = lab_config.priority_max - lab_config.priority_min
TypeError: unsupported operand type(s) for -: 'str' and 'str'
```

Then it's working with today's staging:
```
$ git checkout staging-20220417.0
$ ./kci_test generate --lab-config=lab-collabora --install-path=data/x86 --output=data/jobs
data/jobs/kernelci-staging-mainline-staging-mainline-20220418.0-x86_64-x86_64_defconfig-gcc-10-no-dtb-minnowboard-turbot-E3826-baseline.yaml
data/jobs/kernelci-staging-mainline-staging-mainline-20220418.0-x86_64-x86_64_defconfig-gcc-10-no-dtb-qemu_x86_64-uefi-baseline_qemu.yaml
data/jobs/kernelci-staging-mainline-staging-mainline-20220418.0-x86_64-x86_64_defconfig-gcc-10-no-dtb-qemu_x86_64-baseline_qemu.yaml
data/jobs/kernelci-staging-mainline-staging-mainline-20220418.0-x86_64-x86_64_defconfig-gcc-10-no-dtb-qemu_x86_64-uefi-mixed-baseline_qemu.yaml
data/jobs/kernelci-staging-mainline-staging-mainline-20220418.0-x86_64-x86_64_defconfig-gcc-10-no-dtb-qemu_x86_64-smc_qemu.yaml
data/jobs/kernelci-staging-mainline-staging-mainline-20220418.0-x86_64-x86_64_defconfig-gcc-10-no-dtb-qemu_x86_64-ltp-timers_qemu.yaml
```

This is the diff between the 2 tags, which fixed the issue:
```diff
diff --git a/kernelci/config/lab.py b/kernelci/config/lab.py
index ce19c6cea083..9e666a453644 100644
--- a/kernelci/config/lab.py
+++ b/kernelci/config/lab.py
@@ -67,7 +67,7 @@ class LabAPI(Lab):
 
 class Lab_LAVA(LabAPI):
 
-    def __init__(self, priority_min='50', priority_max='50',
+    def __init__(self, priority_min=50, priority_max=50,
                  queue_timeout=None, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self._priority_min = priority_min
@@ -96,6 +96,8 @@ class Lab_LAVA(LabAPI):
                 priority = 50
             elif priority == 'high':
                 priority = 100
+            else:
+                priority = int(priority)
 
             # If min/max are specified these will be overridden
             kw['priority_min'] = priority
@@ -104,10 +106,10 @@ class Lab_LAVA(LabAPI):
         # 0 is a valid value so explicitly check for None
         priority_min = lab.get('priority_min')
         if priority_min is not None:
-            kw['priority_min'] = priority_min
+            kw['priority_min'] = int(priority_min)
         priority_max = lab.get('priority_max')
         if priority_max is not None:
-            kw['priority_max'] = priority_max
+            kw['priority_max'] = int(priority_max)
 
         queue_timeout = lab.get('queue_timeout')
         if queue_timeout:
```

The real problem when not specifying min/max was that the default values in the constructor were strings:
```
    def __init__(self, priority_min='50', priority_max='50',
                 queue_timeout=None, *args, **kwargs):
```
Now they're integers so everyting is integers.  It always worked for `lab-broonie` because it had min/max values and they were defined as integers in the YAML:
```yaml
  lab-broonie:
    lab_type: lava.lava_xmlrpc
    url: 'https://lava.sirena.org.uk/RPC2/'
    priority_min: 0
    priority_max: 40
```
However, converting the values with `int()` is still useful as if the numbers have quotes in YAML they're parsed as strings.  This would have caused the issue too:
```yaml
    priority_min: '0'
    priority_max: '40'
```

The PR as it currently stands is now working fine as far as I can tell so I think it's all good in the end.  Having unit tests would be nice, but that's not a requirement.

gctucker (approved):

PR #1113, opened 2022-04-07
Support use of Debian armel port on systems that require it

broonie:
This board has ethernet and can run NFS root (my own testing
includes something based on the KernelCI NFS tests as a regular
feature).


gctucker (approved):
This is not causing any regressions so merging it now so we can move on. Some parts might need to be refactored to enable picking a different toolchain based on the defconfig, but that can be done as a follow-up.

PR #1114, opened 2022-04-08
test-configs: Separate out selftests that don't have a config dependency

broonie:
There are a number of kselftest targets which do not require any extra
configuration, typically those that exercise core functionality that is
expected to always be enabled like futexes and those that test devices
like RTC which are expected to be part of the platform defconfig. Add a
separate kselftest-basic base class for those kselftests we already have
defined and use it.

This is useful on platforms that have problems with the kselftest
configuration fragment for whatever reason (eg, due to the extra memory
consumption) and lets these tests run in a more normal environment.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -248,6 +248,12 @@ test_plans:
     filters:
       - passlist: {defconfig: ['kselftest']}
 
+  kselftest-basic: &kselftest-basic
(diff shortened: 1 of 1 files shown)


gctucker (commented):
One issue though is that only the builds with the `kselftest` fragment enabled will have kselftest binaries built.  So scheduling a kselftest job using a plain defconfig won't work as things currently stand, since the tests won't be provided with the build.

PR #1115, opened 2022-04-11
test-configs: Add ltp-syscalls

aliceinwire:
Enabling LTP syscalls test

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -368,6 +368,12 @@ test_plans:
     params:
       <<: *ltp-params
       tst_cmdfiles: "pty"
+  
+  ltp-syscalls:
+    <<: *ltp
+    params:
+      <<: *ltp-params
+      tst_cmdfiles: "syscalls"
 
   ltp-timers:
     <<: *ltp
@@ -1847,6 +1853,7 @@ test_configs:
       - ltp-ipc
       - ltp-mm
       - ltp-pty
+      - ltp-syscalls
       - ltp-timers
       - preempt-rt
       - sleep_mem
@@ -1995,6 +2002,7 @@ test_configs:
       - ltp-ipc
       - ltp-mm
       - ltp-pty
+      - ltp-syscalls
       - ltp-timers
       - preempt-rt
       - sleep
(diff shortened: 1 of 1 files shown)


gctucker (commented):
A number of jobs have been run but it looks like they all timed out:
https://staging.kernelci.org/test/plan/id/625d2a7121b7239ac205f4da/

Taking a look to see if it's because the kernel hanged or if it's really because the syscalls tests are taking more than 10min to complete.

gctucker:
So I've re-run the job with a much longer timeout and it completed on the arm64 `hana` Chromebook in 1h41min:
https://staging.lava.collabora.dev/scheduler/job/133369

Running jobs that take that long is potentially an issue as it can impact availability of some devices for other projects that require high responsiveness such as Mesa CI.  So I think we could enable it only for linux-next as we know there isn't more than one new revision per 24h and we can tell how many jobs would be run.  As long as the number of jobs doesn't exceed 50% of the available devices in the labs then this should be OK as an experiment.

We do need to change templates to accommodate for longer timeouts, and adjust filters in YAML to only run this test to limit it to linux-next.  I'm adding the `staging-skip` label for now to avoid scheduling jobs that will always timeout until this has been addressed.

nuclearcat:
Any update on this?

gctucker:
@aliceinwire I can see 2 options here:

1. close this now and reopen when we have some more dynamic scheduling logic to only run long tests when devices are available in labs
2. only run `ltp-syscalls` on a small number of platforms that are fast and can cope with the load easily (e.g. some x86 Chromebooks or arm64 servers)

gctucker:
No updates so closing this now.

gctucker:
@broonie Why did you reopen this?

broonie:
I moved the suite to a different, probably faster, board. It might also be a good fit for the QDF2400 but there's the whole thing with the filter on that not doing the right thing for no reason I've been able to discern.

gctucker:
This PR doesn't apply, has had the skip label for a long time and there hasn't been any update for almost a year.  If you want to pick some changes from it to use with your own PR to enable `ltp-syscalls` then please do that.  But this particular PR should be closed as there's no sign it's going to get fixed and merged any time soon.

broonie:
Sorry, got this confused with a very similar PR of mine that you closed.

PR #1116, opened 2022-04-11
Allow selftests to run without the kselftest fragment

broonie:
Many kselfests have no dependency on the kselftest fragment, such as those for core kernel features which are always enabled and those for drivers which require very device specific configuration expected to come from the defconfg. Even where there is a need for a config fragment the tests should handle it being missing gracefully at runtime. Remove the dependency on building kselftests and remove the runtime dependency from those selftests where it isn't needed.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -255,6 +255,8 @@ test_plans:
     params:
       job_timeout: '10'
       kselftest_collections: "alsa"
+    filters: &kselftest_no_fragment
+      - combination: *arch_defconfig_filter
 
   kselftest-cpufreq:
     <<: *kselftest
@@ -267,12 +269,14 @@ test_plans:
     params:
       job_timeout: '10'
       kselftest_collections: "filesystems"
+    filters: *kselftest_no_fragment
 
   kselftest-futex:
     <<: *kselftest
     params:
       job_timeout: '10'
       kselftest_collections: "futex"
+    filters: *kselftest_no_fragment
 
   kselftest-lib:
     <<: *kselftest
@@ -297,6 +301,7 @@ test_plans:
     params:
       job_timeout: '10'
       kselftest_collections: "rtc"
+    filters: *kselftest_no_fragment
 
   kselftest-seccomp:
     <<: *kselftest
@@ -309,6 +314,7 @@ test_plans:
     params:
       job_timeout: '10'
       kselftest_collections: "tpm2"
+    filters: *kselftest_no_fragment
 
   kselftest-vm:
     <<: *kselftest
--- a/kernelci/build.py
+++ b/kernelci/build.py
@@ -1454,16 +1454,6 @@ class MakeSelftests(Step):
     def name(self):
         return 'kselftest'
 
-    def is_enabled(self):
-        """Check whether the kselftest fragment is enabled
-
-        Return True if the kselftest config fragment is enabled in the build
-        meta-data, or False otherwise.
-        """
-        return 'kselftest' in self._meta.get(
-            'bmeta', 'kernel', 'defconfig_extras'
-        )
-
     def run(self, jopt=None, verbose=False, opts=None):
         """Make the kernel selftests
 


gctucker:
This is currently conflicting with another PR, so we should try to resolve and merge them then this will need to be rebased.

broonie:
It's quite possibly another of mine that it's conflicting against - I'm building up a bit of a backlog here. I could pull some of them together into composite pull requests if that would help, I'm trying to avoid spurious dependencies holding things up by sending conceptually separate things separately but that increases the chances of overlap? github really doesn't deal well with this sort of situation - it either wants fast review cycles or no dependencies.

gctucker:
One way to avoid conflicts is like you did with #1121 by including the changes from earlier PRs it depends upon (i.e. #1119). The only issue is that every time the earlier PR gets updated as part of the review process, the other ones need to be rebased. I've done that before with 3 or 4 PRs and it's usually fine, at least it means all the changes will typically get deployed on staging for testing with no Git conflicts.

Using `staging-skip` it's also possible to focus testing on some PRs to get them merged first and unblock things. I think it's reasonable enough for dealing with this. We might start running staging jobs more often if the rate of changes increases to speed things up too.

gctucker:
Please fix the issue found by the GitHub checks.

broonie:
The github checks are unresolveale as far as I can see - the alphabetic sorting check which for some reason is only applied on v3.10 requires an invalid YAML file with a forward reference. It seems better to apply the sorting manually here TBH.

gctucker:
How about this instead?
```diff
diff --git a/config/core/test-configs.yaml b/config/core/test-configs.yaml
index cbd0f20559c9..a6d94d76594c 100644
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -255,6 +255,8 @@ test_plans:
     params:
       job_timeout: '10'
       kselftest_collections: "alsa"
+    filters: &kselftest-no-fragment
+      - combination: *arch_defconfig_filter
 
   kselftest-cpufreq:
     <<: *kselftest
@@ -267,12 +269,14 @@ test_plans:
     params:
       job_timeout: '10'
       kselftest_collections: "filesystems"
+    filters: *kselftest-no-fragment
 
   kselftest-futex:
     <<: *kselftest
     params:
       job_timeout: '10'
       kselftest_collections: "futex"
+    filters: *kselftest-no-fragment
 
   kselftest-lib:
     <<: *kselftest
@@ -297,6 +301,7 @@ test_plans:
     params:
       job_timeout: '10'
       kselftest_collections: "rtc"
+    filters: *kselftest-no-fragment
 
   kselftest-seccomp:
     <<: *kselftest
@@ -309,6 +314,7 @@ test_plans:
     params:
       job_timeout: '10'
       kselftest_collections: "tpm2"
+    filters: *kselftest-no-fragment
 
   kselftest-vm:
     <<: *kselftest
```
This means only the defconfigs would be tested with the `kselftest-no-fragment` filter, and all the others with the fragment as before.

We might also need to add some filters to the `kselftest-no-fragment` filters to only run on recent kernels as I think support for running individual collections was introduced at some point after 5.10.  This was implicit until now because only mainline, linux-next and the kselftest tree build configs have the kselftest fragment enabled.

broonie:
That seems to DTRT - thanks!

config/core/test-configs.yaml
```diff
@@ -255,6 +255,8 @@ test_plans:
     params:
       job_timeout: '10'
       kselftest_collections: "alsa"
+    filters: &kselftest_no_fragment
+      - combination: *arch_defconfig_filter
```
gctucker, line 259: OK https://staging.kernelci.org/test/plan/id/62613a3fdc4d6831fb05f4cd/

config/core/test-configs.yaml
```diff
@@ -267,12 +269,14 @@ test_plans:
     params:
       job_timeout: '10'
       kselftest_collections: "filesystems"
+    filters: *kselftest_no_fragment
```
gctucker, line 272: OK https://staging.kernelci.org/test/plan/id/62612a687ddcf0c87c05f512/

gctucker, line 272: OK https://staging.kernelci.org/test/plan/id/62612a687ddcf0c87c05f512/

config/core/test-configs.yaml
```diff
@@ -267,12 +269,14 @@ test_plans:
...
 
   kselftest-futex:
     <<: *kselftest
     params:
       job_timeout: '10'
       kselftest_collections: "futex"
+    filters: *kselftest_no_fragment
```
gctucker, line 279: OK https://staging.kernelci.org/test/plan/id/6261314487edc449cc05f4d4/

gctucker, line 279: OK https://staging.kernelci.org/test/plan/id/6261314487edc449cc05f4d4/

config/core/test-configs.yaml
```diff
@@ -297,6 +301,7 @@ test_plans:
     params:
       job_timeout: '10'
       kselftest_collections: "rtc"
+    filters: *kselftest_no_fragment
```
gctucker, line 304: OK https://staging.kernelci.org/test/plan/id/6261412fc1c26b7dc305f4cf/

gctucker, line 304: OK https://staging.kernelci.org/test/plan/id/6261412fc1c26b7dc305f4cf/

config/core/test-configs.yaml
```diff
@@ -309,6 +314,7 @@ test_plans:
     params:
       job_timeout: '10'
       kselftest_collections: "tpm2"
+    filters: *kselftest_no_fragment
```
gctucker, line 317: OK https://staging.kernelci.org/test/plan/id/62614262372063e30705f4c4/

Some tests failed but they were run correctly:

https://storage.staging.kernelci.org/kernelci/staging-mainline/staging-mainline-20220421.0/x86_64/x86_64_defconfig+x86-chromebook/gcc-10/lab-collabora/kselftest-tpm2-hp-11A-G6-EE-grunt.html

Also there's a tracked regression in production so it's not related to this PR:
https://linux.kernelci.org/test/job/next/branch/master/kernel/next-20220420/plan/kselftest-tpm2/

gctucker, line 317: OK https://staging.kernelci.org/test/plan/id/62614262372063e30705f4c4/

Some tests failed but they were run correctly:

https://storage.staging.kernelci.org/kernelci/staging-mainline/staging-mainline-20220421.0/x86_64/x86_64_defconfig+x86-chromebook/gcc-10/lab-collabora/kselftest-tpm2-hp-11A-G6-EE-grunt.html

Also there's a tracked regression in production so it's not related to this PR:
https://linux.kernelci.org/test/job/next/branch/master/kernel/next-20220420/plan/kselftest-tpm2/

gctucker (approved):

gctucker:
The other kselftest collections that use the fragment are still producing results like before:
https://staging.kernelci.org/test/job/kernelci/branch/staging-mainline/kernel/staging-mainline-20220421.0/

PR #1118, opened 2022-04-12
Enable various kselftests

broonie:
This series adds just the basic stanzas for some reasonably reliable kselftests which seem useful and which I've been using to have more kselftest fragments to work with for other development. 

It doesn't include any board enabements due to the concerns about picking systems to run things on, I will turn them on for AT91SAM9G20-EK once https://github.com/kernelci/kernelci-core/pull/1097 is in (and there's some overlap with https://github.com/kernelci/kernelci-core/pull/1116 too).

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -299,6 +299,13 @@ test_plans:
       kselftest_collections: "kcmp"
     filters: *kselftest_no_fragment
 
+  kselftest-kvm:
+    <<: *kselftest
+    params:
+      job_timeout: '10'
+      kselftest_collections: "kvm"
+    filters: *kselftest_no_fragment
+
   kselftest-lib:
     <<: *kselftest
     params:
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -303,6 +303,12 @@ test_plans:
       job_timeout: '10'
       kselftest_collections: "lkdtm"
 
+  kselftest-net:
+    <<: *kselftest
+    params:
+      job_timeout: '10'
+      kselftest_collections: "net"
+
   kselftest-openat2:
     <<: *kselftest
     params:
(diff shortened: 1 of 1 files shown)


gctucker:
This is currently conflicting with another PR, so we should try to resolve and merge them then this will need to be rebased.

broonie:
Rebased against main (resolving a conflict with the ALSA enablement) and added a stanza for the arm64 selftests which will hopefully help minimise further conflicts.

config/core/test-configs.yaml
```diff
@@ -296,6 +345,33 @@ test_plans:
...
+    params:
+      job_timeout: '10'
+      kselftest_collections: "mincore"
+    filters: *kselftest_no_fragment
+
+  kselftest-net:
+    <<: *kselftest
+    params:
+      job_timeout: '10'
+      kselftest_collections: "net"
```
gctucker, line 366: I think we might not be able to run this one due to the debug network interface causing issues with kernel-side DHCP.  See this commit https://github.com/kernelci/kernelci-core/commit/ddedee95b11175e9e95941523d0872ce78c0392b

gctucker, line 366: I think we might not be able to run this one due to the debug network interface causing issues with kernel-side DHCP.  See this commit https://github.com/kernelci/kernelci-core/commit/ddedee95b11175e9e95941523d0872ce78c0392b

broonie: I have a plan for bringing up an environment for that using virtualization which would avoid that issue. In any case, this is just adding the fragment and not enabling running it on any platform.

gctucker: Sure, but I'm not convinced of the value of adding dead code or config data especially when we know it can't be used.

broonie: My goal here is to get the basic stanzas in for the kselftests since I've been finding they're a frequent source of conflicts when enabling things - for things like this where the test clearly exists upstream and is unlikely to go away it seems reasonable to have the stanza there ready for people to build on.

gctucker (approved):

PR #1119, opened 2022-04-13
test-configs: Enable baseline-nfs for Pine64+

broonie:
The board has built in ethernet which is well supported by the kernel.

Signed-off-by: Mark Brown <broonie@kernel.org>

PR #1121, opened 2022-04-13
test-configs: Enable more tests for Pine64+

broonie:
The board has suitable hardware to support the cpufreq and RTC tests and has built in ethernet so can NFS boot.

This builds on top of https://github.com/kernelci/kernelci-core/pull/1119 which is in the history, hopefully github handles that well...

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -2587,6 +2587,7 @@ test_configs:
       # ltp-mm - runs system out of memory
       - ltp-pty
       - ltp-timers
+      - smc
 
   - device_type: sun50i-h5-libretech-all-h3-cc
     test_plans:
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -2580,6 +2580,13 @@ test_configs:
     test_plans:
       - baseline
       - baseline-nfs
+      - ltp-crypto
+      - ltp-fcntl-locktests
+      - ltp-ima
+      - ltp-ipc
+      # ltp-mm - runs system out of memory
+      - ltp-pty
+      - ltp-timers
 
   - device_type: sun50i-h5-libretech-all-h3-cc
     test_plans:


broonie:
Needs a device-type tweak in LAVA to handle bloated kernels like this - eg, https://lava.sirena.org.uk/scheduler/job/64757

I'd expect https://github.com/kernelci/kernelci-core/pull/1116 should also enable those of the tests that don't need the config fragment like RTC and arm64 to work with the existing configuration.

montjoie:
I have some fixes in progress for LAVA:
- to detect detect thoses error:
- change addresses to load big kernel
You can find them at https://github.com/montjoie/lava-docker/tree/lab-2022.03
Like https://github.com/montjoie/lava-docker/commit/d2a9a3e4462735a61614e6da4242e17222885ae4
or https://github.com/montjoie/lava-docker/commit/63a6366d9ddc8689a344c95cc310bb3f5548e5f4
I need to add the sunxi addresses change patch

gctucker:
> This builds on top of https://github.com/kernelci/kernelci-core/pull/1119 which is in the history, hopefully github handles that well...

Yes that worked.  It's not really a GitHub thing but just how Git can pull branches on top of each other with no conflicts as done by [`pending.py`](https://github.com/kernelci/kernelci-deploy/blob/main/pending.py) for creating the staging branch.

I think we'll get #1119 merged today and this one should merge cleanly on top of it with GitHub doing the rebase.

gctucker:
I confirm the kselftest kernels aren't booting on this platform at the moment:
https://lava.sirena.org.uk/scheduler/job/64784

The #1116 PR is conflicting with an earlier one, let's see if we can get that resolved to unblock the situation here and run kselftest with plain defconfig builds.  Or if the device dict gets updated and the big kernels start booting then that should work too.

broonie:
@montjoie Ah, nice! 

@gctucker The question with github handling pull requests stacked on top of each other is if it can cope with situations like what happens when, eg, pull request A is merged but pull request B which it incorporates is not merged and if it otherwise manages to link the two - as far as I can see it doesn't do anything of the sort.

broonie:
I've split the kselftests into a separate pull request (https://github.com/kernelci/kernelci-core/pull/1124) for now.

broonie:
The SMC test definition appears to have generic problems on arm64 FWIW:

https://storage.kernelci.org/next/master/next-20220414/arm64/defconfig+arm64-chromebook/clang-13/lab-cip/smc-r8a774a1-hihope-rzg2m-ex.html

gctucker:
> @gctucker The question with github handling pull requests stacked on top of each other is if it can cope with situations like what happens when, eg, pull request A is merged but pull request B which it incorporates is not merged and if it otherwise manages to link the two - as far as I can see it doesn't do anything of the sort.

If PR A was merged with a rebase, then PR B still appears to have the commits since they have different SHA1. Doing a rebase-and-merge of PR B as-is typically works since Git will then see that the same commits have already been merged (if everything applied cleanly). But the safest thing to do is to just rebase PR B on top of main once PR A has been merged. This approach also ensures there aren't conflicts between PR A and B so I think it's less work and issues overall.

gctucker:
Well, in this case there's a conflict because the first commit has already been merged so the branch needs to be rebased interactively with just the last 2 commits.

broonie:
Ugh, right - so github doesn't cope with this at all (especially with rebases getting added on which breaks things further). It *should* be able to figure out that the base commits are the same even when rebasing, though obviously the disadvantage of rebasing is that it makes that harder.

broonie:
I guess in this situation the best thing to do is to work backwards from the latest branch to the earliest, that way the maximum set can go in at once?

gctucker (commented):
For some reason there aren't any results for this platform on staging even though there are online platforms in `lab-baylibre` and `lab-baylibre`.  This might be an issue with the arm64 build status causing Jenkins to not schedule LAVA jobs or something like that.  Investigating...

broonie:
The board is also online in my lab FWIW - I had noticed that no jobs were being scheduled from staging, I wasn't sure if that was a deliberate staging configuration thing or something.

broonie:
The board is also online in my lab FWIW - I had noticed that no jobs were being scheduled from staging, I wasn't sure if that was a deliberate staging configuration thing or something.

broonie:
It's not clear to me what the situation with ltp-pty is but it looks like a generic issue rather than a problem with this board - we do have a similar log showing up in production on most boards - eg: https://storage.kernelci.org/next/master/next-20220421/arm64/defconfig/gcc-10/lab-cip/ltp-pty-r8a774a1-hihope-rzg2m-ex.txt The only board where it seems to work fine is rk3399-gru-kevin. Without investigating further it's not clear to me if this is just because pty07 takes a while to run and is marginal for timeouts or if there's some bug here, either way it looks like there's nothing board specific going on and the test needs fixing. 

My guess at the minute is timeout given the logging pty06 is doing and similar logging that pty07 is doing. I've submitted a new job with larger timeouts at https://lava.sirena.org.uk/scheduler/job/78548 which might give some clue when it runs.

I do note that LKFT doesn't seem to run that test as part of ltp-pty - see eg https://lkft.validation.linaro.org/scheduler/job/4933791 - it's stopping at pty05.c and we're into pty07.c here and the logs on test pty06 aren't super reassuring about expected stability

gctucker:
OK thanks, so this should be good to go for now. We might tweak timeouts or skiplists as a follow-up to resolve the ltp-pty issue.

gctucker (approved):

broonie:
That job with increased timeouts ran in 17 minutes so looks like it is indeed a timeout issue - the job timeout is 15 minutes.

PR #1123, opened 2022-04-14
Fix pkg-config

broonie:
This installs a copy of pkg-config we can actually use and ensures it finds libasound2 in cross builds.

### DIFF
--- a/config/docker/clang-base/Dockerfile
+++ b/config/docker/clang-base/Dockerfile
@@ -32,7 +32,7 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
    libhugetlbfs-dev:arm64 \
    libpopt-dev:arm64 \
    libasound2-dev:arm64 \
-   pkg-config:arm64
+   pkg-config
 
 # kselftest arm
 RUN dpkg --add-architecture armhf
@@ -44,6 +44,6 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
    libhugetlbfs-dev:armhf \
    libpopt-dev:armhf \
    libasound2-dev:armhf \
-   pkg-config:armhf
+   pkg-config
 
 RUN apt-get autoremove -y gcc
--- a/config/docker/gcc-10_arm/Dockerfile
+++ b/config/docker/gcc-10_arm/Dockerfile
@@ -27,4 +27,4 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
    libnuma-dev:armhf \
    libpopt-dev:armhf \
    libasound2-dev:armhf \
-   pkg-config:armhf
+   pkg-config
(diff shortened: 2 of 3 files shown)


config/docker/clang-base/Dockerfile
```diff
@@ -32,7 +32,7 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
    libhugetlbfs-dev:arm64 \
    libpopt-dev:arm64 \
    libasound2-dev:arm64 \
-   pkg-config:arm64
+   pkg-config
```
gctucker, line 35: I guess we don't strictly need to list `pkg-config` multiple times. Is this more to keep the list self-contained in case we want to split the Dockerfile into per-arch fragments? In this case, maybe `libasound2-dev` should be listed again as well?

config/docker/gcc-10_arm/Dockerfile
```diff
@@ -27,4 +27,5 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
    libnuma-dev:armhf \
    libpopt-dev:armhf \
    libasound2-dev:armhf \
-   pkg-config:armhf
+   libasound2-dev \
```
gctucker, line 30: So is the host package needed in order to get the header files installed, and the arch-specific one for linkage?

gctucker (approved):
LGTM - tested OK on staging:
```
make[1]: Entering directory '/tmp/kci/linux/tools/testing/selftests/alsa'
ccache aarch64-linux-gnu-gcc     mixer-test.c -lasound -o /tmp/kci/linux/build/kselftest/alsa/mixer-test
make[1]: Leaving directory '/tmp/kci/linux/tools/testing/selftests/alsa'
```
https://storage.staging.kernelci.org/kernelci/staging-next/staging-next-20220414.0/arm64/defconfig+kselftest/gcc-10/logs/kselftest.log

gctucker:
Also confirmed the `alsa` tests are now listed in the artifacts:
```json
    "kselftest": [
        {
            "contents": [
                "alsa",
                "arm64",
                "breakpoints",
```
https://storage.staging.kernelci.org/kernelci/staging-next/staging-next-20220415.0/arm64/defconfig%2Barm64-chromebook%2Bkselftest/gcc-10/artifacts.json

As opposed to current production builds:
```json
    "kselftest": [
        {
            "contents": [
                "arm64",
                "breakpoints",
```
https://storage.kernelci.org/next/master/next-20220414/arm64/defconfig%2Barm64-chromebook%2Bkselftest/gcc-10/artifacts.json

config/docker/clang-base/Dockerfile
broonie: I'm not sure of the rationale TBH. As things stand we should be able to just merge everything.

config/docker/gcc-10_arm/Dockerfile
broonie: The host package is needed to drop a configuration file in the pkg-config directory so it knows about the library and can produce appropriate flags without error - otherwise the target -dev package should be enough.

PR #1130, opened 2022-04-15
test-configs.yaml: Add a machine type for Juno with U-Boot firmware

broonie:
The standard LAVA interpretation of Juno includes support for rewriting
the firmware as part of the test job but not all labs will want to do
that since it makes jobs take longer to run and be more hassle to
specify. Add a separate device type for Juno r0 systems with fixed
u-boot firmware.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -967,6 +967,12 @@ device_types:
       - blocklist: *device_config_filter
       - blocklist: {defconfig: ['multi_v7_defconfig+kselftest']}
 
+  juno-uboot:
+    mach: vexpress
+    class: arm64-dtb
+    dtb: 'arm/juno.dtb'
+    boot_method: uboot
+
   kirkwood-db-88f6282:
     mach: mvebu
     class: arm-dtb
@@ -2170,6 +2176,24 @@ test_configs:
     test_plans:
       - baseline
       - baseline-nfs
+
+  - device_type: juno-uboot
+    test_plans:
+      - baseline
+      - baseline-nfs
+      - kselftest-cpufreq
+      - kselftest-futex
+      - kselftest-lib
+      - kselftest-lkdtm
+      - kselftest-rtc
+      - kselftest-seccomp
+      - ltp-crypto
+      - ltp-fcntl-locktests
+      - ltp-ipc
+      # ltp-mm - runs system out of memory
+      - ltp-pty
+      - ltp-timers
+      - smc
 
   - device_type: kirkwood-db-88f6282
     test_plans:


config/core/test-configs.yaml
```diff
@@ -959,6 +959,12 @@ device_types:
       - blocklist: *device_config_filter
       - blocklist: {defconfig: ['multi_v7_defconfig+kselftest']}
 
+  juno-r0-uboot:
```
gctucker, line 962: The LAVA device type in upstream LAVA is [`juno-uboot`](https://git.lavasoftware.org/lava/lava/-/blob/master/etc/dispatcher-config/device-types/juno-uboot.jinja2) so you either need to add `name: juno-uboot` to the device type definition here or send a `juno-r0-uboot` variant upstream.

broonie: The upstream LAVA device type appears to be both broken and to look like it's got stuff to reflash the firmware in there which I'm actively trying to avoid. The whole point with this type as with the -acpi version is that there is upstream support for these devices but it pulls in flashing the firmware which both complicates generating the job definitions and slows everything down.

It's not clear why we have this requirement for some devices but not others - there wasn't an issue with the ACPI version of this?

broonie: Ah, no - the upstream LAVA type does seem to be able to avoid reflashing. Doesn't really help with the bugginess though (or the fact that it's independent of revision of the board which means that there is no way to figure out the correct DT).

broonie: Never mind, looks like upstream uses no revision as a synonym for r0... will switch to that.

gctucker: > It's not clear why we have this requirement for some devices but not others - there wasn't an issue with the ACPI version of this?

It was an oversight on my part in the previous PR with ACPI, I just assumed that Juno device types were all good in upstream LAVA.

gctucker: > Never mind, looks like upstream uses no revision as a synonym for r0... will switch to that.

OK thanks, that should simplify things a bit. I just wonder how we might cope with different Juno revisions if we have some online at some point.

config/core/test-configs.yaml
```diff
@@ -967,6 +967,12 @@ device_types:
       - blocklist: *device_config_filter
       - blocklist: {defconfig: ['multi_v7_defconfig+kselftest']}
 
+  juno-uboot:
+    mach: vexpress
+    class: arm64-dtb
+    dtb: 'arm/juno.dtb'
+    boot_method: u-boot
```
gctucker, line 974:
```suggestion
    boot_method: uboot
```

gctucker (commented):
With the small fix above I was able to generate and submit jobs manually:
https://staging.kernelci.org/test/plan/id/626106eff771e3cb3605f4c5/

config/core/test-configs.yaml
```diff
@@ -2171,6 +2177,24 @@ test_configs:
       - baseline
       - baseline-nfs
 
+  - device_type: juno-uboot
+    test_plans:
+      - baseline
```
gctucker, line 2182: OK https://staging.kernelci.org/test/plan/id/62665c5a225c5bb8441786a0/

config/core/test-configs.yaml
```diff
@@ -2171,6 +2177,24 @@ test_configs:
...
+      - baseline-nfs
```
gctucker, line 2183: OK https://staging.kernelci.org/test/plan/id/62666197e290f2bfb61786d5/

gctucker, line 2183: OK https://staging.kernelci.org/test/plan/id/62666197e290f2bfb61786d5/

config/core/test-configs.yaml
```diff
@@ -2171,6 +2177,24 @@ test_configs:
...
+      - kselftest-cpufreq
```
gctucker, line 2184: OK https://staging.kernelci.org/test/plan/id/62665b7e42a95fe54c178696/

gctucker, line 2184: OK https://staging.kernelci.org/test/plan/id/62665b7e42a95fe54c178696/

config/core/test-configs.yaml
```diff
@@ -2171,6 +2177,24 @@ test_configs:
...
+      - kselftest-futex
```
gctucker, line 2185: Failed, looks like something got stuck:
https://lava.sirena.org.uk/scheduler/job/84875

Also it seems to only be an issue with Clang builds, the GCC ones appear to be completing:
https://lava.sirena.org.uk/scheduler/job/84868

gctucker, line 2185: Failed, looks like something got stuck:
https://lava.sirena.org.uk/scheduler/job/84875

Also it seems to only be an issue with Clang builds, the GCC ones appear to be completing:
https://lava.sirena.org.uk/scheduler/job/84868

config/core/test-configs.yaml
```diff
@@ -2171,6 +2177,24 @@ test_configs:
...
+      - kselftest-lib
```
gctucker, line 2186: OK https://staging.kernelci.org/test/plan/id/6266633bf3b69989fb17869e/

gctucker, line 2186: OK https://staging.kernelci.org/test/plan/id/6266633bf3b69989fb17869e/

config/core/test-configs.yaml
```diff
@@ -2171,6 +2177,24 @@ test_configs:
...
+      - kselftest-lkdtm
```
gctucker, line 2187: OK with some failing test cases: https://staging.kernelci.org/test/plan/id/626667b2a74b776ad4178697/

gctucker, line 2187: OK with some failing test cases: https://staging.kernelci.org/test/plan/id/626667b2a74b776ad4178697/

config/core/test-configs.yaml
```diff
@@ -2171,6 +2177,24 @@ test_configs:
...
+      - kselftest-rtc
```
gctucker, line 2188: OK https://staging.kernelci.org/test/plan/id/62666996dc4068e1521786a5/

gctucker, line 2188: OK https://staging.kernelci.org/test/plan/id/62666996dc4068e1521786a5/

config/core/test-configs.yaml
```diff
@@ -2171,6 +2177,24 @@ test_configs:
...
+      - kselftest-seccomp
```
gctucker, line 2189: OK https://staging.kernelci.org/test/plan/id/626665bc8290fd9a561786f9/

gctucker, line 2189: OK https://staging.kernelci.org/test/plan/id/626665bc8290fd9a561786f9/

config/core/test-configs.yaml
```diff
@@ -2171,6 +2177,24 @@ test_configs:
...
+      - ltp-crypto
```
gctucker, line 2190: OK https://staging.kernelci.org/test/plan/id/6265be67b474599a2d178696/

gctucker, line 2190: OK https://staging.kernelci.org/test/plan/id/6265be67b474599a2d178696/

config/core/test-configs.yaml
```diff
@@ -2171,6 +2177,24 @@ test_configs:
...
+      - ltp-fcntl-locktests
```
gctucker, line 2191: OK https://staging.kernelci.org/test/plan/id/6265a2e10608c745d0178696/

gctucker, line 2191: OK https://staging.kernelci.org/test/plan/id/6265a2e10608c745d0178696/

config/core/test-configs.yaml
```diff
@@ -2171,6 +2177,24 @@ test_configs:
...
+      - ltp-ipc
```
gctucker, line 2192: OK https://staging.kernelci.org/test/plan/id/6265942d1d9ddfbe0d17869a/

gctucker, line 2192: OK https://staging.kernelci.org/test/plan/id/6265942d1d9ddfbe0d17869a/

config/core/test-configs.yaml
```diff
@@ -2171,6 +2177,24 @@ test_configs:
...
+      # ltp-mm - runs system out of memory
+      - ltp-pty
```
gctucker, line 2194: OK https://staging.kernelci.org/test/plan/id/626592ed6c9a8ff8e7178696/

gctucker, line 2194: OK https://staging.kernelci.org/test/plan/id/626592ed6c9a8ff8e7178696/

config/core/test-configs.yaml
```diff
@@ -2171,6 +2177,24 @@ test_configs:
...
+      - ltp-timers
```
gctucker, line 2195: OK with some failing cases but that's not specific to Juno: https://staging.kernelci.org/test/plan/id/6265a0b02ab8578665178696/

gctucker, line 2195: OK with some failing cases but that's not specific to Juno: https://staging.kernelci.org/test/plan/id/6265a0b02ab8578665178696/

config/core/test-configs.yaml
```diff
@@ -2171,6 +2177,24 @@ test_configs:
...
+      - smc
```
gctucker, line 2196: The test wasn't actually run due to a known issue in test-definitions:
https://lava.sirena.org.uk/scheduler/job/84922

So this is "OK" as far as this PR is concerned, although we don't know if it would have produced useful results. I think we can leave it here and review things later when the smc test has been fixed.

gctucker, line 2196: The test wasn't actually run due to a known issue in test-definitions:
https://lava.sirena.org.uk/scheduler/job/84922

So this is "OK" as far as this PR is concerned, although we don't know if it would have produced useful results. I think we can leave it here and review things later when the smc test has been fixed.

gctucker (commented):
OK with maybe just one problem with Clang build for `kselftest-futex`.

gctucker:
Looks like GitHub got confused somewhere and posted all my comments 3 times or something... Sorry about the noise.

config/core/test-configs.yaml
broonie: This looks like a build issue with clang - the same thing is happening for the same test on existing platforms:

https://storage.staging.kernelci.org/kernelci/staging-next/staging-next-20220425.0/arm64/defconfig/clang-15/lab-collabora/kselftest-futex-meson-g12b-a311d-khadas-vim3.txt

and we're generally thin on kselftest results built with clang-15 on arm - if you look at 

https://storage.staging.kernelci.org/kernelci/staging-next/staging-next-20220425.0/arm64/defconfig/clang-15/lab-broonie/kselftest-alsa-meson-gxl-s905x-libretech-cc.txt

it looks an awful lot like the program hasn't actually been cross compiled when building for clang.

gctucker: Right, yes that makes sense. So I think this PR is good to go then, and we should chase up the kselftest build issue separately.

gctucker (approved):

PR #921, opened 2021-12-01
test-configs: Use implementation defined FEAT_PAUTH with qemu on arm64

broonie:
The implementation of the architected pointer authentication implementation
in qemu is very slow so qemu provides optional support for an additional
implementation defined algorithm which runs much faster. This is not so
noticeable when userspace doesn't make use of pointer authentication but
once it does the performance improvement is dramatic. Switch our virtual
arm64 targets over to this, there's no real coverage impact for the kernel.

Signed-off-by: Mark Brown <broonie@kernel.org>

PR #1131, opened 2022-04-16
config/docker: Use a current version of qemu

broonie:
Half the point of running all docker jobs in qemu is to allow us
to make use of features of and get coverage of architecture
emulation in current versions of qemu. Unfortunately we're just
using the default version from Debian which is v5.2, already more
than a year old and only going to get futher and further behind.
Change the container to pick up the verison of qemu from
backports which should be a lot more current - at time of writing
it's got the latest release.

### DIFF
--- a/config/docker/qemu/Dockerfile
+++ b/config/docker/qemu/Dockerfile
@@ -1,7 +1,10 @@
 FROM debian:bullseye
 
-RUN apt-get update && \
-	apt-get install --yes --no-install-recommends \
+RUN echo deb http://deb.debian.org/debian bullseye-backports main \
+	> /etc/apt/sources.list.d/backports.list
+
+RUN apt-get update && apt-get install -t bullseye-backports --yes
+	--no-install-recommends \
 	qemu-system \
 	qemu-system-arm \
 	qemu-system-mips \


config/docker/qemu/Dockerfile
```diff
@@ -1,7 +1,11 @@
 FROM debian:bullseye
 
-RUN apt-get update && \
-	apt-get install --yes --no-install-recommends \
+RUN echo deb http://deb.debian.org/debian bullseye-backports main \
+	> /etc/apt/sources.list.d/backports.list
+
+RUN apt-get update
+
+RUN apt-get install -t bullseye-backports --yes --no-install-recommends \
```
gctucker, line 8: Please keep `update` and `install` on the same line. This is to avoid bad surprises with how Docker layer caching works: if the list of packages gets updated, `apt-get update` won't be run and the install might fail.

https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#apt-get

gctucker (dismissed):

gctucker:
@mgalka I think you can just the "resolve conversation" button for each thing that has been resolved.  Then it's more something for the reviewer to change the review status. I thought there was a way to request a new review, but even that shouldn't clear the "changes required" flag.

gctucker:
This got built OK on staging, here's the new QEMU version:
```
$ docker pull kernelci/staging-qemu
$ docker run -it kernelci/staging-qemu qemu-system-x86_64 --version
QEMU emulator version 6.2.0 (Debian 1:6.2+dfsg-2~bpo11+1)
Copyright (c) 2003-2021 Fabrice Bellard and the QEMU Project developers
```

gctucker:
Tested OK on staging (there's an infrastructure problem in `lab-collabora` which doesn't have QEMU enabled in KernelCI production yet):
https://staging.kernelci.org/test/job/kernelci/branch/staging-mainline/kernel/staging-mainline-20220418.0/plan/baseline/

Also confirmed this fixes the issue with #921.

PR #1132, opened 2022-04-16
test-configs: Don't use KVM acceleration for arm64 virtual platforms

broonie:
Using KVM acceleration means that when running natively instead
of emulating whatever qemu supports we directly virtualise
whatever CPU the host has. This results in a less consistent
target platform between test runs and means we miss coverage of
new features only available in emulation which is one of the
benefits of using qemu. It's reasonable to want to test
virtualisation but we should specifically do so as a separate
target.

This is unlikely to have had any effect at present since
currently I think all our host systems are x86 based.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -1307,7 +1307,7 @@ device_types:
       arch: arm64
       cpu: 'max'
       guestfs_interface: 'virtio'
-      machine: 'virt,gic-version=2,mte=on,accel=kvm:tcg'
+      machine: 'virt,gic-version=2,mte=on,accel=tcg'
     filters:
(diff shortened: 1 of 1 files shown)


gctucker (approved):
Tested OK on staging:
https://staging.kernelci.org/test/plan/id/625d28082e4e25267905f4c8/
https://storage.staging.kernelci.org/kernelci/staging-mainline/staging-mainline-20220418.0/arm64/defconfig+arm64-chromebook/gcc-10/lab-baylibre/baseline-qemu_arm64-virt-gicv3-uefi.html#L214
```
-machine virt,gic-version=3,mte=on,accel=tcg
```

PR #1133, opened 2022-04-18
CHROMEOS Allow to modify root device prefix

nuclearcat:
As some chromebooks might have mmcblk1, some nvme storage device,
we must allow to specify flashing device by environment variable FLASH_DEVICE.
Also allow special case "detect" when script will detect mmcblk0 OR mmcblk1,
as on some chromebooks device number change randomly on boot.

### DIFF
--- a/config/rootfs/debos/overlays/chromeos-flash/opt/chromeos/flash
+++ b/config/rootfs/debos/overlays/chromeos-flash/opt/chromeos/flash
@@ -38,12 +38,13 @@ flash_chromeos()
     mount_special_fs "$IMAGE_MOUNTPOINT"
     mount --bind . "$mount_dir"/mnt/empty
     ls -l "$mount_dir"/mnt/empty/"$IMAGE_NAME"
-    ls -l "$mount_dir"/dev/mmc*
+    ls -l "$mount_dir"/dev/{mmc*,nvme*} || true
 
     echo "Starting to flash..."
     chroot "$mount_dir" \
          /usr/sbin/chromeos-install \
-        --dst /dev/mmcblk0 \
+        --debug \
+        --dst /dev/${FLASH_DEVICE} \
         --payload_image /mnt/empty/"$IMAGE_NAME" \
         --yes
     sync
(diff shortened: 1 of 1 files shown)


config/rootfs/debos/overlays/chromeos-flash/opt/chromeos/flash
```diff
@@ -97,6 +97,16 @@ fixup_root()
 commands="${*:-flash_chromeos enable_rw fixup_root}"
 IMAGE_NAME="${IMAGE_NAME:-octopus-chromiumos-test-image.bin}"
 IMAGE_MOUNTPOINT="${IMAGE_MOUNTPOINT:-/root/chromeos}"
+FLASH_DEVICE="${FLASH_DEVICE:-mmcblk0}"
+
+if [ "${FLASH_DEVICE}" == "detect" ]; then
+    echo "Detecting storage device"
+    if [ -e "/dev/mmcblk0" ]; then
+       FLASH_DEVICE="mmcblk0"
+    elif [ -e "/dev/mmcblk1" ]; then
```
gctucker, line 106: Surely if `mmcblk1` exists it means there is also a `mmcblk0` and this branch is never reached.  Inverting the tests could work, but then there are other issues with doing this.

Is there any guarantee that the same device number will always map to the same device?  Typically this depends on the order they get initialised by the kernel, which may vary following random factors.  Then if they're always the same for each device type, we could pass it as a parameter in the flashing job based on some YAML config parameter.

nuclearcat: It will sound surprising, but thats the case with grunt: https://lava.collabora.co.uk/scheduler/job/6097834
mmcblk0 doesnt exist, but mmcblk1 does.
Worse, sometimes it books in ChromiumOS as mmcblk0, sometimes as mmcblk1, and that forced me to search alternative for tast tests root=/dev/mmcblkX as device name is not predictable (PARTLABEL should work).

nuclearcat:
Rebased

config/rootfs/debos/overlays/chromeos-flash/opt/chromeos/flash
```diff
@@ -97,6 +99,16 @@ fixup_root()
...
+       FLASH_DEVICE="mmcblk1"
+    fi
```
mgalka, line 110: Maybe `case` could be more readable than if..elif. We could also think of making a function out of it.

mgalka (commented):
a minior comment

nuclearcat:
Built: https://bot.staging.kernelci.org/job/chromeos/job/rootfs-builder/23/console
Available at: https://storage.chromeos.kernelci.org/images/rootfs/chromeos/20220502.1/bullseye-chromeos-flash/amd64/
Tested: https://lava.collabora.co.uk/scheduler/job/6234073

gctucker:
Please rebase

nuclearcat:
Rebased

mgalka (approved):
LGTM

PR #1134, opened 2022-04-18
Enable test coverage on Le Potato

broonie:
Another board in my lab that's able to run more tests than we currently do. This could use some more passes once some of my other pull requests have gone in.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -2258,6 +2258,10 @@ test_configs:
     test_plans:
       - baseline
       - baseline-nfs
+      - kselftest-cpufreq
+      - kselftest-futex
+      - kselftest-lib
+      - kselftest-seccomp
 
   - device_type: meson-gxm-khadas-vim2
     test_plans:
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -2257,6 +2257,7 @@ test_configs:
   - device_type: meson-gxl-s905x-libretech-cc
     test_plans:
       - baseline
+      - baseline-nfs
 
   - device_type: meson-gxm-khadas-vim2
     test_plans:
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -2258,10 +2258,17 @@ test_configs:
     test_plans:
       - baseline
       - baseline-nfs
+      - kselftest-alsa
       - kselftest-cpufreq
       - kselftest-futex
       - kselftest-lib
       - kselftest-seccomp
+      - ltp-crypto
+      - ltp-fcntl-locktests
+      - ltp-ipc
+      # ltp-mm - Runs board out of memory
+      - ltp-pty
+      - ltp-timers
 
   - device_type: meson-gxm-khadas-vim2
     test_plans:


config/core/test-configs.yaml
```diff
@@ -2248,6 +2248,12 @@ test_configs:
       - kselftest-futex
       - kselftest-lib
       - kselftest-seccomp
+      - ltp-crypto
+      - ltp-fcntl-locktests
+      - ltp-ipc
+      # ltp-mm - Runs board out of memory
```
gctucker, line 2254: If the memory size is a hard requirement then this will never be able to run, in which case adding a comment isn't useful.  I think we need other ways of tracking things that don't get documented well in YAML.  There's a section in the documentation about tests run by KernelCI, and there's GitHub issues for things that can be fixed.

Maybe with a better configuration system we would just add a filter to block `ltp-mm` on this platform, so the config would be both functional and self-documenting.

broonie: I'd say this is a bug in the testsuite - it should be skipping tests that are going to run the system out of memory, if it's just exploding easily that's not really usable.

broonie: I do also find it useful to have comments in the file explaining why something's not enabled - especially with the way we're currently mixing lab config in with board capability stuff. I know there's a few things that I've gone and looked at enabling multiple times which I'd not have done with a note present.

gctucker: OK fair enough. As discussed in the meeting today, we can use comments for the time being and think of a better way of dealing with this kind of issues in the longer term.

config/core/test-configs.yaml
```diff
@@ -2257,6 +2257,18 @@ test_configs:
   - device_type: meson-gxl-s905x-libretech-cc
     test_plans:
       - baseline
+      - baseline-nfs
```
gctucker, line 2260: @montjoie Failed in `lab-baylibre`: https://staging.kernelci.org/test/plan/id/62612c99a70a240a9905f4d4/

https://storage.staging.kernelci.org/kernelci/staging-mainline/staging-mainline-20220421.0/arm64/defconfig/gcc-10/lab-baylibre/baseline-nfs-meson-gxl-s905x-libretech-cc.html

However it passed in `lab-broonie`: https://staging.kernelci.org/test/plan/id/6261322de464679a3205f544/

config/core/test-configs.yaml
```diff
@@ -2257,6 +2257,18 @@ test_configs:
...
+      - kselftest-alsa
```
gctucker, line 2261: OK https://staging.kernelci.org/test/plan/id/62612f0cfdf2daf9f005f4c4/

config/core/test-configs.yaml
```diff
@@ -2257,6 +2257,18 @@ test_configs:
...
+      - kselftest-cpufreq
```
gctucker, line 2262: OK https://staging.kernelci.org/test/plan/id/62613dc08d96887fac05f4d1/

config/core/test-configs.yaml
```diff
@@ -2257,6 +2257,18 @@ test_configs:
...
+      - kselftest-futex
```
gctucker, line 2263: OK https://staging.kernelci.org/test/plan/id/6261318ceaba57607805f4d6/

config/core/test-configs.yaml
```diff
@@ -2257,6 +2257,18 @@ test_configs:
...
+      - kselftest-lib
```
gctucker, line 2264: OK https://staging.kernelci.org/test/plan/id/62614b5b536cf9c7b905f4c4/

config/core/test-configs.yaml
```diff
@@ -2257,6 +2257,18 @@ test_configs:
...
+      - kselftest-seccomp
```
gctucker, line 2265: OK with a few failing test cases https://staging.kernelci.org/test/plan/id/62613ba30706d8e88f05f4dd/

config/core/test-configs.yaml
```diff
@@ -2257,6 +2257,18 @@ test_configs:
...
+      - ltp-crypto
```
gctucker, line 2266: OK https://staging.kernelci.org/test/plan/id/6261448fb7bfe0768505f4ec/

config/core/test-configs.yaml
```diff
@@ -2257,6 +2257,18 @@ test_configs:
...
+      - ltp-fcntl-locktests
```
gctucker, line 2267: OK https://staging.kernelci.org/test/plan/id/6261309cda84da457505f4d4/

config/core/test-configs.yaml
```diff
@@ -2257,6 +2257,18 @@ test_configs:
...
+      - ltp-ipc
```
gctucker, line 2268: OK https://staging.kernelci.org/test/plan/id/62612fd6b88abd9f9105f523/

config/core/test-configs.yaml
```diff
@@ -2257,6 +2257,18 @@ test_configs:
...
+      # ltp-mm - Runs board out of memory
+      - ltp-pty
```
gctucker, line 2270: Same problem as mentioned earlier:
https://staging.kernelci.org/test/plan/id/626139aeed990d896505f4d4/
https://storage.staging.kernelci.org/kernelci/staging-mainline/staging-mainline-20220421.0/arm64/defconfig/gcc-10/lab-broonie/ltp-pty-meson-gxl-s905x-libretech-cc.html

So it's not "OK" but the issue is not related to this PR.

config/core/test-configs.yaml
```diff
@@ -2257,6 +2257,18 @@ test_configs:
...
+      - ltp-timers
```
gctucker, line 2271: OK with a couple of failing test cases https://staging.kernelci.org/test/plan/id/626135ed6dcd090e2605f4c4/

gctucker (approved):
It's failing to run in `lab-baylibre` but it's a lab issue to fix, not a kernel problem and not an issue with this PR so it can be merged now.

PR #1139, opened 2022-04-19
test-configs: Add multi_v5_defconfig as a defconfig for 32 bit arm

broonie:
The arm port has two main defconfigs - multi_v7_defconfig for the more
modern hardware and multi_v5_defconfig for older system. Currently we
only consider the former where we need a list of defconfigs but we do
have some older platforms so consider multi_v5_defconfig part of our
standard architecture defconfig list.


mgalka (approved):
Test job definitions generate correctly.

PR #1140, opened 2022-04-20
setup.py: add LAVA template files to package

gctucker:
Add the LAVA template files to the kernelci package so they get
distributed with the code.  The idea is to only include generic config
files that can be reused in any installation.  Also fix the path to
where the documentation files are getting installed as a side-effect.


gctucker (commented):
Tested OK on staging with the new pipeline.

PR #1142, opened 2022-04-21
T34835 lab config path

gctucker:
Add `config_path` attribute to the lab configuration to know where to find things such as Jinja2 templates. Also populate the attribute with all the existing lab configs.


gctucker:
This is actually still a bit experimental and I'm considering alternatives where the path to the template is constructed differently without the need for this extra config option.

gctucker:
I think this is ready now, it's simply about adding the path where to find the base templates or potentially other config files for each lab.

gctucker (commented):
Tested OK on staging

PR #1144, opened 2022-04-22
test-configs: Increase timeout for LTP pty tests

broonie:
The LTP pty tests have been timing out in production, with only one
board managing to complete the tests recently. The default timeout is 15
minutes but running with a larger timeout showed that the test job took
17 minutes to complete on Pine64 Plus. Raise the timeout to 25 minutes,
covering that with some additional headroom for slower systems.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -383,6 +383,7 @@ test_plans:
     params:
       <<: *ltp-params
       tst_cmdfiles: "pty"
+      job_timeout: '25'
 
   ltp-timers:
     <<: *ltp


alexandrasp:
I would say it is good to be merged, but not sure if we have any hardware limitation, could you confirm please @gctucker ?

alexandrasp (approved):

broonie:
There *is* an issue with this test in terms of lab capacity, especially given that it gets scheduled for every defconfig+ variant, which is only going to get worse with fixing it so it runs successfully. However that seems a separate issue to fixing it so that it runs whenever it does get scheduled.

gctucker:
OK, still I think this should be fine to merge now as while it does add a bit of load to the test labs it shouldn't be big enough to cause real congestion.  Optimising the filters to run with fewer builds should still be done like you said, as a follow-up PR.

gctucker (approved):

PR #1145, opened 2022-04-22
T34238 kubernetes templates

gctucker:
Add templates for generating Kubernetes jobs with Python inline.

Depends on #1141 for the setup.py part


gctucker (commented):

gctucker (commented):

PR #1148, opened 2022-04-25
CHROMEOS LAVA flash improvements

nuclearcat:
Based on real testing, to improve reliability of flashing following changes are done:
1)Accept ChromiumOS as full URL, as it might be placed on various resources, not only internally
2)Assume image is compressed (significant space gains)
3)Use compressed scp (few minutes win over 100Mbps-forced port)
4)Flash over ssh, as serial console very unreliable
5)Specify root device for each device type flashing job
6)Include examples for known device types that was flashed successfully

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -173,7 +173,6 @@ test_plans:
     params:
       job_timeout: '60'
       docker_image: 'kernelci/chromeos-flash'
-      chromeos_image: 'chromiumos_test_image-R94-14150.19.0-octopus.bin'
       flashing_kernel: http://storage.chromeos.kernelci.org/images/kernel/v5.17.2-x86_64+x86_chromebook/kernel/bzImage
       flashing_modules: http://storage.chromeos.kernelci.org/images/kernel/v5.17.2-x86_64+x86_chromebook/modules.tar.xz
       flashing_rootfs: http://storage.chromeos.kernelci.org/images/rootfs/chromeos/20220412.0/bullseye-chromeos-flash/amd64/full.rootfs.tar.xz
@@ -653,6 +652,9 @@ device_types:
     arch: x86_64
     boot_method: depthcharge
     filters: *x86-chromebook-filters
+    params:
+      chromeos_image: 'https://storage.chromeos.kernelci.org/images/rootfs/chromeos/crostest007/chromiumos-hatch/amd64/chromiumos_test_image.bin.gz'
+      flashing_device: nvme0n1
 
   asus-C523NA-A20057-coral:
     mach: x86
@@ -823,6 +825,9 @@ device_types:
     arch: x86_64
     boot_method: depthcharge
     filters: *x86-chromebook-filters
+    params:
+      chromeos_image: 'https://storage.chromeos.kernelci.org/images/rootfs/chromeos/crostest007/chromiumos-grunt/amd64/chromiumos_test_image.bin.gz'
+      flashing_device: detect
 
   # Alphabetical order exception: 14 is an int but 12b is a string
   hp-x360-14-G1-sona:
@@ -836,6 +841,9 @@ device_types:
     arch: x86_64
     boot_method: depthcharge
     filters: *x86-chromebook-filters
+    params:
+      chromeos_image: 'https://storage.chromeos.kernelci.org/images/rootfs/chromeos/crostest007/chromiumos-octopus/amd64/chromiumos_test_image.bin.gz'
+      flashing_device: mmcblk0
 
   hp-x360-12b-n4000-octopus: *octopus
 
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -173,8 +173,8 @@ test_plans:
     params:
       job_timeout: '60'
       docker_image: 'kernelci/chromeos-flash'
-      flashing_kernel: http://storage.chromeos.kernelci.org/images/kernel/v5.17.2-x86_64+x86_chromebook/kernel/bzImage
-      flashing_modules: http://storage.chromeos.kernelci.org/images/kernel/v5.17.2-x86_64+x86_chromebook/modules.tar.xz
+      flashing_kernel: https://storage.chromeos.kernelci.org/images/kernel/v5.10.112-x86-chromebook/kernel/bzImage
+      flashing_modules: https://storage.chromeos.kernelci.org/images/kernel/v5.10.112-x86-chromebook/modules.tar.xz
       flashing_rootfs: http://storage.chromeos.kernelci.org/images/rootfs/chromeos/20220412.0/bullseye-chromeos-flash/amd64/full.rootfs.tar.xz
       flashing_ramdisk: http://storage.chromeos.kernelci.org/images/rootfs/chromeos/20220412.0/bullseye-chromeos-flash/amd64/initrd.cpio.gz
 
--- a/config/lava/chromeos/chromeos-flash-template.jinja2
+++ b/config/lava/chromeos/chromeos-flash-template.jinja2
@@ -46,39 +46,16 @@ actions:
     - repository:
         metadata:
           format: Lava-Test Test Definition 1.0
-          name: docker-network
+          name: docker-network-flashing
         run:
           steps:
             - ping -c 1 172.17.0.1
-            - lava-test-case wget --shell  "wget http://172.17.0.1/chromeos/{{ chromeos_image }}"
-            - lava-test-case scp --shell "scp -i /keys/id_rsa_chromeos_flash -o ConnectTimeout=3 -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null {{ chromeos_image }} root@$(lava-target-ip):/root"
+            - lava-test-case wget --shell "wget --no-check-certificate -O chromiumos_test_image.bin.gz {{ chromeos_image }}"
+            - lava-test-case unpack --shell "gzip -d chromiumos_test_image.bin.gz || mv chromiumos_test_image.bin.gz chromiumos_test_image.bin"
+            - lava-test-case scp --shell "scp -C -i /keys/id_rsa_chromeos_flash -o ConnectTimeout=3 -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null chromiumos_test_image.bin root@$(lava-target-ip):/root"
+            - ssh -i /keys/id_rsa_chromeos_flash -o ConnectTimeout=3 -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null root@$(lava-target-ip) "IMAGE_NAME=chromiumos_test_image.bin FLASH_DEVICE={{ flashing_device }} /bin/bash /opt/chromeos/flash"
       from: inline
-      name: docker-network
+      name: docker-network-flashing
       path: inline/docker-network.yaml
-
-- test:
-    timeout:
-      minutes: 60
-    definitions:
-    - repository:
-        metadata:
-          format: Lava-Test Test Definition 1.0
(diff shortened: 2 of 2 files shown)


config/lava/chromeos/chromeos-flash-template.jinja2
```diff
@@ -50,35 +50,12 @@ actions:
         run:
           steps:
             - ping -c 1 172.17.0.1
-            - lava-test-case wget --shell  "wget http://172.17.0.1/chromeos/{{ chromeos_image }}"
```
gctucker, line 53: The reason for using `172.17.0.1` is to flash proprietary Chrome OS images that aren't stored on storage.chromeos.kernelci.org.  I don't think we've completely confirmed yet whether we'll be able to use the Chromium OS builds and not the ones with proprietary components, so we probably should keep this for the time being.

It could be done with a different PR later on, and this PR could be just about using SSH rather than the serial console.

config/core/test-configs.yaml
```diff
@@ -158,16 +158,41 @@ test_plans:
...
       flashing_rootfs: http://storage.chromeos.kernelci.org/images/rootfs/chromeos/20220412.0/bullseye-chromeos-flash/amd64/full.rootfs.tar.xz
       flashing_ramdisk: http://storage.chromeos.kernelci.org/images/rootfs/chromeos/20220412.0/bullseye-chromeos-flash/amd64/initrd.cpio.gz
+      flashing_device: detect
+
+  chromeos-flash-hatch:
+    pattern: 'chromeos/chromeos-flash-template.jinja2'
+    params:
+      job_timeout: '60'
+      docker_image: 'kernelci/chromeos-flash'
+      chromeos_image: 'https://storage.chromeos.kernelci.org/images/rootfs/chromeos/crostest007/chromiumos-hatch/amd64/chromiumos_test_image.bin.gz'
```
gctucker, line 178: How about adding these parameters to the device type definitions instead? That way we would keep a single `chromeos-flash` test plan and the parameters would be directly tied to each Chromebook.

config/lava/chromeos/chromeos-flash-template.jinja2
nuclearcat: With my PR it is still possible to fetch private image, just instead of using hardcoded http://172.17.0.1/chromeos/ we specify full url.
It can be: 
chromeos_image: 'http://172.17.0.1/chromeos/chromiumos_test_image-R94-14150.19.0-octopus.bin'
or
chromeos_image: 'https://storage.chromeos.kernelci.org/images/rootfs/chromeos/crostest007/chromiumos-hatch/amd64/chromiumos_test_image.bin.gz'

I am trying to make templates such way, that we can avoid changing them later. 
But then i need to find way to remove hardcoded filename chromiumos_test_image.bin.gz, or internal image can be moved to something like 'http://172.17.0.1/chromeos/chromiumos_test_image-R94-14150.19.0-octopus/chromiumos_test_image.bin.gz'

config/core/test-configs.yaml
nuclearcat: I will check how to do that

config/lava/chromeos/chromeos-flash-template.jinja2
gctucker: OK, I see.  Maybe use `basename` to get the file name from the URL, or some Python code to handle this with `urllib`?

config/core/test-configs.yaml
gctucker: With the `params` attribute, see this one for example:

```yaml
  meson-g12a-sei510:
    mach: amlogic
    class: arm64-dtb
    boot_method: uboot
    flags: ['fastboot']
    params:
      kernel_addr: '0x01080000'
      dtb_addr: '0x1000000'
      ramdisk_addr: '0x6000000'
      cmdline: '"console=ttyAML0,115200 consoleblank=0 earlycon"'
```

config/lava/chromeos/chromeos-flash-template.jinja2
nuclearcat: I will try to figure that out, so it will work with any url. Filename part is easiest , i can use -O option of wget, but about compressed and uncompressed part it is a bit more tricky

config/core/test-configs.yaml
nuclearcat: Thank you! That will help a lot.

nuclearcat: This part probably improved as suggested

config/lava/chromeos/chromeos-flash-template.jinja2
nuclearcat: Sorted out too. Trick is simple, i just wget -o chromiumos_test_image.bin.gz, then attempt to unpack and if it fails, i assume it is not compressed and just rename it.

gzip -d chromiumos_test_image.bin.gz || mv chromiumos_test_image.bin.gz chromiumos_test_image.bin

Compressed:
https://lava.collabora.co.uk/scheduler/job/6187231
Uncompessed (manually moved to my host)
https://lava.collabora.co.uk/scheduler/job/6187470

nuclearcat:
Second commit tested in LAVA: https://lava.collabora.co.uk/scheduler/job/6187817

config/lava/chromeos/chromeos-flash-template.jinja2
```diff
@@ -50,35 +50,12 @@ actions:
...
-            - lava-test-case scp --shell "scp -i /keys/id_rsa_chromeos_flash -o ConnectTimeout=3 -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null {{ chromeos_image }} root@$(lava-target-ip):/root"
+            - lava-test-case wget --shell "wget --no-check-certificate -O chromiumos_test_image.bin.gz {{ chromeos_image }}"
+            - lava-test-case unpack --shell "gzip -d chromiumos_test_image.bin.gz || mv chromiumos_test_image.bin.gz chromiumos_test_image.bin"
```
mgalka, line 54: I think it's fine for now, but in future we could come up with some more explicit solution for checking if the image is compressed. Maybe it's worth filing an issue about it.

mgalka (approved):

gctucker:
Please rebase

nuclearcat:
Rebased

config/lava/chromeos/chromeos-flash-template.jinja2
gctucker, line 54: That's not bullet-proof but I guess it's OK for now in practice.

config/lava/chromeos/chromeos-flash-template.jinja2
```diff
@@ -50,35 +50,12 @@ actions:
...
+            - lava-test-case scp --shell "scp -C -i /keys/id_rsa_chromeos_flash -o ConnectTimeout=3 -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null chromiumos_test_image.bin root@$(lava-target-ip):/root"
+            - ssh -i /keys/id_rsa_chromeos_flash -o ConnectTimeout=3 -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null root@$(lava-target-ip) "IMAGE_NAME=chromiumos_test_image.bin FLASH_DEVICE={{ flashing_device }} /bin/bash /opt/chromeos/flash"
       from: inline
       name: docker-network
```
gctucker, line 58:
```suggestion
      name: chromeos-flash
```

gctucker (commented):
A minor glitch with the test names, it should be `chromeos-flash`

config/lava/chromeos/chromeos-flash-template.jinja2
nuclearcat: Done

config/lava/chromeos/chromeos-flash-template.jinja2
```diff
@@ -50,35 +50,12 @@ actions:
...
-            - ls -l /root
-            - IMAGE_NAME={{ chromeos_image }}  /bin/bash /opt/chromeos/flash
-      lava-signal: kmsg
+            - lava-test-case wget --shell "wget --no-check-certificate -O chromiumos_test_image.bin.gz {{ chromeos_image }}"
+            - lava-test-case unpack --shell "gzip -d chromiumos_test_image.bin.gz || mv chromiumos_test_image.bin.gz chromiumos_test_image.bin"
+            - lava-test-case scp --shell "scp -C -i /keys/id_rsa_chromeos_flash -o ConnectTimeout=3 -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null chromiumos_test_image.bin root@$(lava-target-ip):/root"
+            - ssh -i /keys/id_rsa_chromeos_flash -o ConnectTimeout=3 -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null root@$(lava-target-ip) "IMAGE_NAME=chromiumos_test_image.bin FLASH_DEVICE={{ flashing_device }} /bin/bash /opt/chromeos/flash"
       from: inline
       name: chromeos-flash
-      path: inline/chromeos-flash.yaml
```
gctucker, line 82: Looks like this was not intended?

gctucker: The "repository" name is also still called `docker-network` instead of `chromeos-flash` (but I can't put a comment as it's outside the diff).

nuclearcat: This whole section is removed, because i changed flashing from serial to ssh.
As we have hanging console issue, ssh is way more reliable.

nuclearcat: I will update names in patch to make it less confusing

config/lava/chromeos/chromeos-flash-template.jinja2
```diff
@@ -55,7 +55,7 @@ actions:
             - lava-test-case scp --shell "scp -C -i /keys/id_rsa_chromeos_flash -o ConnectTimeout=3 -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null chromiumos_test_image.bin root@$(lava-target-ip):/root"
             - ssh -i /keys/id_rsa_chromeos_flash -o ConnectTimeout=3 -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null root@$(lava-target-ip) "IMAGE_NAME=chromiumos_test_image.bin FLASH_DEVICE={{ flashing_device }} /bin/bash /opt/chromeos/flash"
       from: inline
-      name: docker-network
+      name: chromeos-flash
```
gctucker, line 58: This should have been in the first patch.

config/lava/chromeos/chromeos-flash-template.jinja2
nuclearcat: Now i fixed both patches properly. I removed whole section chromeos-flash that works over console, and added "             - ssh -i /keys/id_rsa_chromeos_flash -o ConnectTimeout=3 -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null root@$(lava-target-ip) "IMAGE_NAME=chromiumos_test_image.bin FLASH_DEVICE={{ flashing_device }} /bin/bash /opt/chromeos/flash"" to docker-network, so it does flashing too, and to remove confusion renamed it to docker-network-flashing

gctucker: But it should just be `chromeos-flash`, right? The `docker-network` thing is just some leftover from initial debugging things to check whether Docker on the dispatcher could reach the Chromebook.

gctucker:
Alright never mind, let's merge this and clean things up later on if need.

gctucker:
Right, there was still a minor issue so #1166

PR #1150, opened 2022-04-25
T34238 kernelci Docker image

gctucker:
Add a `kernelci/kernelci` Docker image with the `kernelci` Python package installed with its dependencies as well as the config files copied to `/etc/kernelci`.

### DIFF
--- a/config/docker/kernelci/Dockerfile
+++ b/config/docker/kernelci/Dockerfile
@@ -1,0 +1,31 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+#
+# Copyright (C) 2021, 2022 Collabora Limited
+# Author: Guillaume Tucker <guillaume.tucker@collabora.com>
+
+FROM python:3.10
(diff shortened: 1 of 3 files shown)


config/docker/kernelci/Dockerfile
```diff
@@ -0,0 +1,31 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+#
+# Copyright (C) 2021, 2022 Collabora Limited
```
nuclearcat, line 3: Is it 2021?

nuclearcat (approved):
Builds fine locally for me

config/docker/kernelci/Dockerfile
gctucker: I should have mentioned that this Dockerfile was based on the one used by kernelci-pipeline which was written in 2021.  I'll add that in the commit message.

gctucker:
> Builds fine locally for me

This has also been built and used with the new pipeline on staging.

PR #1151, opened 2022-04-26
T34835 rename kernelci.data to kernelci.db

gctucker:
Rename `kernelci.data` to `kernelci.db` to avoid ambiguity with config data files and make it more consistent with the `--db-config` command line arguments.


nuclearcat:
Small typo in "[rename from kernrelci.config.data]" commit header

gctucker:
OK I think I'm done with rewriting this one. Rather than renaming `kernelci.config` to install config files in the kernelci package directory, only `kernelci.data` is renamed for consistency.  The config files are now being installed in the Dockerfile in `/etc/kernelci` outside of `setup.py` with #1150.

PR #1152, opened 2022-04-27
test-configs: Constrain defconfigs ltp-pty test is run on

broonie:
The PTY tests are quite slow to run, taking over 15 minutes on most systems,
but we schedule them for every defconfig variant we build which consumes a
lot of time on boards and leads to huge backlogs. Since this suite is not
expected to have much config dependency this is very wasteful add a
passlist which restricts it to just core defconfigs.

We probably want to look at doing this at a higher level, though for
quicker to run tests there's a tradeoff where we might catch issues and
some of the config variants like page size changes are likely to have an
impact on some test suites so let's do some focused restrictions for now.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -537,6 +537,12 @@ test_plans:
       tst_cmdfiles: "pty"
       job_timeout: '25'
       priority: 0
+    filters:
+      - passlist:
+          defconfig:
+          - 'defconfig'
+          - 'multi_v5_defconfig'
+          - 'multi_v7_defconfig'
 
   ltp-sched:
     <<: *ltp


gctucker:
Please rebase.

Note that the `passlist` filter will match if the string in the filter is part of the full defconfig string.  To only run with exact defconfig and no added fragments, a `regex` or `combination` filter should be used.  What's strange is that the default filters for tests already only allow the main defconfig for each arch so I'm not sure where the other jobs are coming from.  I'll probably take a closer look at that next week.

nuclearcat:
Probably filter doesnt work as designed?
For example it looks like scheduled to lot of configurations: https://staging.kernelci.org/test/job/kernelci/branch/staging-mainline/kernel/staging-mainline-20220628.0/plan/ltp-pty/

gctucker:
The `passlist` defconfig will match if at least one if the values listed is present in the input string. So, anything with `config` in it will match. To force an exact match with `config` no more no less then a `regex` filter can be used instead.

That feels like writing documentation. Oh wait: https://kernelci.org/docs/core/config/filters/
...time to review and improve that page.

broonie:
We do still seem to be running too many of these.

gctucker:
Probably because `defconfig` will essentially match all the config file names.  Maybe you need to use `regex` filters or something else.  You can get the list of jobs run with a particular build JSON using `kci_test list_jobs` to debug this locally without actually running anything.

gctucker:
Closing as no update.  I've created #1883 to keep track of the issue.

PR #1157, opened 2022-04-29
config/scripts: set exit status to 0 even it test failed

gctucker:
Set the exit status to 0 in base-python.jinja2 even if the test case
failed as the script ran successfully.

An exit status of 1 should only mean that the script failed to run,
for example if no test result could be sent to the API.  Some
follow-up changes might be required to handle error propagation
correctly if defaulting to Python exceptions is not enough.


mgalka (approved):
LGTM

gctucker:

Issue #880, opened 2021-11-17
Add shell runtime


gctucker:
Add a way to generate jobs to run in a local shell and execute them.

This would then be used to build kernels, run static checks and other things that could be run in any environment.  The next steps would be to have a Docker runtime, to generate some commands and run them in a container, and a Kubernetes runtime to do the same but submit the YAML definition to a cluster.

gctucker:
All the changes related to this issue are being first added to the `api-pipeline` topic branch:
https://github.com/kernelci/kernelci-core/tree/api-pipeline

PR #1162, opened 2022-05-03
kernelci.lab.shell: drop get_process

gctucker:
Drop the get_process argument in .submit() as it's specific to the
shell runtime environment and the default behaviour should be to
return the process object.

Fixes: #880


gctucker (commented):
Tested OK on staging combined with https://github.com/kernelci/kernelci-pipeline/pull/58

PR #1163, opened 2022-05-03
Lower priorities for longer running tests

broonie:
This branch aims to mitigate the impact of longer running tests on labs which are overloaded by configuring lower priorities for jobs which are expected to have longer runtimes, trying to help us get a wider range of coverage more promptly with the bandwidth on the boards we have.

This won't fix anything if the lab is just completely overwhelmed but for cases where there are peaks in load it should help us get a wider set of results through more promptly as we catch up and if the lab has a queue timeout it should be these slower jobs that get timed out.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -474,6 +474,7 @@ test_plans:
       <<: *ltp-params
       tst_cmdfiles: "pty"
       job_timeout: '25'
+      priority: 0
 
   ltp-timers:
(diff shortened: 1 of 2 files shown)


gctucker:
OK, this is the kind of issue that will eventually get resolved by using dynamic job scheduling based on the labs' utilisation. I think it's fine to tweak priorities and timeout settings in the meantime to experiment.

gctucker (approved):

PR #1164, opened 2022-05-04
test-configs.yaml: Enable hardware coverage for BeagleBone Black

broonie:
The BeagleBone Black has both a sound card and a crypto engine so
enable both the ALSA kselftests and the LTP crypto tests to cover
both of these.  Since these tests run fairly quickly the load on
labs should be reasonably light.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -1934,6 +1934,8 @@ test_configs:
       - baseline
       - baseline-cip-nfs
       - baseline-nfs
+      - kselftest-alsa
+      - ltp-crypto
       - ltp-ipc
 
   - device_type: cubietruck


broonie:
ALSA job running OK in staging https://staging.kernelci.org/test/plan/id/6279bca05aabdecb4be12bce/ (I have a patch in flight on the kernel side to enable the drivers by default in multi_v7_defconfig).
ltp-crypto running OK in staging: https://staging.kernelci.org/test/plan/id/6279b8f3a2b3d98714e12bd6/

broonie:
ALSA job now that the drivers are enabled in the defconfig: https://staging.kernelci.org/test/plan/id/627bfbfc4cd9c017b8e12bce/

gctucker (approved):

PR #1165, opened 2022-05-06
 build-configs.yaml: Add SERIAL_8250 to x86-chromebook fragment #1165

nuclearcat:
For initial testing of real hardware and qemu we need serial console support,
through which we can check if the OS has booted, if there is a login prompt,
and it is better to carry out initial diagnostics over the serial console,
since the network connection may not be available.

### DIFF
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -245,6 +245,7 @@ fragments:
   x86-chromebook:
     path: "kernel/configs/x86-chromebook.config"
     configs:
+      - 'CONFIG_SERIAL_8250=y'
       - 'CONFIG_SERIAL_8250_DW=y'
       - 'CONFIG_X86_AMD_PLATFORM_DEVICE=y'
       - 'CONFIG_MFD_INTEL_LPSS_PCI=y'


config/core/build-configs.yaml
```diff
@@ -276,6 +276,7 @@ fragments:
       - 'CONFIG_PINCTRL_AMD=y'
       - 'CONFIG_NVME_CORE=y'
       - 'CONFIG_BLK_DEV_NVME=y'
+      - 'CONFIG_SERIAL_8250=y'
```
gctucker, line 279: Maybe add it at the top of the list before `CONFIG_SERIAL_8250_DW=y`?

gctucker:
This config is already enabled in mainline though, it's in `x86_64_defconfig`:
https://storage.kernelci.org/mainline/master/v5.18-rc5-115-gfe27d189e3f4/x86_64/x86_64_defconfig+x86-chromebook/gcc-10/config/kernel.config

Is some Chrome OS config fragment disabling it?

nuclearcat:
It looks like when we have builds for ChromeOS, for example "cros://chromeos-5.10/x86_64/chromiumos-x86_64.flavour.config" - it doesnt apply x86_64_defconfig, it just apply fragments from chromiumos and runs olddefconfig on top of it.

gctucker:
OK I see. So please explain in the git commit message why adding this here is needed even though it's already in `x86_64_defconfig`.  Also, I think the git commit subject is a bit ambiguous.  Something like `add SERIAL_8250 to x86-chromebook fragment` would be clearer imho.

config/core/build-configs.yaml
nuclearcat: Done

PR #1170, opened 2022-05-09
CHROMEOS test-configs-chromeos.yaml: split tests from main file

gctucker:
Create a test-configs-chromeos.yaml config file with all the test
configs related to Chrome OS to avoid conflicts with main branch and
simplify maintenance.

Signed-off-by: Guillaume Tucker <guillaume.tucker@collabora.com>

PR #1172, opened 2022-05-09
T35268 tast test plans

gctucker:
Simplify the YAML syntax and template name for the QEMU Tast tests.
This should result in the same tests being run.

Depends on #1170


gctucker:
Tested OK on staging: https://chromeos.kernelci.org/test/plan/id/627968adf7af7ee87406ce8e/

nuclearcat (approved):
LGTM, just naming is changed, i am not sure how strictly we need to follow naming convention. Usually opensource builds called ChromiumOS and proprietary builds are ChromeOS.
But i dont want to become too pedantic like Richard Stallman with GNU(!!!)/Linux :)

Issue #235, opened 2019-11-08
Please remove non-working builds from android-3.18


lag-linaro:
None of the builds below are supported. Please prevent them from being initiated.

allnoconfig ‐ arc
bigsur_defconfig ‐ mips
fpga_defconfig ‐ arc
fpga_noramfs_defconfig ‐ arc
nlm_xlr_defconfig ‐ mips
sb1250_swarm_defconfig ‐ mips
sead3micro_defconfig ‐ mips
tinyconfig ‐ arc

lag-linaro:
Please. :smiley:

lag-linaro:
These non-legitimate builds are still failing.

Please could someone either remove them, or show me how.

gctucker:
I wonder if `lee_android_3.18` is still being used.

nuclearcat:
Didn't noticed that, i will check who is author and will ask

lag-linaro:
You can strip out all of my `*android_3.18` builds, thanks.

gctucker:
> You can strip out all of my `*android_3.18` builds, thanks.

Ah OK, thanks for confirming.

PR #1176, opened 2022-05-10
build-configs.yaml: remove lee_android_3.18 build config

nuclearcat:
Remove lee_android_3.18 as requested in https://github.com/kernelci/kernelci-core/issues/235#issuecomment-1122124032

### DIFF
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -61,9 +61,6 @@ trees:
   kselftest:
     url: 'https://git.kernel.org/pub/scm/linux/kernel/git/shuah/linux-kselftest.git'
 
-  lee:
-    url: "https://git.kernel.org/pub/scm/linux/kernel/git/lee/linux.git"
-
   linusw:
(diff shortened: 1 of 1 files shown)


gctucker:
This is not the same issue - please see my comments.  The lee_android_3.18 build config is not an Android tree.

gctucker:
> This is not the same issue - please see my comments. The lee_android_3.18 build config is not an Android tree.

As confirmed by Lee in an other comment, this can be dropped.

config/core/build-configs.yaml
```diff
@@ -706,11 +706,6 @@ build_configs:
     <<: *kselftest-tree
     branch: 'next'
 
-  lee_android_3.18:
-    tree: lee
```
gctucker, line 710: This branch was the only one using the `lee` tree, so we might also drop it.

@lag-linaro Are there other branches in your tree you would like to have covered by KernelCI?

lag-linaro: No, not right now, thank you.

nuclearcat: I will update PR accordingly.

gctucker (approved):

PR #1181, opened 2022-05-11
lava_rest.py: Add debug in case LAVA request failed

nuclearcat:
The current way of showing error on invalid http reply status
provides insufficient information about errors,
in some cases it is important to know why LAVA did not like the request,
and which function failed.


kernelci/lab/lava/lava_rest.py
```diff
@@ -38,6 +38,8 @@ class LavaRest(LavaAPI):
 
     def _get_response(self, url):
         resp = self._server.session.get(url)
+        if not resp.ok:
+            print(f'_get_response() failed, code: {resp.status_code} body: {resp.text}')
```
gctucker, line 42: We shouldn't be printing things in the library methods, instead we should be returning the string with the error or raising an exception with it.  The issue with `resp.raise_for_status()` is that it doesn't include the detailed error message from the response, just a generic one.

nuclearcat: Updated, can you please check if it is better now?

gctucker (commented):
I've tried it locally by submitting a job with a device type which doesn't exist in the lab:
```
$ ./kci_test submit --lab-config=lab-collabora-staging --jobs=data/jobs/*
HTTPError: 400 Client Error: Bad Request for url: https://staging.lava.collabora.dev/api/v0.2/jobs/?format=json&limit=256
Reply body: {"message":"Devices unavailable: No devices of type hp-x360-12b-ca0010nr-n4020-octopus are available."}:
```
So it worked, but maybe this would have been easier to read:
```
$ ./kci_test submit --lab-config=lab-collabora-staging --jobs=data/jobs/*
HTTPError: 400 Client Error: Bad Request for url: https://staging.lava.collabora.dev/api/v0.2/jobs/?format=json&limit=256
Devices unavailable: No devices of type hp-x360-12b-ca0010nr-n4020-octopus are available.
```
We could be printing the message if there is one in the request object.  Also, errors should be printed on `stderr` with `print("blah", file=sys.stderr)`.

nuclearcat:
I added handling of json if it contains message, so it will display it is as you suggested, and few other following situations:
- in some cases message key doesn't exist in json, for example 403, then i do json.dumps {"detail": "Invalid token."}
 - display if json cannot be parsed (sometimes happens) 
- non-json content (for example in case LAVA is down, gateway timeout or php error) - will be displayed old way as in first version

gctucker:
Please fix the GitHub checks.

nuclearcat:
> Please fix the GitHub checks.

pylint... :facepalm: sorry :(
fixing

kci_test
```diff
@@ -284,8 +284,22 @@ class cmd_submit(Command):
             try:
                 job_id = lab_api.submit(path)
                 print("{} {}".format(job_id, path))
+            except requests.exceptions.HTTPError as e:
+                print(f'HTTPError: {e}')
+                r = e.response
+                if r.headers.get('content-type') == 'application/json':
+                    try:
+                        jobj = json.loads(r.text)
+                        if "message" in jobj.keys():
```
gctucker, line 293:
```suggestion
                        if "message" in jobj:
```

kci_test
```diff
@@ -284,8 +284,22 @@ class cmd_submit(Command):
             try:
                 job_id = lab_api.submit(path)
                 print("{} {}".format(job_id, path))
+            except requests.exceptions.HTTPError as e:
+                print(f'HTTPError: {e}')
```
gctucker, line 288: Do we still need this with the extra prints we're now adding below?

kci_test
gctucker: Or:
```
msg = jobj.get('message')
if msg:
  # print something with msg
```

kci_test
nuclearcat: Yes, in some cases it might be non-json reply gateway error (then will go over Reply body), might be truncated/broken json, and non-HTTPError (tcp connection failure, DNS failure). Better to have max verbose errors (considering risks, for example to not leak secrets).
But what i forgot - to set res/return code, added now.

kci_test
nuclearcat: Done

gctucker:
Well, it didn't quite work:
```
$ ./kci_test submit --lab-config=lab-collabora --jobs=data/jobs/*
HTTPError: 400 Client Error: Bad Request for url: https://lava.collabora.co.uk/api/v0.2/jobs/?format=json&limit=256
  File "/home/gtucker/project/kernelci/kernelci-core/./kci_test", line 285, in __call__
    job_id = lab_api.submit(path)
  File "/home/gtucker/project/kernelci/kernelci-core/kernelci/lab/lava/__init__.py", line 70, in submit
    job_id = self._submit(job)
  File "/home/gtucker/project/kernelci/kernelci-core/kernelci/lab/lava/lava_rest.py", line 107, in _submit
    resp.raise_for_status()
  File "/usr/lib/python3/dist-packages/requests/models.py", line 943, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 400 Client Error: Bad Request for url: https://lava.collabora.co.uk/api/v0.2/jobs/?format=json&limit=256

  File "/home/gtucker/project/kernelci/kernelci-core/./kci_test", line 315, in <module>
  File "/home/gtucker/project/kernelci/kernelci-core/./kci_test", line 294, in __call__
    print(jobj[message], file=sys.stderr)
NameError: name 'message' is not defined
```
To reproduce this error I'm generating a test definition with a device type which doesn't exist in the lab.

nuclearcat:
Created test web server and simulated all scenarios, sorry for silly mistake, will try to test in future PR better.
Unit tests will take too much time,so i just made fake web server and simulated lab.
Added also handling if returned content type is application/json and is "null".

gctucker:
FYI I tested it with the real Collabora LAVA lab using an API token.

nuclearcat:
I just did that too with similar approach:
`./kci_test submit --lab-config=lab-collabora --jobs=jobs/*
HTTPError: 400 Client Error: Bad Request for url: https://lava.collabora.co.uk/api/v0.2/jobs/?format=json&limit=256
Devices unavailable: Device type 'hp-x360-12b-ca0010nr-n4020-octopus1' is unavailable.
`

gctucker (commented):
OK I've tested this again locally and it's working well:
```
HTTPError: 400 Client Error: Bad Request for url: https://lava.collabora.co.uk/api/v0.2/jobs/?format=json&limit=256
Devices unavailable: Not enough devices available matching the requested tags.
```
The commit message should be changed to `kci_test:` as the changes are not in `lava_rest.py` any more.

nuclearcat:
Updated

gctucker (approved):

PR #1183, opened 2022-05-12
CHROMEOS config/lava: wget -q the Chrome OS image when flashing

gctucker:
Add wget -q option when downloading a Chrome OS image to flash as this
otherwise causes a lot of noise in the LAVA logs and it's not useful
as progress isn't shown dynamically anyway.


mgalka (approved):
This is actually a very good idea.

PR #1184, opened 2022-05-12
CHROMEOS Add test definition to install modules from the fixed reference kernel

gctucker:
Add a test plan variant `chromeos-install-modules-fixed` to install modules from the fixed reference kernel as this is needed to run the `-boot-fixed` and `-tast-fixed` test plans.


gctucker:
Tested OK: https://lava.collabora.co.uk/scheduler/job/6349778

gctucker:
I also manually checked the modules were correctly installed over SSH.

gctucker:
There appears to be an issue when booting Chrome OS but it seems to be related to how the image is flashed rather than how the modules are installed.  So merging this now to avoid conflicts with next PRs.

PR #1185, opened 2022-05-12
test-configs: Add etnaviv GPU test stanza

broonie:
Add some IGT testing for the Vivante GPUs used in many SoCs, including some
i.MX generations. There aren't any etnaviv specific tests in upstream igt
yet.

I'll add these to the Udoo devices once support for those is merged, they
could also run this on other i.MX6 devices.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -199,6 +199,18 @@ test_plans:
     filters:
       - passlist: {defconfig: ['amdgpu']}
 
+  igt-gpu-etnaviv:
+    <<: *igt
+    params:
+      drm_driver: "etnaviv"
+      tests: >
+        core_auth
+        core_getclient
+        core_getstats
+        core_getversion
+        core_setmaster
+        core_setmaster_vs_auth
+
   igt-gpu-i915:
     <<: *igt
     params:
@@ -2247,6 +2259,7 @@ test_configs:
     test_plans:
       - baseline
       - baseline-nfs
+      - igt-gpu-etnaviv
       - kselftest-alsa
       - ltp-crypto
 
@@ -2279,6 +2292,7 @@ test_configs:
     test_plans:
       - baseline
       - baseline-nfs
+      - igt-gpu-etnaviv
       - kselftest-alsa
       - ltp-crypto
 


nuclearcat (approved):
Manually tested, added to one of devices and generated job definition.

gctucker:
> Manually tested, added to one of devices and generated job definition.

Maybe make a PR for that to actually enable the tests to run if we already have a capable platform online somewhere?

nuclearcat:
> > Manually tested, added to one of devices and generated job definition.
> 
> Maybe make a PR for that to actually enable the tests to run if we already have a capable platform online somewhere?

My guess it is question to @broonie , i am not aware if we have such GPU anywhere and if all support is merged.

gctucker:
> > > Manually tested, added to one of devices and generated job definition.
> > 
> > 
> > Maybe make a PR for that to actually enable the tests to run if we already have a capable platform online somewhere?
> 
> My guess it is question to @broonie , i am not aware if we have such GPU anywhere and if all support is merged.

Oh so you just generated a dummy job, right. Well I guess this will remain stale for a while now, either a dust-gathering PR or dust-gathering YAML config entry if we merge it.

nuclearcat:
Maybe i can add skip-staging so other PR will be tested on staging?

gctucker:
Yes, if it's causing some conflicts then I think we can just skip it.

gctucker:
FYI the Udoo boards have been enabled already. So now we need to wait for the IGT tests to be added, and then this should all fall in place nicely.

gctucker:
@nuclearcat I've double checked and this PR is not conflicting with any other PR so while it's not doing anything yet we don't need to add the `staging-skip` label.

nuclearcat:
I think similar to https://github.com/kernelci/kernelci-core/pull/1186 this test need to be added to lab configs passlist

config/core/test-configs.yaml
```diff
@@ -2247,6 +2259,7 @@ test_configs:
     test_plans:
       - baseline
       - baseline-nfs
+      - igt-gpu-etnaviv
```
gctucker, line 2262: OK: https://staging.kernelci.org/test/plan/id/62b800c71f232e49cd74f1ea/

config/core/test-configs.yaml
```diff
@@ -2279,6 +2292,7 @@ test_configs:
...
```
gctucker, line 2295: https://staging.kernelci.org/test/plan/id/62b800c6c71eb4519c74f306/

gctucker (approved):

PR #1186, opened 2022-05-12
test-configs: Provide igt-kms coverage for lima driver

broonie:
Enable igt-gpu for systems with the lima driver, enabled on pine64-plus
initially. There are no lima specific tests in upstream igt so it's just
the core tests.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -212,6 +212,18 @@ test_plans:
         gem_create
         i915_selftest
 
+  igt-gpu-lima:
+    <<: *igt
+    params:
+      drm_driver: "lima"
+      tests: >
+        core_auth
+        core_getclient
+        core_getstats
+        core_getversion
+        core_setmaster
+        core_setmaster_vs_auth
+
   igt-gpu-panfrost:
     <<: *igt
     params:
@@ -2821,6 +2833,7 @@ test_configs:
     test_plans:
       - baseline
       - baseline-nfs
+      - igt-gpu-lima
       - kselftest-arm64
       - kselftest-cpufreq
       - kselftest-filesystems


gctucker:
Please rebase.

nuclearcat:
Please rebase, so we can test it in staging

broonie:
Rebased, though the rebase didn't hit any conflicts...

nuclearcat:
> Is there any prospect of this getting merged? This has been sitting for more than a month and I've had to rebase twice now...

Yes, after it hits staging and get tested. It was failing to pull to staging cause of conflicts, we are trying to shorten PR queue now.

nuclearcat:
I think we need to add in lab-configs.yaml / labs in lab-baylibre or/and lab-broonie in filter passlist test igt-gpu-lima, otherwise it will not generate test definition for for this lab.

Something like
'''
diff --git a/config/core/lab-configs.yaml b/config/core/lab-configs.yaml
index 3a71df4e..5cb9c1e5 100644
--- a/config/core/lab-configs.yaml
+++ b/config/core/lab-configs.yaml
@@ -14,6 +14,7 @@ labs:
             - baseline
             - kselftest
             - preempt-rt
+            - igt-gpu-lima
 
   lab-broonie:
     lab_type: lava.lava_xmlrpc
'''

broonie:
My lab is not supposed to have any lab wide filters other than excluding the Android trees?

nuclearcat:
I guess its there since while, as i see similar way other tests was added:
https://github.com/kernelci/kernelci-core/commit/38d927bec4b60e0b1bc87ac3a21d1257f2a6c5c7

broonie:
Oh, we need to list every test suite for every lab that can run it. Wonderful :(

nuclearcat:
I think if you want to block just android, you can do similar to lab-collabora. @gctucker please correct me if i am wrong.

broonie:
https://github.com/kernelci/kernelci-core/pull/1272

broonie:
I'm confused here. The only reason I would have a list at all would be because I had to - looking at the history I've only ever added things.

gctucker:
> Oh, we need to list every test suite for every lab that can run it. Wonderful :(

You need to do that with a `passlist`, since that's what it's for. But you don't have to use a `passlist` filter.

gctucker:
> My lab is not supposed to have any lab wide filters other than excluding the Android trees?

Then drop the test plan filters and just have a `blocklist` for the tree.  Like @nuclearcat wrote, it's like this for the Collabora labs:
```yaml
    filters: &collabora-filters
      - blocklist:
          tree: [android]
```
(The `&collabora-filters` anchor is there to be able to reuse it with `lab-collabora-staging`, but you can just copy-paste these filters for your lab)

gctucker:
> I'm confused here. The only reason I would have a list at all would be because I had to - looking at the history I've only ever added things.

You've mostly been adding `kselftest` plans and `kselftest` is already in your `passlist` filter, which is why it just worked before.

Mind you, if you drop the `passlist` with test plans then you might have more tests scheduled in your lab. But then you can add a `blocklist` if there are any tests you definitely don't want to have running in your lab (say, some long-running ones... I can't think of any that would be an issue right now).

broonie:
No, I mean that the only reason I've added a passlist for my lab at all is so that testsuites can run - looking at the first commit adding the passlist it was to enable kselftest and ltp, previously there was no filtering at all for my lab. Is there some default list for labs that don't have any filters at all lurking somewhere? That would explain why the collabora lab is working...

config/core/test-configs.yaml
```diff
@@ -2821,6 +2833,7 @@ test_configs:
     test_plans:
       - baseline
       - baseline-nfs
+      - igt-gpu-lima
```
gctucker, line 2836: OK: https://staging.kernelci.org/test/plan/id/62b8b51f18005ae9f474f1cc/

gctucker (approved):

PR #1187, opened 2022-05-13
CHROMEOS add tags for octopus devices

gctucker:
Add `device_id` template parameter and define it for octopus devices so Chrome OS jobs run on only particular devices.

Depends on #1195

PR #1192, opened 2022-05-13
CHROMEOS Add --device-ids argument

mgalka:
Add --device-id option to kci_test commandline tool to add explicitly devices 
on which the test plan should be executed. 
The option produces device_ids argument being a list of device 
identifiers.
The --device-id commandline option can be repeated multiple times
providing single device id per occurrence.

Add support to specify device ids list to the Chrome OS related job templates based on device_ids list provided.

### DIFF
--- a/config/lava/chromeos/chromeos-boot-chromebook-template.jinja2
+++ b/config/lava/chromeos/chromeos-boot-chromebook-template.jinja2
@@ -3,6 +3,9 @@
 {{ super () }}
 context:
   extra_kernel_args: console_msg_format=syslog cros_secure cros_debug root=/dev/{{ flashing_device }}p3
+{%- if device_id %}
+tags: ['{{ device_id }}']
+{%- endif %}
 {% endblock %}
 
 {% block actions %}
--- a/config/lava/chromeos/chromeos-flash-template.jinja2
+++ b/config/lava/chromeos/chromeos-flash-template.jinja2
@@ -3,6 +3,9 @@
 {{ super () }}
 context:
   extra_kernel_args: console_msg_format=syslog cros_secure cros_debug
+{%- if device_id %}
+tags: ['{{ device_id }}']
+{%- endif %}
 {% endblock %}
 
 {% block actions %}
--- a/config/lava/chromeos/chromeos-install-modules-template.jinja2
+++ b/config/lava/chromeos/chromeos-install-modules-template.jinja2
@@ -3,6 +3,9 @@
 {{ super () }}
 context:
   extra_kernel_args: console_msg_format=syslog cros_secure cros_debug
+{%- if device_id %}
+tags: ['{{ device_id }}']
+{%- endif %}
 {% endblock %}
 
 {% block actions %}
--- a/config/lava/chromeos/chromeos-tast-qemu.jinja2
+++ b/config/lava/chromeos/chromeos-tast-qemu.jinja2
@@ -1,6 +1,9 @@
 {% extends 'base/kernel-ci-base.jinja2' %}
 {% block main %}
 {{ super () }}
+{%- if device_id %}
+tags: ['{{ device_id }}']
+{%- endif %}
 {% endblock %}
 
 {% block actions %}
--- a/config/lava/chromeos/chromeos-tast-template.jinja2
+++ b/config/lava/chromeos/chromeos-tast-template.jinja2
@@ -2,6 +2,9 @@
 {% block main %}
 {% do context.update({"extra_kernel_args": "console_msg_format=syslog earlycon cros_debug cros_secure root=/dev/mmcblk0p3" }) %}
 {{ super() }}
+{%- if device_id %}
+tags: ['{{ device_id }}']
+{%- endif %}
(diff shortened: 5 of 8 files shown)


mgalka:
Tested with `kci_test generate`

```bash
$ ./kci_test generate --plan chromeos-install-modules --install-path /path/to/chromeos-install/ --output /output/path --target hp-x360-12b-n4000-octopus --device-id cbg-1
```

```
...
context:
  extra_kernel_args: console_msg_format=syslog cros_secure cros_debug

tags: ['cbg-1']




actions:
- deploy:
    namespace: install_modules
    kernel:
      url: http://storage.chromeos.kernelci.org/images/kernel/v5.17.2-x86_64+x86_chromebook/kernel/bzImage
...
```


 ```bash
$  ./kci_test generate --plan chromeos-install-modules --install-path /path/to/chromeos-install/ --output /output/path --target hp-x360-12b-n4000-octopus --device-id my-device-1 --device-id my-device-2
```

```
...
context:
  extra_kernel_args: console_msg_format=syslog cros_secure cros_debug

tags: ['my-device-1', 'my-device-2']




actions:
- deploy:
    namespace: install_modules
    kernel:
      url: http://storage.chromeos.kernelci.org/images/kernel/v5.17.2-x86_64+x86_chromebook/kernel/bzImage
...
```

**Review by gctucker — COMMENTED** [review]
I think there's the problem with calling the option "device id" and then allowing several of them, as an identifier is expected to be unique for each device.  So either we have `--device-id` and accept only one string rather than an iterable, or we have something else such as `--device-attrs` with a list of attributes to match some device.  The latter would allow things such as matching an SoC family etc. which we can't deal with at the moment.  So I believe it's better to have `--device-id` with a single value for the time being as that's not ambiguous.

File: config/lava/chromeos/chromeos-boot-chromebook-template.jinja2 [hunk 1]
```diff
@@ -3,6 +3,9 @@
 {{ super () }}
 context:
   extra_kernel_args: console_msg_format=syslog cros_secure cros_debug root=/dev/{{ flashing_device }}p3
+{% if device_ids %}
```
gctucker, line 6:
```suggestion
{%- if device_ids %}
```

config/lava/chromeos/chromeos-flash-template.jinja2
```diff
@@ -3,6 +3,9 @@
 {{ super () }}
 context:
   extra_kernel_args: console_msg_format=syslog cros_secure cros_debug
+{% if device_ids %}
```
gctucker, line 6:
```suggestion
{%- if device_ids %}
```

config/lava/chromeos/chromeos-install-modules-template.jinja2
```diff
@@ -3,6 +3,9 @@
 {{ super () }}
 context:
   extra_kernel_args: console_msg_format=syslog cros_secure cros_debug
+{% if device_ids %}
```
gctucker, line 6:
```suggestion
{%- if device_ids %}
```

config/lava/chromeos/chromeos-tast-qemu.jinja2
```diff
@@ -1,6 +1,9 @@
 {% extends 'base/kernel-ci-base.jinja2' %}
 {% block main %}
 {{ super () }}
+{% if device_ids %}
```
gctucker, line 4:
```suggestion
{%- if device_ids %}
```

config/lava/chromeos/chromeos-tast-template.jinja2
```diff
@@ -2,6 +2,9 @@
 {% block main %}
 {% do context.update({"extra_kernel_args": "console_msg_format=syslog earlycon cros_debug cros_secure root=/dev/mmcblk0p3" }) %}
 {{ super() }}
+{% if device_ids %}
```
gctucker, line 5:
```suggestion
{%- if device_ids %}
```

gctucker (commented):
Minor tweak to reduce the amount of blank lines in the generated YAML

gctucker:
Tested OK by hand: https://lava.collabora.co.uk/scheduler/job/6402082

kernelci/test.py
```diff
@@ -174,5 +175,6 @@ def _get_compression(url):
 
     params.update(plan_config.params)
     params.update(target.params)
+    params.update({'device_id': device_id})
```
gctucker, line 178: Or simply `params['device_id'] = device_id` since it's a single value.

gctucker: Actually, this is better:
```suggestion
    if device_id:
        params['device_id'] = device_id
```
This way, if there is already a `device_id` set by `target.params` or something else then it only get overridden if there was also a `device_id` provided explicitly.  Hard-coding a `device_id` is actually useful and part of #1187 to only run Chrome OS tests on particular devices.  I've tested the change above by hand locally.

gctucker (approved):
Tested OK again locally by generating jobs with and without `device_id`.

PR #1193, opened 2022-05-16
test-configs.yaml: Enable KVM selftests on qdf2400

broonie:
For arm64 the KVM selftests really want to be run on a system with a single
CPU type, additional configuration is required for big.LITTLE systems. The
QDF2400 is interesting for this since it is a unique CPU implementation and
a server machine where KVM is one of the intended applications so let's
enable KVM there. Unfortunately the platform is also very slow to boot
which in conjunction with the ~6 minute runtime for the tests means that
we are on the edge of the current 10 minute job timeout, raise that a bit
to give some headroom.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -336,7 +336,7 @@ test_plans:
   kselftest-kvm:
     <<: *kselftest
     params:
-      job_timeout: '10'
+      job_timeout: '15'
       kselftest_collections: "kvm"
     filters: *kselftest_no_fragment
 
@@ -2534,6 +2534,7 @@ test_configs:
       - baseline
       - kselftest-filesystems
       - kselftest-futex
+      - kselftest-kvm
       - kselftest-lib
       - kselftest-lkdtm
       - kselftest-seccomp


broonie:
This is now running OK when it gets to running the kselftest:

https://staging.kernelci.org/test/plan/id/628bc42ec394eb703be12bce/

I'm seeing a lot of timeouts that don't look to be related to the job though:

https://staging.kernelci.org/test/plan/id/628bed63e66d7330d8e12bce/

with jobs like 

https://lkft.validation.linaro.org/scheduler/job/5076518

It looks awfully like export-device-env is interacting poorly with the 200ms character delay the device has and the log generated NFS root path so it's timing out due to the time taken to write the data out rather than any device issue. I can't see where the timeout in question is coming from. This is affecting other tests too, eg:

https://lkft.validation.linaro.org/scheduler/job/5077860
https://lkft.validation.linaro.org/scheduler/job/5077834

so this is fine in and of itself but we need a generic fix for qdf2400 timeouts.

gctucker:
The issue is that it takes about 5min to boot and the boot timeout is 5min. I've resubmitted it with a 10min boot timeout and it's fine:
https://lkft.validation.linaro.org/scheduler/job/5101122

So we should add a parameter for devices that take a very long time to boot.

gctucker:
See #1217 for the `boot_timeout` parameter.

gctucker:
And #1218 to set the timeout to 10min for `qdf2400`.

gctucker:
The timeout fix seems to be working, but this platform isn't coping with the load in the LKFT lab and most jobs end up timing out in the queue after 24h.  Maybe we should add a filter to reduce the number of branches being tested, for example blocking the stable-rc tree should reduce the load quite significantly.

gctucker (approved):
Tested OK in combination with #1218.

However, see my previous comment about the load being too high for a single board online.  We can still merge this now and then come up with some fine-tuning to avoid filling the queue with jobs that never get run.

PR #1199, opened 2022-05-18
test-configs.yaml: drop ltp-mm from rock2-square

gctucker:
Drop the ltp-mm test plan from rk3288-rock2-square as there is only
one board online in kernelci.org at the moment and it can't cope with
the test load.


gctucker (commented):
Tested OK on staging.

PR #1200, opened 2022-05-18
CHROMEOS build_board.sh: Build octopus with ttyS1 serial console

nuclearcat:
As investigated, octopus are different from QEMU by using second serial,
so slight adjustments for build script required.
It is done such way that it will be easier to add similar Chromebooks in
future.
Patch tested by manual build and verified on LAVA for console functionality.

### DIFF
--- a/config/rootfs/chromiumos/scripts/build_board.sh
+++ b/config/rootfs/chromiumos/scripts/build_board.sh
@@ -2,6 +2,7 @@
 set -e
 BOARD=$1
 BRANCH=$2
+SERIAL=$3
 DATA_DIR=$(pwd)
 USERNAME=$(/usr/bin/id -run)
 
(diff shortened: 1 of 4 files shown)


config/rootfs/chromiumos/scripts/build_board.sh
```diff
@@ -16,6 +16,17 @@ function cleanup()
...
+
+    octopus)
+        SERIAL="ttyS1"
+        ;;
+
+    *)
+        SERIAL="ttyS0"
+        ;;
+esac
+
```
gctucker, line 29: How about adding a parameter to `chromiumos-octopus` in `rootfs-configs.yaml` instead?

nuclearcat: Updated according suggestion

gctucker: Ah I thought there was some generic `params` attribute for rootfs config objects but it looks like there isn't. Maybe that feature should be moved to the base YAML config object. But that's fine, it's some improvement for later. The approach of adding an attribute explicitly in the config object works as well.

nuclearcat:
Tested manually for octopus (as it needs ttyS1) - now serial console login works fine, tested on Chromebook.
Also built and tested for amd64-generic, everything remain the same, as it was value by default (ttyS0).

PR #1201, opened 2022-05-18
config/scripts: align base templates with kubernetes

gctucker:
Align the base templates in config/scripts with the ones in
config/kubernetes so that test plans templates such as check-describe
(in kernelci-pipeline) can include either base-python.jinja2 or
base-k8s-python.jinja2 depending on the runtime environment.


gctucker:
Tested OK in combination with `kernelci-pipeline`: https://github.com/kernelci/kernelci-pipeline/pull/61#issuecomment-1130471181

PR #1204, opened 2022-05-20
T35399 Create new user utility

JenySadadia:
Add command-line utility to `kci_data` to create new users.

Tested on local:
```
./kci_data create_user --username jeny --password jeny --db-config api --db-token eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJib2IiLCJzY29wZXMiOlsiYWRtaW4iXX0.t3bAE-pHSzZaSHp7FMlImqgYvL6f_0xDUD-nQwxEm3k --is-admin 1
{'_id': '62873c99c7fa5742896cc432', 'username': 'jeny', 'hashed_password': '$2b$12$rL9zT6Xtj1sgFpoZVNrNZu3WE.MBwDSNBXD/nondmY3ZORIc61IYu', 'active': True, 'is_admin': True}
```

### DIFF
--- a/kernelci/cli.py
+++ b/kernelci/cli.py
@@ -170,6 +170,12 @@ class Args:
         "Path to the install directory, or _install_ inside kdir by default",
     }
 
+    is_admin = {
+        'name': '--is-admin',
+        'help': "Create an admin user",
+        'action': 'store_true',
+    }
+
     j = {
         'name': '-j',
         'help': "Number of parallel build processes",
@@ -236,6 +242,11 @@ class Args:
     output = {
         'name': '--output',
         'help': "Path the output directory",
+    }
+
+    password = {
+        'name': '--password',
+        'help': "Password of a new user",
     }
 
     plan = {
@@ -310,6 +321,11 @@ class Args:
         'name': '--user',
         'help': "Test lab user name",
         'section': SECTION_LAB,
+    }
+
+    username = {
+        'name': '--username',
+        'help': "Name of a new user",
     }
 
     variant = {
--- a/kernelci/db/kernelci_api.py
+++ b/kernelci/db/kernelci_api.py
@@ -51,9 +51,10 @@ class KernelCI_API(Database):
         resp.raise_for_status()
         return resp
 
-    def _post(self, path, data=None):
+    def _post(self, path, data=None, params=None):
         url = self._make_url(path)
-        resp = requests.post(url, headers=self._headers, data=data)
+        resp = requests.post(url, headers=self._headers, params=params,
+                             data=data)
         resp.raise_for_status()
         return resp
 
@@ -62,6 +63,18 @@ class KernelCI_API(Database):
         resp = requests.put(url, headers=self._headers, data=data)
         resp.raise_for_status()
         return resp
+
+    def create_user(self, username, password, is_admin, verbose=False):
+        """Create new user"""
+        path = '/'.join(['user', username])
+        params = {'is_admin': is_admin}
+        data = {'password': password}
+        try:
+            resp = self._post(path, data=json.dumps(data), params=params)
+        except requests.exceptions.HTTPError as ex:
+            self._print_http_error(ex, verbose)
+            raise(ex)
+        return resp.json()
 
     def subscribe(self, channel):
         resp = self._post(f'subscribe/{channel}')
--- a/kernelci/db/kernelci_api.py
+++ b/kernelci/db/kernelci_api.py
@@ -39,6 +39,12 @@ class KernelCI_API(Database):
     def _make_url(self, path):
         return urllib.parse.urljoin(self.config.url, path)
 
+    def _print_http_error(self, http_error, verbose=False):
+        print(http_error)
+        if verbose:
+            print("Error:",
+                  json.loads(http_error.response.content).get("detail", []))
+
     def _get(self, path, params=None):
         url = self._make_url(path)
         resp = requests.get(url, params=params, headers=self._headers)
--- a/kci_data
+++ b/kci_data
@@ -97,6 +97,22 @@ class cmd_submit_test(Command):
         return db.submit_test(data, args.verbose)
 
 
+class cmd_create_user(Command):
+    help = "Create new user"
+    args = [Args.db_config, Args.username, Args.password]
+    opt_args = [Args.is_admin, Args.db_token, Args.verbose]
+
+    def __call__(self, config_data, args):
+        config = config_data['db_configs'][args.db_config]
+        db = kernelci.db.get_db(config, args.db_token)
+        resp = db.create_user(
+            args.username, args.password, args.is_admin, args.verbose
+        )
+        if args.verbose:
+            print(resp)
+        return resp
+
+
 if __name__ == '__main__':
     opts = parse_opts("kci_data", globals())
     configs = kernelci.config.load(opts.yaml_config)


gctucker:
> Add command-line utility to `kci_data` to create new users.
> 
> Tested on local:
> 
> ```
> ./kci_data create_user --username jeny --password jeny --db-config api --db-token eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJib2IiLCJzY29wZXMiOlsiYWRtaW4iXX0.t3bAE-pHSzZaSHp7FMlImqgYvL6f_0xDUD-nQwxEm3k --is-admin 1
> {'_id': '62873c99c7fa5742896cc432', 'username': 'jeny', 'hashed_password': '$2b$12$rL9zT6Xtj1sgFpoZVNrNZu3WE.MBwDSNBXD/nondmY3ZORIc61IYu', 'active': True, 'is_admin': True}
> ```

Note: It's better to store the API token in the `kernelci.conf` file to not leak it on the command line.  As this is for debugging and the API is your local instance it's not an issue in practice here.

JenySadadia:
> > Add command-line utility to `kci_data` to create new users.
> > Tested on local:
> > ```
> > ./kci_data create_user --username jeny --password jeny --db-config api --db-token eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJib2IiLCJzY29wZXMiOlsiYWRtaW4iXX0.t3bAE-pHSzZaSHp7FMlImqgYvL6f_0xDUD-nQwxEm3k --is-admin 1
> > {'_id': '62873c99c7fa5742896cc432', 'username': 'jeny', 'hashed_password': '$2b$12$rL9zT6Xtj1sgFpoZVNrNZu3WE.MBwDSNBXD/nondmY3ZORIc61IYu', 'active': True, 'is_admin': True}
> > ```
> 
> Note: It's better to store the API token in the `kernelci.conf` file to not leak it on the command line. As this is for debugging and the API is your local instance it's not an issue in practice here.

Yes, I have added the below section to `kernelci.conf`:
```
[db:api]

db_token: 

```

Moved `db_token` to `opt_args` in `kci_data`.

Now, the below command is working:
```
./kci_data create_user --username jeny --password jeny --db-config api --is-admin 1
```

gctucker:
This PR uses a branch from the main kerneli-core repository. The regular workflow is to use branches in people's forks to avoid filling the main repository with work topic branches. I think we can keep this one, but please keep that in mind for future PRs.

JenySadadia:
> 

Oh, by mistake I cloned the main repository instead of forked one on local.
I will correct this.

config/core/db-configs.yaml
```diff
@@ -1,5 +1,9 @@
 db_configs:
 
+  api:
+    db_type: kernelci_api
+    url: http://localhost:8001/
+
```
gctucker, line 6: I think it's a bit too early to add this, which is why it hasn't been added yet. The pipeline has its own YAML config to define it instead.

kernelci/cli.py
```diff
@@ -165,6 +165,11 @@ class Args:
         "Path to the install directory, or _install_ inside kdir by default",
     }
 
+    is_admin = {
+        'name': '--is-admin',
+        'help': "Provide 1 for superuser and 0 for regular user",
```
gctucker, line 170: Add this to make it a boolean flag:
```python
        'action': 'store_true'
```
Then the `--is-admin` option can be used with no argument.

kernelci/cli.py
```diff
@@ -165,6 +165,11 @@ class Args:
         "Path to the install directory, or _install_ inside kdir by default",
     }
 
+    is_admin = {
```
gctucker, line 168: Please mention in the commit subject that the arguments are for creating API users.

kernelci/db/kernelci_api.py
```diff
@@ -57,6 +57,17 @@ def _put(self, path, data=None):
         resp.raise_for_status()
         return resp
 
+    def create_user(self, username, password, is_admin):
+        """Create new user"""
+        path = '/'.join(['user', username])
+        if is_admin:
+            path = '?'.join([path, 'is_admin='+str(is_admin)])
+
+        data = {'password': password}
```
gctucker, line 66: I think this should ideally be done with `urllib`

kernelci/db/kernelci_api.py
```diff
@@ -57,6 +57,17 @@ def _put(self, path, data=None):
...
+        resp = self._post(path, json.dumps(data))
+        print(resp.json())
```
gctucker, line 68: Please drop this, the calling code can decide to print some information if needed.

gctucker (commented):
A few things to tweak

gctucker:
I've made a PR to tweak the documentation accordingly:
https://github.com/kernelci/kernelci-core/pull/1270

kernelci/db/kernelci_api.py
```diff
@@ -45,9 +45,10 @@ def _get(self, path):
         resp.raise_for_status()
         return resp
 
-    def _post(self, path, data=None):
+    def _post(self, path, params=None, data=None):
```
gctucker, line 48: The method is used in one place with data as a positional argument:
```python
                    resp = self._post(path, json.dumps(item))
```

So please keep the order of the arguments and add `params` at the end:
```suggestion
    def _post(self, path, data=None, params=None):
```

gctucker (commented):
One more small thing to fix

kernelci/db/kernelci_api.py
JenySadadia: I think `params` is a better option.

kernelci/db/kernelci_api.py
JenySadadia: Done.

kernelci/cli.py
JenySadadia: Done.

kernelci/cli.py
JenySadadia: Done.

config/core/db-configs.yaml
JenySadadia: But it needs to be added to run the command:
```
./kci_data create_user --username admin --password admin --db-config api --is-admin
```

gctucker: Yes but that's already defined in `kernelci-pipeline`, so it's a grey area. My point is that support for KernelCI API is not actually ready yet in `kernelci-core`. Calling it `api` is also not descriptive enough, so I think it will lead to issues later when we have an actual production one online. You can provide `--yaml-config` to specify where the YAML configuration is when running `kci_data`, and use the YAML file from `kernelci-pipaline` which has an entry for the local API.

kernelci/db/kernelci_api.py
gctucker: Ah yes, that's even better with `requests.post()` directly.

config/core/db-configs.yaml
JenySadadia: Okay. But then it will be dependent on `pipeline`. Why would we do that when we can directly add configurations in `core`? And what about when we just need `core` utilities without setting up the `pipeline`?

gctucker: The code in `kernelci-core` is not ready for communicating with the API, we only have a few things in place that only work with the code in `kernelci-pipeline` and it's likely to have to be rewritten as the API is evolving and support for `kernelci-backend` will be deprecated and removed. So yes I think we might ultimately have entries in `api-configs.yaml` or something for the new API server but right now I don't feel comfortable with adding it here. It seems too early, some people might try to use it and hit some issues. So it's a grey area, and because we don't really need it I think it's safer to leave it out for now. If in practice it's causing too many issues then I guess we can have it merged.

JenySadadia: Okay. Let's leave it out as of now and I'll drop the commit.

JenySadadia:
Now, the revised command for creating a regular user is:
```
./kci_data --yaml-config ../kernelci-pipeline/config/pipeline.yaml create_user -db-config docker-host --username jeny --password jeny -
```
To create an admin user:
```
./kci_data --yaml-config ../kernelci-pipeline/config/pipeline.yaml create_user --db-config docker-host  --username jeny --password jeny --is-admin
```

gctucker:
Yes, and that aligns with the updated documentation:
https://static.staging.kernelci.org/docs/core/kci_data/#creating-new-api-user-experimental

Now the "getting started" section might need to be updated, I'll take a look.

JenySadadia:
> Yes, and that aligns with the updated documentation: https://static.staging.kernelci.org/docs/core/kci_data/#creating-new-api-user-experimental
> 
> Now the "getting started" section might need to be updated, I'll take a look.

I think we will need to update it after I change the code for verbose mode. I will commit the changes soon.

kernelci/db/kernelci_api.py
```diff
@@ -57,6 +58,16 @@ def _put(self, path, data=None):
...
         return resp
 
+    def create_user(self, username, password, is_admin, verbose=False):
+        """Create new user"""
+        path = '/'.join(['user', username])
+        params = {'is_admin': is_admin}
+        data = {'password': password}
+        resp = self._post(path, data=json.dumps(data), params=params)
+        if verbose:
+            print(resp.json())
```
gctucker, line 68: I think the method should return the response and then kci_data can print something with verbose flag enabled instead.

JenySadadia: Done.

JenySadadia:
@gctucker Does this look good to you?

mgalka:
@gctucker Is this [conversation](https://github.com/kernelci/kernelci-core/pull/1204#discussion_r905785039) resolved?

@JenySadadia Have you tested it with current main?

JenySadadia:
> @gctucker Is this [conversation](https://github.com/kernelci/kernelci-core/pull/1204#discussion_r905785039) resolved?
> 
> @JenySadadia Have you tested it with current main?

I encountered an issue of YAML configuration load failure when I tested with the current main.
Created a GH issue: https://github.com/kernelci/kernelci-core/issues/1528
Also, sent a PR to fix it.

After the fixing above issue, `create_user` is working as expected.
Also, I have added a few tweaks to print HTTP errors from the new API.

mgalka:
It looks some checks are not passing. @JenySadadia could you have a look at it?

JenySadadia:
> It looks some checks are not passing. @JenySadadia could you have a look at it?

Fixed it.

mgalka:
As a follow up we can add an option to enter password interactively from the command line. Otherwise it looks good

nuclearcat (approved):
LGTM

mgalka (approved):

PR #1207, opened 2022-05-23
Add Bullseye rootfs image for flashing Chrome OS on Chromebook devices

gctucker:
Add a rootfs configuration to flash Chrome OS images on Chromebook devices in LAVA.

PR #1208, opened 2022-05-23
Fix a couple of issues with the kci_* validate commands

gctucker:
Fix a typo in the code and also align `kci_rootfs validate` with the other command line tools.


gctucker:
OK this fixes the validation errors seen with #1207

PR #1209, opened 2022-05-23
CHROMEOS rework cros-flash job templates

gctucker:
Rework the cros-flash job templates to rely on the Chrome OS to be included in the NFS root tarball.  This simplifies things greatly as we're only using Chromium OS images at the moment.


gctucker:
Tested OK: https://lava.collabora.co.uk/scheduler/job/6448600

nuclearcat (commented):
Maybe instead of multiple parameters, we can have just directory to current cros release?
cros_path: 'https://storage.chromeos.kernelci.org/images/rootfs/chromeos/chromiumos-octopus/20220520.0/amd64/'
Where we assume by suffix path to other components, tast.tgz, cros-flash.tar.gz, initrd and so on.

gctucker:
> Maybe instead of multiple parameters, we can have just directory to current cros release? cros_path: 'https://storage.chromeos.kernelci.org/images/rootfs/chromeos/chromiumos-octopus/20220520.0/amd64/' Where we assume by suffix path to other components, tast.tgz, cros-flash.tar.gz, initrd and so on.

It's a bit too early to assume things and rely on conventions, the cros-flash tarball was created by hand. Once we have this stabilisied then we can potentially optimise the configuration and reduce repetition if it doesn't make things too implicit.

PR #1212, opened 2022-05-24
cros-boot rework

gctucker:
Rework the `cros-boot` test plan to install modules in 2-stage jobs and use correct kernel command line arguments and login prompt.


nuclearcat (approved):
Looks ok to me, just only a bit confusing thing is naming: cros-boot is relevant only to Chrome(ium)-Chromebook variant, cros-qemu-boot expected to be different.

gctucker:
> 

Thanks for the review. Yes there will be a PR to align QEMU test plans too a bit later.

gctucker:
Also tested the `cros-boot` plan with actual 5.10 kernel: https://lava.collabora.co.uk/scheduler/job/6457566

PR #1213, opened 2022-05-24
CHROMEOS cros-tast rework

gctucker:
Rework the `cros-tast` test plan to inherit `cros-boot` and avoid duplication of the common parts to install modules and reboot with Chrome OS on eMMC.


gctucker (commented):
Tested OK: https://chromeos.kernelci.org/test/job/kernelci/branch/chromeos-stable/kernel/chromeos-stable-20220525.0/plan/cros-tast-fixed/

PR #1214, opened 2022-05-24
CHROMEOS cros-boot_qemu and cros-tast_qemu rework

gctucker:
Rework the boot and Tast test plans for QEMU to align them with the Chromebook ones.

Depends on #1213

PR #1215, opened 2022-05-25
CHROMEOS drop old chromeos test plans

gctucker:
Drop test plans that aren't being used any more now that the modules are installed by `cros-boot.jinja2` and the QEMU templates have been consolidated.

Depends on #1214


gctucker (commented):
Tested OK: https://chromeos.kernelci.org/test/job/kernelci/branch/chromeos-stable/kernel/chromeos-stable-20220526.2/

PR #1217, opened 2022-05-26
config/lava: add boot_timeout template variable

gctucker:
Add a boot_timeout template variable which can be used to override the
boot timeout.  This can typically be set in the device type parameters
for devices that take longer to boot.


gctucker (commented):
Tested OK on staging.

PR #1220, opened 2022-05-26
CHROMEOS kernelci.lab.lava: shorten Chrome OS job names

gctucker:
Add extra processing step before generating LAVA test job definitions
to shorten the job name using a 1-to-1 mapping between long and short
strings for the Chrome OS configs and device type names.  Strings that
aren't in the mapping won't be shortened so it won't fail but might
result in longer names until they get added too.

This is required in order to avoid hitting the limit of 256 characters
for LAVA job names.  For example, without this change a job name could
have been (217 characters):

  kernelci-chromeos-stable-chromeos-stable-20220526.0-x86_64-cros---chromeos-5.10-x86_64-chromeos-intel-denverton.flavour.config+x86-chromebook-clang-14-no-dtb-hp-x360-12b-ca0500na-n4000-octopus_chromeos-cros-tast-fixed

With the shortening done by this change we now have (136 characters):

  kernelci-chromeos-stable-chromeos-stable-20220526.0-x86_64-x86_64-denv-5.10-x86-chromebook-clang-14-no-dtb-octopus-n4000-cros-tast-fixed

### DIFF
--- a/kernelci/lab/lava/__init__.py
+++ b/kernelci/lab/lava/__init__.py
@@ -21,7 +21,29 @@
 
 from jinja2 import Environment, FileSystemLoader
 import os
+import re
 from kernelci.lab import LabAPI
+
+CROS_CONFIG_RE = re.compile(r'cros://chromeos-([0-9.]+)/([a-z0-9_]+)/([a-z-.]+).flavour.config(\+[a-z0-9-+]+)?')  # noqa
+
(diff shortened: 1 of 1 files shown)


gctucker (commented):
Tested OK: https://lava.collabora.co.uk/scheduler/alljobs?length=100&search=chromeos-stable-20220526.2#table

PR #1228, opened 2022-05-27
CHROMEOS test-configs-chromeos.yaml: update cros-flash rootfs URL

gctucker:
Update the URL of the bullseye-cros-flash rootfs with the production
build rather than the initial staging one.

### DIFF
--- a/config/core/test-configs-chromeos.yaml
+++ b/config/core/test-configs-chromeos.yaml
@@ -77,7 +77,7 @@ device_types:
       - passlist: {defconfig: ['x86-chromebook']}
     params: &octopus-params
       cros_flash_nfs:
-        base_url: 'https://storage.staging.kernelci.org/images/rootfs/debian/bullseye-cros-flash/20220523.1/amd64/'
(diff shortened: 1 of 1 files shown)


config/core/test-configs-chromeos.yaml
```diff
@@ -77,7 +77,7 @@ device_types:
       - passlist: {defconfig: ['x86-chromebook']}
     params: &octopus-params
       cros_flash_nfs:
-        base_url: 'https://storage.staging.kernelci.org/images/rootfs/debian/bullseye-cros-flash/20220523.1/amd64/'
+        base_url: 'https://storage.kernelci.org/images/rootfs/debian/bullseye-cros-flash/20220527.0/amd64/'
```
mgalka, line 80: @gctucker Is there any reason why the URL is using HTTPS?

gctucker: @mgalka So do you think it should be HTTP?

mgalka: I don't mind. I was just curious.

mgalka (approved):
LGTM

PR #1230, opened 2022-05-30
kernelci.test: use describe_verbose for kernel filters

gctucker:
Use the describe_verbose meta-data attribute when applying filters
based on the kernel revision.  This attribute is formatted to always
rely on standard kernel release tags e.g. v5.4.123 rather than any
other tree-specific tags set by maintainers or linux-next etc.

### DIFF
--- a/kernelci/test.py
+++ b/kernelci/test.py
@@ -43,7 +43,7 @@ def match_configs(configs, meta, lab):
     filters = {
         'arch': arch,
         'defconfig': defconfig,
-        'kernel': rev['describe'],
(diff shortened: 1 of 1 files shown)


gctucker:
Tested OK on staging:
https://bot.staging.kernelci.org/job/test-runner/32212/console

Before this change, the jobs for `hip07-d05` were not being generated as the kernel tags used on staging don't include the kernel version (only `describe_verbose` does):
https://bot.staging.kernelci.org/job/test-runner/32211/console

PR #1231, opened 2022-05-31
kernelci/config: add email configuration to kernelci-core

alexandrasp:
Include a class in KernelCI core to load a basic set of email configurations.

### DIFF
--- a/kernelci/config/__init__.py
+++ b/kernelci/config/__init__.py
@@ -23,6 +23,7 @@ import kernelci
 import kernelci.config
 import kernelci.config.build
 import kernelci.config.db
+import kernelci.config.email
 import kernelci.config.lab
 import kernelci.config.rootfs
 import kernelci.config.test
@@ -90,6 +91,7 @@ def from_data(data):
     filters = kernelci.config.base.default_filters_from_yaml(data)
     config.update(kernelci.config.build.from_yaml(data, filters))
     config.update(kernelci.config.db.from_yaml(data, filters))
+    config.update(kernelci.config.email.from_yaml(data))
     config.update(kernelci.config.lab.from_yaml(data, filters))
     config.update(kernelci.config.rootfs.from_yaml(data, filters))
     config.update(kernelci.config.test.from_yaml(data, filters))
--- a/kernelci/config/email.py
+++ b/kernelci/config/email.py
@@ -1,0 +1,49 @@
+# Copyright (C) 2022 Collabora Limited
+# Author: Alexandra Pereira <alexandra.pereira@collabora.com>
+#
+# This module is free software; you can redistribute it and/or modify it under
+# the terms of the GNU Lesser General Public License as published by the Free
+# Software Foundation; either version 2.1 of the License, or (at your option)
+# any later version.
+#
+# This library is distributed in the hope that it will be useful, but WITHOUT
+# ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS
+# FOR A PARTICULAR PURPOSE.  See the GNU Lesser General Public License for more
+# details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with this library; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+
+import yaml
+
+from kernelci.config.base import YAMLObject
+
+
+class Email(YAMLObject):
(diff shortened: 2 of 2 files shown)


mgalka:
@alexandrasp Please, add a comment about testing

mgalka:
@alexandrasp regarding the commit message 

```
kernelci/config: add email configurations to kernelci-core
include a class in KernelCI core to load a basic set
of email configurations.
```

I think it should be "configuration" rather than "configurations"

alexandrasp:
> Include a class in KernelCI core to load basic e-mail configuration.

Sure, updated ;)

mgalka:
> > Include a class in KernelCI core to load basic e-mail configuration.
> 
> Sure, updated ;)

The commit header still says "add email configurations"

kernelci/config/__init__.py
```diff
@@ -26,6 +26,7 @@
 import kernelci.config.lab
 import kernelci.config.rootfs
 import kernelci.config.test
+import kernelci.config.email
```
gctucker, line 29: Please keep in alphabetical order

kernelci/config/__init__.py
```diff
@@ -93,6 +94,7 @@ def from_data(data):
     config.update(kernelci.config.lab.from_yaml(data, filters))
     config.update(kernelci.config.rootfs.from_yaml(data, filters))
     config.update(kernelci.config.test.from_yaml(data, filters))
+    config.update(kernelci.config.email.from_yaml(data))
```
gctucker, line 97: Please keep in alphabetical order

kernelci/config/email.py
```diff
@@ -0,0 +1,60 @@
...
+    def email_send_to(self):
+        return self._email_send_to
+
+    @property
+    def email_send_from(self):
+        return self._email_send_from
+
+    @property
+    def email_subject(self):
+        return self._email_subject
```
gctucker, line 49: The subject shouldn't be part of the config as it should be set by the template and the rendering code.

kernelci/config/email.py
```diff
@@ -0,0 +1,60 @@
...
+    def email_type(self):
+        return self._email_type
+
+    @property
+    def email_send_to(self):
+        return self._email_send_to
+
+    @property
+    def email_send_from(self):
+        return self._email_send_from
```
gctucker, line 45: I don't think the "from" field should be configurable in YAML, it's a system-wide parameter like the SMTP port etc.

kernelci/config/email.py
```diff
@@ -0,0 +1,60 @@
...
+        self._email_send_from = email_send_from
+        self._email_subject = email_subject
+
+    @classmethod
+    def from_yaml(cls, email_type, kw):
+        return cls(email_type, **kw)
+
+    @property
+    def email_type(self):
+        return self._email_type
```
gctucker, line 37: What is the email "type" supposed to be?

gctucker (commented):
I think that while we will need some YAML configuration, this hasn't been completely thought through so there are a few things that will need to be redesigned.

kernelci/config/email.py
alexandrasp: It makes sense. I removed subject as well.. at this moment not sure about what we should have in this YAML other than recipients list and the environment. Should I consider putting anything else?

kernelci/config/__init__.py
alexandrasp: updated

kernelci/config/__init__.py
alexandrasp: updated

kernelci/config/email.py
mgalka: @gctucker Do you have any thoughts on how this value should be provided to the script which renders the e-mail from template?

alexandrasp: @mgalka maybe could be like this: https://github.com/kernelci/kernelci-pipeline/pull/72/files#diff-a356af3cdd34dc1001fadc4e67566b93959d3e9c1d50946c3032112b186557d0R149 as in the legacy back-end.

nuclearcat:
ping, is any work planned on this PR or it will be made as new PR?

gctucker:
We do need YAML-based email configuration and code to handle that and replace `kernel-arch-complete.sh`, but this PR is not implementing the actual features we need.  So I think we can just close it.

nuclearcat:
Thank you, closing

PR #1242, opened 2022-06-06
T35078: Get all child nodes from parent

JenySadadia:
Send GET request to /trigger_completed_event endpoint of API.

### DIFF
--- a/kernelci/db/kernelci_api.py
+++ b/kernelci/db/kernelci_api.py
@@ -108,6 +108,14 @@ class KernelCI_API(Database):
         resp = self._get('?'.join(['regressions', 'name=' + regression_name]))
         return json.loads(resp.text)
 
+    def get_child_nodes_from_parent(self, node_id):
+        """Get child nodes from parent"""
+        params = {
+            "parent": node_id
+        }
+        resp = self._get('nodes', params=params)
+        return resp.json()
+
     def pubsub_event_filter(self, sub_id, event):
         """Filter Pub/Sub events
 
--- a/kernelci/db/kernelci_api.py
+++ b/kernelci/db/kernelci_api.py
@@ -39,9 +39,9 @@ class KernelCI_API(Database):
     def _make_url(self, path):
         return urllib.parse.urljoin(self.config.url, path)
 
-    def _get(self, path):
+    def _get(self, path, params=None):
         url = self._make_url(path)
-        resp = requests.get(url, headers=self._headers)
+        resp = requests.get(url, params=params, headers=self._headers)
         resp.raise_for_status()
         return resp
 


kernelci/db/kernelci_api.py
```diff
@@ -103,6 +103,16 @@ def get_root_node(self, node_id):
         resp = self._get('/'.join(['get_root_node', node_id]))
         return json.loads(resp.text)
 
+    def get_completed_event(self, node_id, hours=0, minutes=0, seconds=0):
+        """Get an event when all child nodes are complete of a given node"""
+        resp = self._get(f'trigger_completed_event/{node_id}?wait_time_hours=\
+{hours}&wait_time_minutes={minutes}&wait_time_seconds={seconds}')
```
mgalka, line 109: When using query parameters we should consider using params kyeword argument for requests.get(), which is used by _get method.
https://requests.readthedocs.io/en/latest/api/#requests.get

kernelci/db/kernelci_api.py
```diff
@@ -108,6 +108,20 @@ def get_regressions(self, regression_name):
         resp = self._get('?'.join(['regressions', 'name=' + regression_name]))
         return json.loads(resp.text)
 
+    def get_completed_event(self, node_id, hours=0, minutes=0, seconds=0):
+        """Get an event when all child nodes are complete of a given node"""
+        params = {
+            'wait_time_hours': hours,
+            'wait_time_minutes': minutes,
+            'wait_time_seconds': seconds
+        }
```
gctucker, line 117: Sorry I didn't notice this before, and I see the API parts have been merged already.  I think we should remove this completely, we can't have a query that waits on the API side when we should just have a timer on the client side.

gctucker (changes requested):

kernelci/db/kernelci_api.py
JenySadadia: So what I need here is to move the logic of `Node.wait_for_node` in the pipeline.
The endpoint will just check whether the node is completed or not and trigger event.

JenySadadia: Setting node status to `timeout` will be handled by `set_timeout` script.
I will update the `trigger_completed` pipeline script to wait for the child nodes to get completed.
When all child nodes are completed, the root status will be set to completed and the API endpoint will be called only to trigger `completed` event at the end.

JenySadadia:
Removed methods related to `completed` event as we have discarded it from the API.

JenySadadia:
Function from https://github.com/kernelci/kernelci-core/pull/1265 can be used to get child nodes from parent node id. Now we don't need any special method for this. Hence, closing this PR.

PR #1244, opened 2022-06-06
Add cros-flash test plan with parameters for octopus

gctucker:
Add a `cros-flash` test plan to flash Chrome OS images on Chromebooks, starting with `octopus`.

### DIFF
--- a/config/core/test-configs-chromeos.yaml
+++ b/config/core/test-configs-chromeos.yaml
@@ -1,0 +1,12 @@
+test_plans:
+
+  cros-flash:
+    pattern: 'chromeos/cros-flash.jinja2'
+    params:
+      job_timeout: '60'
+      docker_image: 'kernelci/chromeos-flash'
+      chromeos_image: 'chromiumos_test_image-R94-14150.19.0-octopus.bin'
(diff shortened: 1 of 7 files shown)


gctucker:
Tested OK by hand:
https://lava.collabora.co.uk/scheduler/job/6566238

Then Chrome OS booted OK with reference kernel:
https://lava.collabora.co.uk/scheduler/job/6566596

nuclearcat (approved):
Did generated manually job definition for both octopus and tested (but cancelled) in LAVA.
I think in future we need to add or mention somehow that tag should be added, otherwise it will generate job definition that flash random device.

gctucker:
If I remove the conditional around the `device_id` tag thing in `cros-flash.jinja2` and submit a job with no tag then I get this error:
```
<h1>500 Internal Server Error</h1>
<h2>Device tag &#39;&#39; does not exist in the database.</h2>
<h3></h3>
<p>
  Oops, something has gone wrong!
</p>
```
So it's a bit of a hack but at least jobs aren't run with no tag.

I'll still see if there's a way to "assert" that a value is populated in Jinja2.

gctucker:
OK the last 2 patches should fix that, now an exception is raised from Jinja2 when there's no device id.

gctucker:
I've submitted a flash job again on cbg-4 with the latest version of this PR to double check:
https://lava.collabora.co.uk/scheduler/job/6569219

Will take a look at the results later.  This device is the one used by KernelCI so we should have results on chromeos.kernelci.org for it.

PR #1247, opened 2022-06-07
T35800 Allow multivalued PubSub filter

JenySadadia:
Fixes https://github.com/kernelci/kernelci-api/issues/130

Support multivalued PubSub filer by providing tuple values.
A PubSub event will be received even if any one value of the multivalued filter matches the event data.

Now we can provide filters like below:
{"op":("updated","created")}
{"revision":{"commit": ("a", "b")}, "status": ("pass","fail"), "op": "created"}


gctucker (approved):
I think we could use regex matching instead of a tuple, but that's something we can rework later. Also we should remove the special case for sub-values and have the filter truly recursive to simplify the implementation.

PR #1249, opened 2022-06-08
HACK use 20220608.1/buildroot-baseline from staging

gctucker:
This should include bootrr changes from https://github.com/kernelci/bootrr/pull/16 with tests for Kontron KSwitch D10 boards


gctucker:
Tested OK.

PR #1258, opened 2022-06-10
CHROMEOS test-configs-chromeos.yaml: drop device_id from octopus

gctucker:
Drop the device_id parameters in the octopus device types as all
devices should be flashed with the same R100 image in production now.


mgalka (approved):
LGTM

PR #1261, opened 2022-06-16
CHROMEOS docker/cros-tast: Add dmesg baseline tests

nuclearcat:
To improve tests information, it is useful to add test similar to other kernelci
baseline tests that show kernel messages with warn/error/crit levels,
mark test as fail if they are present, and highlight number of such messages.

### DIFF
--- a/config/core/test-configs-chromeos.yaml
+++ b/config/core/test-configs-chromeos.yaml
@@ -1,32 +1,61 @@
 test_plans:
 
-  cros-boot: &cros-boot
-    pattern: 'chromeos/cros-boot.jinja2'
+  cros-baseline: &cros-baseline
+    pattern: 'chromeos/cros-baseline.jinja2'
     filters: &cros-filters
       - blocklist: { defconfig: ['kselftest'] }
+    params:
+      docker_image: 'kernelci/cros-baseline'
 
-  cros-boot-fixed:
-    <<: *cros-boot
+  cros-baseline-fixed:
+    <<: *cros-baseline
     filters: &cros-filters-fixed
       - blocklist: { defconfig: ['kselftest'] }
       - blocklist: { defconfig: ['cros'] }
+    params:
+      docker_image: 'kernelci/cros-baseline'
+      fixed_kernel: true
+
+  cros-baseline_qemu: &cros-baseline_qemu
+    base_name: cros-baseline
+    pattern: 'chromeos/cros-baseline.jinja2'
+    filters: &cros-qemu-filters
+      - blocklist: { defconfig: ['kselftest'] }
+      - passlist: { lab: ['lab-collabora-staging', 'lab-collabora'] }
+    params: &cros-baseline-qemu-params
+      base_template: 'chromeos/cros-boot-qemu.jinja2'
+      docker_image: 'kernelci/cros-baseline'
+
+  cros-baseline_qemu-fixed:
+    <<: *cros-baseline_qemu
+    base_name: cros-baseline-fixed
+    filters: &cros-qemu-filters-fixed
+      - blocklist: { defconfig: ['kselftest'] }
+      - blocklist: { defconfig: ['cros'] }
+      - passlist: { lab: ['lab-collabora-staging', 'lab-collabora'] }
+    params:
+      <<: *cros-baseline-qemu-params
+      fixed_kernel: true
+
+  cros-boot: &cros-boot
+    pattern: 'chromeos/cros-boot.jinja2'
+    filters: *cros-filters
+
+  cros-boot-fixed:
+    <<: *cros-boot
+    filters: *cros-filters-fixed
     params:
       fixed_kernel: true
 
   cros-boot_qemu: &cros-boot_qemu
     base_name: cros-boot
     pattern: 'chromeos/cros-boot-qemu.jinja2'
-    filters: &cros-qemu-filters
-      - blocklist: { defconfig: ['kselftest'] }
-      - passlist: { lab: ['lab-collabora-staging', 'lab-collabora'] }
+    filters: *cros-qemu-filters
 
   cros-boot_qemu-fixed:
     <<: *cros-boot_qemu
     base_name: cros-boot-fixed
-    filters: &cros-qemu-filters-fixed
-      - blocklist: { defconfig: ['kselftest'] }
-      - blocklist: { defconfig: ['cros'] }
-      - passlist: { lab: ['lab-collabora-staging', 'lab-collabora'] }
+    filters: *cros-qemu-filters-fixed
     params:
       fixed_kernel: true
 
@@ -156,6 +185,8 @@ test_configs:
 
   - device_type: hp-x360-12b-ca0010nr-n4020-octopus_chromeos
     test_plans:
+      - cros-baseline
+      - cros-baseline-fixed
       - cros-boot
       - cros-boot-fixed
       - cros-tast
@@ -163,6 +194,8 @@ test_configs:
 
   - device_type: hp-x360-12b-ca0500na-n4000-octopus_chromeos
     test_plans:
+      - cros-baseline
+      - cros-baseline-fixed
       - cros-boot
       - cros-boot-fixed
       - cros-tast
@@ -170,6 +203,8 @@ test_configs:
 
   - device_type: qemu_x86_64-uefi-chromeos
     test_plans:
+      - cros-baseline_qemu
+      - cros-baseline_qemu-fixed
       - cros-boot_qemu
       - cros-boot_qemu-fixed
       - cros-tast_qemu
--- a/config/docker/cros-baseline/Dockerfile
+++ b/config/docker/cros-baseline/Dockerfile
@@ -1,0 +1,41 @@
+FROM debian:bullseye-slim
+MAINTAINER "KernelCI TSC" <kernelci-tsc@groups.io>
+
+ARG DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update && apt-get install --no-install-recommends -y \
+    apt-transport-https \
+    ca-certificates 
+
+RUN apt-get update && apt-get install -y \
+    git \
+    inetutils-ping \
+    ssh
+
+RUN \
+  mkdir -p /home/cros && \
+  useradd cros -d /home/cros && \
+  chown cros: -R /home/cros && \
+  adduser cros sudo && \
+  echo "cros ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers
+USER cros
+ENV HOME=/home/cros
+WORKDIR $HOME
+
+# All this just to extract rsa key to access DUT
+# Adding private key "as is" might trigger github audit alert
+RUN mkdir -p $HOME/trunk
+RUN git clone \
+    https://chromium.googlesource.com/chromiumos/chromite \
+    $HOME/trunk/chromite
+ENV PATH=$PATH:$HOME/trunk/chromite/scripts
+
+# This SSH key is only used with Chrome OS test images.
+RUN mkdir "$HOME/.ssh"
+RUN cp "$HOME/trunk/chromite/ssh_keys/testing_rsa" "$HOME/.ssh/id_rsa"
+RUN chmod 0600 "$HOME/.ssh/id_rsa"
+
+ADD https://raw.githubusercontent.com/kernelci/kernelci-core/chromeos.kernelci.org/config/rootfs/debos/overlays/baseline/opt/kernelci/dmesg.sh /home/cros/dmesg.sh
+
+# Needed by LAVA to install the overlay
+USER root
--- a/config/lava/chromeos/cros-baseline.jinja2
+++ b/config/lava/chromeos/cros-baseline.jinja2
@@ -1,0 +1,30 @@
+{% extends base_template|default('chromeos/cros-boot.jinja2') %}
+{% block actions %}
+{{ super() }}
+- test:
+    namespace: chromeos
+    timeout:
+      minutes: 5
+    docker:
+      image: {{ docker_image }}
+      wait:
+        device: true
+    results:
+      location: /home/cros/lava
+    definitions:
+    - repository:
+        metadata:
+          format: Lava-Test Test Definition 1.0
+          name: cros-baseline
+        run:
+          steps:
+            - cd /home/cros
+            - while ! ping -c 1 -w 1 $(lava-target-ip); do sleep 1; done
+            - >-
+              KERNELCI_LAVA=y
+              cmdwrapper="ssh -i /home/cros/.ssh/id_rsa -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null root@$(lava-target-ip)"
+              /bin/sh dmesg.sh
+      from: inline
+      name: cros-baseline
+      path: inline/cros-baseline.yaml
+{% endblock %}


config/docker/cros-tast/baseline_cros.py
```diff
@@ -0,0 +1,64 @@
...
+import subprocess
+import sys
+
+loglevels = ["crit", "alert", "emerg", "err", "warn"]
+keyfile = '/home/cros-tast/.ssh/id_rsa'
+
+
+def report_lava(testname, result, count):
+    """
+    Report test state/result to LAVA,
```
mgalka, line 29:
```suggestion
    """ Report test state/result to LAVA
```

Short description should be put right after the triple quotes.

config/docker/cros-tast/baseline_cros.py
```diff
@@ -0,0 +1,64 @@
...
+    Arguments:
+    result: string, should contain fail or pass
+    count: integer, number of error lines
```
mgalka, line 32: I'm not insisting on it, but maybe it's worth using type hints?

config/docker/cros-tast/baseline_cros.py
```diff
@@ -0,0 +1,64 @@
...
+    client = paramiko.client.SSHClient()
+    client.set_missing_host_key_policy(paramiko.AutoAddPolicy())
+    client.load_system_host_keys()
+    client.connect(host, username="root", key_filename=keyfile)
+    for lvl in loglevels:
+        _stdin, _stdout, _stderr = client.exec_command(" dmesg --level="
+                                                       +lvl+" --notime -x -k")
+        buffer = _stdout.read().decode()
+        count = len(buffer)
+        if not count:
```
mgalka, line 55: as count is an `int` I'd rather put here `if count == 0`

mgalka (commented):
some minor tweaks

config/docker/cros-tast/baseline_cros.py
nuclearcat: Good catch, added also space after short summary, thanks!

config/docker/cros-tast/baseline_cros.py
nuclearcat: Added type hints and forgot to document one of arguments, added too

config/docker/cros-tast/baseline_cros.py
```diff
@@ -0,0 +1,64 @@
...
+    result: string, should contain fail or pass
+    count: integer, number of error lines
+    """
+    opts = ['lava-test-case', testname, '--result', result, '--units',
+            'lines', '--measurement', str(count)]
+    subprocess.run(opts)
+
+
+def fetch_dut():
+    output = subprocess.check_output("lava-target-ip", shell=True).strip()
```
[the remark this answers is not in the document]
nuclearcat: I think it is not necessary, removed

gctucker:
@nuclearcat Why re-implement this in Python when we already have a basic shell script? We can just run ssh to dump the log from `dmesg` remotely and pipe that into the `dmesg.sh` script. The advantage is to avoid repeating the same logic so we actually have the same results as the generic baseline tests (which needs to be shell to fit in the small buildroot ramdisk with no Python).

nuclearcat:
We can, but I think python scripts are much more elegant than shell magic and easier to maintain. It's a lot nicer to deal with paramiko than the shell constructs with redirection, as we have docker image with python anyway.
In addition, this script is already different from the standard dmesg.sh, since it works via ssh to DUT, and it may be possible to add to it some additional baseline tests specific to ChromeOS.
But if you think it's better to update and use dmesg.sh, I can add logic to it to check for the existence of a dmesg.$level files (which may have been obtained via ssh before running it, in case of remote DUT).

mgalka (approved):
Looks good from the code point of view. I guess it's @gctucker  to decide wether it should be merged.

gctucker:
My point is that we need to have the `dmesg.sh` script anyway to avoid adding Python to the small rootfs images. So it's not making things simpler to have to maintain a Python implementation in parallel. It's easy enough to include `dmesg.sh` in the Docker image, scp it to the device and run it there over ssh. I think we should consider doing this first, and if there's some issue with it then look for an alternative with a duplicate implementation in Python or any other language.

nuclearcat:
We can't just scp it, it needs lava test helper scripts (lava-test-case, lava-common-functions and related stuff) and in most cases distro specific support in lava (https://github.com/Linaro/lava/tree/master/lava_dispatcher/lava_test_shell/distro) , if we install it as per LAVA design, as overlay.
I tried installed them by scp, and it is a bit headache, as ChromeOS is different from typical distribution, and read-only on path where they want to be installed, create temporary directories and etc, for example had issues when i try to run them, as by default they try create /lava-NNNNN/ directory for results, which wont be created, as it is read-only partition, and etc).

gctucker:
OK so we can pull the dmesg logs over ssh to create the local files:
```sh
    dmesg --level=$level --notime -x -k > dmesg.$level
```
would become
```sh
    ssh root@192.something dmesg --level=$level --notime -x -k > dmesg.$level
```

I guess the `ssh root@something` bit could be passed as a command line argument to the script:
```sh
    wrapper=$1
    ${wrapper} dmesg --level=$level --notime -x -k > dmesg.$level
```
Then we would run `dmesg.sh "ssh root@something"` for Chrome OS and nothing changes for buildroot.

nuclearcat:
Added variant as suggested over shell script with wrapper variable.
Slightly modified test (added err and warn as other levels dont have any errors) results: https://lava.collabora.co.uk/scheduler/job/6666737

nuclearcat:
> Please don't add `err` and `warn`... we should keep the results consistent with the main version
This is only for tests (not in git), to have at least some fail result, because other levels have zero messages.

gctucker (commented):
Please split this into 2 commits, changing the dmesg.sh script is unrelated to the cros-tast change and they may get merged later into `main` separately. Also I guess you can add commits to actually run `dmesg.sh` over SSH from the Docker container, in the LAVA template.

nuclearcat:
One more important question, where to add dmesg.sh (i guess as one more commit in this PR):
1)Add one more test cros-something
2)Add it to cros-boot (just extend and add "baseline" docker there)
3)Add to cros-tast in existing docker commands

gctucker:
I think we basically need `cros-baseline`, then once it's merged and running fine we'll drop `cros-boot`. Well we might keep the cros-boot definitions for manual submissions but stop listing it in the test_configs.

nuclearcat:
Minor problem i am facing, staging will fetch dmesh.sh from chromeos.kernelci.org branch (over github url), which wont contain patch at moment of staging tests

gctucker:
> Minor problem i am facing, staging will fetch dmesh.sh from chromeos.kernelci.org branch (over github url), which wont contain patch at moment of staging tests

Create a draft PR with the dmesg.sh changes for the `chromeos` branch, then it'll be there.

config/core/test-configs-chromeos.yaml
```diff
@@ -1,32 +1,62 @@
...
+  cros-baseline_qemu-fixed:
+    <<: *cros-baseline_qemu
+    base_name: cros-baseline-fixed
     filters: &cros-qemu-filters-fixed
       - blocklist: { defconfig: ['kselftest'] }
       - blocklist: { defconfig: ['cros'] }
       - passlist: { lab: ['lab-collabora-staging', 'lab-collabora'] }
+    params:
+      base_template: 'chromeos/cros-boot-qemu.jinja2'
+      docker_image: 'kernelci/cros-baseline'
```
gctucker, line 38: This could be inherited from `cros-baseline_qemu` to avoid repeating it.

config/docker/cros-baseline/Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+    apt-transport-https \
+    ca-certificates \
+    ssh
+
+RUN apt-get update && apt-get install -y \
+    curl \
+    git \
+    inetutils-ping \
+    vim \
+    xz-utils
```
gctucker, line 16: Do we really need all this, or even any of this?

gctucker (commented):
I can see how we might group things between cros-baseline and cros-tast into a parent image e.g. cros-ssh and make the others inherit from it as they've very similar.

config/docker/cros-baseline/Dockerfile
nuclearcat: git yes, to extract private key from ChromiumOS repository.
inetutils yes, we are using to wait DUT to come up.
Rest i will try to remove.

nuclearcat: That will not work well, as Linux have SYN retransmit interval is arithmetic progression. E.g. first retry X sec, second X*2, X*3.

gctucker: Well, works for me:
```
$ while ! ssh staging.kernelci.org ls /tmp/nope; do date; sleep 1; done
ls: cannot access '/tmp/nope': No such file or directory
Tue 28 Jun 07:27:59 BST 2022
ls: cannot access '/tmp/nope': No such file or directory
Tue 28 Jun 07:28:04 BST 2022
ls: cannot access '/tmp/nope': No such file or directory
Tue 28 Jun 07:28:08 BST 2022
ls: cannot access '/tmp/nope': No such file or directory
Tue 28 Jun 07:28:11 BST 2022
ls: cannot access '/tmp/nope': No such file or directory
Tue 28 Jun 07:28:15 BST 2022
ls: cannot access '/tmp/nope': No such file or directory
Tue 28 Jun 07:28:19 BST 2022
ls: cannot access '/tmp/nope': No such file or directory
Tue 28 Jun 07:28:22 BST 2022
ls: cannot access '/tmp/nope': No such file or directory
Tue 28 Jun 07:28:26 BST 2022
ls: cannot access '/tmp/nope': No such file or directory
Tue 28 Jun 07:28:29 BST 2022
/tmp/nope
```
(I created the file eventually to stop the loop)

nuclearcat: In this particular case it looks ok because host is reachable. 
In our case DUT sometimes is unreachable for 10-30 seconds (especially QEMU that is always "first boot").

With exponential syn retry backoff TCP retries look like this (10.255.255.153 host, 1.2.3.4 is imaginary, unreachable DUT)
'''
09:37:37.147578 IP 10.255.255.153.44016 > 1.2.3.4.23: Flags [S], seq 1734573241, win 64240, options [mss 1460,sackOK,TS val 4052930222 ecr 0,nop,wscale 7], length 0
09:37:38.173547 IP 10.255.255.153.44016 > 1.2.3.4.23: Flags [S], seq 1734573241, win 64240, options [mss 1460,sackOK,TS val 4052931248 ecr 0,nop,wscale 7], length 0
09:37:40.193530 IP 10.255.255.153.44016 > 1.2.3.4.23: Flags [S], seq 1734573241, win 64240, options [mss 1460,sackOK,TS val 4052933268 ecr 0,nop,wscale 7], length 0
09:37:44.285616 IP 10.255.255.153.44016 > 1.2.3.4.23: Flags [S], seq 1734573241, win 64240, options [mss 1460,sackOK,TS val 4052937360 ecr 0,nop,wscale 7], length 0
09:37:52.477535 IP 10.255.255.153.44016 > 1.2.3.4.23: Flags [S], seq 1734573241, win 64240, options [mss 1460,sackOK,TS val 4052945552 ecr 0,nop,wscale 7], length 0
09:38:08.609560 IP 10.255.255.153.44016 > 1.2.3.4.23: Flags [S], seq 1734573241, win 64240, options [mss 1460,sackOK,TS val 4052961684 ecr 0,nop,wscale 7], length 0
09:38:41.121525 IP 10.255.255.153.44016 > 1.2.3.4.23: Flags [S], seq 1734573241, win 64240, options [mss 1460,sackOK,TS val 4052994196 ecr 0,nop,wscale 7], length 0
'''
That means if DUT will boot just after 09:38:08, it will have to wait next SYN packet at 09:38:41 (extra up to 33 seconds).
Thats actually reason i added this ping loop in first place, to speed up process.

config/core/test-configs-chromeos.yaml
nuclearcat: Updated

config/docker/cros-baseline/Dockerfile
nuclearcat: Removed all except ping, as i prefer to discuss it further.
Also ping proven to be useful in identifying and troubleshooting LAVA failures, as it provides extended network information (for example incorrect gateway address and etc).

gctucker: OK, how about ping and then ssh? That way you know the host is up, and then you can wait for the SSH server to be ready.

nuclearcat: Thats exactly what i am doing now (in other tests, such as tast - too):
https://github.com/kernelci/kernelci-core/pull/1261/commits/bce31bdcd87373bc9d6e866e251c9a1f31559d8f#diff-2bc116dc65df2b517c2e3af68f4b6e7f6307477639f85515be49300a23b4f021R22

gctucker: I see. Then we might add `-o ConnectTimeout` with a reasonable value to ensure the ssh client timeout is long enough (not sure what the default is and that might change anyway). Maybe 5min, so 300 seconds.

config/docker/build-and-push.sh
```diff
@@ -54,6 +54,7 @@ all_targets="\
 build-base \
 buildroot \
 compilers \
+cros-baseline \
```
gctucker, line 57: Please drop this.

config/docker/cros-baseline/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+
+ARG DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update && apt-get install --no-install-recommends -y \
+    apt-transport-https \
+    ca-certificates \
+    ssh
+
+RUN apt-get update && apt-get install -y \
+    inetutils-ping
```
gctucker, line 12: Add git:
```
Step 11/17 : RUN git clone     https://chromium.googlesource.com/chromiumos/chromite     $HOME/trunk/chromite
 ---> Running in 31f0ad9fb74d
/bin/sh: 1: git: not found
The command '/bin/sh -c git clone     https://chromium.googlesource.com/chromiumos/chromite     $HOME/trunk/chromite' returned a non-zero code: 127
```

gctucker (changes requested):

config/docker/cros-baseline/Dockerfile
nuclearcat: Fixed that, sorry forgot to run manual build

config/docker/build-and-push.sh
nuclearcat: done

config/docker/cros-baseline/Dockerfile
```diff
@@ -0,0 +1,41 @@
+FROM debian:bullseye-slim
+MAINTAINER "KernelCI TSC" <kernelci-tsc@groups.io>
+
+ARG DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update && apt-get install --no-install-recommends -y \
+    apt-transport-https \
+    ca-certificates \
+    ssh
```
gctucker, line 9: I think `ssh` should ideally be in the list below, with `git` and `ping`. This part about certificates is more about being able to use HTTPS apt sources.  It's not a big issue, but keeping it the same as other Docker images can help with caching.

config/lava/chromeos/cros-baseline.jinja2
```diff
@@ -0,0 +1,30 @@
...
+        metadata:
+          format: Lava-Test Test Definition 1.0
+          name: cros-baseline
+        run:
+          steps:
+            - cd /home/cros
+            - while ! ping -c 1 -w 1 $(lava-target-ip); do sleep 1; done
+            - >-
+              KERNELCI_LAVA=y
+              cmdwrapper="ssh -i /home/cros/.ssh/id_rsa -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null root@$(lava-target-ip)"
```
gctucker, line 25: So how about adding the timeout option, to be sure it's long enough?

nuclearcat: If host is pingable, it should not fail if ssh up too, and will be long enough by default, as i recall it is based on default SYN timeout + number of retries.
We might face case when ssh is not up yet, and connection refused, but as i recall ssh don't have mechanism to retry on such case.

config/docker/cros-baseline/Dockerfile
nuclearcat: Moved

nuclearcat:
https://chromeos.kernelci.org/test/job/kernelci/branch/chromeos-stable/kernel/chromeos-stable-20220709.0/plan/cros-baseline-fixed/
cros-baseline-fixed passing tests

config/docker/cros-baseline/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+  adduser cros sudo && \
+  echo "cros ALL=(ALL) NOPASSWD:ALL" >> /etc/sudoers
+USER cros
+ENV HOME=/home/cros
+WORKDIR $HOME
+
+# All this just to extract rsa key to access DUT
+# Adding private key "as is" might trigger github audit alert
+RUN mkdir -p $HOME/trunk
+RUN git clone \
```
gctucker, line 28: Guess it could have been a shallow clone?

gctucker (approved):

config/docker/cros-baseline/Dockerfile
nuclearcat: I think yes, i will test it and if shallow clone works, will send update as it will make things faster

PR #1265, opened 2022-06-21
T35078 Get all nodes matching attributes from API

JenySadadia:
Implement KernelCI_API.get_nodes
to get all the nodes from API.

Used this in https://github.com/kernelci/kernelci-pipeline/pull/71

### DIFF
--- a/kernelci/db/kernelci_api.py
+++ b/kernelci/db/kernelci_api.py
@@ -39,9 +39,9 @@ class KernelCI_API(Database):
     def _make_url(self, path):
         return urllib.parse.urljoin(self.config.url, path)
 
-    def _get(self, path):
+    def _get(self, path, params=None):
         url = self._make_url(path)
-        resp = requests.get(url, headers=self._headers)
+        resp = requests.get(url, params=params, headers=self._headers)
         resp.raise_for_status()
         return resp
 
@@ -87,6 +87,11 @@ class KernelCI_API(Database):
     def get_node(self, node_id):
         resp = self._get('/'.join(['node', node_id]))
         return json.loads(resp.text)
+
+    def get_nodes(self, attributes: dict = None):
+        """Get all nodes matching attributes"""
+        resp = self._get('nodes', params=attributes)
+        return resp.json()
 
     def get_nodes_by_commit_hash(self, commit_hash):
         """
--- a/kernelci/db/kernelci_api.py
+++ b/kernelci/db/kernelci_api.py
@@ -93,14 +93,6 @@ class KernelCI_API(Database):
         resp = self._get('nodes', params=attributes)
         return resp.json()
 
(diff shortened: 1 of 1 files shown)


gctucker (commented):
Sorry we can't do that, it will not scale and eventually break the server when we get to a real number of nodes.  We might be able to add pagination to receive nodes in chunks, but that's something we should be adding to the existing methods that already retrieve nodes with some filters.

JenySadadia:
> Sorry we can't do that, it will not scale and eventually break the server when we get to a real number of nodes. We might be able to add pagination to receive nodes in chunks, but that's something we should be adding to the existing methods that already retrieve nodes with some filters.

Makes sense. I can change the method to get only pending nodes. That is what we need in the `set_timeout` script.

gctucker:
We could also just have a `get_nodes()` method with some filters that would be passed to the `/nodes` endpoint.  We should also set a maximum number of nodes until we have support for paginated results.

JenySadadia:
> We could also just have a `get_nodes()` method with some filters that would be passed to the `/nodes` endpoint. We should also set a maximum number of nodes until we have support for paginated results.

Updated the method. I will also send a PR to drop the `get_nodes_by_commit_hash()` method as now `get_nodes()` can be used to get nodes by commit hash as well.

JenySadadia:
I think `get_child_nodes_from_parent` can be dropped too. What do you suggest? @gctucker

kernelci/db/kernelci_api.py
```diff
@@ -88,6 +88,12 @@ def get_node(self, node_id):
         resp = self._get('/'.join(['node', node_id]))
         return json.loads(resp.text)
 
+    def get_nodes(self, attributes: dict = None):
+        """Get all nodes matching attributes"""
+        resp = self._get('nodes', params=attributes)
+        # Maximum 10 nodes will be returned
+        return resp.json()[:10]
```
gctucker, line 95: I think the limit should be an argument of the method, and ideally this should be implemented as pagination in the API. Then we should also have a `skip` argument to get the next "page" of nodes.

JenySadadia: Implemented pagination in the API https://github.com/kernelci/kernelci-api/pull/164.
`skip` and `limit` arguments can be sent from client scripts using `attributes` dictionary.

gctucker (approved):
Tested OK on staging.

Got the email report:
```
kernelci/staging-mainline staging-mainline-20220804.0: 2 runs 0 fails

Tests Summary
-------------

Tree:     kernelci
Branch:   staging-mainline
Describe: staging-mainline-20220804.0
URL:      https://github.com/kernelci/linux.git
SHA1:     0b109692b29381c4e3d9d1adff0d4f120ee78f84

test            | result
----------------+-------
kver            | pass
kunit           | pass
```

PR #1270, opened 2022-06-23
doc: update kci_data section about adding API users

gctucker:
Update the documentation section about how to create API users with
kci_data.  Add relative links to other parts of the documentation,
wrap long lines, mention it's still under development and simplify the
sample commands.

### DIFF
--- a/doc/kci_data.md
+++ b/doc/kci_data.md
@@ -125,34 +125,32 @@ having an API token for the KernelCI staging API.  This was useful to be able
 to show the results on the web dashboard.  The same command can be run with a
 local instance using `--db-config=localhost`.
 
-## Creating user
+## Creating new API user (experimental)
 
-We can create new users of KernelCI-API using `kci_data`.
+We can create users for the new [KernelCI API](/docs/api) using `kci_data`.
+This is still under development and it's not used in production by KernelCI.
 
-We need to provide an API token to access `/user` endpoint of API.
-Please refer the instructions [here](https://kernelci.org/docs/api/api-details/#create-an-api-token-with-security-scopes) to create token for admin user.
+An API token with admin permissions is required in order to do this, please see
+the documentation about [how to create
+one](/docs/api/api-details/#create-an-api-token-with-security-scopes).
 
-For security purpose, instead of providing `--db-token` from the comman line we can add `db_token` in settings file i.e. `kernelci.conf`.
+For security reasons, instead of providing `--db-token` from the command line
+we can add `db_token` in the [settings file](../settings) e.g. `kernelci.conf`:
 
-```
+```ini
 [db:api]
 
 db_token: XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX
 ```
 
-> **Note** Only admin users are permitted to use this utility of creating users.
-
-
-Command to create a regular user would be:
+Here's a sample command to create a regular user:
 
 ```
 ./kci_data create_user --username test --password test --db-config api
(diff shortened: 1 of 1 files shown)


doc/kci_data.md
```diff
@@ -125,34 +125,32 @@ having an API token for the KernelCI staging API.  This was useful to be able
...
 
-> **Note** Only admin users are permitted to use this utility of creating users.
-
-
-Command to create a regular user would be:
+Here's a sample command to create a regular user:
 
 ```
 ./kci_data create_user --username test --password test --db-config api
-{'_id': '628c81e641616ced4e31a10b', 'username': 'test', 'hashed_password': '$2b$12$Ev72ythf0jViigUOtHuifOW6Mtrr0RmwQx0zO1CIy5XMwrN57bhuW', 'active': True, 'is_admin': False}
```
JenySadadia, line 150: Shouldn't we include what the response would look like?

gctucker: No because I don't think `kci_data` should be printing this, it looks more like debugging data to me. `curl` prints it because it's a low-level tool for doing HTTP queries, but `kci_data` is a command line tool for KernelCI users.

gctucker: Typically, if it worked then nothing needs to be printed. That's how Unix command line tools work. Maybe we can have a verbose mode to print something like "User created.". Then if there's an error then yes it should be printed of course.

JenySadadia (changes requested):
Need to add `--yaml-config` flag to use the configuration file from the pipeline.
Also, need to change the value of `--db-config`.

doc/kci_data.md
```diff
@@ -125,34 +125,32 @@ having an API token for the KernelCI staging API.  This was useful to be able
...
+the documentation about [how to create
+one](/docs/api/api-details/#create-an-api-token-with-security-scopes).
 
-For security purpose, instead of providing `--db-token` from the comman line we can add `db_token` in settings file i.e. `kernelci.conf`.
+For security reasons, instead of providing `--db-token` from the command line
+we can add `db_token` in the [settings file](../settings) e.g. `kernelci.conf`:
 
-```
+```ini
 [db:api]
```
JenySadadia, line 141:
```suggestion
[db:docker-host]
```

gctucker: This is coming from the `kernelci-pipeline` YAML configuration, which is not something `kernelci-core` depends on. The dependency goes the other way round. The code in `kernelci-core` for supporting the new API is still incomplete, and so the documentation will also have some gaps for a while.

As discussed in a meeting today, providing sample commands that work with the pipeline YAML config should be part of the "getting started" pipeline documentation page. The standard `kci_data` documentation page should have generic examples that are agnostic of the code in `kernelci-pipeline`. So I don't think these latest suggestions should be applied. Some tweaks could (should) be made in the pipeline documentation to provide similar commands that actually work though, so I'll take a look at that next.

JenySadadia: But we should provide all the information needed to use this particular functionality of kci_data.

gctucker: I think the page explains how to use the `kci_data create_user`, or is it missing something? The pipeline "getting started" guide should explain how to do this in the particular case of the pipeline which provides a YAML file with `docker-host`. I've sent a PR to cover that, it's now on staging: https://static.staging.kernelci.org/docs/api/getting-started/#create-an-admin-user-account

JenySadadia: Okay, so I think we need to provide a link here to pipeline documentation containing configurations related to this functionality.

gctucker: Again, this is going in the wrong direction. The kernelci-core tools don't depend on kernelci-pipeline. There is a link from the "getting started" page to the `kci_data` page as a reference, in addition to sample commands. The code in `kernelci-core` is not complete yet and so is the documentation, but that's fine.

JenySadadia (approved):

gctucker:
Thanks for the reviews. I guess this can be merged now it has been approved, then it will need to be updated as the code evolves.

Issue #948, opened 2021-12-16
LTP POSIX test results are not parsed


gctucker:
So far, KernelCI LTP tests have been using using the `20200515` version of LTP which is working with the current LAVA test-definitions.  However, when using the latest version on the `master` branch, the POSIX test results don't appear to be parsed correctly and nothing gets reported to LAVA.  Here's an example job:
https://lava.collabora.co.uk/scheduler/job/5217395#L3219

Other LTP tests are producing results correctly, such as IPC etc.

PR #1279, opened 2022-07-01
rootfs/ltp: add workaround patch for #948

gctucker:
The commit introducing the issue #948 can't be reverted any more due
to conflicts with more recent changes.  Add a patch to do the same
thing with the latest code instead.


gctucker:
Tested OK on staging: https://lava.collabora.co.uk/scheduler/job/6723326

PR #1280, opened 2022-07-01
CHROMIUMOS test-configs-chromeos.yaml: Remove blocklist for cros

nuclearcat:
As we need cros://chromeos-5.10/x86_64/chromiumos-x86_64.flavour.config
for proper qemu support, we should remove "cros" from blocklist.
Device already have proper filter.
It will solve issue with missing tests for -fixed tests.

### DIFF
--- a/config/core/test-configs-chromeos.yaml
+++ b/config/core/test-configs-chromeos.yaml
@@ -9,7 +9,6 @@ test_plans:
     <<: *cros-boot
     filters: &cros-filters-fixed
       - blocklist: { defconfig: ['kselftest'] }
-      - blocklist: { defconfig: ['cros'] }
     params:
       fixed_kernel: true
 
@@ -25,7 +24,6 @@ test_plans:
     base_name: cros-boot-fixed
     filters: &cros-qemu-filters-fixed
       - blocklist: { defconfig: ['kselftest'] }
-      - blocklist: { defconfig: ['cros'] }
       - passlist: { lab: ['lab-collabora-staging', 'lab-collabora'] }
     params:
       fixed_kernel: true


gctucker (changes requested):
This is for the `-fixed` tests, which don't use the kernel built by KernelCI. The point of the filters was to only run these tests once since all of them use the same kernel binaries anyway.

gctucker:
So I think this PR should be closed.

nuclearcat:
I suggest to keep it, because of 2 reasons:
1)
> This is for the `-fixed` tests, which don't use the kernel built by KernelCI. The point of the filters was to only run these tests once since all of them use the same kernel binaries anyway.
-fixed tests are trigged now by kernel build, with this filter, by any kernel that doesnt contain "cros" (even they dont use this kernel build), which likely means either vanilla kernel or just x86_chromebook fragment. If both are built, i assume fixed test might be fired twice for same kernel release.
-hp-x360-12b-ca0010nr-n4020-octopus_chromeos and qemu_x86_64-uefi-chromeos have filters for chromeos kernel that require to be built with "cros://" fragments.
That means -fixed tests will never run, which is happening and reason why this PR made.

2)It is good to have same kernel triggering in jenkins -fixed and kernel tests, for better consistency and timing. Sometimes pineview for example is not built, and we will have cros-tast-fixed results, but we will not have matching cros-tast results.

gctucker:
> I suggest to keep it, because of 2 reasons: 1)
> 
> > This is for the `-fixed` tests, which don't use the kernel built by KernelCI. The point of the filters was to only run these tests once since all of them use the same kernel binaries anyway.
> 
> -fixed tests are trigged now by kernel build, with this filter, by any kernel that doesnt contain "cros" (even they dont use this kernel build), which likely means either vanilla kernel or just x86_chromebook fragment. If both are built, i assume fixed test might be fired twice for same kernel release. -hp-x360-12b-ca0010nr-n4020-octopus_chromeos and qemu_x86_64-uefi-chromeos have filters for chromeos kernel that require to be built with "cros://" fragments. That means -fixed tests will never run, which is happening and reason why this PR made.

I guess the difference now is that devices have more filters to run with only one kernel, including QEMU. So dropping some filters in the test plans may not mean more jobs will be run.

> 2)It is good to have same kernel triggering in jenkins -fixed and kernel tests, for better consistency and timing. Sometimes pineview is not built, and we have cros-tast-fixed results, but we wont have matching cros-tast results.

There is really no point in running the `-fixed` tests with more than one kernel on each platform. If some kernel builds are missing that's another issue to fix.


There appear to be lots of issues with running Chrome OS tests now as there's only a couple of results and no `-fixed` results at all:
https://chromeos.kernelci.org/test/job/kernelci/branch/chromeos-stable/kernel/chromeos-stable-20220706.0/plan/cros-boot/

The regular 5.10 job got more results a few days ago:
https://chromeos.kernelci.org/test/job/stable/branch/linux-5.10.y/kernel/v5.10.128/

I guess we can merge this and when we have jobs working again and producing all the results then we can double check that none of the `-fixed` tests are run twice to avoid wasting lab capacity.

PR #1288, opened 2022-07-07
CHROMEOS Dockerfile: Add custom repository to build-base

nuclearcat:
As ChromeOS builds require compressed modules and debian kmod by default
doesn't support zlib compression, we add repository with kmod compiled
with proper compression support.
Without compression support knod cannot generate proper dependencies
during build, which can be seen as several dep files
in /lib/modules/kver/ will have 0 size.

### DIFF
--- a/config/docker/build-base/Dockerfile
+++ b/config/docker/build-base/Dockerfile
@@ -9,7 +9,14 @@ RUN apt-get update && apt-get install --no-install-recommends -y procps
 # SSL / HTTPS support
 RUN apt-get update && apt-get install --no-install-recommends -y \
     apt-transport-https \
-    ca-certificates
+    ca-certificates \
+    wget
+
+# Add custom build tools repository
+RUN echo 'deb https://repositories.collabora.com/kernelci/ bullseye tools' \
+   >> /etc/apt/sources.list.d/tools.list
+RUN wget https://repositories.collabora.com/kernelci/pool/tools/c/collabora-archive-keyring/collabora-archive-keyring_0.3+debian11.3_all.deb && \
+    dpkg -i collabora-archive-keyring_0.3+debian11.3_all.deb
 
 # Host build tools
 RUN apt-get update && apt-get install --no-install-recommends -y \


gctucker:
I see the issue here, the `chromeos.kernelci.org` jobs use `kernelci/staging-` as a base like `staging.kernelci.org` but this PR is only merged into the `chromeos.kernelci.org` branch. So with how things are currently set up, we can have it only for `chromeos.` and not for `staging.`.

We probably need to start creating a separate set of Docker images for `chromeos.` with a different prefix e.g. `kernelci/cros-` to be able to deal with this kind of situation.

gctucker:
@nuclearcat Please rebase.

I guess we then have two options here: either get the patch merged in the upstream Debian package so it directly supports compression or have a separate set of Docker images for the Chrome OS builds. Or maybe we could also have a KernelCI deb package repository as a third alternative. But I don't think we should merge this as-is in the main branch with the Collabora deb repository.

nuclearcat:
I think it is unlikely that this patch will be accepted in the bullseye, at best in the next release.
Moreover, we are using llvm.org apt repository for clang, so maybe using Collabora apt repository for a while is ok, until we find more permanent solution. 
Ideally, if we will convince Debian maintainer to include compression support.

As another option, if we really cannot use repository of Collabora, we can simply host deb files on storage.kernelci and install them via wget/dpkg -i

Or as you suggested separate set of Docker images, we add the prefix cross- for all build packages. cros-build-base, cros-clang-11, and if they do not differ from the original package, these may be just a symlinks.

nuclearcat:
Yes, i just applied trivial patch to debian-specific files
```
--- kmod-28/debian/rules	2021-01-08 02:46:40.000000000 +0200
+++ kmod-28-kernelci/debian/rules	2022-06-13 14:29:07.712238830 +0300
@@ -21,6 +21,7 @@
   --enable-gtk-doc \
   --with-openssl \
   --with-xz \
+  --with-zlib \
   --enable-debug
 CONFFLAGS_udeb = $(CONFFLAGS) \
   $(subst -O2,-Os -fomit-frame-pointer,$(shell DEB_BUILD_MAINT_OPTIONS="hardening=-all" \
```
Maybe i should create directory with script to rebuild this deb files, probably inside docker image, similar way as we build kernels?

nuclearcat:
I guess this PR can be dropped in favor of https://github.com/kernelci/kernelci-core/pull/1318 ?

PR #1289, opened 2022-07-07
T36372 Update base template for new node schema

JenySadadia:
Need to update the value of `Node.status` and add `Node.result` to accommodate new node schema from API.


gctucker (commented):
LGTM, let's see if we can get it tested on staging.

gctucker:
Please change the commit subjects to something more descriptive. "Modify" doesn't mean much, each git commit is modifying something.

JenySadadia:
> Please change the commit subjects to something more descriptive. "Modify" doesn't mean much, each git commit is modifying something.

Updated the commits.

gctucker:
Tested OK on staging.

PR #1297, opened 2022-07-14
CHROMEOS rootfs/chromiumos: fix hard-coded user in build_board.sh

gctucker:
Use the ${USERNAME} variable in build_board.sh rather than the
hard-coded "user" name.


nuclearcat (approved):
LGTM

PR #1300, opened 2022-07-15
T33737 update kselftest make commands

gctucker:
Call `make headers` first for kernels earlier than v5.20 and just call `make kselftest-gen_tar` for `v5.20` onwards. This relies on the changes described in https://github.com/kernelci/kernelci-project/issues/92 to have been merged in mainline (which should happen during the v5.20 merge window).

We may drop the 2-step make approach once the next LTS has been declared and stop building kselftest for earlier kernels (say, `v6.0` or something).

### DIFF
--- a/kernelci/build.py
+++ b/kernelci/build.py
@@ -1490,8 +1490,16 @@ class MakeSelftests(Step):
         opts = {
             'FORMAT': '.xz',
         }
-        res = self._make('gen_tar', jopt, verbose, opts,
-                         'tools/testing/selftests')
+        if self._check_min_kver(5, 20):
+            # Once v5.20 has been released and a new LTS has been declared then
+            # we can always run this command and bump the v5.10 version test
+            # for kselftest altogether.
+            res = self._make('kselftest-gen_tar', jopt, verbose, opts)
+        else:
+            res = self._make('headers', jopt, verbose)
+            if res:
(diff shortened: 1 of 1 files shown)


nuclearcat:
I think there is typo and it is breaking staging:
https://github.com/kernelci/kernelci-core/issues/1323
as there is defined
```
KVER_RE = re.compile(r'^v([\d]+)\.([\d]+)')
```
but removed
```
-        self._kver_re = re.compile(r'^v([\d]+)\.([\d]+)')
```

gctucker:
Ah yes, thanks.  Will fix that.

gctucker:
Tested OK on staging:
https://staging.kernelci.org/build/id/62e71dca626fc02769e19dbd/
https://storage.staging.kernelci.org/kernelci/staging-mainline/staging-mainline-20220801.0/x86_64/x86_64_defconfig+kselftest/gcc-10/logs/kselftest.log

https://staging.kernelci.org/build/id/62e722047c90c320e1e19edd/
https://storage.staging.kernelci.org/kernelci/staging-mainline/staging-mainline-20220801.0/x86_64/x86_64_defconfig+crypto/gcc-10/logs/kselftest.log

gctucker:
@nuclearcat Would you like to take another look now this appears to have been fixed?  Ideally this should be in production on Monday as the changes in the kselftest Makefiles have been merged in mainline now.

PR #1301, opened 2022-07-16
config/runtime: consolidate shell and kubernetes templates

gctucker:
config/runtime: consolidate shell and kubernetes templates

Consolidate the templates for running Python tests in either a local
shell or in Kubernetes by moving using a variable to extend a base
template depending on the runtime name and grouping all the base
templates under config/runtime/base.

Update the default paths for the lab.shell and lab.kubernetes
accordingly to 'config/runtime'.

Actual runtime "payload" templates can then be added to config/runtime
and extend a base template such as base/python.jinja2 for payloads
written in Python.


gctucker (commented):
Tested OK on staging with new pipeline, and no issues reported so far.

PR #1303, opened 2022-07-17
CHROMEOS cros-snapshot.xml: Add manifest snapshot support

nuclearcat:
As we experienced problems with rootfs builds due rolling release
nature of ChromeOS tags, we introduce repo snapshot feature that
will allow us to freeze source tree to specific snapshot,
and build rootfs based on this snapshot

### DIFF
--- a/config/rootfs/chromiumos/scripts/build_board.sh
+++ b/config/rootfs/chromiumos/scripts/build_board.sh
@@ -5,6 +5,9 @@ BRANCH=$2
 SERIAL=$3
 DATA_DIR=$(pwd)
 USERNAME=$(/usr/bin/id -run)
+SCRIPT=$(realpath "$0")
+SCRIPTPATH=$(dirname "$SCRIPT")
+MANIFESTDIR=$(dirname "$SCRIPTPATH")
 
 function cleanup()
 {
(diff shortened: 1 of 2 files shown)


gctucker:
I think the snapshot manifest file name should include the revision it was taken from, e.g. R100 or the branch name. Then the manifest file name and git URL can be passed as arguments to `build_board.sh` which would pass it on to `repo init`.

nuclearcat:
I added current branch as parameter and variable in filename "construct". Still need to test build by hand, is it better like this?

config/rootfs/chromiumos/scripts/build_board.sh
```diff
@@ -25,7 +28,14 @@ git config --global user.email "bot@kernelci.org"
 git config --global user.name "KernelCI Bot"
 git config --global color.ui false
 repo init --repo-url https://chromium.googlesource.com/external/repo --manifest-url https://chromium.googlesource.com/chromiumos/manifest --manifest-name default.xml --manifest-branch ${BRANCH}
```
gctucker, line 30: How about using parameters for the git URL, git branch and manifest file name here? Then we could use the kernelci-core URL with the chromeos.kernelci.org branch and the snapshot name including the release version (R100 etc.).

nuclearcat: Definitely, great idea, it looks way more neat like this.
Done, updated, will do testing by hand when branch get updated

nuclearcat:
Build tested on jenkins: https://bot.staging.kernelci.org/job/denysf/job/cros-rootfs-build/10/

nuclearcat:
Should be fine to merge.

mgalka (approved):

PR #1313, opened 2022-07-20
CHROMEOS test-configs-chromeos.yaml: add chromiumos-grunt

alexandrasp:
Enable grunt device to run cros-flash test-plan.

### DIFF
--- a/config/core/test-configs-chromeos.yaml
+++ b/config/core/test-configs-chromeos.yaml
@@ -148,14 +148,29 @@ device_types:
         image: 'https://storage.chromeos.kernelci.org/images/rootfs/chromeos/chromiumos-hatch/20220621.0/amd64/bzImage'
         modules: 'https://storage.chromeos.kernelci.org/images/rootfs/chromeos/chromiumos-hatch/20220621.0/amd64/modules.tar.xz'
 
-  # Placeholder
-  # hp-11A-G6-EE-grunt_chromeos:
-  #   base_name: hp-11A-G6-EE-grunt
-  #   mach: x86
-  #   arch: x86_64
-  #   boot_method: depthcharge
-  #   params:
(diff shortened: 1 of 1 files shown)


alexandrasp:
@nuclearcat @gctucker With this changes we can but lava templates to run cros-flash at grunt, but It still missing the filters part, not sure about if it will cause any problem, so I put the `staging-skip` tag.

I already did some tests:
https://lava.collabora.dev/scheduler/job/6854818
https://lava.collabora.dev/scheduler/job/6854757

gctucker (commented):
With these changes you should be able to drop the `staging-skip` label.

alexandrasp:
PR updated, I also removed `kernel_reference` in order to add in the next PR that includes baseline test_plan.
Re-tested: https://lava.collabora.dev/scheduler/job/6863323

alexandrasp:
wow, I will rebase my branch..

gctucker:
I think it would be good to have this merged with the known issue around the block device number.  As `detect` isn't working, we could just set it to `mmcblk1` since it's the one that comes up the most often and keep track of an issue on the ChromeOS GitHub workboard about how to fix this properly.

Having the changes merged now will simplify rebase and checking the issue gets resolved.

alexandrasp:
> I think it would be good to have this merged with the known issue around the block device number. As `detect` isn't working, we could just set it to `mmcblk1` since it's the one that comes up the most often and keep track of an issue on the ChromeOS GitHub workboard about how to fix this properly.
> 
> Having the changes merged now will simplify rebase and checking the issue gets resolved.

@gctucker updated to `mmcblk1`.

gctucker (approved):
Thanks, let's merge this now and fix issues with the block device as a follow-up.

PR #1316, opened 2022-07-22
Add hatch and grunt config for Chromium OS

gctucker:
Add config entries for Chromium OS R100 rootfs images for the hatch and grunt Chromebooks.


nuclearcat (approved):
LGTM

PR #1317, opened 2022-07-22
Revert "kernelci.config.lab: add config_path attribute"

gctucker:
This reverts commit add1ffa86a8658c9134f19020395d61fbd588c01.

It's not necessary to have a configured template path as all the base
templates should be in runtime/base.  Then they can extend other
templates in sub-directories if needed e.g. "extends
base/foo-bar/something.jinja2" without any YAML config.


nuclearcat (approved):
LGTM
( did few runs with this patch, but not sure where to dig for possible failures :) )

PR #1324, opened 2022-07-27
test-configs.yaml: Add mt8192-asurada-spherion-r0

nfraprado:
Add an entry for the mt8192-asurada-spherion-r0 device and enable the
baseline tests for it.

At this moment the devicetree for this machine is only available on linux-next.

baseline test LAVA run: https://lava.collabora.dev/scheduler/job/6896370
baseline-nfs test LAVA run: https://lava.collabora.dev/scheduler/job/6900379

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -1324,6 +1324,12 @@ device_types:
     boot_method: depthcharge
     filters: *arm64-chromebook-filters
 
+  mt8192-asurada-spherion-r0:
+    mach: mediatek
+    class: arm64-dtb
+    boot_method: depthcharge
+    filters: *arm64-chromebook-filters
+
   mustang:
     mach: apm
     class: arm64-dtb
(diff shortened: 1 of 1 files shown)


nfraprado:
Oh, and by the way, if I could get CC'ed in the email notifications when a regression for this platform is detected that would be pretty handy.

nuclearcat:
We are getting test results on staging: https://staging.kernelci.org/test/plan/id/6308176b50731923dce19db1/

gctucker:
> Oh, and by the way, if I could get CC'ed in the email notifications when a regression for this platform is detected that would be pretty handy.

That's an interesting "user story", something to keep in mind for the new dashboard / user experience stuff.  Basically, individual developers would be able to subscribe to particular notifications like this.

At the moment, email reports are sent to some fixed recipients based on the git tree name and type of test.  Bisection emails have some logic to dynamically work out the recipients based on the commit found, but the emails are not sent automatically at the moment to avoid causing any spam (issues that don't get fixed will keep having new emails generated every time a new revision is pushed on the same branch e.g. linux-next).

In the meantime, we have some people with eyeballs who can ping you when they notice anything going wrong with this platform ;)

nuclearcat:
Device passing baseline tests:
https://staging.kernelci.org/test/plan/id/6314319d2b630bea78985621/
And baseline-nfs:
https://staging.kernelci.org/test/plan/id/631431d70c0777293898569f/

nuclearcat (approved):
LGTM

nuclearcat:
Device works fine in tests and passing necessary baseline tests

PR #1325, opened 2022-07-27
Enable kernel configs to get full Spherion support

nfraprado:
This is based on #1324.

The logs on #1324 may give the impression that everything was already working, but with https://github.com/kernelci/bootrr/pull/17
applied, it becomes clear a whole lot of devices aren't probing: [LAVA run](https://lava.collabora.dev/scheduler/job/6896547).

This PR enables a bunch of kernel configs on the arm64-chromebook fragment in order to get everything probing: [LAVA run](https://lava.collabora.dev/scheduler/job/6896700). I'll also try to get these configs into the defconfig upstream. With the exception of `CONFIG_SND_DYNAMIC_MINORS`, there shouldn't be any issues getting them merged, but it will probably take some time.

It also gets most of the tests added in https://github.com/kernelci/bootrr/pull/20 passing: [LAVA run](https://lava.collabora.dev/scheduler/job/6896805). Display and audio aren't enabled in mainline for spherion yet, so those tests are failing still. There's also a small variance in device naming that can be fixed with [this commit](https://gitlab.collabora.com/nfraprado/linux/-/commit/6371e9f15506a5fad3a53ed454e772daddc81fbc). I'll send it upstream, but it could also be applied here in the mean time. Result: [LAVA run](https://lava.collabora.dev/scheduler/job/6900267)

Finally, there's the decoder tests failing, due to missing firmware:
```
remoteproc remoteproc0: Direct firmware load for mediatek/mt8192/scp.img failed with error -2
```

This firmware is already in linux-firmware, but the question is if we want to have a firmware in the baseline test.
I see three options:
1. add the scp firmware to the baseline test so that the decoder probes and the check passes
2. remove the decoder probe check in the baseline test to avoid adding a firmware to baseline
3. don't add the firmware nor remove the test, let it fail for now

TL;DR: with these configs and the additional kernel patch, most bootrr tests pass for spherion. The ones that don't either haven't been enabled in mainline yet (display and audio) and will automatically pass once they are, or they're just missing a firmware (decoder).


nuclearcat:
Regarding firmware, we have option **extra_firmware** for rootfs that allow to include required firmware.
Question is that baseline is really generic test, so it is important to know @mgalka and @gctucker opinion about possibility of including firmware in it. Because if we go such way, we need to include then also AMD Stoneyridge chipset GPU firmware too.
Maybe too much peripheral specific tests should be separate, like bullseye-cros-ec and bullseye-igt?

nuclearcat:
@nfraprado can you please rebase PR and make it ready for review? I think we can test it

nuclearcat:
https://storage.staging.kernelci.org/kernelci/staging-mainline/staging-mainline-20230111.0/arm64/defconfig+arm64-chromebook/gcc-10/lab-collabora/baseline-mt8192-asurada-spherion-r0.html
Looks like more devices detected, at least can spot "mtk-vcodec-enc" and stuff, and probably more tests passing.
Most important doesnt introduce any regressions or issues, thanks!

nuclearcat (approved):
LGTM

PR #1327, opened 2022-07-28
config/runtime: add /dev/shm to kubernetes.jinja2

gctucker:
Add a tmpfs volume mounted in /dev/shm as this is required for running
KUnit tests in particular.


gctucker (commented):
Tested OK on staging.

PR #1330, opened 2022-07-28
config/runtime: add missing requests import

gctucker:
Add missing Python import for the requests package in the
base/python.jinja2 template.


gctucker (commented):
Tested OK on staging.

PR #1320, opened 2022-07-26
kci_docker

gctucker:
Add a `kci_docker` tool to produce a new set of Docker images using Jinja2 templates.  This is needed in order to provide the flexibility required when running builds and tests in Kubernetes, and to facilitate producing images with special combinations of tools for local development.

See the README on the incoming branch: https://github.com/gctucker/kernelci-core/tree/kci_docker/config/docker-new#readme

The tree of files looks like this:
```
docker-new
├── base
│   ├── clang.jinja2
│   ├── debian.jinja2
│   ├── host-tools.jinja2
│   └── python.jinja2
├── clang-11.jinja2
├── clang-12.jinja2
├── clang-13.jinja2
├── clang-14.jinja2
├── clang-15.jinja2
├── fragment
│   ├── azure-cli.jinja2
│   ├── gcloud.jinja2
│   ├── kernelci.jinja2
│   ├── kselftest.jinja2
│   ├── kubernetes.jinja2
│   └── kunit.jinja2
├── gcc-10-arc.jinja2
├── gcc-10-arm64.jinja2
├── gcc-10-arm.jinja2
├── gcc-10-armv5.jinja2
├── gcc-10-mips.jinja2
├── gcc-10-riscv64.jinja2
├── gcc-10-x86.jinja2
├── k8s.jinja2
├── kci_docker
├── kernelci.jinja2
├── README.md
└── requirements.txt
```

PR #1331, opened 2022-07-28
config/docker: drop old kernelci and k8s Dockerfiles

gctucker:
Drop the "old" kernelci and k8s Dockerfiles as this can now be built with `kci_docker`

Depends on #1320


gctucker (commented):
Tested OK on staging with new API and pipeline using `kver` and `kunit` test plans.

PR #1338, opened 2022-08-03
config/docker: fix kci_docker with no build args

gctucker:
Fix the case where there is no build argument as the args.build_arg
variable is then None.

Fixes: df6106e71add ("config/docker: add --build-arg option to kci_docker")


gctucker:
Tested OK locally.

PR #1340, opened 2022-08-03
CHROMEOS config/lava/chromeos: rename cros-baseline and cros-tast tests

mgalka:
Correct wrong test names. "dmesg" test in cros-baseline.jinja2 was named
"cros-baseline", which is incorrect. Renamed it to "dmesg".
"cros-tast" test renamed to "tast".


nuclearcat (approved):
LGTM

gctucker:
Tested OK on staging: https://chromeos.kernelci.org/test/job/kernelci/branch/chromeos-stable/kernel/chromeos-stable-20220803.1/plan/cros-baseline/

Let's keep an eye on the dashboard results in case some results are still missing, but in principle this should fix it as having the sub-group name being the same as the test plan name was a special case we don't have with other test plans.

PR #1341, opened 2022-08-04
config/docker: add kci_docker build --push option

gctucker:
Add the --push command line argument to push the image to the Docker
Hub after building it.  This is only used by the kci_docker build
command.


gctucker (commented):
Tested OK on staging

PR #1345, opened 2022-08-04
kci_docker patch files

gctucker:
Copy some patch files used in Docker images to the new config directory so they can be used by kci_docker.


nuclearcat (approved):
Looks good

gctucker:
FYI: Once this is merged I'll make another PR to update the URLs in the Dockerfile templates to use the new location.  It's possible to provide a Docker context but kci_docker normally generates the Dockerfile on the fly using templates and when doing that there's no context available.  So the patch files are downloaded from GitHub instead.

PR #1346, opened 2022-08-04
config/docker: Update clang-15 Dockerfile

mgalka:
Clang 15 got moved to llvm-toolchain-15 so the clang-15 Dockerfile needs
to be updated accordingly.


gctucker:
Nope:
```
E: The repository 'http://apt.llvm.org/bullseye llvm-toolchain-15 Release' does not have a Release file.
```
Closing in favour of #1348 which actually works and also updates the YAML config accordingly.

PR #1349, opened 2022-08-05
config/docker-new: update URLs for patch files

gctucker:
Update the URLs of the patch files used for the kmod and gcc-10 x86 images to download them from the config/docker-new/data directory on the main branch now that they have been merged.


gctucker (commented):
Tested OK on staging

PR #1352, opened 2022-08-05
build-configs.yaml: Add kernel configs for juniper

nfraprado:
Enable all kernel configs required to get all devices on
mt8183-kukui-jacuzzi-juniper-sku16 probing.

A lot of these configs are the same as in #1325. After one of the PRs is merged I'll rebase the other to have only the missing ones. I'll also send these configs upstream.

### DIFF
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -148,6 +148,24 @@ fragments:
       # This is required for sc7180-trogdor-lazor-limozeen and it hasn't been
       # merged upstream yet.
       - 'CONFIG_INTERCONNECT_QCOM_SC7180=y'
+      - 'CONFIG_MTK_CMDQ=m'
+      - 'CONFIG_DRM_ANALOGIX_ANX7625=m'
+      - 'CONFIG_SND_SOC_MT8183=y'
+      - 'CONFIG_SND_SOC_MT8183_MT6358_TS3A227E_MAX98357A=y'
+      - 'CONFIG_SND_SOC_TS3A227E=m'
+      - 'CONFIG_SPI_MT65XX=y'
+      - 'CONFIG_TOUCHSCREEN_ELAN=m'
+      - 'CONFIG_MTK_SCP=m'
+      - 'CONFIG_VIDEO_MEDIATEK_JPEG=m'
+      - 'CONFIG_TCG_TIS=m'
+      - 'CONFIG_TCG_TIS_SPI=m'
+      - 'CONFIG_TCG_TIS_SPI_CR50=y'
+      - 'CONFIG_TCG_TIS_I2C_CR50=m'
+      - 'CONFIG_ARM_MEDIATEK_CCI_DEVFREQ=m'
+      - 'CONFIG_MEDIATEK_MT6577_AUXADC=m'
+      - 'CONFIG_GENERIC_ADC_THERMAL=m'
+      - 'CONFIG_MTK_SVS=m'
+      - 'CONFIG_ATH10K_SDIO=m'
 
   crypto:
     path: "kernel/configs/crypto.config"


config/core/build-configs.yaml
```diff
@@ -145,6 +145,25 @@ fragments:
       - 'CONFIG_REGULATOR_DA9211=y'
```
aliceinwire, line 145: probably is better to add this setting to the juniper device only as they are device specific not to every arm64-chromebook unless arm64-chromebook use only juniper.

aliceinwire: you are adding this settings that are specifics to juniper Chromebook to all the arm64-chromebook not only juniper Chromebook. is this the correct behavior?

gctucker: We could have a separate fragment just for this Chromebook, but that would mean more kernel builds.  If these extra config options don't have any side-effects on the other arm64 Chromebooks then it's fine.  The only issue is that with every additional config option to the fragment, the arm64 Chromebook build becomes a bit less like the standard mainline build.  Ideally these should be merged in the arm64 defconfig upstream, but that's not always possible.

nuclearcat:
Can you rebase it please?

nfraprado:
@nuclearcat  Rebased. And also dropped `CONFIG_SPI_MTK_NOR` since it was added by mistake and is needed by mt8192, but not mt8183.

I also sent the defconfig changes upstream: https://lore.kernel.org/all/20221108232228.1177199-1-nfraprado@collabora.com

And the sound configs
```
       - 'CONFIG_SND_SOC_MT8183=y'
       - 'CONFIG_SND_SOC_MT8183_MT6358_TS3A227E_MAX98357A=y'
```

can be changed to modules after https://lore.kernel.org/all/20221107160437.740353-1-nfraprado@collabora.com is merged.

nfraprado:
The defconfig changes were merged to the mediatek tree, and should find their way to mainline for v6.2: https://lore.kernel.org/linux-mediatek/610edfca-0a3a-fd41-5208-71978866be4f@gmail.com/

But I think adding these configs to the fragment still makes sense so we can test juniper with older stable kernels.

nuclearcat (changes requested):
At the moment seems device is broken and doesnt boot in mainline: https://storage.staging.kernelci.org/kernelci/staging-mainline/staging-mainline-20221130.0/arm64/defconfig+arm64-chromebook/gcc-10/lab-collabora/baseline-mt8183-kukui-jacuzzi-juniper-sku16.html
It doesnt boot in -next either: https://storage.staging.kernelci.org/kernelci/staging-next/staging-next-20221129.1/arm64/defconfig+arm64-chromebook/gcc-10/lab-collabora/baseline-mt8183-kukui-jacuzzi-juniper-sku16.html

I checked also stable, but it doesn't look like options are working, i dont see any ALSA devices detected for example.


Also i think there is certain number of warnings appeared after enabling audio driver, for example:
```
 1371 16:46:54.629362  <3>[   11.975754] debugfs: Directory '11220000.audio-controller:mt8183-afe-pcm' with parent 'mt8183_mt6358_ts3a227_max98357' already present!
 1372 16:46:54.646699  <3>[   11.992695] mt8183_mt6358_ts3a227 mt8183-sound: ASoC: driver name too long 'mt8183_mt6358_ts3a227_max98357' -> 'mt8183_mt6358_t'
 1373 16:46:54.653364  <3>[   12.005202] debugfs: File 'Playback' in directory 'dapm' already present!
 1374 16:46:54.659919  <3>[   12.012287] debugfs: File 'Capture' in directory 'dapm' already present!
```

nfraprado:
Indeed looks like mt8183-jacuzzi-juniper is hanging during boot on next right now. From a quick check, seems to be related to the GPU, since disabling the GPU node gets it booting fine again. I know there were some recent clock and DVFS changes for the GPU on mt8183, so maybe some of that is causing issues here. I'll investigate it shortly so we can get this out of the way. (To be clear the issue is unrelated to this PR, but of course we need the machine working again before we can test and merge this PR)

Regarding the audio warnings, the driver name too long is something that needs to be taken care of before upstreaming the UCM for this machine, and I can do that soon, but otherwise is not an issue. Since this PR enables a bunch of the devices on the machine, it's somewhat expected that new warnings and errors will appear. And as long as they don't make the machine unbootable/unusable, by having the drivers be frequently tested, we can work on fixing any warnings/errors upstream and see them disappearing from these logs over time.

nfraprado:
The boot hang issue on mt8183 is due to a series being only partially applied. [I've replied to the missing patch](https://lore.kernel.org/all/20221206183037.3jnssar4yxcx5hpb@notapiano/) so soon it should be applied and the issue fixed.

nuclearcat:
Device is booting. 
https://storage.staging.kernelci.org/kernelci/staging-next/staging-next-20221219.2/arm64/defconfig+arm64-chromebook/gcc-10/lab-collabora/baseline-mt8183-kukui-jacuzzi-juniper-sku16.html

One of devices i can see properly detected:
```

[ 1388] 10:06:29.080177  <6>[   14.253299] ALSA device list:
[ 1389] 10:06:29.086606  <6>[   14.256586]   #0: mt8183_mt6358_ts3a227_max98357
```

PR #1358, opened 2022-08-09
docs: use gcc-10 instead of gcc-8 in examples

gctucker:
Use gcc-10 which is the default toolchain now instead of gcc-8 in the
documentation examples so they should actually work with the current
YAML config.


nuclearcat (approved):
LGTM

PR #1359, opened 2022-08-09
config/docker: add i386 packages to clang.jinja2

gctucker:
Add the 32-bit i386 packages to clang.jinja2 to build kselftest on
i386 and some 32-bit host tools.


gctucker (commented):
Tested OK on staging although there are still some kselftest build errors that will need to be addressed as a follow-up:
https://bot.staging.kernelci.org/job/kernel-build/77584/console

PR #1361, opened 2022-08-10
WIP: CHROMEOS: set a static part-uuid for grunt

alexandrasp:
Grunt is having a variation on block_device when booted, so the idea of this fix is to set a static part-uuid
independent of the block_device used.


alexandrasp:
hey @gctucker and @nuclearcat sorry for the delay to open this PR. 

As I reported some days ago during the daily meetings that we had I got the problem and see that is possible to solve fixing the [part-uuid](https://lava.collabora.dev/scheduler/job/6977030)! But the problem is how to set and keep the same part-uuid or even generate this during the runtime because we have to set it into the template to be used in the second boot action, I tried to update the disc information using [tune2fs](https://lava.collabora.dev/scheduler/job/6982023#L2592) and [sgdisk](https://lava.collabora.dev/scheduler/job/6982421#L2597)/[gdisk](https://lava.collabora.dev/scheduler/job/6982461#L2675) also including new libraries, but it is not working as expected for all devices automatically yet..

config/lava/chromeos/cros-boot.jinja2
```diff
@@ -94,7 +94,11 @@ actions:
       minutes: 5
     method: depthcharge
     commands: emmc
+{%- if block_device == 'detect' %}
+    extra_kernel_args: cros_secure cros_debug root=/dev/2dc32764-6303-9e41-8ee0-164634dd1da2 rootwait ro
```
gctucker, line 98:
```suggestion
    extra_kernel_args: cros_secure cros_debug root=PARTUUID=2dc32764-6303-9e41-8ee0-164634dd1da2 rootwait ro
```

config/rootfs/debos/overlays/cros-flash/opt/chromeos/install-modules
```diff
@@ -56,6 +73,12 @@ install_modules()
     cd "$root_path"
     umount "$mount_dir"
     sync
+
+    if [ $BLOCK_DEVICE_DETECT==1 ]; then
+        blkid
+        sfdisk --part-uuid /dev/${block_device}p3 39ea80c4-e748-47eb-835c-64025de53e26
```
gctucker, line 79: Looks like a different UUID from the one used with `root=`

gctucker (commented):
OK thank you, a few things to fix I think.  Maybe we could drop the logic about `detect` at some point and treat all the Chromebooks in the same way.  At least we could always boot with `PARTUUID` and maybe still have the option to manually specify the block device used for flashing.

@nuclearcat maybe you could take a look?

nuclearcat:
I think it will be good to set UUID on all Chromebooks as @gtucker suggested, so we can use same boot commandline
```
"root=PARTUUID=2dc32764-6303-9e41-8ee0-164634dd1da2" for all Chromebooks.
```
To avoid unnecessary writes in modules install we can check if it is already exist, for example :
```
blkid -t PARTUUID="2dc32764-6303-9e41-8ee0-164634dd1da2"
```
Then check return code, if it is non-zero, then we need to run part of script that install UUID.

gctucker:
> To avoid unnecessary writes in modules install we can check if it is already exist, for example :

The UUID should be set by the flashing job, not the modules installer job.  Why check during modules install?

nuclearcat:
> > To avoid unnecessary writes in modules install we can check if it is already exist, for example :
> 
> The UUID should be set by the flashing job, not the modules installer job. Why check during modules install?
At current patch i saw it is done in modules installer.
If by flasher job then yes, everything is simpler.

nuclearcat:
Taken as reference and completed in https://github.com/kernelci/kernelci-core/pull/1381

PR #1363, opened 2022-08-10
config/k8s: run builds as root user

gctucker:
Run the builds as root user with the new Docker images that set the
default user to kernelci (1000).  The previous images used root by
default, and the build artifacts were all owned by root as a result.
It's important to keep that as kernel files are expected to be owned
by root.

A follow-up improvement could be to run the build with the regular
user and then have a step in another container to change the ownership
of the artifacts to root before uploading them.

### DIFF
--- a/config/k8s/job-build.jinja2
+++ b/config/k8s/job-build.jinja2
@@ -49,9 +49,14 @@ spec:
         - mountPath: "/scratch"
           name: scratch-volume
 
+        # Run as root so the build artifacts have root uid:gid set on them
+        securityContext:
+          runAsUser: 0
+          runAsGroup: 0
+
         command: ["/bin/bash", "-x", "-c"]
         args: ["\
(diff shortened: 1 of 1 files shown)


nuclearcat (approved):
whoami shows root in https://bot.staging.kernelci.org/job/kernel-build/76466/console

gctucker:
Tested OK on staging:
```
$ wget https://storage.staging.kernelci.org/kernelci/staging-mainline/staging-mainline-20220810.0/arm/multi_v7_defconfig+crypto/gcc-10/modules.tar.xz
$ sudo tar xf modules.tar.xz
$ ls -l
drwxr-xr-x 3 root    root       4096 Aug 10 13:11 lib
```
Whereas before this change, the `lib` directory had uid 1000.

gctucker:
Merging now so we can rebase the `chromeos` branch on top of it.

PR #1366, opened 2022-08-11
CHROMEOS config/core/build-configs-chromeos.yaml: Rework extra_configs

hardboprobot:
The current ChromeOS tests for kernel v5.15 use a v5.15 kernel built with ChromeOS kernel config fragments from v5.10. I ran the same tests on a v5.15 kernel with v5.15 config fragments and it fared a bit better, it could at least reach a login prompt (https://lava.collabora.dev/scheduler/job/7012884).

This commit reworks the variants definitions for ChromeOS builds. I have only tested this manually, though. `kci_build` seems to ignore the variants definitions and the processing of ChromeOS-specific config fragments is still done in a separate step from the rest of fragments defined in `build-configs.yaml`, so proper testing within the test pipeline would be necessary.

============================================================

Split the build variants between the different kernel versions (5.10 and
5.15 for now). For each version define two variants: one with module
compression support and another without it.

### DIFF
--- a/config/core/build-configs-chromeos.yaml
+++ b/config/core/build-configs-chromeos.yaml
@@ -1,37 +1,11 @@
 trees:
-
   kernelci:
     url: "https://github.com/kernelci/linux.git"
 
+# Common blocks
+###############
 
-chromeos_variants: &chromeos-variants
-  chromeos-clang-14:
-    build_environment: clang-14
-    architectures: &variants_architectures
-      arm:
-        base_defconfig: 'cros://chromeos-5.10/armel/chromiumos-arm.flavour.config'
-        extra_configs:
-          - 'cros://chromeos-5.10/armel/chromiumos-rockchip.flavour.config'
-        filters: &cros-filters
-          - regex: { defconfig: 'cros' }
-      arm64:
-        base_defconfig: 'cros://chromeos-5.10/arm64/chromiumos-arm64.flavour.config'
-        fragments: [arm64-chromebook]
-        extra_configs:
-          - 'cros://chromeos-5.10/arm64/chromiumos-mediatek.flavour.config+arm64-chromebook'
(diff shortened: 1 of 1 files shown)


nuclearcat:
At moment this changes causing both compressed and non-compressed 5.10 kernels built.
https://chromeos.kernelci.org/test/job/kernelci/branch/chromeos-stable/kernel/chromeos-stable-20220812.0/plan/cros-tast/
Which doubles number of builds and tests.
I am not sure there is any point to do tests with compressed modules.

nuclearcat:
I think there is definitely something wrong, as it causes lot of more than necessary configs built than before even on 5.15.y:
https://chromeos.kernelci.org/job/stable/branch/linux-5.15.y/

nuclearcat:
Probably i found why, did "unmerge" by simple script:
```python
#!/usr/bin/env python3
import yaml

class NoAliasDumper(yaml.SafeDumper):
    def ignore_aliases(self, data):
        return True

with open("build-configs-chromeos.yaml") as f:
    y = yaml.safe_load(f)
with open(r'merged.yaml', 'w') as file:
    yaml.dump(y, file, Dumper=NoAliasDumper)
```
and compared old and new version and noticed, for example, in new version:
"chromeos-next - variants - chromeos-clang-14 " missing filters, because they are overwritten by merge operation.
This is similar issue i faced recently (and even argued about with one of linters authors), and only this small snippet helped to understand logic how python merge keys in yaml works and how final config looks after merging.

nuclearcat:
I am working on this PR (https://github.com/kernelci/kernelci-core/pull/1400), it seems problem is deeper than we tought. 5.15 kernels doesnt work compressed too.

nuclearcat:
I suggest i continue on my PR, but if you dont mind, i add your signed-off-by as i based my PR on your idea and draft.

PR #1368, opened 2022-08-13
Add Job class in python.jinja2

gctucker:
Add a `Job` class in `python.jinja2` to be able to override how jobs are run and how results are submitted.  Also print call stack when an exception is caught.

### DIFF
--- a/config/runtime/base/python.jinja2
+++ b/config/runtime/base/python.jinja2
@@ -28,53 +28,79 @@ TARBALL_URL = '{{ tarball_url }}'
 WORKSPACE = '/tmp/kci'
 {%- endblock %}
 
-{% block python_body %}{% endblock %}
+{% block python_base_job -%}
+class BaseJob:
+
+    def __init__(self, workspace):
+        self._workspace = workspace
+
+    def _get_db(self, db_config_yaml):
+        db_config = kernelci.config.db.DatabaseFactory.from_yaml(
+            'db', yaml.load(db_config_yaml, Loader=yaml.CLoader)
+        )
+        if not db_config:
+            return None
+
+        api_token = os.getenv('API_TOKEN')
+        if not api_token:
+            return None
+
+        return kernelci.db.get_db(db_config, api_token)
+
+    def _get_source(self, url):
+        resp = requests.get(url, stream=True)
+        resp.raise_for_status()
+        tarball_name = os.path.basename(urllib.parse.urlparse(url).path)
+        base, ext = tarball_name.split('.tar.')
+        with tarfile.open(fileobj=resp.raw, mode=f'r|{ext}') as tarball:
+            tarball.extractall(path=self._workspace)
+        return os.path.join(self._workspace, base)
+
+    def _run(self, src_path):
+        raise NotImplementedError("_run() method required to run job")
+
+    def _submit(self, results, node_id, db):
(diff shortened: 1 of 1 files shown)


config/runtime/base/python.jinja2
```diff
@@ -62,7 +63,8 @@ if __name__ == '__main__':
     try:
         src_path = _get_source(TARBALL_URL)
         res = main(sys.argv + [src_path])
-    except:
+    except Exception as e:
```
nuclearcat, line 92: Do we need "as e" if we are not using it?

config/runtime/base/python.jinja2
```diff
@@ -27,52 +28,79 @@ TARBALL_URL = '{{ tarball_url }}'
...
+        base, ext = tarball_name.split('.tar.')
+        with tarfile.open(fileobj=resp.raw, mode=f'r|{ext}') as tarball:
+            tarball.extractall(path=self._workspace)
+        return os.path.join(self._workspace, base)
 
-def _get_source(url):
-    resp = requests.get(url, stream=True)
-    resp.raise_for_status()
-    tarball_name = os.path.basename(urllib.parse.urlparse(url).path)
-    base, ext = tarball_name.split('.tar.')
```
mgalka, line 54: I'd use `os.splitext()` for getting file extenstion. It makes sure you'll get the part after the last dot. It handles the corner cases for you.

mgalka (commented):
small changes

config/runtime/base/python.jinja2
gctucker: Not really, I'll drop this.

config/runtime/base/python.jinja2
```diff
@@ -27,52 +28,79 @@ TARBALL_URL = '{{ tarball_url }}'
...
+
+    if NODE_ID and DB_CONFIG_YAML:
+        job.submit(results, NODE_ID, DB_CONFIG_YAML)
+
+    return results
+
+
+if __name__ == '__main__':
+    results = main(sys.argv)
+    sys.exit(1 if results is None else 1)
```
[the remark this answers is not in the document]
gctucker: Yes that's a typo

config/runtime/base/python.jinja2
```diff
@@ -27,52 +28,79 @@ TARBALL_URL = '{{ tarball_url }}'
...
+            return None
 
-    return kernelci.db.get_db(db_config, api_token)
+        return kernelci.db.get_db(db_config, api_token)
 
+    def _get_source(self, url):
+        resp = requests.get(url, stream=True)
+        resp.raise_for_status()
+        tarball_name = os.path.basename(urllib.parse.urlparse(url).path)
+        base, ext = tarball_name.split('.tar.')
```
gctucker, line 54: OK I'll fix that as a follow-up as it's not related to the refactoring with the `BaseJob` class.

gctucker: Actually it's not quite the same thing as I want the root of the name:
`file.tar.gz` -> `file`, `gz`

With `os.splitext` I get `file.tar`, `gz` instead.  So then I need to split `file.tar` again.

gctucker:
Merging this now to unblock work on `fstests`.  If we need to adjust anything in the code this can be done as a follow-up.

PR #1375, opened 2022-08-16
CHROMEOS rootfs-configs-chromeos.yaml: Add missing serial to qemu build

nuclearcat:
Serial is required for qemu too, as it is necessary to enable serial
console.
rootfs built without serial support will not pass LAVA tests, example:
https://lava.collabora.dev/scheduler/job/7046848


nuclearcat:
This commit will be added to R102 version bump PR

PR #1382, opened 2022-08-18
config/docker: add coccinelle and cvehound templates

gctucker:
Add a fragment/coccinelle.jinja2 template that can be reused by images
that need coccinelle to be installed.  Also add a coccinelle.jinja2
template to build an image with just coccinelle, and cvehound to build
an image with cvehound which includes the coccinelle fragment.


gctucker:
Built the images OK manually and on staging.

Issue #1372, opened 2022-08-15
ChromeOS: New SELinux watches failing due to missing policies


10ne1:
Starting with newer Linux kernels, SELinux has added some new watches [like this one for v5.3](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=5825a95fe92566ada2292a65de030850b5cff1da) which require additional policies. 

This causes failures [like this one](https://storage.chromeos.kernelci.org/kernelci/chromeos-stable/chromeos-stable-20220815.0/x86_64/cros---chromeos-5.10-x86_64-chromiumos-x86_64.flavour.config/clang-14/lab-collabora/cros-baseline-qemu_x86_64-uefi-chromeos.txt) because the CrOS image does not define the respective policies. The officially supported downstream kernel [does not have this problem](https://storage.chromeos.kernelci.org/kernelci/chromeos-stable/chromeos-stable-20220815.0/x86_64/cros---chromeos-5.10-x86_64-chromiumos-x86_64.flavour.config/clang-14/lab-collabora/cros-baseline-fixed-qemu_x86_64-uefi-chromeos.txt) because it's using v4.14.

v4.14:
```
[    2.194077] SELinux:  Class infiniband_pkey not defined in policy.
[    2.194077] SELinux:  Class infiniband_pkey not defined in policy.
[    2.195663] SELinux:  Class infiniband_endport not defined in policy.
[    2.195663] SELinux:  Class infiniband_endport not defined in policy.
[    2.197489] SELinux: the above unknown classes and permissions will be denied
[    2.197489] SELinux: the above unknown classes and permissions will be denied
```

v5.15:
```
[    5.617960] SELinux:  Permission watch in class filesystem not defined in policy.
[    5.628387] SELinux:  Permission watch_mount in class file not defined in policy.
[    5.638720] SELinux:  Permission watch_sb in class file not defined in policy.
[    5.648661] SELinux:  Permission watch_with_perm in class file not defined in policy.
[    5.659414] SELinux:  Permission watch_reads in class file not defined in policy.
[    5.669910] SELinux:  Permission watch_mount in class dir not defined in policy.
[    5.680626] SELinux:  Permission watch_sb in class dir not defined in policy.
[    5.690548] SELinux:  Permission watch_with_perm in class dir not defined in policy.
[    5.701486] SELinux:  Permission watch_reads in class dir not defined in policy.
[...]

PR #1387, opened 2022-08-22
CHROMEOS lava/cros-boot[-qemu]: do not enforce selinux policies

10ne1:
Latest Linux kernels have more selinux watches which do not
have policies defined in the CrOS images as documented in
https://github.com/kernelci/kernelci-core/issues/1372

CrOS images do not offically support the latest kernels
tested by KernelCI, some even ship kernels as old as 4.14,
so instead of downstream hacking the CrOS images to define
the missing policies, we put selinux in non-enforcing mode.

Another factor for this decision is that selinux is used
for securing the Android Runtime for Chrome which is not a
testing target for KernelCI.

### DIFF
--- a/config/lava/chromeos/cros-boot-qemu.jinja2
+++ b/config/lava/chromeos/cros-boot-qemu.jinja2
@@ -49,7 +49,7 @@ actions:
         image_arg: -drive format=raw,file={rootfs},id=lavatest2,if=none
         url: downloads://chromiumos_test_image.bin
       kernel:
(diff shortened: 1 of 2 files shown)


nuclearcat:
I will add skip-staging label, so it wont break staging testing, if you dont mind. Feel free to remove it as soon as PR ready to be tested on staging

nuclearcat:
Can you rebase please? We have newest R106 build without selinux, so we can give it a try (if it is still required if image built without selinux)

nuclearcat:
Let's test it now :)

nuclearcat:
This PR is still WIP, we are testing it on new ChromiumOS build.

nuclearcat:
As update, ChromeOS fails to boot with this PR:
```

<6>[    5.602722] Run /sbin/init as init process
init: Unable to mount /sys/fs/selinux filesystem: No such file or directory
```
Maybe we can drop PR for now and reopen when we have solution?

PR #1388, opened 2022-08-22
CHROMEOS build_board.sh: disable selinux use flags

10ne1:
The CrOS userpace does not support the latest kernels
used by KernelCI in the sense that they miss selinux
policies for newer watches which fail as described in
https://github.com/kernelci/kernelci-core/issues/1372

We can't just disable selinux at boot via the kernel
cmdline selinux=0 because that will cause packages like
the upstart init system to fail, so we need to actually
disable them during image builds.


nuclearcat:
Temporary adding skip-staging to merge other PR

nuclearcat:
Flashed image built with this patch: https://lava.collabora.dev/scheduler/job/7539862
Tast tested: https://lava.collabora.dev/scheduler/job/7540084 it works as intended

nuclearcat (approved):
LGTM

PR #1389, opened 2022-08-23
CHROMEOS config/lava/chromeos: Add more tast tests (kernel + platform)

mgalka:
Extend cros-tast test plan with some kernel and platform tast tests.


mgalka:
Changes included in #1397

PR #1393, opened 2022-08-23
test-configs: hifive-unleashed: fix blocklist

khilman:
Having "v5.1" in the kernel blockist actually blocks v5.1*, which
explains why this board has not been booting on v5.10+ either. :(

Since we have not been build/boot testing v5.1 for a long time, just
removing this is an OK fix, but this is just a workaround for deeper
problem when kernel minor versions have two digits.  For example, the
v5.2 block would have also  blocked v5.20, but fortunately (for us),
the kernels past v5.19 are v6.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -896,7 +896,7 @@ device_types:
     # no console in tinyconfig
       - blocklist:
           defconfig: ['tinyconfig']
-          kernel: ['v4.', 'v5.1', 'v5.2']
+          kernel: ['v4.']
 
   hip07-d05:
     mach: hisilicon


config/core/test-configs.yaml
```diff
@@ -896,7 +896,7 @@ device_types:
     # no console in tinyconfig
       - blocklist:
           defconfig: ['tinyconfig']
-          kernel: ['v4.', 'v5.1', 'v5.2']
+          kernel: ['v4.', 'v5.2']
```
gctucker, line 899: I think you can just drop the `v5.1` or `v5.2` as the LTS versions are v4.9 which is already blocked, and then v5.4.  So there just won't be kernels between v5.1 and v5.4 tested on kernelci.org.

Also if the filter had to be really for blocking anything up to v5.1 it could be changed to a `regex` instead of a `blocklist`.

khilman:
@gctucker patch updated to drop both v5.1 and v5.2.

nuclearcat (approved):
LGTM

gctucker:
Note: To only block say v5.1 kernels we could use `v5.1.` in a blocklist filter.  That's probably a safer way to specify versions to block everywhere in the filters anyway.  We already have some `v3.` filters following the same logic.

The only issue is with mainline versions, which will last until they move on and we then only have stable versions with 3 numbers.  But maybe it's just a non-issue, we can let the mainline tests fail until some fixes or missing patches get merged.

PR #1394, opened 2022-08-24
CHROMEOS ssh_retry.sh: Add ssh retry mechanism

nuclearcat:
As discovered during lava tests, network stack of ChromeOS might
be pingable, but ssh daemon is not up yet (Connection refused).
As ssh don't have internal retry mechanism for such situation,
we add simple helper script for that.
This script will retry 10 times with delay, until it is able to connect
to DUT over ssh.

Example of failure
```
+ ping -c 1 -w 1 192.168.201.13
PING 192.168.201.13 (192.168.201.13): 56 data bytes
64 bytes from 192.168.201.13: icmp_seq=0 ttl=63 time=0.529 ms
--- 192.168.201.13 ping statistics ---
1 packets transmitted, 1 packets received, 0% packet loss
round-trip min/avg/max/stddev = 0.529/0.529/0.529/0.000 ms
+ lava-target-ip
+ lava-test-case os-release --shell ssh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i /home/cros-tast/.ssh/id_rsa root@192.168.201.13 cat /etc/os-release
ssh: connect to host 192.168.201.13 port 22: Connection refused
```

### DIFF
--- a/config/docker/cros-tast/ssh_retry.sh
+++ b/config/docker/cros-tast/ssh_retry.sh
@@ -1,0 +1,24 @@
+#!/usr/bin/env bash
+let tries=10
+
+if [ -z "$*" ] ; then
+    echo "Need to specify ssh options"
+    exit 1
+fi
+
+while [[ $tries -ne 0 ]] ; do
+    ssh $*
+    retcode=$?
+    if [[ $retcode -eq 0 ]] ; then
+        break
+    else
+        echo "Failed, retrying..."
+        sleep 1
+    fi
+    let tries--
+done
+
+if [[ $retcode -ne 0 ]] ; then
+    echo "Connection to $* failed, giving up"
+    exit 1
+fi
--- a/config/docker/cros-tast/Dockerfile
+++ b/config/docker/cros-tast/Dockerfile
@@ -43,6 +43,7 @@ RUN cp "$HOME/trunk/chromite/ssh_keys/testing_rsa" "$HOME/.ssh/id_rsa"
 RUN chmod 0600 "$HOME/.ssh/id_rsa"
 
 ADD tast_parser.py /home/cros-tast/tast_parser.py
+ADD ssh_retry.sh /home/cros-tast/ssh_retry.sh
 
 # Needed by LAVA to install the overlay
 USER root
--- a/config/lava/chromeos/cros-tast.jinja2
+++ b/config/lava/chromeos/cros-tast.jinja2
@@ -27,7 +27,7 @@
             - while ! ping -c 1 -w 1 $(lava-target-ip); do sleep 1; done
             - >-
               lava-test-case os-release --shell
-              ssh
+              ./ssh_retry.sh
               -o StrictHostKeyChecking=no
               -o UserKnownHostsFile=/dev/null
               -i /home/cros-tast/.ssh/id_rsa


nuclearcat:
This is link to particular failed test case: https://chromeos.kernelci.org/test/case/id/6305905e3305bf060830d7c4/

gctucker:
Great, yes we knew that was a potential problem since the beginning so it's good to address it now.

The issue was that the "exponential backoff" time in the SSH client would cause it to wait for potentially a very long time if the SSH server wasn't up, I guess that's still the case?  Presumably we're just preempting things up a bit with the ping step first.

nuclearcat:
Exponential backoff case (for TCP) we handled by ping, it happens when network stack is slow to come up, it is sorted it out.

Now we have new issue, when ChromiumOS bring up interface and ip, so ping is up, but it didnt run ssh daemon yet.
When ssh tries to connect to ChromiumOS, port is closed, it get connection refused, so it just return error and there is no meaningful retry mechanisms for that, except such wrappers.

gctucker:
Yes I get the part about the SSH server only being up later than when ping returns a reply, that's the issue we've known since the beginning but wasn't addressed until now.

> Exponential backoff case (for TCP) we handled by ping

As far as I know, ping doesn't do exponential backoff, so are you just confirming what I wrote in my previous comment?

nuclearcat:
Sorry, i think i understand now. 
Yes, i confirm, we are preempting things, but even if we remove ping and rely on ssh client exponential backoff, still we have high chance to get false positive (connection refused).

nuclearcat:
Updating, return code in case of failure

config/docker/cros-tast/ssh_retry.sh
```diff
@@ -0,0 +1,24 @@
...
+if [ -z "$*" ] ; then
+    echo "Need to specify ssh options"
+    exit 1
+fi
+
+while [[ $tries -ne 0 ]] ; do
+    ssh $*
+    retcode=$?
+    if [[ $retcode -eq 0 ]] ; then
+        let tries=1
```
mgalka, line 13: I'd consider using `break` here. It looks more clear to me.

mgalka (commented):
Small thing to consider

config/docker/cros-tast/ssh_retry.sh
nuclearcat: Done! Thanks, no idea why i did like that, probably forgot that bash has break too

mgalka:
Script works fine on staging
https://staging.lava.collabora.dev/scheduler/job/213718#L5640

```
+ lava-test-case os-release --shell ./ssh_retry.sh -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i /home/cros-tast/.ssh/id_rsa root@192.168.201.11 cat /etc/os-release
<LAVA_SIGNAL_STARTTC os-release>
Received signal: <STARTTC> os-release
Listened to connection for namespace 'modules' done
Listened to connection for namespace 'chromeos' done
Warning: Permanently added '192.168.201.11' (ED25519) to the list of known hosts.
NAME=Chromium OS
ID=chromiumos
HOME_URL=https://www.chromium.org/chromium-os
BUG_REPORT_URL=https://crbug.com/new
VERSION=102
VERSION_ID=102
BUILD_ID=14695.119.0-d2022_08_11_053039
<LAVA_SIGNAL_ENDTC os-release>
<LAVA_SIGNAL_TESTCASE TEST_CASE_ID=os-release RESULT=pass>
```

mgalka (approved):

PR #1395, opened 2022-08-24
test.py: Remove dtb from job name and truncate below 200 chars

nuclearcat:
LAVA doesn't accept job names with more than 200 characters, drop dtb
from job name to reduce size and truncate job name below 200 characters


gctucker:
I've pushed this commit to the `chromeos` branch so it'll get tested with the Chrome OS jobs as well.

nuclearcat:
As far as i can see we are not facing issues with job names anymore.
@mgalka if no objections, is it possible to merge this PR?

mgalka (approved):
lgtm

PR #1398, opened 2022-08-25
Revert "rootfs-configs.yaml: Remove sid rootfs image"

khilman:
This reverts commit 496880b32f07e96547869f5fd9604a942ff6283f.

Debian support for riscv64 is not available in bullseye, but is in
sid/debian-ports.

### DIFF
--- a/config/core/rootfs-configs.yaml
+++ b/config/core/rootfs-configs.yaml
@@ -253,3 +253,24 @@ rootfs_configs:
       - e2fsprogs
     script: "scripts/bullseye-v4l2.sh"
     test_overlay: "overlays/v4l2"
+
+  sid:
+    rootfs_type: debos
+    debian_release: sid
+    debian_mirror: http://deb.debian.org/debian-ports
+    keyring_package: debian-ports-archive-keyring
+    keyring_file: /usr/share/keyrings/debian-ports-archive-keyring.gpg
+    arch_list:
+      - riscv64
+    extra_packages_remove: &extra_packages_remove_sid
+      - bash
+      - e2fslibs
+      - e2fsprogs
+      - klibc-utils
+      - libext2fs2
+      - libgnutls30
+      - libklibc
+      - libncursesw6
+      - libp11-kit0
+      - sensible-utils
+    extra_files_remove: *extra_files_remove_bullseye


gctucker:
Out of interest, what are you planning to test with riscv64 using Debian?

Also, building with sid is very unstable so I'm not sure we can actually use it at all.  That's why it was removed.

khilman:
The goal is to run kselftest, LTP etc. on riscv, so for that we need debian rootfs images just like the other arches.

gctucker:
Currently the rootfs images are rebuilt once a week in production.  With sid, this is likely to break a good portion of the time, making things more complicated during the production update.  We would need to treat it as a special case and only upgrade it manually with "lucky" builds that fall on a good day etc.  So please take this into consideration, it would be good to suggest a way to manage that without causing any disruption.

khilman:
I've seen breakage on sid during the debootstrap, but if it builds it mostly boots.  So as long as the pointers to rootfs images are not updated unless the rootfs build fails, that's likely to be good enough for now.

In parallel, I'm also exploring how to do the cross-`debootstrap` equivalent for Fedora[1].  Since Fedora has official support for riscv64, this might be the more stable path going forward.  But this will require adding some docker images with a Fedora environment to do the debos equivalent, so that's a more involved task.

[1] https://dvdhrm.github.io/2018/01/09/cross-bootstrap-fedora/

gctucker:
Right now the production update starts a rootfs build for all the images, and if they all succeed it creates a commit automatically (the vast majority of the time).  If one fails, then it needs to be done by hand.  So adding `sid` will have a maintenance cost.

Adding Fedora builds means adding support for it as another distro as you explained, and so far there hasn't been a need for that as everything could be done with buildroot and debian.

Is there a place where the missing bits for RISC-V in Debian are tracked, at least to have it working in Testing?

khilman:
From what I understand from debian devs, what's blocking riscv from official support is the lack of hardware to pass all the full debian tests.  

Re: Fedora, it appears I spoke to soon.  It's not actually officially supported in Fedora either, it's requires using a special "rawhide" version, which likely has the same issues as using debian/unstable. :(

khilman:
> 
> Would you be OK with doing that?

Unfortunately, I'm not very familiar enough with the kci_rootfs/kci-deploy process for me to add this feature easily, and don't have the time to ramp up on that code right now.

gctucker:
OK, then unless someone has the skills and time to address this issue I'm afraid we won't be able to enable Debian sid builds again.

gctucker:
@mgalka Please follow-up on this now that we should be able to deal with failing rootfs builds in production.

khilman:
This was folded into a larger series working improving riscv support: https://github.com/kernelci/kernelci-core/pull/1460

PR #1399, opened 2022-08-26
CHROMEOS test-configs-chromeos.yaml: Simplify device_type configuration

nuclearcat:
After adding several device types we can optimize config, to make
configuration less complex and reduce probability of mistakes while
adding new devices in future.

### DIFF
--- a/config/core/test-configs-chromeos.yaml
+++ b/config/core/test-configs-chromeos.yaml
@@ -119,14 +119,14 @@ test_plans:
 
 device_types:
 
-  asus-C436FA-Flip-hatch_chromeos:
+  asus-C436FA-Flip-hatch_chromeos: &chromebook-generic-type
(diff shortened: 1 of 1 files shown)


gctucker:
Let's not forget to add Chrome OS related PRs to the Chrome OS work board.  It can't really be done automatically, I've tried some GitHub apps but couldn't find one that worked well enough.  Maybe we could fix / improve / implement one in the longer term to save some effort, or have a bot that periodically goes through all the PRs via the API (like the `pending.py` script) and sets the project field automatically based on some criteria (project name, target branch, area of the code being updated...).

nuclearcat:
Tested at: https://chromeos.kernelci.org/test/job/kernelci/branch/chromeos-stable/kernel/chromeos-stable-20220902.0/
Working fine for several last days

mgalka (approved):

PR #1400, opened 2022-08-26
CHROMEOS build-configs-chromeos.yaml: Use correct fragment and set default builds to 5.15

nuclearcat:
For kernels 5.15 we need to use correct cros:// config fragments, also it is preferable to
test more(daily) ChromiumOS rootfs with kernel 5.15.


nuclearcat:
Builds are successfully tested and used for about a week: https://chromeos.kernelci.org/build/kernelci/branch/chromeos-stable/kernel/chromeos-stable-20220902.0/

mgalka (approved):
LGTM

PR #1411, opened 2022-09-06
kernelci_api: return only 'items' from 'get_nodes'

JenySadadia:
Instead of returning entire paginated response
received from API, return only list of node
objects by fetching value of 'items' key.

### DIFF
--- a/kernelci/db/kernelci_api.py
+++ b/kernelci/db/kernelci_api.py
@@ -91,7 +91,7 @@ class KernelCI_API(Database):
     def get_nodes(self, attributes: dict = None):
         """Get all nodes matching attributes"""
         resp = self._get('nodes', params=attributes)
-        return resp.json()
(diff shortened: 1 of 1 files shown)


kernelci/db/kernelci_api.py
```diff
@@ -91,7 +91,7 @@ def get_node(self, node_id):
     def get_nodes(self, attributes: dict = None):
         """Get all nodes matching attributes"""
         resp = self._get('nodes', params=attributes)
-        return resp.json()
+        return (resp.json()).get('items')
```
gctucker, line 94:
```suggestion
        return resp.json().get('items')
```
In fact, could we just use `['items']` instead?  The API has the return type set to `PageModel` so it should be something we can rely on.

gctucker (approved):
Tested OK locally

PR #1417, opened 2022-09-08
Drop old dockerfiles except cros-* and build-base

gctucker:
Drop all the old Dockerfiles except the `cros-*` ones which aren't yet covered by `kci_docker` and `build-base` which is still used by some Jenkins jobs.


gctucker:
@mgalka This was tested OK on staging, I haven't seen any issues as side-effects of this PR.

mgalka (approved):
LGTM

Issue #1412, opened 2022-09-06
Add pagination arguments to KernelCI_API.get_nodes()


gctucker:
Now that we're adding support for pagination, we need to add the `limit` and `offset` arguments to the `KernelCI_API.get_nodes()` method so the client code can manage pagination.  When these arguments are not provided, we can implement some logic to retrieve all the nodes (with some default pagination value) and return a flat list as a convenience to keep the client code simple when pagination is not required.

PR #1420, opened 2022-09-09
T37626 kernelci_api: add pagination params and use it in `get_nodes`

JenySadadia:
Fixes https://github.com/kernelci/kernelci-core/issues/1412

Need to add parameters `_limit` and `_offset` to `KernelCI_API` class specific to pagination.
Use the default values of params to return nodes when client doesn't provide them externally. Otherwise use the values provided by the client to return nodes matching attributes.

### DIFF
--- a/kernelci/db/kernelci_api.py
+++ b/kernelci/db/kernelci_api.py
@@ -26,7 +26,7 @@ from kernelci.db import Database
 
 class KernelCI_API(Database):
 
-    def __init__(self, config, token):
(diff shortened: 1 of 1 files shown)


gctucker:
There were several problems with this PR and it caused the pipeline to fail to start due to an infinite loop so I made #1421 to unblock the situation.

gctucker:
Merging together with #1421 now that the API changes for pagination have been merged.

gctucker (approved):

PR #1421, opened 2022-09-09
get_nodes() pagination fixup

gctucker:
This fixes some issues with #1420

### DIFF
--- a/kernelci/db/kernelci_api.py
+++ b/kernelci/db/kernelci_api.py
@@ -26,7 +26,7 @@ from kernelci.db import Database
 
 class KernelCI_API(Database):
 
-    def __init__(self, config, token):
+    def __init__(self, config, token, limit=100, offset=0):
         super().__init__(config, token)
         if self._token is None:
             raise ValueError("API token required for kernelci_api")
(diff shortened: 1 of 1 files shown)


kernelci/db/kernelci_api.py
```diff
@@ -88,10 +88,32 @@ def get_node(self, node_id):
...
+                  offset: int = None, limit: int = None):
         """Get all nodes matching attributes"""
-        resp = self._get('nodes', params=attributes)
-        return resp.json()['items']
+        params = attributes.copy() if attributes else {}
+
+        if any((offset, limit)):
+            params.update({
+                'offset': offset or 0,
+                'limit': limit or 0,
```
JenySadadia, line 99: The limit must be greater than 0.

gctucker: Well I think the parameters should actually only be added to the data if they're not `None`.  I'll rework that.

gctucker: OK I've changed it to `or None` as that means the parameter won't be in the data if it's not provided, and the API will be using its own default value.

PR #1424, opened 2022-09-09
config/k8s: drop kernelci-core git clone

gctucker:
Drop the git clone of kernelci-core as the Docker image already has it installed system-wide with the configuration files in /etc/kernelci.


gctucker:
Tested OK on staging: https://bot.staging.kernelci.org/job/kernel-build/81421/consoleFull

PR #1425, opened 2022-09-09
Support for using kernelci-core tools within other CI pipelines

eds-collabora:
This provides 4 changes that assist in building CI pipelines around kernelci tools:

- Add a flag to kernelci to be able to insert authorisation headers into the requests made from LAVA jobs. It's used like: `--storage-auth=TOKEN=example-token-value`
- Change uses of urljoin to preserve any query parameters present in the storage URL. This means that the additional parts of the path are added in the middle (before the query parameters). For example, if our storage URL was `https://storage.example.com/?token=example-token` then this allows our kernel URL to end up being `https://storage.example.com/next/5.10/blah/arm64/defconfig/gcc-10/Image?token=example-token`.
- Honor `--install-path` within `kci_build`. This means that you can pick where the install directory is made, rather than having it always be `_install_` within the kernel build. This makes it easier to control the URLs that are generated, e.g. in gitlab, where you can't remove leading parts of the path from the artifact paths.
- Permit the publish path to be omitted from the generate storage URLs. Using `--omit-publish-path`, the LAVA jobs will be generated containing URLs like `https://storage.example.com/kernel/Image` rather than `https://storage.example.com/next/5.10/blah/arm64/defconfig/gcc-10/kernel/Image`

I can split this PR if that would be more convenient, I wasn't sure which was preferable.

### DIFF
--- a/config/lava/base/kernel-ci-base-tftp-deploy.jinja2
+++ b/config/lava/base/kernel-ci-base-tftp-deploy.jinja2
@@ -22,6 +22,12 @@ actions:
     to: tftp
     kernel:
       url: {{ kernel_url }}
+{%- if storage_headers %}
+      headers:
+{%- for header, value in storage_headers.items() %}
+        {{ header }}: {{ value }}
+{%- endfor %}
+{%- endif %}
 {%- block kernel_image_type %}
       type: {{ kernel_image.lower() }}
 {%- endblock %}
@@ -42,6 +48,12 @@ actions:
 {%- if modules_url %}
     modules:
       url: {{ modules_url }}
+{%- if storage_headers %}
+      headers:
+{%- for header, value in storage_headers.items() %}
+        {{ header }}: {{ value }}
+{%- endfor %}
+{%- endif %}
       {%- if modules_compression %}
       compression: {{ modules_compression }}
       {%- endif %}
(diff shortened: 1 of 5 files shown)


gctucker:
This is causing some issues on the staging.kernelci.org builds as the `install_path` variable gets used as `None` when it shouldn't.

Adding the `staging-skip` label to stop deploying it on staging until that's fixed.

Someone should be doing an actual review of the code soon to highlight the problem.

eds-collabora:
> This is causing some issues on the staging.kernelci.org builds as the `install_path` variable gets used as `None` when it shouldn't.

I believe the issue was a typo in `kci_data`, which my pipeline doesn't use. At the very least, there was such a typo, which I've fixed.

nuclearcat (commented):
I think when you make job definition like this, secrets are not really secrets and will be visible in job definition and probably logs.
Maybe it is better to use secrets section in LAVA?
Publishing artifacts[¶](https://docs.lavasoftware.org/lava/publishing-artifacts.html#publishing-artifacts)

gctucker:
Adding `staging-skip` to avoid conflict with #1573 on staging

gctucker:
@eds-collabora Yes I think it would be good to split this PR into smaller ones.  I don't think we'll be able to get this merged without a lot of rework which would take a long time if this can be done at all.  There is now `kernelci.storage` which can be used as an abstraction layer for uploading artifacts to storage, maybe that would work for you?

mgalka:
@eds-collabora Any thoughts on @gctucker 's suggestion?

gctucker:
OK let's wait a couple more weeks or so before closing this one.  I think it will most likely have to be recreated as a series of smaller PRs anyway.

nuclearcat:
I will close this PR as legacy tools are being deprecated.

PR #1431, opened 2022-09-15
docs: Update docker image names

nuclearcat:
As we have new docker image building system in place, and naming changed as well, update documentation for correct docker image names.


gctucker:
Looks OK on staging: https://static.staging.kernelci.org/docs/core/kci_build/

PR #1434, opened 2022-09-15
Enable more kselftests for Chromebooks

hardboprobot:
Huge amount of commits to enable kselftests for all current KernelCI Chromebooks. Each commit covers one kselftest collection (suite). [This is a list](https://gitlab.collabora.com/rcn/rcn/-/blob/master/tasks/KernelCI/T37646__check_missing_kselftests/doc.org) of the current status of each test (in a very sketchy status). The initial idea is to enable as many tests as possible for as many different Chromebook types as possible. Then, once they're running we can check the results and find out possible problems with individual board types and fix them or disable them selectively.

Up to @nuclearcat , @mgalka and @gctucker to decide which ones to integrate and which ones to keep off.

### DIFF
--- a/config/core/rootfs-configs-chromeos.yaml
+++ b/config/core/rootfs-configs-chromeos.yaml
@@ -22,6 +22,14 @@ rootfs_configs:
     arch_list:
       - amd64
 
+  chromiumos-coral:
+    rootfs_type: chromiumos
+    board: coral
+    branch: release-R100-14526.B
+    serial: ttyS2
+    arch_list:
+      - amd64
+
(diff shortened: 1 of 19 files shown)


gctucker:
Please change the target branch to `chromeos`.

nuclearcat:
I believe we need to go one by one with kselftests, considering PR backlog and capacity of reviewers.

nuclearcat:
I think we need to close this PR, as it is almost impossible to get success on all tests, and we need to split and  open several PR for small "batches" of tests.

PR #1435, opened 2022-09-15
kci_test update_rootfs_urls

gctucker:
Add a command to automatically generate the YAML configuration with the rootfs images URLs when they exist and leave them to their previous URLs when getting HTTP 404.  For example:
```
kci_test update_rootfs_url --release=20220913.0 --output=config/core/rootfs-image.yaml
```

### DIFF
--- a/kci_test
+++ b/kci_test
@@ -21,8 +21,10 @@ import argparse
 import glob
 import json
 import os
+import re
 import requests
 import sys
+import yaml
 
 from kernelci.cli import Args, Command, parse_opts
 import kernelci
@@ -31,6 +33,8 @@ import kernelci.config.test
 import kernelci.build
 import kernelci.lab
 import kernelci.test
+
+RELEASE_RE = re.compile(r'(.*)([0-9.]{10})(.*)')
 
 
 # -----------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


kci_test
```diff
@@ -73,42 +77,114 @@ class cmd_validate(Command):
...
+        yield fs, config, name, root_type
+
+
+def _check_rootfs_urls(fs_config, rootfs_config, root_type, verbose):
+    url_formats = list(
+        url_fmt for url_fmt in (
+            fs_config.get_url_format(fmt_type)
+            for fmt_type in set({root_type}).union({'nfs', 'ramdisk'})
+        ) if url_fmt
+    )
```
mgalka, line 100: I think we can just use a normal list comprehension here

```suggestion
    url_formats = [
        url_fmt for url_fmt in (
            fs_config.get_url_format(fmt_type)
            for fmt_type in set({root_type}).union({'nfs', 'ramdisk'})
        ) if url_fmt
    ]
```

The internal one can be also converted to list comprehension or you can leave it as a generator expression as it is now. Having said that there's not much gain from using genexp there.

mgalka (commented):
one small comment

kci_test
gctucker: Yes I think that's some leftover from previous experiments, I'll change it to `[]`

gctucker:
@aliceinwire FYI this is moving all the CIP test configuration to a separate `test-configs-cip.yaml` file as the CIP rootfs config can't be automatically generated like the regular buildroot and debos ones which are built by KernelCI.  It might cause some trivial conflicts with other PRs, I hope that's OK.

mgalka (approved):

PR #1436, opened 2022-09-16
Drop Docker build-base image

gctucker:
Drop the `build-base` Dockerfile as it's not used any more.  Rename `docker-new` to `docker` and the old `docker` to `docker-old`.


nuclearcat:
I guess we need also updates in kernelci-deploy?

gctucker:
> I guess we need also updates in kernelci-deploy?

Just pushed that.

gctucker:
@mgalka It would be good to have this merged before the prod update.

mgalka (approved):

PR #1442, opened 2022-09-20
config/docker: add kci_docker --no-cache argument

### LINKED ISSUE #1441: Add --no-cache to kci_docker build command
At the moment `kci_docker build` doesn't provide a way to not use cache during image builds. It should be possible to add similar functionality to the one offered by `docker image build --no-cache`.

Docker SDK for Python [`client.images.build()`](https://docker-py.readthedocs.io/en/stable/images.html#docker.models.images.ImageCollection.build) function can take a `nocache` parameter which disables cache usage for current build.
`--no-cache` command line parameter could be added to `kci_docker`. Its presence should set the `nocache` parameter to `True`.

gctucker:
Add --no-cache argument to kci_docker to disable the Docker cache when building images.

Fixes #1441

### DIFF
--- a/config/docker/kci_docker
+++ b/config/docker/kci_docker
@@ -32,11 +32,11 @@ def _gen_dockerfile(img_name, params):
     return template.render(params)
 
 
-def _build_image(dockerfile, tag_name, buildargs):
+def _build_image(dockerfile, tag, buildargs, nocache):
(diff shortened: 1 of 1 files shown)


nuclearcat (approved):
Tested locally, LGTM

PR #1443, opened 2022-09-20
CHROMEOS docker/data: Copy CrOS related files for kci_docker

nuclearcat:
Copy CrOS docker-related files so it can be used with kci_docker.
Similar to gcc-10-x86.jinja2 and reference to gcc-header-fix.patch,
we need to have files in place beforehand to reference them in Dockerfiles.

### DIFF
--- a/config/docker/data/add_modules.sh
+++ b/config/docker/data/add_modules.sh
@@ -1,0 +1,11 @@
+#!/bin/sh
+# In case you experience any problems with guestfish,
+# you can enable extended debug:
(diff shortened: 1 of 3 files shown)


mgalka:
@nuclearcat  I think it would be beneficial to get a bit more elaborate description why we need to place them in this location. Mentioning that there are links to the github content necessary during Dockerfile generation, hence these files need to be there beforehand.

nuclearcat:
@mgalka done, updated

PR #1444, opened 2022-09-20
config/docker: split lines in debos.jinja2

gctucker:
Split the "go get" and "go install" commands into two lines as it's not useful to have them on the same line and now we can better tell when something fails what is causing it.


nuclearcat (approved):
LGTM

PR #1448, opened 2022-09-21
riscv/clang build updates

khilman:
Update riscv/clang builds to work with recent clang for mainline & linux-next, removing some workarounds put in place for older kernels/toolchains and also enabling builds for all upstream riscv defconfigs.

### DIFF
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -380,8 +380,6 @@ build_environments:
       riscv:
         <<: *riscv_params
         name:
-        opts:
-          LLVM_IAS: '1'
 
   clang-16:
     cc: clang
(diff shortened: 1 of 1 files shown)


khilman:
FYI @atishp04

khilman:
@mgalka ping?

mgalka:
> Working successfuly on staging. All upstream defconfigs are now being built for riscv for both gcc and clang.

At the moment I can only see gcc-10 builds for riscv. I think I need to fiddle with staging builds config to confirm it works fine.

khilman:
> 
> At the moment I can only see gcc-10 builds for riscv. I think I need to fiddle with staging builds config to confirm it works fine.

See: https://staging.kernelci.org/build/next/branch/master/kernel/next-20221104/

TBH, I'm not sure what the difference between `next` and `staging-next` is, but in this one you can see multiple clang builds for riscv defconfigs.

nuclearcat (commented):
defconfig built ok with clang-16: https://staging.kernelci.org/build/id/6379c5e88881e00700ddf76c/
Is it expected result or anything else expected?

PR #1455, opened 2022-09-28
config/runtime: enable 'src_dir' debug option in python.jinja2

hardboprobot:
Allow building kernels that are already extracted to a local dir instead of downloading a new tarball every time.

The default behavior if this debug option is not enabled is still the same.

Related: ~https://github.com/kernelci/kernelci-pipeline/pull/155~ https://github.com/kernelci/kernelci-pipeline/pull/164


nuclearcat:
I guess it is tested on staging and not introducing any issues with existing pipelines?

PR #1456, opened 2022-09-29
CHROMEOS config/lava/cros*: Update home path in tast tests

nuclearcat:
As in new Docker images we are using unified username for CrOS, we need to update some of LAVA templates that are using old username 'cros-tast' to new - 'cros'.

### DIFF
--- a/config/lava/chromeos/cros-tast.jinja2
+++ b/config/lava/chromeos/cros-tast.jinja2
@@ -10,7 +10,7 @@
       wait:
         device: true
     results:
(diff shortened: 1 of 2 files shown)


nuclearcat:
ChromeOS patches are being used successfully in staging/chromeos according to new commands in kernelci-deploy

nuclearcat:
For example: https://lava.collabora.dev/scheduler/job/7488498
Using new /home/cros path

mgalka (approved):

PR #1467, opened 2022-10-11
Fix the build of some kselftest testsuites for riscv64

nkbelin:
This PR fixes the build of some kselftest testsuites for riscv64 platforms by using the cross compiled libc6 package.
Also install clang in order to build kselftest testsuites that require it.
Note that some kselftest testsuites are still failing to build due to the fact that several libs that are needed are not available yet for riscv64.

### DIFF
--- a/config/docker/fragment/kselftest.jinja2
+++ b/config/docker/fragment/kselftest.jinja2
@@ -1,10 +1,16 @@
 # kselftest
 {%- if sub_arch %}
-{%- set pkgarch = ':' + sub_arch %}
-RUN dpkg --add-architecture {{ sub_arch }}
+    {%- if sub_arch == 'riscv64' %}
+        {%- set libc6_pkgarch = '-' + sub_arch + '-cross' %}
+        {%- set pkgarch = '' %}
+    {% else %}
+        {%- set libc6_pkgarch = ':' + sub_arch %}
+        {%- set pkgarch = libc6_pkgarch %}
+        RUN dpkg --add-architecture {{ sub_arch }}
+    {%- endif %}
 {%- endif %}
 RUN apt-get update && apt-get install --no-install-recommends -y \
-   libc6-dev{{ pkgarch }} \
+   libc6-dev{{ libc6_pkgarch }} \
    libcap-dev{{ pkgarch }} \
    libcap-ng-dev{{ pkgarch }} \
    libelf-dev{{ pkgarch }} \
--- a/config/docker/gcc-10-riscv64.jinja2
+++ b/config/docker/gcc-10-riscv64.jinja2
@@ -1,3 +1,4 @@
+{%- set sub_arch = 'riscv64' %}
 {% extends 'base/host-tools.jinja2' %}
 
 {% block packages %}


broonie:
The changelog/commit descriptions could do with some description of why there's something special about RISC-V here - we're already installing development packages for the target architecture...

config/docker/fragment/kselftest.jinja2
```diff
@@ -22,4 +22,7 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
    libasound2-dev{{ pkgarch }} \
    libasound2-dev \
    patch \
-   pkg-config
+   pkg-config \
+   clang \
```
broonie, line 26: This seems to be rather counter to the idea of having the GCC container - it should be fixed at the build-config.yaml level where we decide which toolchains we're going to build kselftest with, if it just can't be fixed to work with GCC then we should probably just disable kselftest for RISC-V GCC builds

config/docker/fragment/kselftest.jinja2
```diff
@@ -1,10 +1,16 @@
 # kselftest
 {%- if sub_arch %}
-{%- set pkgarch = ':' + sub_arch %}
-RUN dpkg --add-architecture {{ sub_arch }}
+    {%- if sub_arch == 'riscv64' %}
+        {%- set libc6_pkgarch = '-' + sub_arch + '-cross' %}
+        {%- set pkgarch = '' %}
+    {% else %}
+        {%- set libc6_pkgarch = ':' + sub_arch %}
+        {%- set pkgarch = ':' + sub_arch %}
```
khilman, line 8: nit: this would be a little clearer to read if you just `set pkgarch = libc6_pkgarch`

config/docker/fragment/kselftest.jinja2
khilman: note this isn't riscv specific.  clang is required to build this test for all arches.

broonie: We should probably be skipping the test if we don't have clang then (really, fixing upstream to do so) shouldn't we?

nkbelin: I agree with the fact having clang in a GCC container seems wrong. But I was thinking that it would be nice to have most of the kselftest testsuites built ok in a single package (BTW this issue is not RISCV specific: it happens for the other arches).
On the other side, I tried building kselftest with a "clang" docker and I can see there are a lot of failures. As a consequence, I don't think we can consider using clang (only) to build kselftest for now.

broonie: Yes, clang has a lot of trouble with kselfest at the minute - I only just fixed it a release or two ago to actually try to build with clang when asked to but there's a whole bunch more work needed.

khilman: I think the (short-term) solution here is to just leave out clang and accept that this test will not be run in KernelCI for any arches.  That's what is happening with the other arches.

@broonie your other idea of only building kselftsest with clang is a possibility, but currently we are only doing kselftest builds under gcc (managed by `build-configs.yaml`) but TBH, I don't know if anyone has done any work to be sure that all of kselftests build under clang.

broonie: I did a very small amount of initial work on it, up until a couple of releases ago it used to just not even try to cross compile when using clang which just didn't work at all.

nkbelin: I just dropped this part as discussed

config/docker/fragment/kselftest.jinja2
nkbelin: changed

nkbelin:
@broonie I just updated the commit message of the remaining commit. Let me know if it looks clearer to you, thanks

broonie:
LGTM.

khilman (changes requested):
Please describe a bit more about why packages are "not available."   The background is that riscv support is only available in debian unstable (sid), and in side the full set of pacakages for riscv isn't ready, so the only thing available is the cross-compilation package.

nkbelin:
@khilman added more background

khilman (approved):

khilman:
FYI @atishp04

nuclearcat:
Excellent, thank you for build results, thats what i need. I checked also staging, everything looks fine.

PR #1468, opened 2022-10-12
config: add rust build configurations

10ne1:
config: add rust build configurations
    
    Starting with the upcoming v6.1, mainline Linux has merged the
    initial Rust infrastructure so this adds some configs for build
    testing it together with some sample modules.
    
    The kernel requires a specific version of rustc, so we add the
    rustc-1.62 build environment which derives from clang-15, since
    a C compiler is still required to build the kernel and the
    supported kernel version is 15 (we might bump this later).
    
    Obviously GCC can be used as well but for now testing all the
    toolchain combinations does not add significant value. In the
    future more toolchain combinations can be used as needed.
    
    The official "offline" toolchain installation method is used as
    documented at [1] with sha256sum and because some distros like
    Debian stable might not provide up to date toolchain and crates
    to keep up with the mainline kernel.
    
    Only the x86_64 architecture is supported by the kernel for now.
    
    We also add the Rust-for-Linux kernel maintainer trees which
    contain additional modules and bindings.
    
### DIFF
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -109,6 +109,9 @@ trees:
   rt-stable:
     url: "https://git.kernel.org/pub/scm/linux/kernel/git/rt/linux-stable-rt.git"
 
+  rust-for-linux:
+    url: "https://github.com/Rust-for-Linux/linux.git"
+
   samsung:
     url: "https://git.kernel.org/pub/scm/linux/kernel/git/kgene/linux-samsung.git"
 
@@ -227,6 +230,42 @@ fragments:
       - 'CONFIG_EXPERT=y'
       - 'CONFIG_PREEMPT_RT=y'
       - 'CONFIG_PREEMPT_RT_FULL=y' # <= v4.19
+
+  rust:
+    path: "kernel/configs/rust.config"
+    configs:
+      - 'CONFIG_RUST=y'
+
+  rust-for-linux-samples:
+    path: "kernel/configs/rust-for-linux-samples.config"
+    configs:
+      - 'CONFIG_SAMPLES=y'
+      - 'CONFIG_SAMPLES_RUST=y'
+      - 'CONFIG_SAMPLE_RUST_MINIMAL=m'
+      - 'CONFIG_SAMPLE_RUST_HOSTPROGS=m'
+      - 'CONFIG_SAMPLE_RUST_PRINT=m'
+      - 'CONFIG_SAMPLE_RUST_MODULE_PARAMETERS=m'
+      - 'CONFIG_SAMPLE_RUST_SYNC=m'
+      - 'CONFIG_SAMPLE_RUST_CHRDEV=m'
+      - 'CONFIG_SAMPLE_RUST_MISCDEV=m'
+      - 'CONFIG_SAMPLE_RUST_STACK_PROBING=m'
+      - 'CONFIG_SAMPLE_RUST_SEMAPHORE=m'
+      - 'CONFIG_SAMPLE_RUST_SEMAPHORE_C=m'
+      - 'CONFIG_SAMPLE_RUST_RANDOM=m'
+      - 'CONFIG_SAMPLE_RUST_PLATFORM=m'
+      - 'CONFIG_SAMPLE_RUST_FS=m'
+      - 'CONFIG_SAMPLE_RUST_NETFILTER=m'
+      - 'CONFIG_SAMPLE_RUST_ECHO_SERVER=m'
+      - 'CONFIG_SAMPLE_RUST_HOSTPROGS=m'
+      - 'CONFIG_SAMPLE_RUST_SELFTESTS=m'
+
+  rust-samples:
+    path: "kernel/configs/rust-samples.config"
+    configs:
+      - 'CONFIG_SAMPLES=y'
+      - 'CONFIG_SAMPLES_RUST=y'
+      - 'CONFIG_SAMPLE_RUST_HOSTPROGS=y'
+      - 'CONFIG_SAMPLE_RUST_MINIMAL=m'
 
   tinyconfig:
     path: "kernel/configs/tiny.config"
@@ -395,6 +434,14 @@ build_environments:
     cc_version: 16
     arch_params:
       <<: *clang_15_arch_params
+
+  rustc-1.62:
+    cc: clang
+    cc_version: 15
+    arch_params:
+      x86_64:
+        <<: *x86_64_params
+        name:
 
 # Default config with full build coverage
 build_configs_defaults:
@@ -959,6 +1006,26 @@ build_configs:
     tree: rt-stable
     branch: 'v5.15-rt'
     variants: *preempt_rt_variants
+
+  rust:
+    tree: mainline
+    branch: 'master'
+    variants:
+      rustc-1.62:
+        build_environment: rustc-1.62
+        fragments: [rust, rust-samples, kselftest]
+        architectures:
+          x86_64: *x86_64_arch
+
+  rust-for-linux:
+    tree: rust-for-linux
+    branch: 'rust'
+    variants:
+      rustc-1.62:
+        build_environment: rustc-1.62
+        fragments: [rust, rust-for-linux-samples, kselftest]
+        architectures:
+          x86_64: *x86_64_arch
 
   samsung:
     tree: samsung
--- a/config/docker/rustc-1.62.jinja2
+++ b/config/docker/rustc-1.62.jinja2
@@ -1,0 +1,39 @@
+{% extends 'clang-15.jinja2' %}
+
+{% block packages %}
+{{ super() }}
+
+ARG RUST_VER=1.62.0
+ARG BINDGEN_VER=0.56.0
+
+ARG RUST_TRIPLE=rust-${RUST_VER}-x86_64-unknown-linux-gnu
+
+ENV CARGO_HOME=/home/kernelci/.cargo
+ENV PATH=/usr/${RUST_TRIPLE}/bin:${CARGO_HOME}/bin:${PATH}
+
+ARG SHA256SUM=4172d3cb316498025410bdf33a4d0f756f5e77fbaee1fb042ccdef78239be1db
+
+# fetch, verify the toolchain
+RUN wget https://static.rust-lang.org/dist/${RUST_TRIPLE}.tar.gz && \
+    echo "${SHA256SUM} ${RUST_TRIPLE}.tar.gz" | sha256sum --check --quiet
+
+# install & cleanup tmp files
+RUN tar -xf ${RUST_TRIPLE}.tar.gz  -C /tmp/ && \
+    /tmp/${RUST_TRIPLE}/install.sh --prefix=/usr/${RUST_TRIPLE} \
+        --components=rustc,cargo,rust-std-x86_64-unknown-linux-gnu && \
+    rm -rf /tmp/${RUST_TRIPLE} && \
+    rm /${RUST_TRIPLE}*
+
+# This image is based on clang-*jinja2 which only has clang installed but rustc
+# defaults to linker "cc" which calls the non-existing GNU stack (gcc, libgcc).
+# So we point cargo/rustc to use the LLVM stack via the clang linker entrypoint.
+RUN mkdir -p ${CARGO_HOME} && \
+    printf '[build]\nrustflags = ["-C", "linker=clang"]' > ${CARGO_HOME}/config.toml
+
+RUN git clone --recurse-submodules --branch $RUST_VER \
+        https://github.com/rust-lang/rust \
+        $(rustc --print sysroot)/lib/rustlib/src/rust
+
+RUN cargo install --locked --version ${BINDGEN_VER} bindgen
+
+{%- endblock %}


nuclearcat:
Testing it, might take a while, i will report progress here.
Docker image rustc-1.62:kernelci built successfully

config/core/build-configs.yaml
```diff
@@ -225,6 +228,42 @@ fragments:
...
+  rust-samples:
+    path: "kernel/configs/rust-samples.config"
+    configs:
+      - 'CONFIG_SAMPLES=y'
+      - 'CONFIG_SAMPLES_RUST=y'
+      - 'CONFIG_SAMPLE_RUST_MINIMAL=m'
+      - 'CONFIG_SAMPLE_RUST_HOSTPROGS=m'
+
+  rust-for-linux-samples:
+    path: "kernel/configs/rust-samples.config"
```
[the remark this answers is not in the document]
mgalka: I think, I'd go with two different filenames for config. Their content seems different, so I guess separate names could let us avoid confusion.

mgalka (commented):
In general it looks good. I'd probably go with 2 separate names for the .config files (mentioned above)

config/core/build-configs.yaml
```diff
@@ -384,6 +423,10 @@ build_environments:
     arch_params:
       <<: *clang_12_arch_params
 
+  rustc-1.62:
+    cc: clang
+    cc_version: 15
```
mgalka, line 428:
```suggestion
  rustc-1.62:
    cc: clang
    cc_version: 15
    arch_params:
      x86_64: 
        <<: *x86_64_params
        name:
```
I think we can explicitly set an empty architecture name here. Otherwise jenkins build job will be looking for an `rustc-1.62-x86_64` docker image to get a proper toolchain. As long as we have just one docker image template for this config, I think we don't need to suffix it with an arch. 
Having said that, if you think it's worth adding an arch suffix to the docker image then `rustc-1.62.jinja2` template should be renamed to `rustc-1.62-x86_64.jinja2`.

mgalka (changes requested):
Some changes are required to make the image build by kernelci Jenkins.

gctucker:
See also the email thread with @ojeda from January 2022 https://groups.io/g/kernelci/message/1336

ojeda:
Thanks for the Cc!

> The official "rustup" toolchain installation method is used as
> documented in the upstream kernel

The current commits seem to be using the standalone installers, not `rustup`. Is the commit description correct?

mgalka:
Tested builds on staging:
https://bot.staging.kernelci.org/job/kernel-build/91146/console
https://bot.staging.kernelci.org/job/kernel-build/91160/consoleFull

Builds OK

config/docker/rustc-1.62.jinja2
```diff
@@ -0,0 +1,42 @@
...
+# fetch, validate and import the rust team key
+RUN wget https://static.rust-lang.org/rust-key.gpg.ascii
+RUN gpg --import --import-options show-only rust-key.gpg.ascii | \
+        grep -q 108F66205EAEB0AAA8DD5E1C85AB96E6FA1BE5FE
+RUN gpg --import rust-key.gpg.ascii
+
+# fetch, verify the toolchain
+RUN wget https://static.rust-lang.org/dist/${RUST_TRIPLE}.tar.gz
+RUN wget https://static.rust-lang.org/dist/${RUST_TRIPLE}.tar.gz.asc
+RUN gpg --verify ${RUST_TRIPLE}.tar.gz.asc ${RUST_TRIPLE}.tar.gz
```
ojeda, line 24: If the version needs to be fixed, then an easier/faster way is to download the `.tar.gz` and compare its sha256 against one stored here.

config/docker/rustc-1.62.jinja2
```diff
@@ -0,0 +1,42 @@
...
+
+{% block packages %}
+{{ super() }}
+
+ENV RUST_VER=1.62.0
+ENV BINDGEN_VER=0.56.0
+
+ENV RUST_TRIPLE=rust-${RUST_VER}-x86_64-unknown-linux-gnu
+
+ENV CARGO_HOME=/usr/$RUST_TRIPLE/cargo
```
ojeda, line 11: Should these (and the above ones) be `ARG`?

config/docker/rustc-1.62.jinja2
```diff
@@ -0,0 +1,42 @@
...
+
+# install & cleanup tmp files
+RUN tar -xf ${RUST_TRIPLE}.tar.gz  -C /tmp/
+RUN /tmp/${RUST_TRIPLE}/install.sh --prefix=/usr/${RUST_TRIPLE} \
+        --components=rustc,cargo,rust-std-x86_64-unknown-linux-gnu
+RUN rm -rf /tmp/${RUST_TRIPLE}
+RUN rm /${RUST_TRIPLE}*
+
+RUN mkdir $CARGO_HOME
+RUN printf '[build]\nrustflags = ["-C", "linker=clang"]' > $CARGO_HOME/config.toml
```
ojeda, line 34: Why is this needed? A comment would be useful here.

config/docker/rustc-1.62.jinja2
ojeda: My pleasure!

config/docker/rustc-1.62.jinja2
```diff
@@ -0,0 +1,42 @@
+{% extends 'clang-15.jinja2' %}
+
+{% block packages %}
+{{ super() }}
+
+ENV RUST_VER=1.62.0
+ENV BINDGEN_VER=0.56.0
```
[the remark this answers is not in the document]
gctucker: It's possible to set build configurations with kernel version filters to only build particular kernel versions with particular toolchain versions.  However this will require manual maintenance in the YAML configuration.

config/docker/rustc-1.62.jinja2
ojeda: The signature way should be fine since you were checking the full fingerprint above (I think it would require a preimage attack on SHA-1?), but yeah, using a hardcoded thing is simpler (though requires a bit more updating when changing the version, of course). For similar reasons I suggested the same for `tc-build`.

Also, speaking of simplicity: you can probably combine some of these steps into a single `RUN`, since I don't think it buys much having several layers per step?

config/docker/rustc-1.62.jinja2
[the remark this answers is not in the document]
ojeda: Yeah, I imagined something like that. Then it is perfectly fine creating sets of build images :)

> We could add a test in KernelCI for the kernel builds which verifies what specific Rust toolchain version is required and fail the build if there is a toolchain version mismatch, though I don't know if that makes sense,

The kernel already checks if the version is recent enough, so it should be fine. It also checks if it is too old, but that is a warning only.

> depending on what the plans are for Rust toolchain backwards compatibility in the kernel? Maybe it's still too early to know this?

The goal is to provide a minimum Rust supported version like for GCC and Clang (that can be upgraded from time to time like any other), but before that happens some unstable features need to be stabilized on the Rust side (or at least "semi-stabilized", even if not fully stable for general use).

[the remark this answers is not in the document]
ojeda: > It's possible to set build configurations with kernel version filters to only build particular kernel versions with particular toolchain versions. However this will require manual maintenance in the YAML configuration.

Sounds good to me (or do you mean this is an alternative vs. something else?)

How should we coordinate the changes? i.e. if we submit an upgrade to mainline, should we submit it here before Linus merges the PR? (setting it to the next release? Or per -rc?)

config/docker/rustc-1.62.jinja2
ojeda: `PATH` for sure, that is why I didn't include it.

As for `CARGO_HOME`, is there a reason to not let it use the default at the home folder? It feels weird to put it inside `/usr`, since it is meant to be a writable cache.

config/docker/rustc-1.62.jinja2
ojeda: > but that cc is being used as the linker

Yeah, that is what I meant by `cc` instead of `ld`.

> I tried setting for eg linker-flavor=ld.lld but got the following error because libgcc is still required.

As far as I know, that is because `linker-flavor` will make it call `lld` directly, while C compilers take care of adding some extra flags to the linker, such as to find the system libraries.

config/docker/rustc-1.62.jinja2
ojeda: Are you sure Cargo home supports being shared across users?

config/docker/rustc-1.62.jinja2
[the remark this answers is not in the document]
ojeda: The kernel already checks the version, so I assume you will get a fairly clear build error if you are using the lowest version previously supported before a given upgrade. But yeah, adding an specific test sounds good too if that gives better reports.

As for intervention from the kernel maintainers side, @gctucker asked us in the mailing list whether we would help with this, so I assumed we would have to submit the updates or something like that. I would be happy to do so if needed; but if things are more streamlined, even better!

config/docker/rustc-1.62.jinja2
ojeda: How did you verify it? i.e. note that it is not just that it works, it would need to be designed for it. For instance, if it uses lock files or similar, does it properly work for several users? Or, since the cache is writable, does it have security implications? etc.

Considering the default is per-user, unless the docs state it is supported, I would be very cautious about it.

nuclearcat:
I'm seeing some problems with rust on staging:
```
+ kci_build pull_tarball --url http://storage.staging.kernelci.org/rust-for-linux/rust/v6.1-rc1-1486-g729c03011c60/linux-src_rust-for-linux.tar.gz --retries 3 --delete
+ kci_build generate_fragments --build-config=rust-for-linux
kernel/configs/rust.config
kernel/configs/rust-for-linux-samples.config
kernel/configs/amdgpu.config
kernel/configs/crypto.config
kernel/configs/ima.config
kernel/configs/kvm_guest.config
kernel/configs/xfstests.config
+ kci_build init_bmeta --build-config=rust-for-linux --commit=729c03011c60382030520f84904a85e8a5b3cfb9 --describe=v6.1-rc1-1486-g729c03011c60 --describe-verbose=v6.1-rc1-1486-g729c03011c60
Initialising build meta-data in /tmp/kci/linux/build
Install directory: /tmp/kci/linux/build/_install_
+ kci_build make_config --defconfig=x86_64_defconfig+x86-chromebook+kselftest
  File "/usr/local/bin/kci_build", line 4, in <module>
    __import__('pkg_resources').run_script('kernelci==1.1', 'kci_build')
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 651, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 1448, in run_script
    exec(code, namespace, namespace)
  File "/usr/local/lib/python3.9/dist-packages/kernelci-1.1-py3.9.egg/EGG-INFO/scripts/kci_build", line 451, in <module>
  File "/usr/local/lib/python3.9/dist-packages/kernelci-1.1-py3.9.egg/EGG-INFO/scripts/kci_build", line 344, in __call__
    res = step.run(args.j, args.verbose, opts)
  File "/usr/local/lib/python3.9/dist-packages/kernelci-1.1-py3.9.egg/kernelci/build.py", line 1204, in run
    self._gen_kci_frag(configs, fragments, kci_frag_name)
  File "/usr/local/lib/python3.9/dist-packages/kernelci-1.1-py3.9.egg/kernelci/build.py", line 1118, in _gen_kci_frag
    with open(path) as frag:
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/kci/linux/kernel/configs/kselftest.config'
Build result: 1

bmeta.json
```
Several rust builds failed.

config/docker/rustc-1.62.jinja2
```diff
@@ -0,0 +1,39 @@
...
+
+# install & cleanup tmp files
+RUN tar -xf ${RUST_TRIPLE}.tar.gz  -C /tmp/
+RUN /tmp/${RUST_TRIPLE}/install.sh --prefix=/usr/${RUST_TRIPLE} \
+        --components=rustc,cargo,rust-std-x86_64-unknown-linux-gnu
+RUN rm -rf /tmp/${RUST_TRIPLE}
+RUN rm /${RUST_TRIPLE}*
+
+# This image is based on clang-*jinja2 which only has clang installed but rustc
+# defaults to linker "cc" which calls the non-existing GNU stack (gcc, libgcc, ld).
```
ojeda, line 28: Nit: I think binutils is installed? So `ld` would be there.

ojeda:
> can you please leave a comment so I can add a reviewed-by tag for you?

I wrote a nit above and also I left a comment a few days ago above on minimizing the `RUN` layers, but that is just an optimization, and if you are not doing it in other files, I would ignore it here (I saw it in `qemu.ninja2`, but not in others, so I don't know if you are following that practice in KernelCI or not...).

Other than those, it looks good to me!

One final note: the `rust` branch will likely go away or be repurposed at some point in the future -- I assume this is not a problem, but just so that you are aware :)

config/core/build-configs.yaml
```diff
@@ -939,6 +986,26 @@ build_configs:
...
+    branch: 'master'
+    variants:
+      rustc-1.62:
+        build_environment: rustc-1.62
+        fragments: [rust, rust-samples]
+        architectures:
+          x86_64: *x86_64_arch
+
+  rust-for-linux:
+    tree: rust-for-linux
```
ojeda, line 1000: Is it possible to add rust-next too (i.e. is the one that is in linux-next)? If it is a matter of resources, it is probably better to replace this one (`rust`) with it.

mgalka: I think it's  possible, however I'd rather have it added with another PR.
@gctucker What do you recon?

nuclearcat:
It seems build still failed recently on staging due kfselftest fragment missing (not sure why):
```
+ kci_build init_bmeta --build-config=rust-for-linux --commit=729c03011c60382030520f84904a85e8a5b3cfb9 --describe=v6.1-rc1-1486-g729c03011c60 --describe-verbose=v6.1-rc1-1486-g729c03011c60
Initialising build meta-data in /tmp/kci/linux/build
Install directory: /tmp/kci/linux/build/_install_
+ kci_build make_config --defconfig=x86_64_defconfig+x86-chromebook+kselftest
  File "/usr/local/bin/kci_build", line 4, in <module>
    __import__('pkg_resources').run_script('kernelci==1.1', 'kci_build')
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 651, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 1448, in run_script
    exec(code, namespace, namespace)
  File "/usr/local/lib/python3.9/dist-packages/kernelci-1.1-py3.9.egg/EGG-INFO/scripts/kci_build", line 451, in <module>
  File "/usr/local/lib/python3.9/dist-packages/kernelci-1.1-py3.9.egg/EGG-INFO/scripts/kci_build", line 344, in __call__
    res = step.run(args.j, args.verbose, opts)
  File "/usr/local/lib/python3.9/dist-packages/kernelci-1.1-py3.9.egg/kernelci/build.py", line 1204, in run
    self._gen_kci_frag(configs, fragments, kci_frag_name)
  File "/usr/local/lib/python3.9/dist-packages/kernelci-1.1-py3.9.egg/kernelci/build.py", line 1118, in _gen_kci_frag
    with open(path) as frag:
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/kci/linux/kernel/configs/kselftest.config'
+ export KERNEL_BUILD_RESULT=1
+ KERNEL_BUILD_RESULT=1
+ set +x
Build result: 1

bmeta.json
{
    "build": {
        "duration": 0.03600597381591797,
        "status": "PASS"
    },
    "environment": {
        "arch": "x86_64",
        "compiler": "clang",
        "compiler_version": "15",
        "compiler_version_full": "Debian clang version 15.0.4-++20221102053030+5c68a1cb1231-1~exp1~20221102053120.92",
        "cross_compile": "",
        "cross_compile_compat": "",
        "make_opts": {
            "KBUILD_BUILD_USER": "KernelCI"
        },
        "name": "rustc-1.62",
        "platform": {
            "uname": [
                "Linux",
                "build-j92564-x86-64-rustc-1-62-x86-64-defconfig-x86-chromed54hv",
                "5.10.127+",
                "#1 SMP Fri Jul 8 00:39:24 UTC 2022",
                "x86_64",
                ""
            ]
        },
        "use_ccache": true
    },
    "revision": {
        "branch": "rust",
        "commit": "729c03011c60382030520f84904a85e8a5b3cfb9",
        "describe": "v6.1-rc1-1486-g729c03011c60",
        "describe_verbose": "v6.1-rc1-1486-g729c03011c60",
        "tree": "rust-for-linux",
        "url": "https://github.com/Rust-for-Linux/linux.git"
    }
}
_install_
```

nuclearcat:
I am not sure why it is hardcoded, @mgalka @gctucker ?

mgalka:
> I am not sure why it is hardcoded, @mgalka @gctucker ?

I don't know why `+keselftest` is part of the `x86_64_arch`. I think we can just add config without kselftest for rust builds.

```yaml
  rust:
    tree: mainline
    branch: 'master'
    variants:
      rustc-1.62:
        build_environment: rustc-1.62
        fragments: [rust, rust-samples]
        architectures:
          x86_64: &x86_64_arch_nokselftest
            extra_configs:
              - 'allmodconfig'
              - 'allnoconfig'
              - 'x86_64_defconfig+x86-chromebook'
            fragments: [amdgpu, crypto, ima, x86_kvm_guest, x86-chromebook]

  rust-for-linux:
    tree: rust-for-linux
    branch: 'rust'
    variants:
      rustc-1.62:
        build_environment: rustc-1.62
        fragments: [rust, rust-for-linux-samples]
        architectures:
          x86_64: *x86_64_arch_nokselftest

```

I'm giving a try to at the moment on staging.

gctucker:
Well staging build configurations are pretty arbitrary and aim at getting the highest tests / build ratio.  Building with kselftest fragment enabled is required to run kselftests, and we need that on staging.  There's no point building without kselftest on staging as that's not adding any real coverage.  If we can't build kselftest with rust then we just need to fix that in the rust build config options, no need to disable kselftest builds when rust is not enabled.

mgalka:
> @mgalka do you agree?

I'm fine with it. Giving it a try on staging.

mgalka:
Tested OK on staging. A build which previously failed, works fine now.

https://bot.staging.kernelci.org/job/kernel-build/95009/console

mgalka (approved):
LGTM

aliceinwire (approved):

aliceinwire:
Thanks LGTM

PR #1473, opened 2022-10-18
qemu: add network to riscV qemu

montjoie:
LAVA use old qemu network options which does not work on riscV qemu. So add network via extra-options with the new full explicit network options.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -1556,7 +1556,7 @@ device_types:
       machine: 'virt'
       no_kvm: True
       memory: 1024
-      extra_options: ['-bios default']
+      extra_options: ['-bios default -device virtio-net,netdev=main -netdev user,id=main']
     filters:
       - passlist: {defconfig: ['defconfig']}
 


nuclearcat:
It looks like network got enabled, but not visible in dmesg:
https://storage.staging.kernelci.org/kernelci/staging-next/staging-next-20221103.0/riscv/defconfig/gcc-10/lab-broonie/baseline-qemu_riscv64.html
Thats strange, because i see: CONFIG_VIRTIO_NET=y in https://storage.staging.kernelci.org/kernelci/staging-next/staging-next-20221103.0/riscv/defconfig/gcc-10/config/kernel.config as well

montjoie:
See qemu start message:
qemu-system-riscv64: warning: requested NIC (#net009, model virtio) was not created (not supported by this machine?)
Legacy network options does not work on qemu riscV.

nuclearcat (approved):
https://storage.staging.kernelci.org/kernelci/staging-mainline/staging-mainline-20221130.0/riscv/defconfig/gcc-10/lab-baylibre/baseline-qemu_riscv64.html
eth0 is working and visible now, but i believe it requires fix in LAVA device dictionary (ip=dhcp) as in https://github.com/kernelci/kernelci-core/pull/1537

PR #1474, opened 2022-10-18
debos: enable systemd-timesync

montjoie:
DUT need to have a correct time and cannot rely on having a RTC for that.
This is needed for lot of tests which need to have a correct time. Moslty all x509/TLS based actions, like doing apt (over https).

Signed-off-by: Corentin Labbe <clabbe@baylibre.com>

### DIFF
--- a/config/rootfs/debos/rootfs.yaml
+++ b/config/rootfs/debos/rootfs.yaml
@@ -36,6 +36,7 @@ actions:
       - usbutils
       - initramfs-tools
       - patch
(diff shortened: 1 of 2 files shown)


mgalka (approved):
Tested OK on staging 

https://bot.staging.kernelci.org/job/rootfs-builder/462/

```
# dpkg -l | grep timesync
ii  systemd-timesyncd         247.3-7+deb11u1                amd64        minimalistic service to synchronize local time with NTP servers
```

PR #1475, opened 2022-10-19
CHROMEOS cros-tast.jinja2: Add new file required by R106 tast

nuclearcat:
As tast tests keep evolving, local docker require new file that seems working as kind of RPC server. Without this file tast tests can be listed, but cannot be executed.
This patch need to be enabled only after all Chromebooks are flashed to R106.


nuclearcat:
Removing staging skip as all chromebooks are updated

nuclearcat:
Results are on dashboard: https://chromeos.kernelci.org/test/job/kernelci/branch/chromeos-stable/kernel/chromeos-stable-20221102.1/

PR #1478, opened 2022-10-19
choose the debian version depending on the plaform arch

nkbelin:
In order to test kselftest and ltp on riscv platforms, we currently have to use debian-sid instead of debian-bullseye.
The reason for that is that bullseye does not support riscv platform and only sid does.
As a consequence, and in order to not duplicate the test plans in test-configs.yaml, the logic to be able to choose the debian version depending on the arch that is being tested has been added.

To do so:
- the mention to bullseye has been removed in rootfs in the test plans defintions
- a image-distros.yaml file has been added and it contains the information to get the rootfs name with the debian version from the name (without the debian version, now used in the test plan) and the architecture.
- a Distro object has been added to contain the information from the image-distros.yaml
- the TestPlan object now has a Distro object that can contain different RootFS objects depending on the architecture instead of a single RootFS object.
- When generating a test, the architecture is known and the correct RootFs object is chosen.

I am not really happy with this PR as:
- it duplicates information like the boot_protocol, the type etc
- the way `get_template_path(self, boot_method)` is modified at the moment is not great
- I feel like I may be missing something !

Maybe I can transfer some parameters from rootfs-images to the file I just created (for example: type and boot_protocol which would make the get_template_path function better)?
Maybe we should just duplicate the testplans for riscv/sid and not do all this?

Let me know what you think about this PR, what could be improved or if there is a completely different solution that would be much better...

Thanks

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -78,7 +78,7 @@ test_plans:
       priority: 90
 
   baseline-nfs:
-    rootfs: debian_bullseye_nfs
+    rootfs: debian_nfs
     pattern: 'baseline/{category}-{method}-{protocol}-nfs-baseline-template.jinja2'
     params:
       priority: 85
@@ -101,10 +101,10 @@ test_plans:
             - 'lab-collabora-staging'
 
   cros-ec:
-    rootfs: debian_bullseye-cros-ec_ramdisk
+    rootfs: debian-cros-ec_ramdisk
 
   igt: &igt
-    rootfs: debian_bullseye-igt_ramdisk
+    rootfs: debian-igt_ramdisk
     pattern: 'igt/{category}-{method}-{protocol}-{rootfs}-igt.jinja2'
 
   igt-gpu-amd:
@@ -209,7 +209,7 @@ test_plans:
       drm_driver: "tegra"
 
   kselftest: &kselftest
-    rootfs: debian_bullseye-kselftest_nfs
+    rootfs: debian-kselftest_nfs
     pattern: 'kselftest/{category}-{method}-{protocol}-{rootfs}-kselftest-template.jinja2'
     filters:
       - passlist: {defconfig: ['kselftest']}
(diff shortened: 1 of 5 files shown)


nkbelin:
@gctucker @mgalka Let me know what you think and if there a better way to handle this, Thanks!

khilman:
Let's also think long term how to generalize this for different versions of other rootfs types (e.g. ubuntu, fedora, etc.) but also for yocto/OE.

gctucker:
I'm pretty sure we can use filters in YAML to define this without adding any code.  Also, supporting distros is not really something KernelCI was set up for.  However we can already do all that if we want to afaict.

nkbelin:
@gctucker Could you show me a simple filter example that would work for that? I am not entirely sure how to use the filters TBH! Then I would create a new PR from there... Thanks!

nkbelin:
@gctucker @mgalka I force pushed a completely different approach: I simply duplicate the test-plans.

For now I only duplicated the test plans that I am planning to run on the hifive-unleased platform.
In the future I will have to do more or less the same for the qemu_riscv64 platform.

This approach has the advantage of being easily "revertable" when riscv is finally supported by the stable debian and it does not change the kernelci internal logic.

Let me know what you think,

Thanks

config/core/test-configs.yaml
```diff
@@ -448,6 +503,56 @@ test_plans:
...
+    pattern: 'ltp/{category}-{method}-{protocol}-{rootfs}-ltp-template.jinja2'
+    filters:
+      - blocklist: *kselftest_defconfig_filter
+    params: &ltp-params_riscv
+      skip_install: "true"
+      tst_cmdfiles: ""
+      job_timeout: '15'
+      skipfile: skipfile-lkft.yaml
+
+  ltp-crypto_riscv:
```
gctucker, line 517: I think you need to add `base_name: ltp-crypto` so the results are tracked as `ltp-crypto` and not a separate test plan just for RISC-V.  See `baseline_qemu` for example, the results are still plain `baseline`.

gctucker: Same thing for the other test plans btw.

gctucker (commented):
That looks a lot more manageable.  However, the test plans aren't being used at all in any test configuration in this PR.  Is there at least `baseline-nfs` working for RISC-V already?

config/core/test-configs.yaml
nkbelin: @gctucker I created a brand new PR for this (https://github.com/kernelci/kernelci-core/pull/1556) as this PR was more about choosing the debian version and not simply duplicating the test plans.

In the new PR, I fixed the base_name and added the test plans that will be initially run on the hifive platform

khilman:
Replaced by: https://github.com/kernelci/kernelci-core/pull/1556

PR #1479, opened 2022-10-20
wait.py: Handle 2 more errors properly

nuclearcat:
If more than 1 pod found - don't fail, try to continue. If pod really failed - wget will fail to fetch data. Another case is when there is networking errors inside kubernetes, and because it generate exceptions specific to urllib3, we cannot handle them easily in our code, so we handle it as generic exception, which is not nice and thing to fix in future. And few pycodestyle fixes i missed.

### DIFF
--- a/config/k8s/wait.py
+++ b/config/k8s/wait.py
@@ -44,6 +44,8 @@ def find_pod(core, args):
         except client.rest.ApiException as e:
             print("ERROR: Exception on list_namespaced_pod, ", e)
             retries -= 1
+            if retries == 0:
+                raise e
             time.sleep(5)
             # One of solutions to reload config
             # https://github.com/DataBiosphere/toil/issues/2867
@@ -51,6 +53,16 @@ def find_pod(core, args):
                                     persist_config=False)
             core = client.CoreV1Api()
             continue
+        # exception might happen deep inside kubernetes, example:
+        # https://github.com/kernelci/kernelci-core/issues/1470#issue-1408983064
+        except Exception as e:
+            print(f"ERROR: find_pod generic exception: {e!r}")
+            retries -= 1
+            if retries == 0:
+                raise e
+            time.sleep(5)
+            continue
+
     return None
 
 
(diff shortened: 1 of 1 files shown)


nuclearcat:
It looks like this PR and previous ones are improving in general availability (as looking to staging), but unfortunately there is no significant improvement in case of "double pods", which i tried to handle as well, in part of this PR. I will attempt to address this later, if issue with duplicate pods are significant enough.

nuclearcat:
Finally got situation where new code is working, while it is not solving situation, it gives interesting insights and allows to say it is tested:
```

WARNING: >1 pod found with job name build-j58089-arm-gcc-10-multi-v7-defconfig-crypto
Found job on pod build-j58089-arm-gcc-10-multi-v7-defconfig-crypto-27tdc
ERROR: initContainer scratch-setup not ready / failed.
Exception: Unable to get pod log (400)
Reason: Bad Request
HTTP response headers: HTTPHeaderDict({'Audit-Id': '2ae4b1e7-43b9-4d4f-b031-161bd6e01d74', 'Cache-Control': 'no-cache, private', 'Content-Type': 'application/json', 'Date': 'Fri, 30 Dec 2022 02:18:56 GMT', 'Content-Length': '246'})
HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"container \"scratch-setup\" in pod \"build-j58089-arm-gcc-10-multi-v7-defconfig-crypto-27tdc\" is waiting to start: PodInitializing","reason":"BadRequest","code":400}


Found job on pod build-j58089-arm-gcc-10-multi-v7-defconfig-crypto-cdvxk
ERROR: initContainer scratch-setup not ready / failed.
Exception: Unable to get pod log (400)
Reason: Bad Request
HTTP response headers: HTTPHeaderDict({'Audit-Id': '57f76e64-c643-4e18-9992-9f9edd1969ee', 'Cache-Control': 'no-cache, private', 'Content-Type': 'application/json', 'Date': 'Fri, 30 Dec 2022 02:18:56 GMT', 'Content-Length': '246'})
HTTP response body: {"kind":"Status","apiVersion":"v1","metadata":{},"status":"Failure","message":"container \"scratch-setup\" in pod \"build-j58089-arm-gcc-10-multi-v7-defconfig-crypto-cdvxk\" is waiting to start: PodInitializing","reason":"BadRequest","code":400}
```

config/k8s/wait.py
```diff
@@ -130,44 +136,53 @@ def main(args):
     # Find pod where job ran
     #
     # Reload credentials to reduce risk of token expiration
-    config.load_kube_config(context=args.context, persist_config=False)
```
mgalka, line 133: Please move linter related changes to a separate commit

config/k8s/wait.py
```diff
@@ -185,14 +200,15 @@ def main(args):
     # Make jenkins fail only if k8s fails.  Kernel build fails will
     # be reported to the backend/
     sys.exit(not k8s_success)
-
```
mgalka, line 188: Please move linter related changes to a separate commit

config/k8s/wait.py
```diff
@@ -185,14 +200,15 @@ def main(args):
...
+
 
 if __name__ == "__main__":
     parser = argparse.ArgumentParser()
     parser.add_argument('--context')
     parser.add_argument('--job-name')
     parser.add_argument('--namespace', default='default')
     parser.add_argument('--sleep', type=int, default=60)
-    parser.add_argument('--no-delete', dest='delete', default=True, action='store_false')
+    parser.add_argument('--no-delete', dest='delete', default=True,
```
mgalka, line 211: Please move linter related changes to a separate commit

config/k8s/wait.py
```diff
@@ -130,44 +136,53 @@ def main(args):
...
+        #
+        # Get logs (from pod)
+        #
+
+        # default: check the main container logs
+        cont_name = jpod.spec.containers[0].name
+
+        # unless the initContainer failed, get that log, because if the
+        # initContainer failed, the main container will not have run
+        if (jpod.spec.init_containers):
```
mgalka, line 169: No need for parenthesis.

```suggestion
        if jpod.spec.init_containers:
```

mgalka (commented):
some remarks

mgalka:
I'll give it some more time on staging. It's not critical, but it'd be good to have it after it's well tested.

config/k8s/wait.py
```diff
@@ -51,6 +51,12 @@ def find_pod(core, args):
                                     persist_config=False)
             core = client.CoreV1Api()
             continue
+        # exception might happen deep inside kubernetes, example:
+        # https://github.com/kernelci/kernelci-core/issues/1470#issue-1408983064
+        except Exception as e:
+            print(f"ERROR: find_pod generic exception: {e!r}")
+            continue
```
gctucker, line 58: How are we going to notice exceptions if we just catch them?  Now we need to dig into the logs to find them.  Is it really better to hide the problem than make it crash and then we can investigate what actually caused it in Kubernetes?

config/k8s/wait.py
```diff
@@ -136,38 +142,47 @@ def main(args):
     if pod is None:
         print("ERROR: Failed to find pod")
         sys.exit(1)
-    if len(pod.items) < 1:
```
gctucker, line 139: FYI this whole `wait.py` thing is going away with the new API as the jobs will be sending their results to the API directly and that will generate an event so nothing will be waiting.  While it's nice to fix things, I would say that if it breaks less than 5% of the time it's not worth the effort.  That time could be much better spent on ramping up the new API instead.

config/k8s/wait.py
nuclearcat: As explained in comment - sometimes exceptions are just temporary networking errors to Kubernetes API or glitch inside Kubernetes node, so by catching exception and just printing it, we are giving find_pod chance to retry and recover. If it is really critical and repeating, it will try 10 times then just return None and whole script will fail, leading to job failure (then logs will be reviewed).

config/k8s/wait.py
nuclearcat: I just took a few minutes to finalize this PR(as its hanging for quite long time due low priority), as I've observed this is all that makes sense to improve at the moment.

nuclearcat:
Suggestion from @eds-collabora to re-raise exception if retries reached 0

mgalka (approved):

PR #1483, opened 2022-10-31
rootfs-configs.yaml: Add bullseye-audio

nfraprado:
Add a rootfs that can provide us the tools to validate a working audio setup. As a first step, the alsactl utility provided by the alsa-utils package can be used for this purpose.

A lot of current devices rely on UCM configuration files to have working audio. Ideally we would just add alsa-ucm-conf to the list of packages and call it a day. Unfortunately that package is very old and missing the files for a lot of devices that were enabled since. Additionally, the libasound package provides a library that is too old to parse most of the newer UCM files. For these reasons, also add a script that builds the latest libasound and downloads the latest UCM configuration files.

LAVA run running the rootfs on mt8173-elm-hana: https://lava.collabora.dev/scheduler/job/7797317

### DIFF
--- a/config/core/rootfs-configs.yaml
+++ b/config/core/rootfs-configs.yaml
@@ -52,6 +52,21 @@ rootfs_configs:
     script: "scripts/install-bootrr.sh"
     test_overlay: "overlays/baseline"
 
+  bullseye-audio:
+    rootfs_type: debos
+    debian_release: bullseye
+    arch_list:
+      - amd64
+      - arm64
+      - armhf
+    extra_packages:
+      - alsa-utils
+    extra_packages_remove:
+      - bash
+      - e2fslibs
+      - e2fsprogs
+    script: "scripts/bullseye-audio.sh"
+
   bullseye-cros-ec:
     rootfs_type: debos
     debian_release: bullseye
--- a/config/rootfs/debos/scripts/bullseye-audio.sh
+++ b/config/rootfs/debos/scripts/bullseye-audio.sh
@@ -1,0 +1,60 @@
+#!/bin/bash
+
+# Important: This script is run under QEMU
+
+set -e
+
+BUILD_DEPS="\
+      automake \
+      make \
+      libtool \
+      git \
+      openssl \
+      ca-certificates \
+      curl \
+"
+
+apt-get install --no-install-recommends -y  ${BUILD_DEPS}
+
+BUILDFILE=/test_suites.json
+echo '{  "tests_suites": [' >> $BUILDFILE
+
+# Build libasound2
+########################################################################
+
+LIBASOUND_PREFIX_DIR=/usr/local/
+LIBASOUND_URL=https://github.com/alsa-project/alsa-lib.git
+mkdir -p /var/tests/libasound2 && cd /var/tests/libasound2
+
+git clone --depth=1 $LIBASOUND_URL .
+
+echo '    {"name": "libasound2", "git_url": "'$LIBASOUND_URL'", "git_commit": ' \"`git rev-parse HEAD`\" '}' >> $BUILDFILE
+
+./gitcompile --prefix=$LIBASOUND_PREFIX_DIR
+make install
+
+# Make the libasound we built take precendence over the one from the distro
+echo $LIBASOUND_PREFIX_DIR/lib > /etc/ld.so.conf.d/1-libasound.conf
+
+echo '  ]}' >> $BUILDFILE
+
+# Download alsa-ucm-conf
+########################################################################
+
+UCM_CONF_DIR=$LIBASOUND_PREFIX_DIR/share/alsa
+mkdir -p /var/tests/ucm-conf && cd /var/tests/ucm-conf
+mkdir -p $UCM_CONF_DIR
+curl -L -o alsa-ucm-conf.tar.gz https://github.com/alsa-project/alsa-ucm-conf/archive/refs/heads/master.tar.gz
+tar xvzf alsa-ucm-conf.tar.gz -C $UCM_CONF_DIR --strip-components=1 --wildcards "*/ucm" "*/ucm2"
+
+########################################################################
+# Cleanup: remove files and packages we don't want in the images       #
+########################################################################
+rm -rf /var/tests
+
+apt-get remove --purge -y ${BUILD_DEPS}
+apt-get autoremove --purge -y
+apt-get clean
+
+# re-add some stuff that is removed by accident
+apt-get install -y initramfs-tools


broonie:
If you're at the level of using alsactl to validate the rootfs then the mixer-test kselftest ought to be more than enough, that's already supported with the current rootfs. It was rejected for testing on a bunch of machines including IIRC the Chromebooks due to concerns about Collabora's lab capacity. I have a vague plan to try to expand the ALSA selftests with a PCM test that just verifies silence plays in the appropriate amount of time which should mostly be able to run without a UCM config which might also help here.

config/core/rootfs-configs.yaml
```diff
@@ -52,6 +52,21 @@ rootfs_configs:
     script: "scripts/install-bootrr.sh"
     test_overlay: "overlays/baseline"
 
+  bullseye-audio:
+    rootfs_type: debos
+    debian_release: bullseye
+    arch_list:
+      - amd64
+      - arm64
+      - armhf
```
broonie, line 61: What's wrong with x86?

config/rootfs/debos/scripts/bullseye-audio.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
+
+# Important: This script is run under QEMU
```
broonie, line 3: Might it not make sense to get a Debian backport done here if we just need a newer version for newer UCM configs? I can imagine that a bunch of Debian users would benefit and it'd probably be easier to build on top of.

nfraprado:
> If you're at the level of using alsactl to validate the rootfs then the mixer-test kselftest ought to be more than enough, that's already supported with the current rootfs. It was rejected for testing on a bunch of machines including IIRC the Chromebooks due to concerns about Collabora's lab capacity.

Hm, I'm not entirely sure I understand what you mean. Are you suggesting that the mixer-test kselftest would serve the same purpose of this test? I did take a look at that test, but the aim there seemed to be more exercising the full kernel uAPI (similarly to v4l2-compliance) rather verifying correct behavior from a use case point of view (audio actually playing), which is what I was going for here. Both are really important and should be running on these devices, but they have different purposes.

>  I have a vague plan to try to expand the ALSA selftests with a PCM test that just verifies silence plays in the appropriate amount of time which should mostly be able to run without a UCM config which might also help here.

Sounds interesting as a next step for the ALSA selftests. But for this test I wasn't even avoiding the requirement for the UCM config since that's needed to be able to play audio on mt8173-elm-hana and the other devices I was testing on anyway. I was also toying with playing audio on the device's speakers and recording back through the microphone to validate that both speaker and microphone on the device really are working. This is what I was imagining as a second step for this present test, which is just checking the audio setup is usable.

config/core/rootfs-configs.yaml
nfraprado: Oh, I just missed it because most of the other tests weren't running on it, but I can add it. Thanks.

config/rootfs/debos/scripts/bullseye-audio.sh
nfraprado: Good point, I'll take a look into this.

gctucker:
> It was rejected for testing on a bunch of machines including IIRC the Chromebooks due to concerns about Collabora's lab capacity.

Maybe this was true a while ago but there's plenty of capacity now so this should definitely be revisited.

nfraprado:
@broonie Just wanted to confirm that we're on the same page before I move forward with this. Does my previous comment make sense to you? Or do you see a way of validating correct sound setup (including UCM loading) and playback by extending the ALSA kselftest?

In short, the way I see it we would have ALSA selftests running to validate the ALSA kernel uAPI, and alongside that a test to validate that all needed pieces (drivers, UCM, in some cases firmware for SOF) to drive the audio are there by running `alsactl init` (as done in https://github.com/kernelci/kernelci-core/pull/1484, which requires the rootfs in this PR) and later also validating that audio is actually being reproduced through the speaker.

nuclearcat:
Should i proceed with building rootfs in jenkins and making tests as in example? PR ready for review and no objections?

config/core/rootfs-configs.yaml
gctucker: Do you mean i386?

broonie:
The kselftests will verify that the kernel is up and running well - since you originally filed this pull request there's been a pcm-test added which also tests that PCM playback and capture are at least transferring data at an appropriate rate. 

For testing UCM that's getting a bit beyond what KernelCI usually does and testing compatibility of specific userspaces with boards - we could do it but it's a bit of an extra step. If we're doing that we could start off with making sure that all the resources in the UCM (controls, streams and anything else) are there?

If you want to get into trying to test audio in actual use cases that gets really tricky without a lab specifically set up for it. Doing tests playing through speakers to microphones isn't really great for lab setups - you'll have issues with crosstalk between different tests on multiple boards, other sources of background noise in labs, physical setup of the device, and issues with users getting irritated at their labs being noisy and so on. It's a bit more viable if systems can be configured to do an internal loopback but that's far from universal.

BTW please note that there's so much spam from github that things are fairly likely to get lost in the noise.

gctucker:
About testing with real hardware, it's also possible to test headset interfaces with a wired loopback.  For actual speakers, some things can be done to mitigate the issues mentioned above such as modulating the frequency with a different carrier on each device and adding some soundproofing of course.

Since it's all likely to be very much dependent on the hardware setup, maybe part of these tests should rather be done with a separate application rather than kselftest?

(PS @broonie: See also my question about your x86 comment - did you mean i386?  amd64 is already listed there)

broonie:
It's certainly possible to run tests that cover the actual inputs/outputs, but it requires labs to do additional physical setup on the boards that we probably shouldn't be relying on with the model we have without doing a bunch of work to support labs telling us they've done that setup and/or are prepared to tolerate testing with things like speakers and on board microphones.

config/core/rootfs-configs.yaml
broonie: Probably, yeah. That was a month ago.

gctucker:
Right, so this kind of thing could be done on the chromeos.kernelci.org instance initially to at least cover Chromebooks rather than KernelCI "upstream".

For the context of this PR, I guess we need to clarify how much can be defined in kselftest then.

@nfraprado Based on @broonie's comments (silent playback etc.) and the fact we can't assume anything about special hardware features in the lab, what kind of tests are you planning to cover with this new rootfs?

nfraprado:
The pcm-test from the alsa kselftest is indeed a lot more interesting, thank you for mentioning it @broonie. But to give an example, one machine that I wanted to add these tests for is mt8192-asurada-spherion, and for that one a few switches need to be set by userspace to connect the frontend to the backend DAIs. So that test would always fail on this machine without having [the UCM](https://github.com/alsa-project/alsa-ucm-conf/blob/master/ucm2/MediaTek/mt8192/mt6359-rt1015p-rt5682/init.conf) in place:

```
[  852.681388]  Playback_3: ASoC: no backend DAIs enabled for Playback_3
[  852.694751]  Playback_3: ASoC: error at dpcm_fe_dai_prepare on Playback_3: -22
# not ok 37 test.time1.0.3.0.PLAYBACK snd_pcm_hw_params: Invalid argument
# not ok 38 test.time2.0.3.0.PLAYBACK snd_pcm_hw_params: Invalid argument
```

I agree that involving userspace testing in KernelCI is stretching it a bit, but if we want to test functionality, and userspace is required for it, then there isn't  really a way around it.

For validating the UCM what I had in mind was to simply run `alsactl init`, since it reports if a command fails to run (implying the control doesn't exist):
```
	alsa-lib main.c:826:(execute_sequence) unable to execute cset 'name='foo control' 1'
	alsa-lib main.c:2502:(set_boot_user) Unable to execute boot sequence
	Found hardware: "mt8192_mt6359" "" "" "" ""
	Hardware is initialized using a generic method
```

But on closer inspection this only applies to the bootSequence, and not all commands that might be run from the UCM, like the enableSequence for individual devices. So we might need to do this validation manually if there's no other command that already goes through all devices.

Regarding the concerns on testing the audio through a loopback on the devices, I was indeed considering the speaker->microphone path at first to avoid any hardware dependencies (which are there for headphone->headset mic). While I agree the crosstalk and background noise could be a source of issues, by randomizing the carrier frequency like @gctucker mentioned I was able to successfully run some PoC tests with two machines simultaneously locally. Scaling to a lab would probably show more issues but I believe they would be workable. The concern of this test making it annoying to stay in the lab is one that perhaps needs more discussion.

Given the audio loopback test needs some more discussion still, it would be great to enable the alsa-kselftest for mt8192-asurada-spherion in the meantime. But given the UCM dependency I mentioned, I wonder if it would be acceptable to run `alsactl init` to try and load the UCM before running the kselftest, so that it could actually test the relevant PCMs on that machine (If so the packages/script for this rootfs could be added for kselftest, or better yet the backports ones).

gctucker:
> I agree that involving userspace testing in KernelCI is stretching it a bit

As long as the user-space tools are only there to set things up and exercise the kernel then I think it's entirely justified.

broonie:
I don't see a problem with setup that just configures things so audio will clock at all, it's stuff that gets into specific use cases that concerns me - with a lot of devices there's substantial flexibility so actual use cases can differ substantially on the same hardware with the same kernel.

Like I say the issues with testing via speakers/microphones aren't just background noise and crosstalk, labs might also have concerns about these paths being used at all due to noise or privacy concerns so we should arrange for such things to be opt in.

nfraprado:
OK, sounds like applying UCM settings before running the alsa kselftest is uncontroversial, and a good step forward towards audio testing, so I'll tackle that next.

config/rootfs/debos/scripts/bullseye-audio.sh
broonie: No idea, I don't think anyone had any specific plan there. I'd guess once it gets released at least.

nuclearcat:
Any update on PR?

nfraprado:
@nuclearcat no update on this PR yet. The aim for now is to add the alsa kselftest on mediatek platforms and load the UCM beforehand. To achieve this the idea was to backport the UCM-related packages to debian stable, but I haven't heard back from the maintainer, so at this point will run the test using debian testing instead, at least for the time being.

gctucker:
@nfraprado Hello, I thought you were now looking into using Debian Testing to get the packages or getting the UCM binaries via other means.  Has there been any progress on this?

nfraprado:
@gctucker That was done in #1740, and to be enabled in #1741. This PR started with the intention of preparing the way to add a separate test that loopbacks audio. But in the end using alsa kselftest (with the new pcm-test) as in those two PRs covers most if not all of the ground. So I'll just go ahead and close this PR.

PR #1487, opened 2022-11-02
set hifive as riscv64

nkbelin:
Change the hifive-unleashed-a00 class to riscv64-dtb as this platform is a riscv64 platform. This will allow using riscv64 debian images in order to run kselftest or ltp.
However keep using the "riscv" arch for buildroot.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -13,6 +13,7 @@ file_system_types:
       armel:   [{arch: arm}]
       x86:     [{arch: i386}, {arch: x86_64}]
       mipsel:  [{arch: mips}]
+      riscv:  [{arch: riscv64}]
 
   buildroot-staging:
     <<: *buildroot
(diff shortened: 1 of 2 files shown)


nuclearcat:
Tested on staging. No criteria set for successful PR test results,  and as it doesnt break existing builds, i guess it is fine.

gctucker:
I think this needs to be reverted, the convention is to use the architecture names as used in the kernel.  For RISC-V, it's `riscv` regardless of 32-bit or 64-bit as that's set in the kernel config files.

Issues about how to pick the rootfs image should be dealt with differently.  Breaking the convention of the architecture name will lead to more problems in the config and code.

khilman:
Yes, this needs to be reverted @nuclearcat 

With this in place, `kci_test generate` will not create LAVA jobs for hifive-unleashed-a00 at all.  I verified that reverting this allows jobs to be created for hifive-unleashed.

PR #1491, opened 2022-11-03
Add fvp device type

aistcv:
Signed-off-by: Aishwarya TCV <aishwarya.tcv@arm.com>
Change-Id: Id6eae1de9e1a10ff3256fde107afecda023e0a23


config/core/test-configs.yaml
```diff
@@ -3110,3 +3120,7 @@ test_configs:
     test_plans:
       - baseline
       - smc
+
+  - device_type: fvp
+    test_plans:
+      - kselftest-seccomp
```
broonie, line 3126: Probably good to add a baseline test here as well, it helps if we can check if the platform booted before checking individual testsuites.

config/core/test-configs.yaml
```diff
@@ -868,6 +868,14 @@ device_types:
     class: arm64-dtb
     boot_method: uboot
 
+  fvp:
+    mach: vexpress
+    arch: arm64
+    dtb: 'fvp-base-revc.dtb'
```
nuclearcat, line 874: Should be:
   dtb: 'arm/fvp-base-revc.dtb'

nuclearcat (changes requested):

config/core/test-configs.yaml
aistcv: Thanks for the feedback, I have made the changes.

config/core/test-configs.yaml
```diff
@@ -868,6 +868,14 @@ device_types:
     class: arm64-dtb
     boot_method: uboot
 
+  fvp:
+    mach: vexpress
+    arch: arm64
+    dtb: 'arm/fvp-base-revc.dtb'
+    boot_method: uboot
+    filters:
+      - passlist: {defconfig: ['defconfig']}
```
broonie, line 877: This is going to generate standard u-boot type jobs like for a regular physical board but we need custom stuff for the FVPs - it's going to need to be a new set of templates like qemu I think. We need to provide firmware as part of the job definition (and we'll need to supply instructions for building it so kernelci can host it and provide something repeatble).

nuclearcat:
Is there any updates on this PR?

aistcv:
> Is there any updates on this PR?

I am working on this. The plan is to commit the jinja template and the required changes in test-configs.yaml which seems to work fine. The blocker is now to sort out the dependencies required.

Is it possible to use fvp image and firmware binaries required, from shrinkwrap ?
Details of the tool: 
- Documentation: https://shrinkwrap.docs.arm.com/
- Repository v1.0.0 tag: https://gitlab.arm.com/tooling/shrinkwrap/-/tree/v1.0.0

Fvp image: https://hub.docker.com/r/shrinkwraptool/base-slim 
The firmware binaries are produced when we run the tool.

I would be really helpful  if the fvp image and firmware binaries are stored in kernelci storage for access. Could you guide on how to progess to get the dependencies in place, please?

gctucker:
Sorry this would need more discussion, it's not clear with the information provided here what actually needs to be done.

gctucker:
Thanks, so the minimum to be done here would be to provide a set of binaries with instructions to reproduce them in a README and maybe a build script to do like the UEFI binaries:
https://storage.kernelci.org/images/uefi/edk2-stable202005/

If you just make that available somewhere we can copy it to the storage server.  If we also need to build these binaries periodically in KernelCI like we do the rootfs images etc. then that's going to be more complicated, but I don't think that's what you're after here.

gctucker:
Can someone outside of the KernelCI maintainers team be in charge of producing these builds and making them available to KernelCI with release notes?

aistcv:
> @aistcv @broonie Based on last week's discussion, could someone please provide the firmware files?

Sorry for the delay. I have shared a zip file with the binaries.
[fvp_firmware_bin.zip](https://github.com/kernelci/kernelci-core/files/11091988/fvp_firmware_bin.zip)

gctucker:
> Thanks, so the minimum to be done here would be to provide a set of binaries with instructions to reproduce them in a README and maybe a build script to do like the UEFI binaries:
> https://storage.kernelci.org/images/uefi/edk2-stable202005/

@aistcv Thanks.  Would it be possible to also add a README with instructions to build these binaries from source or pointers to the code and any related documentation?

broonie:
I'd expect so but note that @aistcv is on holiday for a few weeks now so it'll take her a little while to update. 

If we need pointers to instructions the EDK2 stuff needs looking at as well...

gctucker:
> If we need pointers to instructions the EDK2 stuff needs looking at as well...

We asked and all we got this readme, which is better than nothing: https://storage.kernelci.org/images/uefi/edk2-stable202005/readme.txt

gctucker:
@broonie In @aistcv's absence, could you please at least provide a link to the source and revision that was used for building these binaries and if possible some description of what the firmware is for?  We could then copy that in a README and resolve this.

broonie:
I don't specifically know what was used to build the binaries. They're firmware for running on the FVP, I'm not sure what exact revisions were used to build so can't provide that information unfortunately.

gctucker:
OK thanks, let's wait for @aistcv to come back then.  We should still be able to get this tested on staging in the meantime with just the binaries in the zip file.

aistcv:
> OK thanks, let's wait for @aistcv to come back then. We should still be able to get this tested on staging in the meantime with just the binaries in the zip file.

I have attached the document below.
[readme.txt](https://github.com/kernelci/kernelci-core/files/11257796/readme.txt)

Along with the binaries, the fvp image also needs to be uploaded to storage. I will get the image and readme.txt for it asap.

broonie:
Is there anything else needed to get this into storage so we can start adding the code changes to generate jobs? They're not going to work without firmware images in storage to reference, and they need to know where things will be hosted to set up URLs appropriately.

gctucker:
Well we need the binaries I guess, unless I've missed something?

gctucker:

gctucker:
@broonie @aistcv Hello, sorry for the delay.  I've now uploaded the binaries and the readme file (with some line wrapping) here:
https://storage.kernelci.org/images/fvp/fvp-1.0.0/

Please let me know if this is all good from your point of view.

aistcv:
> @broonie @aistcv Hello, sorry for the delay. I've now uploaded the binaries and the readme file (with some line wrapping) here: https://storage.kernelci.org/images/fvp/fvp-1.0.0/
> 
> Please let me know if this is all good from your point of view.

Thanks@gctucker. It looks good. I will start with next steps.

gctucker:
@aistcv Can this be closed now that #2013 has been merged?  I believe it was a superset of this one.

PR #1495, opened 2022-11-07
build-configs.yaml: add kernel config fragment for xfstests

hardboprobot:


gctucker (commented):
I thought we were relying on xfstests-bld to generate the kernel config?

gctucker:
@hardboprobot Hello, any update on this?  Is there a particular reason why we would need to add a config fragment rather than rely on the maintainers' tool to generate the full kernel config?

gctucker:
OK, but that doesn't mean we need to add our own kernel fragment.  We can generate the kernel config using `xfstests-bld`, then do the kernel build with make and upload that to the API / storage, then run fstests with `xfstests-bld`.

So I think we can close this PR, and probably create a GitHub issue to keep track of the idea of splitting out the kernel build from the fstests runs.

gctucker:
No update since last time, and here's an example of how the configuration and build is now done for KUnit:
```python
        steps = {
            'config': 'config',
            'build': 'build --jobs=$(nproc)',
            'exec': f'exec --json={kunit_json}',
        }
        step_results = {name: (None, []) for name in steps.keys()}

        for name, command in steps.items():
            res = self._run_kunit(src_path, command, job_log)
            res_str = 'pass' if res is True else 'fail'
            step_results[name] = (res_str, [])
            if res is False:
                break
```
We can follow the same approach for fstests.  Closing this now as it doesn't seem necessary to add a fragment.

PR #1501, opened 2022-11-09
job-build.jinja2: Set image pull policy as Always

nuclearcat:
As reported and observed we have numerous issues with stale docker images on k8s builders. One of possible reasons is that by default pull policy for image with specified tag is **IfNotPresent**, which we need to change to **Always**.


nuclearcat:
It looks like staging still works fine. I verified running jobs definition on staging and can see **imagePullPolicy: Always** applied there

mgalka (approved):
I agree with setting this option to `Always`.
We should also observe how it affects our pull limits.

PR #1503, opened 2022-11-10
config/runtime: allow different kernel source types in python template

hardboprobot:
Make the kernel origin specification more flexible. In addition to source tarballs, make it possible to use a prebuilt kernel image and also to take the kernel from a local source code tree for debugging purposes.

### DIFF
--- a/config/runtime/base/python.jinja2
+++ b/config/runtime/base/python.jinja2
@@ -13,6 +13,7 @@ import tarfile
 import traceback
 import urllib.parse
 import yaml
+import shutil
 {% endblock %}
 
 {%- block python_local_imports %}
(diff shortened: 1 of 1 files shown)


gctucker:
Yes, there should eventually be a build step first to produce kernel binaries and I guess it should apply to fstests as well since kernel builds aren't directly run by xfstests-bld.  My main point is that the runner service shouldn't have any knowledge of that, it should only be defined in YAML: some jobs will depend on other jobs to complete first, anything more specific than that is special to each job / test.  At the moment, the runner service just waits for tarballs to schedule some test jobs.  But ultimately, the dependencies between jobs should be in YAML to describe the pipeline: KUnit would depend just on the source tarball, other things will depend on a build which itself depends on a source tarball etc.  So adding logic about kernel binaries in runner.py won't work when we start doing fully-featured pipelines.

gctucker:
Adding `staging-skip` as this is conflicting with other changes.

mgalka:
@hardboprobot Do you want to work on this or we should close it and re-open once build jobs are implemented in the pipeline/core? Currently the API and pipeline are developed quite heavily atm.

mgalka:
@hardboprobot Ping, btw. #1541 is merged already

gctucker:
Yes I believe this kind of change was useful during development but these pipeline services are evolving very quickly and more sustainable developer tools are starting to take shape.  So the intention is completely valid and should be taken into account so something similar should still be possible with a more definitive implementation.

mgalka:
Closing as per @hardboprobot comment https://github.com/kernelci/kernelci-core/pull/1503#issuecomment-1496012167

PR #1504, opened 2022-11-10
config/lava/base: Support optional recovery image section for TFTP de…

broonie:
…ploys

Some LAVA device types support a recovery image which can be used to get the board into a known state, as well as being used for actual recovery this is sometimes used for switching between multiple firmwares in jobs. The various Arm reference platforms are one example of this.

Add an optional block in our base TFTP deploy template which allows a recovery image to be added to the job if recovery_image_url and recovery_deploy are defined in the parameters generated for the job.

### DIFF
--- a/config/lava/base/kernel-ci-base-tftp-deploy.jinja2
+++ b/config/lava/base/kernel-ci-base-tftp-deploy.jinja2
@@ -7,6 +7,15 @@
 {% endblock %}
 {% block deploy %}
 actions:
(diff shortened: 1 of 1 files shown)


broonie:
This is for internal devices we have in our lab at Arm, I don't have a public device this could actually be tested on unfortunately.

nuclearcat (approved):
I guess it is considered as tested , as it doesn't break any existing devices then.
LGTM

PR #1516, opened 2022-11-17
config/core/test-configs: remove smc test from qemu

aliceinwire:
Remove smc tests on qemu, as it can create false positive.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -2731,7 +2731,6 @@ test_configs:
   - device_type: qemu_arm-virt-gicv3
     test_plans:
       - baseline_qemu
-      - smc_qemu
       - v4l2-compliance-vivid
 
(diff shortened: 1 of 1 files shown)


aliceinwire:
@fbezdeka by the way all cip related issues are tracked by https://github.com/orgs/kernelci/projects/11

fbezdeka:
Can we mention someone who is able to do reviews?

> @fbezdeka by the way all CIP related issues are tracked by https://github.com/orgs/kernelci/projects/11 as per CIP instance documentation https://kernelci.org/docs/instances/cip/

Thanks!

aliceinwire:
@nuclearcat can you review and merge this?

nuclearcat (approved):
Usually i prefer to test on staging, but not sure it is easy with CIP and test removal should be safe, so LGTM

Issue #1508, opened 2022-11-11
Remove armv3 config (rpc_defconfig) from gcc builds


nuclearcat:
According https://www.phoronix.com/news/GCC-9-Dropping-Older-ARM gcc dont have support for armv3 for quite long time.
```
rpc_defconfig (arm, gcc-10) — FAIL, 2 errors, 0 warnings, 0 section mismatches

Errors:
    arm-linux-gnueabihf-gcc: error: unrecognized -march target: armv3
    arm-linux-gnueabihf-gcc: error: missing argument to ‘-march=’
```
It seems rpc_defconfig should be removed from testing by recent versions of gcc

nuclearcat:
Tested, No more failed builds (from email report):
kernelci/staging-mainline build: 50 builds: 0 failed, 50 passed, 50 warnings (staging-mainline-20221130.0)

PR #1517, opened 2022-11-17
build-configs.yaml: Disable ARMv3 platforms from gcc-10

nuclearcat:
ARMv3 was removed in gcc-9, so there is no point trying to compile this platforms with gcc.
Fixes https://github.com/kernelci/kernelci-core/issues/1508


nuclearcat:
Tested, No more failed builds (from email report):
kernelci/staging-mainline build: 50 builds: 0 failed, 50 passed, 50 warnings (staging-mainline-20221130.0)

PR #1527, opened 2022-11-23
baseline.jinja2: Add failure_retry to test

nuclearcat:
We can try to use retry option in case of serial corruption, to improve reliability of jobs on devices with slow or unreliable serial console. It should decrease rate of false positive on baseline test failures.
For example failed test:

```
 1455 17:17:20.117173  + KERNELCI_LAVA=y /bin/sh /opt/kernelci/dmesg.sh
 1456 17:17:20.142263  <8>[   17.024549] <LAVA_SIGNAL_TESTCASE TEST_CASE_ID=crit RESULT=pass UNITS=lines MEASUREMENT=0>
 1457 17:17:20.142531  Received signal: <TESTCASE> TEST_CASE_ID=crit RESULT=pass UNITS=lines MEASUREMENT=0
 1458 17:17:20.142631  Unknown test uuid. The STARTRUN signal for this test action was not received correctly.
 1460 17:17:20.142813  end: 3.1 lava-test-shell (duration 00:00:00) [common]
 1462 17:17:20.143021  lava-test-retry failed: 1 of 1 attempts. 'Invalid TESTCASE signal'
```

### DIFF
--- a/config/lava/baseline/baseline.jinja2
+++ b/config/lava/baseline/baseline.jinja2
@@ -4,6 +4,7 @@
 {%- endif %}
     timeout:
       minutes: 1
+    failure_retry: 5
     definitions:
     - repository:
         metadata:
@@ -30,6 +31,7 @@
 {%- endif %}
     timeout:
       minutes: 1
+    failure_retry: 5
     definitions:
     - repository:
         metadata:
--- a/config/lava/cros-ec/cros-ec.jinja2
+++ b/config/lava/cros-ec/cros-ec.jinja2
@@ -1,6 +1,7 @@
 - test:
     timeout:
       minutes: 5
+    failure_retry: 5
     definitions:
     - repository:
         metadata:


nuclearcat:
It does improve issues with broken serial consoles on baseline tests (at least no false regression registered in this case), but results are complicated:
Example: 
https://storage.staging.kernelci.org/kernelci/staging-mainline/staging-mainline-20221124.0/arm64/defconfig+arm64-chromebook+crypto/gcc-10/lab-collabora/baseline-sc7180-trogdor-lazor-limozeen.html

Job on LAVA shows that some of tests are passing: https://lava.collabora.dev/scheduler/job/8113238

But test **baseline.bootrr.tpm-chip-is-online** got corrupted and might need further work.
https://staging.kernelci.org/test/plan/id/637fae438550e3bb04ddf767/

nuclearcat:
It seems also cros-ec and other tests might need similar fix:
https://staging.kernelci.org/test/case/id/638397c467a11d0a3bddf745/
https://storage.staging.kernelci.org/kernelci/staging-mainline/staging-mainline-20221127.0/arm64/defconfig+arm64-chromebook/gcc-10/lab-collabora/cros-ec-sc7180-trogdor-lazor-limozeen.html

nuclearcat:
Example of job where adding retry helped:
https://lava.collabora.dev/scheduler/job/8182755

gctucker:
The `dmesg` tests should take less than 100ms so adding up to maybe 1s when retrying 5 times should be OK.  The bootrr tests can take quite a bit longer, maybe 5s so that could hit timeout.  Still, it's probably better to retry as there isn't much to lose by hitting timeout if the results were not parsed correctly anyway.  And if retrying until timeout is not enough then I'd argue that's a good case for investigating deeper and fixing some more fundamental problem.  We just need to make sure we have a reliable way of collecting cases like this without crawling through LAVA jobs by hand etc.

Issue #1528, opened 2022-11-23
Failed to load YAML configurations from file


JenySadadia:
I encountered the below issue when I was testing `created_user` utility with `kci_data` command:

Error logs:
```
$ ./kci_data --yaml-config /home/jeny/kernelCI/kernelci-pipeline-env/src/kernelci-pipeline/config/pipeline.yaml create_user --username jeny --password jeny --db-config docker-host --verbose
Traceback (most recent call last):
  File "./kci_data", line 116, in <module>
    configs = kernelci.config.load(opts.yaml_config)
  File "/home/jeny/kernelCI/kernelci-core-env/src/core-main/kernelci-core/kernelci/config/__init__.py", line 110, in load
    data = load_yaml(config_path)
  File "/home/jeny/kernelCI/kernelci-core-env/src/core-main/kernelci-core/kernelci/config/__init__.py", line 77, in load_yaml
    config[k] = v
NameError: name 'v' is not defined
```
```
$ ./kci_data --yaml-config /home/jeny/kernelCI/kernelci-pipeline-env/src/kernelci-pipeline/config/pipeline.yaml list_configs
Traceback (most recent call last):
  File "./kci_data", line 116, in <module>
    configs = kernelci.config.load(opts.yaml_config)
  File "/home/jeny/kernelCI/kernelci-core-env/src/core-main/kernelci-core/kernelci/config/__init__.py", line 110, in load
    data = load_yaml(config_path)
  File "/home/jeny/kernelCI/kernelci-core-env/src/core-main/kernelci-core/kernelci/config/__init__.py", line 77, in load_yaml
    config[k] = v
NameError: name 'v' is not defined
```

PR #1529, opened 2022-11-23
T39140 kernelci/config: fix YAML configuration load failure

JenySadadia:
Fixes https://github.com/kernelci/kernelci-core/issues/1528

Need to fix "NameError: name 'v' is not defined" from `load_yaml` function when try to load YAML configurations from a file.

Fixes: 4f9c0db2 ("kernelci.config: add _iterate_yaml_files() to simplify code")

### DIFF
--- a/kernelci/config/__init__.py
+++ b/kernelci/config/__init__.py
@@ -74,7 +74,7 @@ def load_yaml(config_path, validate_entries=None):
             elif hasattr(config_value, 'extend'):
                 config_value.extend(value)
             else:
-                config[k] = v
+                config[name] = value
     return config
 
 


mgalka:
I think you can get rid of "Need to" in the commit message i.e.

```
Fix  "NameError: name 'v' is not defined"
from `load_yaml` function when trying to load YAML
configurations from a file.
[...]
```

JenySadadia:
> Nice catch! I think it might have slipped during the refactoring.
> 
> I guess you can get rid of "Need to" in the commit message i.e.
> 
> ```
> Fix  "NameError: name 'v' is not defined"
> from `load_yaml` function when trying to load YAML
> configurations from a file.
> [...]
> ```

Yes, I fixed the commit message.
Thanks.

mgalka (approved):
LGTM

PR #1530, opened 2022-11-23
Add docker image versioning

mgalka:
**kci_docker: Enable adding version to image tag**

Introduce `--image-version/-V` to specify image version which will be added at the end of the image tag.
    
**setup.py: Add extras_require with development dependencies**
    
Define extras_require with development packages. It's now possible to use pip to install kernelci package in a regular or development version.

E.g.
```    
pip install kernelci  # install regular version
    
pip install kernelci[dev]  # install development versio
```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -80,5 +80,10 @@ setuptools.setup(
         "pytest",
         "pyyaml",
         "requests",
-    ]
+    ],
+    extras_require={
+        'dev': [
+            'pycodestyle'
+        ]
+    }
 )
--- a/config/docker/kci_docker
+++ b/config/docker/kci_docker
@@ -74,7 +74,10 @@ def _dump_push_log(log):
 def main(args):
     base_name = args.prefix + args.image
     tag_strings = ([args.arch] if args.arch else []) + (args.fragment or [])
+    if args.image_version:
+        tag_strings.append(args.image_version)
     tag_name = '-'.join(tag_strings)
+
     name = ':'.join((base_name, tag_name)) if tag_name else base_name
     if args.command == 'name':
         print(name)
@@ -123,6 +126,8 @@ if __name__ == '__main__':
                         help="Docker image name, e.g. gcc-10:x86")
     parser.add_argument('--prefix', default='kernelci/',
                         help="Docker image tag prefix")
+    parser.add_argument('--image-version', '-V',
+                        help='Docker image version tag e.g. 20221011.0')
     parser.add_argument('--arch',
                         help="CPU architecture, e.g. x86")
     parser.add_argument('--fragment', action='append',


mgalka:
Tested OK locally:

1. Build package with specified version 

```
$ ./kci_docker build --verbose --no-cache --build-arg core_rev=mg-sandbox --prefix "mgalka/" -V 20221123.0 --build-arg core_url=https://github.com/mgalka/kernelci-core kernelci

Building mgalka/kernelci:20221123.0
Successfully tagged mgalka/kernelci:20221123.0
...
```

2. Check version of kernelci package

```
$ pip freeze | grep kernelci
kernelci @ git+https://github.com/mgalka/kernelci-core@003d885e5b5e94f4498034c3e4fee7d424c7c0d6
```

OK - `003d885e5b5e94f4498034c3e4fee7d424c7c0d6` is the commit id of the HEAD of [mg-sandbox branch](https://github.com/kernelci/kernelci-core/commit/003d885e5b5e94f4498034c3e4fee7d424c7c0d6)

3. Check if  `pycodestyle` is present to make sure the `kernelci[dev]` package was installed

```
$ pycodestyle --version
2.8.0
```

OK - version is the same as in `extras_require` parameter in `setup.py` 

```python 
    extras_require={
        'dev': [
            'pycodestyle==2.8.0'
        ]
    }
```

nuclearcat (approved):
Perfect! Thank you

config/docker/fragment/kernelci.jinja2
```diff
@@ -8,12 +8,13 @@ RUN pip3 install --upgrade pip
 ARG core_url=https://github.com/kernelci/kernelci-core.git
 ARG core_rev=main
 RUN apt-get update && apt-get install --no-install-recommends -y git
+RUN pip3 install git+$core_url@$core_rev#egg=kernelci[dev]
```
gctucker, line 11: What's the reason for doing this?  One issue here is that the kernelci-core git repository gets cloned twice which may introduce some discrepancy one day between the package installed and the config files.  Please explain in the commit message what difference it makes to use pip3 rather than setup.py.

setup.py
```diff
@@ -79,5 +79,10 @@ def _list_files(path, match):
         "pytest",
         "pyyaml",
         "requests",
-    ]
+    ],
+    extras_require={
+        'dev': [
+            'pycodestyle==2.8.0'
```
gctucker, line 85: I believe the convention is to not specify package versions in setup.py but to do it in requirements.txt instead.

config/docker/fragment/kernelci.jinja2
mgalka: I added kernelci[dev] package definition in setup.py to reflect what is defined in the `requirements.txt` and `requirements-dev.txt`. Using `pip` allows us also to specify local, remote, repo URLs or package index to install the package as well as specify what `extras_require` sections will be installed. 

While I agree that cloning the repository may lead to some discrepancy, I think that the problem is a bit more broad. In general lack of proper versioning leads to a situation in which we cannot be sure if the version of the kernelci package installed by the kernelci fragment is consistent between 2 docker images build one after another. This actually applies to any situation where kernelci-core repository is cloned more than once and the revision is specified as `main`.

At the moment I changed the kernelci fragment definition to clone the repository just once, but I'd treat it as a temporary solution before the docker images versioning is introduced and configuration and application code are kept in separate repositories.

setup.py
mgalka: I guess that limiting this version to a single one is too tight, I removed the version.

Having said that it's not exactly that specifying package version in setup.py is an anti-pattern. The setup.py should specify a minimal set of requirements to run the application, so should it be running with any version of the dependency there should be no version, but it's perfectly ok to specify a version or a range of versions for a package when e.g. API changed. The dependencies/requirements definition in setup.py doesn't also specify a source of the packages.

gctucker: I mean, all the other packages don't have any version specified.  So if we want to maintain versions both in requirements.txt and setup.py then we should add version information for all the other packages too in setup.py and always ensure they're kept in sync with requirements.txt.  That seems like some overhead to me for not much added value.

config/docker/fragment/kernelci.jinja2
```diff
@@ -12,8 +12,7 @@ RUN git clone --depth=1 $core_url /tmp/kernelci-core
 WORKDIR /tmp/kernelci-core
 RUN git fetch origin $core_rev
 RUN git checkout FETCH_HEAD
-RUN pip3 install -r requirements-dev.txt
-RUN python3 setup.py install
+RUN pip3 install .[dev]
```
gctucker, line 15: Is this equivalent to `pip3 install -r requirements-dev.txt`?  If so, I think it's a bit less explicit and I'm not sure of the advantage.

Also I don't get why the `setup.py install` command was removed, but then I don't remember why it was added either.  Could you please shed some light on this?

setup.py
mgalka: setup.py and requirements.txt serve for a slightly different purposes. It's quite well described [here](https://packaging.python.org/en/latest/discussions/install-requires-vs-requirements/). I think it's better to keep setup.py without version sepcified or with minimum/maximum version specified only when other can break the application.

config/docker/fragment/kernelci.jinja2
gctucker: The commit message says:

> Make use of pip instead of calling setup.py directly.

but it doesn't give the reason for doing this.  Is it fixing or improving anything?

mgalka: It's not the equivalent  to `pip3 install -r requirements-dev.txt`. It's rather the equivalent of

```
pip3 install -r requirements-dev.txt
python3 setup.py install
```

I actually find using `setup.py` and `requrements-dev.txt` a bit confusing. 
Why should we install the application with `setup.py` and some additional packages from `requirements-dev.txt` when there is a mechanism for specifying required and extra packages in `setup.py`?

gctucker: I think the key issue here was about installing the pinned-down package versions in requirements.txt.  Now I'm not sure if we should drop requirements.txt and specify the versions in setup.py, although I believe it's really the norm to only specify pinned down versions in requirements.txt.

mgalka: setup.py and requirements.txt serve a bit different purpose. The setup.py file is more like a package recipe in Linux distributions. It doesn't necessarily need to specify the exact version of the dependencies if it's not absolutely needed to have a specific version to have package up and running.  This is why it's more common to have no version specified (getting the newest available)  or just specify the minimum and/or maximum version of a dependency. 
Basically running `setup.py install` should give us similar effects as running e.g. `apt install` for Debian packages.

So, if kernelci is supposed to run with any version of the packages listed in `install_requires` then there is no need to specify those.

requirements.txt is more of an "environment snapshot", which is why there are usually listed specific versions of the packages.

In my opinion we should use setup.py/pip  to install the kernelci package which should result in having the kernelci package installed in the container together with all its dependencies required for proper run.

mgalka:
@gctucker is it ok to merge it?

config/docker/fragment/kernelci.jinja2
gctucker: I think we agree on this point as it's the established norm, the versions should be in requirements.txt as a snapshot of a working set of dependencies.

Then it's clear that installing requirements.txt with pip installs the versioned packages from the file.  And it's also clear that we need to use `setup.py` to install the kernelci-core package from the git clone.  What's not so clear is what `setup.py` will do with the dependencies since they're already installed using `pip`, and my guess is that it doesn't do anything since they're already installed.

Therefore, doing `pip install -r requirements-dev.txt` for the dependencies and then `python3 setup.py install` for kernelci-core makes complete sense in this case to me.  What's the reason for changing this again?

gctucker: If `pip3 install .[dev]` is really equivalent to doing the 2-step install with `pip` and then `setup.py`, then it's not a functional change but just a syntaxic thing.  I'd rather keep the explicit commands unless the shorter one is really more widespread but I'm not that familiar with it.  But please confirm whether it's actually equivalent or if it does anything different.

gctucker: Having versioned packages means we're in control of when packages get updated, which is very important for the production images.  If we want to depart from that then we can discuss it but it's a larger topic so I guess it shouldn't be part of this PR which is about adding Docker version tags.

Issue #1532, opened 2022-11-25
debos initrams fails, if nfs mount required, but no network devices are ready


nuclearcat:
At current configuration for tests that use NFS debian initramfs scripts doesn't really wait network devices to be detected, unless device name specified in ip= parameter.
As result it might lead to false regressions like here:
```
Loading, please wait...
Starting version 247.3-7+deb11u1
[   15.209757] udevadm (1174) used greatest stack depth: 14000 bytes left
[   15.349713] udevadm (1175) used greatest stack depth: 13992 bytes left
Begin: Loading essential drivers ... done.
Begin: Running /scripts/init-premount ... done.
Begin: Mounting root file system ... Begin: Running /scripts/nfs-top ... done.
Begin: Running /scripts/nfs-premount ... done.
ipconfig: no devices to configure
ipconfig: no devices to configure
ipconfig: no devices to configure
ipconfig: no devices to configure
ipconfig: no devices to configure
ipconfig: no devices to configure

ipconfig: no devices to configure
ipconfig: no devices to configure
ipconfig: no devices to configure
ipconfig: no devices to configure
/init: 333: .: Can't open /run/net-*.conf
[   15.418675] Kernel panic - not syncing: Attempted to kill init! exitcode=0x00000200
[   15.418675] 
[   15.428899] CPU: 0 PID: 1 Comm: init Not tainted 4.4.302-cip70-st25 #1
[   15.436190] Hardware name: Google Coral/Coral, BIOS  09/29/2020
[   15.442795]  0000000000000000 ffff88017abdfe58 ffffffff8192e29f ffff88017abdff00
[   15.451072]  ffffffff81c78e08 ffff88017abdfed0 ffffffff81920a5a ffff880100000010
[   15.459347]  ffff88017abdfee0 ffff88017abdfe80 ffff88017abf8000 0000000000000200
[   15.467627] Call Trace:
[   15.470361]  [<ffffffff8192e29f>] dump_stack+0x57/0x6d
[   15.476100]  [<ffffffff81920a5a>] panic+0xbb/0x1ff
[   15.481450]  [<ffffffff81920d94>] do_exit.cold+0x1e/0xb0
[   15.487379]  [<ffffffff81046e5a>] ? __do_page_fault+0x17a/0x3f0
[   15.493996]  [<ffffffff81056df2>] SyS_exit+0x12/0x20
[...]

PR #1533, opened 2022-11-25
create_initrd_ramdisk.sh: Add script for ethernet device wait

nuclearcat:
If initramfs expects nfs we need to have also ethernet device to be available. In some cases network device are really slow to detect and it cases panic on ipconfig stage
Fixes https://github.com/kernelci/kernelci-core/issues/1532

### DIFF
--- a/config/rootfs/debos/scripts/create_initrd_ramdisk.sh
+++ b/config/rootfs/debos/scripts/create_initrd_ramdisk.sh
@@ -20,6 +20,35 @@ index 4ec926c..ca06c0b 100644
  	# Load ubi with the correct MTD partition and return since fstype
 EOF
 
+patch -p1 << 'EOF'
+--- a/usr/share/initramfs-tools/scripts/nfs-premount/wait_ethernet	1970-01-01 02:00:00.000000000 +0200
++++ b/usr/share/initramfs-tools/scripts/nfs-premount/wait_ethernet	2022-11-25 12:38:46.037498198 +0200
+@@ -0,0 +1,21 @@
++#!/bin/sh
++. /scripts/functions
++
++wait_ethernet() {
++  local netdevwait=60
++  echo "Waiting up to ${netdevwait} secs for any ethernet to become available"
++  while [ "$(time_elapsed)" -lt "$netdevwait" ]; do
++    for device in /sys/class/net/* ; do
++      if [ -f "$device/type" ]; then
++        current_type=$(cat "$device/type")
++        if [ "${current_type}" = "1" ]; then
++          echo "Device ${device} found"
++          return
++        fi
++      fi
++    done
++    sleep 1
++  done
++}
++
++wait_ethernet
+EOF
+
+chmod +x /usr/share/initramfs-tools/scripts/nfs-premount/wait_ethernet
+
 KVER=min
 
 # update-initramfs uses kernel config to decide how to compress ramdisk


gctucker:
Are we sure this won't be hiding some kernel problems?  Why is the ethernet interface so slow to show up?  If it's really only specific to some random USB - Ethernet adapters then that's fine, but if there's a kernel bug then we have a problem and it should keep panicking.

nuclearcat:
Some hardware can be really slow to detect and we should expect that.
Especially as USB host itself and adapter can be modules, so first udev will load usb host drivers, then usb host will see devices connected and kick udev to load network adapter module, and then network adapter will appear.
If device doesn't appear at all - it will eventually timeout and panic.

config/rootfs/debos/scripts/create_initrd_ramdisk.sh
```diff
@@ -20,6 +20,35 @@ index 4ec926c..ca06c0b 100644
...
++  local netdevwait=60
++  echo "Waiting up to ${netdevwait} secs for any ethernet to become available"
++  while [ "$(time_elapsed)" -lt "$netdevwait" ]; do
++    for device in /sys/class/net/* ; do
++      if [ -f "$device/type" ]; then
++	current_type=$(cat "$device/type")
++	if [ "${current_type}" = "1" ]; then
++            echo "Device ${device} found"
++	    return
++	fi
```
gctucker, line 40: Indentation problem

config/rootfs/debos/scripts/create_initrd_ramdisk.sh
```diff
@@ -20,6 +20,35 @@ index 4ec926c..ca06c0b 100644
...
 
+patch -p1 << 'EOF'
+--- a/usr/share/initramfs-tools/scripts/nfs-premount/wait_ethernet	1970-01-01 02:00:00.000000000 +0200
++++ b/usr/share/initramfs-tools/scripts/nfs-premount/wait_ethernet	2022-11-25 12:38:46.037498198 +0200
+@@ -0,0 +1,21 @@
++#!/bin/sh
++. /scripts/functions
++
++wait_ethernet() {
++  local netdevwait=60
```
gctucker, line 31: I believe `local` is a Bashism, to run with any `sh` it should be dropped.

config/rootfs/debos/scripts/create_initrd_ramdisk.sh
```diff
@@ -20,6 +20,35 @@ index 4ec926c..ca06c0b 100644
...
++  echo "Waiting up to ${netdevwait} secs for any ethernet to become available"
++  while [ "$(time_elapsed)" -lt "$netdevwait" ]; do
```
gctucker, line 33: Does `time_elapsed()` get implicitly reset when the script is loaded?  Not sure how it's keeping track of the start time.

config/rootfs/debos/scripts/create_initrd_ramdisk.sh
nuclearcat: I'm just following style of initramfs tools scripts, i think they expect bash and expect this variable to be local
https://salsa.debian.org/kernel-team/initramfs-tools/-/blob/master/scripts/functions#L325

config/rootfs/debos/scripts/create_initrd_ramdisk.sh
nuclearcat: starttime in time_elapsed() is counted from init script start, so it is limit on total delay, which is really convenient to not exceed lava action timeout.
This is where starttime is set:
https://salsa.debian.org/kernel-team/initramfs-tools/-/blob/master/init#L227
then time_elapsed function:
https://salsa.debian.org/kernel-team/initramfs-tools/-/blob/master/scripts/functions#L154

config/rootfs/debos/scripts/create_initrd_ramdisk.sh
nuclearcat: Fixed!

nuclearcat:
Example: https://storage.staging.kernelci.org/kernelci/staging-next/staging-next-20221214.0/x86_64/x86_64_defconfig+x86-chromebook+ima/gcc-10/lab-collabora/baseline-nfs-asus-C436FA-Flip-hatch.html
```
 2850 14:17:58.727934  Begin: Running /scripts/nfs-premount ... Waiting up to 60 secs for any ethernet to become available
 2852 14:17:58.734829  Device /sys/class/net/enx0024325019be found
[ 2853 14:17:58.735317  done.
```

mgalka (approved):
Feel free to merge it

PR #1534, opened 2022-11-26
test-configs.yaml: Enable kselftest-tpm2 on sona

nuclearcat:
The TPM chip here is the same as the one in grunt, the cr50, but on this platform the chip is accessible through a SPI bus instead of an i2c one, thus requiring different drivers.

So while the TPM operation is identical to grunt, adding the test to this platform gets us a second x86 platform running the test, as well as ensuring that the SPI driver for the cr50 works correctly.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -2273,6 +2273,7 @@ test_configs:
       - kselftest-alsa
       - kselftest-cpufreq
       - kselftest-landlock
+      - kselftest-tpm2
 
   - device_type: hp-x360-12b-ca0500na-n4000-octopus
     test_plans:


nuclearcat:
This PR identical to https://github.com/kernelci/kernelci-core/pull/1034 and author is Nícolas F. R. A. Prado <nfraprado@collabora.com> , just rebasing it and trying to test it over weekend to merge at monday, to reduce PR backlog.

nuclearcat:
kselftest-tpm2 is working and giving results on dashboard: https://staging.kernelci.org/test/plan/id/6381d4377d91c25782ddf7ec/
The only interesting issue that only one test passes, and others giving error. It might be related to Chromebooks tricky TPM ownership.
@nfraprado whats your opinion about test results? Is they are anyhow useful?

nuclearcat:
This PR will need further discussion with hardware vendors, as seems tpm statefulness is a problem on some platforms.

PR #1536, opened 2022-11-28
 T39237 db/kernelci_api: drop unused method related to regression

JenySadadia:
Drop `get_regressions_by_node_id` method as it is
not needed anymore for regression tracking.


nuclearcat (approved):
LGTM

PR #1537, opened 2022-11-28
Force bringup of network by default

montjoie:
For more testing, we need to force network to be up. This will permit to detect more breakage.

The easiest to achieve this is to set ip=dhcp on cmdline.

### DIFF
--- a/config/lava/base/kernel-ci-base.jinja2
+++ b/config/lava/base/kernel-ci-base.jinja2
@@ -1,4 +1,4 @@
-{% set extra_kernel_args = "console_msg_format=syslog earlycon" %}
+{% set extra_kernel_args = "console_msg_format=syslog earlycon ip=dhcp" %}
 
 {%- block metadata %}
 metadata:


gctucker:
I think this might conflict with other `ip=` arguments defined at other places.  Have you double checked all use-cases?

nuclearcat:
It was helpful to test https://github.com/kernelci/kernelci-core/pull/1473, but i believe such things need to be fixed in LAVA dictionary, because it might break at least minnowboard, odroid, exynos at Collabora, as they have defined at dictionary ip=:::::eth0:dhcp .

gctucker:
> It was helpful to test #1473, but i believe such things need to be fixed in LAVA dictionary, because it might break at least minnowboard, odroid, exynos at Collabora, as they have defined at dictionary ip=:::::eth0:dhcp .

Exactly, so I don't think this should be merged.

montjoie:
Forcing the network should not LAVA side, but it is probably too late to enforce this.
What about a "network enable flag" per board in kci config ?
I think really that all board should start their network by default, but only if kcijobgenerator decided to do this.

nuclearcat:
Thats possible, is anybody going to implement it in this PR or we open issue?

gctucker:
We shouldn't be doing this as some tests don't require the network interface to be up, typically baseline and all the ones running with ramdisks.  By requiring the network interface, we would be making these tests more fragile and not really add any value to them as there are already tests to cover network functionality (baseline-nfs to start with...).  Also, hard-coding `ip=dhcp` is not always going to work for the reasons mentioned above (when test interfaces also exist etc.).

Then as there hasn't been any follow-up, closing this now.

PR #1545, opened 2022-12-01
kci_test: Add missing return value of cmd_update_rootfs_urls

mgalka:
Missing return value from the __call__ method of cmd_update_rootfs_urls class was causing kci_test update_rootfs_urls command exiting with 1 instead of 0.
This commit adds return value for the method.


nuclearcat (approved):
LGTM, rootfs builds still working fine :)

PR #1546, opened 2022-12-01
Revert "config/docker: Add missing yaml/requests python dependency"

mgalka:
This reverts commit 8d79b262ab977655a69734cf450bc8e2ac73ce5e.

kernelci package which includes the kci_* command line tools is now part of the kernelci fragment added on top of the base debos image and pushed with the "kernelci" tag to the docker hub. PyYAML and requests packages are no longer needed as a part of the base debos image.

To use kci_* tools from the debos docker image use kernelci/debos:kernelci image instead of the vanilla one.


gctucker:
Merged now to unblock today's production update

PR #1548, opened 2022-12-02
CHROMEOS build_board.sh: Rewrite if to switch and add trogdor artifacts

nuclearcat:
As we got quite long list of "if", we can convert them to more elegant switch/case.
Also added extraction of dtb artifacts for arm64 trogdor.

### DIFF
--- a/config/rootfs/chromiumos/scripts/build_board.sh
+++ b/config/rootfs/chromiumos/scripts/build_board.sh
@@ -51,58 +51,44 @@ echo "Board ${BOARD} setup"
 # Future possible option --profile=x, for example kernel-5_15, profiles are at /mnt/host/source/src/overlays/overlay-${BOARD}/profiles/
 cros_sdk setup_board --board=${BOARD}
 
-# Certain boards need temporary fixes/patches to build successully
-if [ "${BOARD}" == "coral" ]; then
-echo "Patching coral specific issues"
-sed ':a;N;$!ba;s/DEPEND="\n\tchromeos-base\/fibocom-firmware\n"/# DEPEND="\n\t# chromeos-base\/fibocom-firmware\n# "/g' -i src/overlays/overlay-coral/chromeos-base/modemfwd-helpers/modemfwd-helpers-0.0.1.ebuild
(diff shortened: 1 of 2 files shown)


config/rootfs/chromiumos/scripts/build_board.sh
```diff
@@ -51,58 +51,44 @@ echo "Board ${BOARD} setup"
...
+    zork)
+    echo 'USE="${USE} -tpm tpm2"' >>src/overlays/overlay-zork/profiles/base/make.defaults
+    ;;
+    grunt)
+    echo 'USE="${USE} -tpm tpm2"' >>src/overlays/baseboard-grunt/profiles/base/make.defaults
+    ;;
+    *)
+    echo "No issues found for this board"
+    ;;
+esac
```
mgalka, line 91: I like the idea of having `switch/case` instead of  the set of ifs. I thought that maybe we could split these into separate files named after the boards and have them run by the `build_board.sh`. 

```
|-build_board.d/
              |_ grunt.sh
              |_ nami.sh
              |_ ocopus.sh
```


On one hand it may be even more clean as we're having here 9 different cases to handle, on the other hand handlers are usually one liners. @nuclearcat What do you think of that?

mgalka:
Why do we have this commit https://github.com/kernelci/kernelci-core/pull/1548/commits/9db4673d8c32ef4dcf54e7f0d0c3a1155f3bc702 in this PR?

config/rootfs/chromiumos/scripts/build_board.sh
nuclearcat: I thought about that, but most of scripts likely will be one-liners, also i might replace those lines by some function for better understanding of similar patches like adding tpm flags.
So it will be something like:
```
volteer)
zork)
grunt)
    add_tpm2_use_flags()
    ;;
```

But probably i will use shell scripts with board name where significant patching required (likely will happen with other arm64 chromebooks and panfrost/mesa).

nuclearcat:
Yes, built rootfs and in progress of testing it now, likely will write report about testing results and completion tomorrow.

nuclearcat:
Flashed trogdors in lab, got successful test results: https://chromeos.kernelci.org/test/plan/id/6399f23ed70514180924cc5c/
https://storage.chromeos.kernelci.org/images/rootfs/chromeos/chromiumos-trogdor/20221212.0/arm64/

mgalka (approved):

PR #1549, opened 2022-12-02
build-configs: add rust next trees

10ne1:
commit 58f7ecbe7574fb1071ac8c92ec645c7ff3402f41                                                                                           
Author: Adrian Ratiu <adrian.ratiu@collabora.com>
Date:   Fri Dec 2 16:03:03 2022 +0200

    build-configs: add rust next trees
    
    This adds a rust build to the linux-next integration tree as well
    as the maintainer 'rust-next' tree.
    
    Suggested-by: Guillaume Tucker <guillaume.tucker@collabora.com>

### DIFF
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -921,6 +921,13 @@ build_configs:
         architectures:
           <<: *arch_clang_configs
 
+      rustc-1.62:
+        build_environment: rustc-1.62
+        fragments: [rust, rust-samples, kselftest]
+        architectures:
+          x86_64:
+            base_defconfig: 'x86_64_defconfig'
+
   next_pending-fixes:
     tree: next
     branch: 'pending-fixes'
@@ -1029,6 +1036,17 @@ build_configs:
       rustc-1.62:
         build_environment: rustc-1.62
         fragments: [rust, rust-for-linux-samples, kselftest]
+        architectures:
+          x86_64:
+            base_defconfig: 'x86_64_defconfig'
+
+  rust-for-linux_rust-next:
+    tree: rust-for-linux
+    branch: 'rust-next'
+    variants:
+      rustc-1.62:
+        build_environment: rustc-1.62
+        fragments: [rust, rust-samples, kselftest]
         architectures:
           x86_64:
             base_defconfig: 'x86_64_defconfig'
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -1029,7 +1029,7 @@ build_configs:
     branch: 'v5.15-rt'
     variants: *preempt_rt_variants
 
-  rust-for-linux:
+  rust-for-linux_rust:
     tree: rust-for-linux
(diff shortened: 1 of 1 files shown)


ojeda:
> Would you like to be added the github kernelci/hackers group? Thanks!

Sure -- or just ping me with the @

This looks fine to me, by the way, thanks! I would maybe put it below the `'rust'` one, but not sure how they are supposed to be sorted in KernelCI:

gctucker:
> Hi @ojeda It was suggested by @gctucker we add a rust build variant directly to the linux-next integration tree config, instead of adding a separate config for the `rust-next` branch because that gets regularly merged into linux-next and will get regularly tested anyway. I've updated the commit to reflect that.
> 
> This is only for the next tree, the maintainer tree (currently the `rust` branch, it can be changed in the future) is still separate.

No, what I meant is that there shouldn't be another build config monitoring the linux-next master branch just for Rust builds, because that will collide with the regular build config entry for linux-next master branch.  Some builds might overwrite each other etc. as the data is stored for each git tree name and branch name, the build config name used in YAML is not representative.

So a `rust-next` branch in the Rust tree does really seem worth having as a separate build config since it'll catch issues before everything gets merged in linux-next.

gctucker:
I saw the issue mentioned above with mainline and @10ne1 has sent a PR to fix it now: #1560.

So to be clear, I think we can add a similar extra block to the existing `next` build config as well as a new build config for the `rust-next` branch from the Rust tree, which should have fewer but more focused builds for Rust coverage.

gctucker:
> > Would you like to be added the github kernelci/hackers group? Thanks!
> 
> Sure -- or just ping me with the @

Invite sent ;)

ojeda:
> Invite sent ;)

Thanks! I think it worked, e.g. I just got one review request for the TSC PR.

aliceinwire:
@10ne1 please rebase this

config/core/build-configs.yaml
```diff
@@ -1033,6 +1040,17 @@ build_configs:
           x86_64:
             base_defconfig: 'x86_64_defconfig'
 
+  rust-next:
```
gctucker, line 1043: BTW the implicit convention for the build config names is to use the `tree_branch` format, so in this case it should be `rust-for-linux_rust-next`.  That's no big deal as these names aren't sent to the API but they may be used in some command line arguments.  Also this isn't really documented, it's just how they've been named until now to make it easier to manage the file.  I don't know if you want to go through the effort of making another PR to tweak that, if so it's better to do it now than later before external dependencies start appearing.

aliceinwire (approved):

gctucker (approved):

ojeda:
LGTM (not sure if I should assign myself to reviews?)

By the way, next week Rust 1.66.0 will be released, and we may end up migrating to it -- if so, it would be the first compiler minimum version upgrade we will have in mainline, so we can use it as a test.

PR #1553, opened 2022-12-05
build-configs.yaml: Block haps_hs_smp_defconfig builds from old kernels

nuclearcat:
Right now attempt to build this config in kernel 4.9 and older causes error and noise in reports:
```
*** Can't find default configuration "arch/arc/configs/haps_hs_smp_defconfig"!
```
Add blocklist entry to not build this config in too old kernels.

### DIFF
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -468,6 +468,8 @@ build_configs_defaults:
                   - 'nsim_700_defconfig'
                   - 'nsimosci_defconfig'
                   - 'tb10x_defconfig'
+            # haps_hs_smp_defconfig not available in some old kernels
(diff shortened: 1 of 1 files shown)


nuclearcat:
I will update PR to explain why it works in 4.19 and doesnt in 4.9

nuclearcat:
https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/arch/arc/configs?h=linux-4.9.y
As we can see haps_hs_smp_defconfig is not available at this kernel, but available in 4.19:
https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/tree/arch/arc/configs/haps_hs_smp_defconfig?h=linux-4.19.y

mgalka (approved):
LGTM

PR #1554, opened 2022-12-05
Add support for jh7100-starfive-visionfive-v1

montjoie:
jh7100-starfive-visionfive-v1 is a RISCV board which is now supported and can be booted from mainline since a few days (but only from linux-next for today).
LAVA already have support for it https://git.lavasoftware.org/lava/lava/-/merge_requests/1878


nuclearcat (approved):
Board successfully passing baseline tests: https://storage.staging.kernelci.org/kernelci/staging-next/staging-next-20221206.0/riscv/defconfig/gcc-10/lab-clabbe/baseline-jh7100-starfive-visionfive-v1.html

PR #1557, opened 2022-12-05
Revert "build.py: Add printout of HTTP error code during last commit …

gctucker:
…retrieval"

This reverts commit 7a1fd0182ab73a476d72e750a3269d6e668975b5.

This commit introduced some error message which gets interpreted as the git sha1, causing the Jenkins monitor jobs to fail with new build configurations that don't have a tree on the production storage yet.

### DIFF
--- a/kernelci/build.py
+++ b/kernelci/build.py
@@ -87,8 +87,6 @@ def get_last_commit(config, storage):
         file_name=_get_last_commit_file_name(config))
     last_commit_resp = requests.get(last_commit_url)
     if last_commit_resp.status_code != 200:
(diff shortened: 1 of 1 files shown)


aliceinwire (approved):

gctucker:
BTW this fix is not in production yet, but that doesn't really matter as there aren't any new trees being added.

@nuclearcat The issue is not about git trees but the files stored on storage with the last git commit that was tested.  This logic is going away with the new API / pipeline as it'll be stored in the database (see how the `trigger.py` service works).  So I think we can just forget about this part.  Reporting errors when fetching the latest revision from the git tree is still relevant though, but it's not related to this issue.

PR #1560, opened 2022-12-06
Misc rust build fixes

10ne1:
commit 7d846c58ce8d82552372213a955bbe9c2fd309fa
Author: Adrian Ratiu <adrian.ratiu@collabora.com>
Date:   Tue Dec 6 15:15:02 2022 +0200

    build-configs: move rust builds under mainline cfg
    
    PR #1468 added the mainline rust build in a separate
    `rust` build configuration instead of adding to the
    already existing mainline config which can cause
    unnecessary builds. Move it in the proper place.
    
    While at it also change the x86_64_arch anchor to
    build just the base x86_64_defconfig to avoid
    unnecessary allmodconfig rebuilds.
    
    Fixes: 43875d0e47f2 ("config: add rust build configurations")
    Suggested-by: Guillaume Tucker <guillaume.tucker@collabora.com>
    Reviewed-by: Alice Ferrazzi <alice.ferrazzi@miraclelinux.com>

commit f5b5769d8de3e4047365fb4676d2c07a8c0ad009
Author: Adrian Ratiu <adrian.ratiu@collabora.com>
Date:   Tue Dec 6 14:43:35 2022 +0200

    config/docker: ensure rustc always uses cc = clang
    
    Even though the cargo rustc invocations got pointed
    to clang via config.toml, the kernel builds can still
    do some rustc invocations outside cargo which would
    fail because cc is not properly set.
    
    Fixes: 43875d0e47f2 ("config: add rust build configurations")
    Reviewed-by: Alice Ferrazzi <alice.ferrazzi@miraclelinux.com>

commit 937c587f1a154083adabd5d47318700250a13250
Author: Adrian Ratiu <adrian.ratiu@collabora.com>
Date:   Tue Dec 6 14:38:22 2022 +0200

    config/docker: add rustfmt to rust images
    
    While testing rust-enabled kernel builds it was noticed that
    some non-fatal errors are printed due to missing rustfmt
    because the build generates code and tries to format it:
    
      BINDGEN rust/bindings/bindings_generated.rs
      BINDGEN rust/bindings/bindings_helpers_generated.rs
    Failed to run rustfmt: No such file or directory (os error 2) (non-fatal, continuing)
      RUSTC L rust/core.o
    Failed to run rustfmt: No such file or directory (os error 2) (non-fatal, continuing)
    
    Silence these by installing rustfmt.
    
    Reviewed-by: Reviewed-by: Miguel Ojeda <ojeda@kernel.org>
    Reviewed-by: Alice Ferrazzi <alice.ferrazzi@miraclelinux.com>

### DIFF
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -834,6 +834,12 @@ build_configs:
       clang-15:
         build_environment: clang-15
         architectures: *arch_clang_configs
+      rustc-1.62:
+        build_environment: rustc-1.62
+        fragments: [rust, rust-samples, kselftest]
+        architectures:
+          x86_64:
+            base_defconfig: 'x86_64_defconfig'
 
   matthiasbgg-for-next:
     tree: matthiasbgg
@@ -1007,16 +1013,6 @@ build_configs:
     branch: 'v5.15-rt'
     variants: *preempt_rt_variants
 
-  rust:
-    tree: mainline
-    branch: 'master'
-    variants:
-      rustc-1.62:
-        build_environment: rustc-1.62
-        fragments: [rust, rust-samples, kselftest]
-        architectures:
-          x86_64: *x86_64_arch
-
   rust-for-linux:
     tree: rust-for-linux
     branch: 'rust'
@@ -1025,7 +1021,8 @@ build_configs:
         build_environment: rustc-1.62
         fragments: [rust, rust-for-linux-samples, kselftest]
         architectures:
-          x86_64: *x86_64_arch
+          x86_64:
+            base_defconfig: 'x86_64_defconfig'
 
   samsung:
     tree: samsung
--- a/config/docker/rustc-1.62.jinja2
+++ b/config/docker/rustc-1.62.jinja2
@@ -36,4 +36,8 @@ RUN git clone --recurse-submodules --branch $RUST_VER \
 
 RUN cargo install --locked --version ${BINDGEN_VER} bindgen
 
+# kernel build generates some rust code and tries to format it, if rustfmt is
+# missing it will print non-fatal error noise
+RUN cargo install rustfmt
+
 {%- endblock %}
--- a/config/docker/rustc-1.62.jinja2
+++ b/config/docker/rustc-1.62.jinja2
@@ -24,11 +24,11 @@ RUN tar -xf ${RUST_TRIPLE}.tar.gz  -C /tmp/ && \
     rm -rf /tmp/${RUST_TRIPLE} && \
     rm /${RUST_TRIPLE}*
 
-# This image is based on clang-*jinja2 which only has clang installed but rustc
-# defaults to linker "cc" which calls the non-existing GNU stack (gcc, libgcc).
-# So we point cargo/rustc to use the LLVM stack via the clang linker entrypoint.
-RUN mkdir -p ${CARGO_HOME} && \
-    printf '[build]\nrustflags = ["-C", "linker=clang"]' > ${CARGO_HOME}/config.toml
+# This image is based on clang-*jinja2 which only has clang installed, but rustc
+# defaults to linker "cc" which usually points to the GNU/GCC stack. Pointing cc
+# to clang ensures both cargo and kernel build rustc invocations use llvm/clang.
+RUN update-alternatives --install /usr/bin/cc cc $(which clang) 30 && \
+    update-alternatives --set cc $(which clang)
 
 RUN git clone --recurse-submodules --branch $RUST_VER \
         https://github.com/rust-lang/rust \


mgalka:
Docker image build went OK on staging.
Rust-for-linux builds completed successfully. 

- https://bot.staging.kernelci.org/job/kernel-build/95745/consoleFull (x86_64_defconfig+rust)
- https://bot.staging.kernelci.org/job/kernel-build/95751/consoleFull (x86_64_defconfig+kselftest)

gctucker:
It's a bit nicer to add `Fixes:` trailers when fixing an earlier commit but that's no big deal.

config/docker/rustc-1.62.jinja2
```diff
@@ -24,11 +24,10 @@ RUN tar -xf ${RUST_TRIPLE}.tar.gz  -C /tmp/ && \
...
 
-# This image is based on clang-*jinja2 which only has clang installed but rustc
-# defaults to linker "cc" which calls the non-existing GNU stack (gcc, libgcc).
-# So we point cargo/rustc to use the LLVM stack via the clang linker entrypoint.
-RUN mkdir -p ${CARGO_HOME} && \
-    printf '[build]\nrustflags = ["-C", "linker=clang"]' > ${CARGO_HOME}/config.toml
+# This image is based on clang-*jinja2 which only has clang installed, but rustc
+# defaults to linker "cc" which usually points to the GNU/GCC stack. Creating
+# this symlink ensures both cargo and kernel build rustc invocations use clang.
+RUN ln -s $(which clang) /usr/bin/cc
```
gctucker, line 30: Sounds like the Debian `update-alternatives` tool should be used for that instead.  I think this is OK, I'm just a bit weary of pulling the rug under the distro's standard tooling.  I mean, normally you get this:

```sh
$ ls -l /usr/bin/cc
lrwxrwxrwx 1 root root 20 May 30  2022 /usr/bin/cc -> /etc/alternatives/cc
$ ls -l /etc/alternatives/cc
lrwxrwxrwx 1 root root 12 May 30  2022 /etc/alternatives/cc -> /usr/bin/gcc
```
which could (should?) be changed to the path to `clang` with `update-alternative install [...exercise for the reader...]`.

config/core/build-configs.yaml
```diff
@@ -834,6 +834,11 @@ build_configs:
       clang-15:
         build_environment: clang-15
         architectures: *arch_clang_configs
+      rustc-1.62:
```
gctucker, line 837: This is a fix so having a Fixes trailer would be good here I think:
```
Fixes: 43875d0e47f2 ("config: add rust build configurations")
```

config/core/build-configs.yaml
```diff
@@ -834,6 +834,11 @@ build_configs:
...
+        build_environment: rustc-1.62
+        fragments: [rust, rust-samples, kselftest]
+        architectures:
+          x86_64: *x86_64_arch
```
gctucker, line 841: Keep in mind that this anchor expands to:
```yaml
        x86_64: &x86_64_arch
          extra_configs:
            - 'allmodconfig'
            - 'allnoconfig'
            - 'x86_64_defconfig+x86-chromebook+kselftest'
```
which includes `allmodconfig`.  Is it really worth doing a huge allmodconfig build with Rust?  If all the Rust-related parts of the code are already turned on with the extra fragments then this should get you covered already.  Maybe it's fine to have allmodconfig builds for mainline and next for completeness, but for the subsystem branches from the `rust-for-linux` tree I think that would be overkill.  Keeping the build configs optimised helps with lower latency for the CI loop and saves some build resources.

mgalka:
mainline build looks good, too:
https://bot.staging.kernelci.org/job/kernel-build/96077/consoleFull

mgalka:
@nuclearcat @gctucker Do you want to add anything?

nuclearcat (approved):
LGTM

aliceinwire:
https://bot.staging.kernelci.org/job/cip-pr_alicef/job/kernel-build/3265/consoleFull

aliceinwire (approved):

config/core/build-configs.yaml
ojeda: Some Rust abstractions may only get enabled if their subsystem is enabled. However, I would expect maintainers that start using Rust in their subsystem to add a `CONFIG_RUST=y` to test that part of their subsystem, rather than us having to add everything here. Does that make sense for KernelCI?

So I think for testing the general Rust infrastructure a `defconfig` is fine. Additionally, we are still upstreaming the basics, so we don't really have Rust-subsystems enabled yet, so I don't want to feel bad spending resources... :) I assume it is not a big deal to change the config in the future, right?

Having said that, of course, an `allmodconfig` somewhere with Rust enabled, in particular for `rust-next` (or `linux-next` as a shared Rust one for everybody), never hurts as a "final check" and to spot edge cases.

gctucker: Thanks for the extra context, yes feel free to submit PRs as and when Rust support expands in the kernel to make the KernelCI coverage grow accordingly.  And yes I think linux-next allmodconfig with Rust enabled would be quite useful at some stage, rather than the `rust-next` branch.  The advantage of keeping the builds for the `rust-for-linux` tree lightweight is that you can more easily do several pushes during the same day and get some quick turnaround, so `allmodconfig` might get in the way rather than help there.  And then `linux-next` builds happen once a day so full coverage there is very useful to everyone, especially once we start seeing patches with Rust code that don't go through the `rust-for-linux` tree (maybe that's already the case?).

PR #1562, opened 2022-12-08
docs: Add contributing guidelines

mgalka:
Add a few guidelines about filing PRs agains kernelci-core.

### DIFF
--- a/doc/contrib.md
+++ b/doc/contrib.md
@@ -1,0 +1,33 @@
+---
+title: "Contributing Guidelines"
+date: 2022-12-08T09:21:56Z
+draft: true
+weight: 3
+---
+
+KernelCI core project is open for contributions. Contributions may consist of
+adding new build, tests and device types as well as features and bugfixes for
+KernelCI core tools.
+The best way to contribute is to send a PR to [kernelci-core](https://github.com/kernelci/kernelci-core).
+When the PR is created, [KernelCI staging](https://kernelci.org/docs/instances/staging)
+instance takes care of updating the [staging.kernelci.org branch](https://github.com/kernelci/kernelci-core/tree/staging.kernelci.org).
+In general the branch is updated every 8h and a limited set of builds and tests
+is run on it. More detailed information about the logic behind staging runs can
+be found [here](https://kernelci.org/docs/instances/staging).
+
+There are several guidelines which can facilitate the PR review process.
+
+1. Make sure the PR is well described
+   1. Describe the purpose of the changes
+   2. Example use cases are welcome
+2. Attach staging build/test results when possible
+   1. Staging instance is updated every 8h. If the PR is expected to produce build/test results
+   check [staging dashboard](https://staging.kernelci.org) and make sure these are mentioned in the PR comment
+   2. If the results are not visible, but they should be, mention it in the comments, too
+3. Make sure that reviewers' comments and questions are addressed
+   1. When there are comments unanswered for more than 1 month the PR will be closed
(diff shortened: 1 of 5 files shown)


doc/contrib.md
```diff
@@ -0,0 +1,33 @@
...
+   1. Describe the purpose of the changes
+   2. Example use cases are welcome
+2. Attach staging build/test results when possible
+   1. Staging instance is updated every 8h. If the PR is expected to produce build/test results
+   check [staging dashboard](https://staging.kernelci.org) and make sure these are mentioned in the PR comment
+   2. If the results are not visible, but they should be, mention it in the comments, too
+3. Make sure that reviewers' comments and questions are addressed
+   1. When there are comments unanswered for more than 1 month the PR will be closed
+4. In case there is a need to consult the PR with KernelCI maintainers join the open hours
+   1. Open hours take place every Thursday at 12:00 UTC at KernelCI [Jitsi](https://meet.kernel.social/kernelci-dev)
```
nuclearcat, line 30: Maybe we should have "unified" place where should publish schedule of current public meetings? (even preferable something auto-generated) So if meeting schedule will change, very likely this link will point to incorrect info.

aliceinwire, line 30: Can we move it to a different day ? Wednesday would be better for me or more early hour.

aliceinwire:
I feel like contributions guideline should be on the README.md file as that information should be easy to find

gctucker:
We could have a paragraph about it in README.md with a link to the documentations once it's in production.  That way we don't have to maintain the same guidelines in 2 places as I think it's still important to have it in the main documentation.

doc/contrib.md
gctucker: That's when the monthly TSC meeting is, right?

gctucker: Yes and I think this is not specific to kernelci-core, but rather it's general for the whole project.  So it should be added under the Organisation section of the main documentation, and we can refer to it here and other places as needed.  Maybe it's fine to leave it here for this PR and then move it as a follow-up.

aliceinwire: no is one our before

aliceinwire: ah right so is 9:00 UTC then

mgalka:
I planned to put a link to it in the README.md once it's landed in the docs.

doc/contrib.md
gctucker: Yes I think we have to keep the existing one as that's been agreed by everyone, and we could have another one at a different time that's more convenient for APAC timezones.  So it might even be much earlier than that in fact.  That's essentially what we did for the hackfests and it worked well (except it was EMEA / AMER rather than EMEA / APAC, but having 3 sessions for EMEA / AMER / APAC might be a bit much...).

aliceinwire: As I'm also a maintainer of kernelci-core my idea was to keep that meeting

gctucker: The Open Hours is not just about kernelci-core, so I don't think that's the point.  But in any case, it sounds like it's worth proposing another Open Hours session at a time that works well for you and APAC timezones, maybe bring this up on the mailing list?

aliceinwire: ok

mgalka: Moved this document to kernelci-project documentation https://github.com/kernelci/kernelci-project/pull/157

mgalka:
As suggested by @gctucker moving this guidelines to kernelci-project respository

https://github.com/kernelci/kernelci-project/pull/157

PR #1563, opened 2022-12-08
Allow configuration to be loaded from multiple sources

eds-collabora:
This permits separating the configuration for a local KernelCI instance used in a conventional CI pipeline into two parts:

- The default configuration data can be taken directly from KernelCI itself (or a fork thereof, as required).
- The local configuration can specified separately, and merged in on top of the defaults KernelCI provides.

For example, in one common case of testing local kernel tree in CI, we can provide our own LAVA lab, own device types, own test suites on top of the KernelCI provided backbone, without maintaining a mandatory fork of KernelCI that contains our configuration data, and adds maintenance burden.

### DIFF
--- a/kernelci/config/__init__.py
+++ b/kernelci/config/__init__.py
@@ -63,7 +63,9 @@ def load_yaml(config_path, validate_entries=None):
     will be merged together under the same top-level dictionary key.
 
     *config_path* is the path to the YAML config directory, or alternative a
-                  single YAML file
+                  single YAML file.
+
+    *validate_entries* is currently unused.
     """
     config = dict()
     for yaml_path, data in _iterate_yaml_files(config_path):
@@ -75,6 +77,64 @@ def load_yaml(config_path, validate_entries=None):
                 config_value.extend(value)
             else:
                 config[k] = v
+    return config
+
+
+def _merge_trees(old, update):
+    """Merge two values loaded from YAML
+
+    This combines two values recursively that have been loaded from
+    YAML.  The data from *update* will be overlaid onto *old*
+    according to the rules:
+
+    - If *old* and *update* are dictionaries, their keys will be
+      unified, with the values for any keys present in both
+      dictionaries being merged recursively.
+
+    - If *old* and *update* are lists, the result is the concatenation
+      of the two lists.
+
+    - Otherwise, *update* replaces *old*.
+
+    Neither *old* nor *update* is modified; any modifications required
+    lead to a new value being returned.
+
+    """
+    if isinstance(old, dict) and isinstance(update, dict):
+        res = dict()
+        for k in (set(old) | set(update)):
+            if (k in old) and (k in update):
+                res[k] = _merge_trees(old[k], update[k])
+            elif k in old:
+                res[k] = old[k]
+            else:
+                res[k] = update[k]
+        return res
+    elif isinstance(old, list) and isinstance(update, list):
+        return old + update
+    else:
+        return update
+
+
+def load_all_yaml(config_paths, validate_entries=None):
+    """Load the YAML configuration
+
+    Load all the YAML files in all the specific configuration
+    directories and aggregate them together. Later directories take
+    precedence over earlier ones. This enables combining sources of
+    configuration data from multiple places.
+
+    *config_paths* is an ordered list of YAML configuration
+                   directories or YAML files, with later entries
+                   having higher priority.
+
+    *validate_entries* is passed to `load_yaml()`
+
+    """
+    config = dict()
+    for path in config_paths:
+        data = load_yaml(path, validate_entries)
+        config = _merge_trees(config, data)
     return config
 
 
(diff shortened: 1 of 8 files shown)


kernelci/config/__init__.py
```diff
@@ -78,6 +78,46 @@ def load_yaml(config_path, validate_entries=None):
     return config
 
 
+def _merge_trees(a, b):
```
mgalka, line 81: Maybe we could have a bit more descriptive names of the arguments?

kernelci/config/__init__.py
```diff
@@ -78,6 +78,46 @@ def load_yaml(config_path, validate_entries=None):
...
+    """Merge two values loaded from YAML"""
```
mgalka, line 82: Please add arguments description to the docstring

kernelci/config/__init__.py
```diff
@@ -78,6 +78,46 @@ def load_yaml(config_path, validate_entries=None):
...
+    if isinstance(a, dict) and isinstance(b, dict):
+        all_keys = set(a.keys())
+        all_keys.update(b.keys())
+        all_keys = list(all_keys)
+        res = dict()
+        for k in all_keys:
```
mgalka, line 88: Maybe we could use set operators here?

```suggestion
        res = dict()
        for k in (set(a) | set(b)):
```

kernelci/config/__init__.py
```diff
@@ -78,6 +78,46 @@ def load_yaml(config_path, validate_entries=None):
...
+    """Load the YAML configuration
+
+    Load all the YAML files in all the specific configuration
+    directories and aggregate them together. Later directories take
+    precedence over earlier ones. This enables combining sources of
+    configuration data from multiple places.
+
+    *config_path* is an ordered list of YAML configuration directories
+                  or YAML files, with later entries having higher
+                  priority
```
mgalka, line 112: I think it is `config_paths` now. 

Can you also add a description of `validate_entries`

kernelci/config/__init__.py
```diff
@@ -78,6 +78,46 @@ def load_yaml(config_path, validate_entries=None):
...
+    """
+    config = dict()
+    for path in config_paths:
+        x = load_yaml(path, validate_entries)
```
mgalka, line 116: maybe we can have here something more descriptive than `x`

kernelci/config/__init__.py
```diff
@@ -102,9 +142,27 @@ def load(config_path):
     Load all the YAML files found in the configuration directory then create
     a dictionary containing the configuration objects and return it.
 
-    *config_path* is the path to the YAML config directory
+    *config_path* is the path to the YAML config directory or a
+    unified file
+
     """
     if config_path is None:
```
mgalka, line 149: Shouldn't that be true for any empty value like `None` or `''`?

```suggestion
    if not config_path:
```

kernelci/config/base.py
```diff
@@ -80,6 +80,24 @@ def match(self, **kw):
...
+        filter matching both our existing data and the new items.
+
+        The *items* can be any data used to filter configurations.
+        Return True if we can combine, False otherwise.
+        """
+        return False
+
+
+def _merge_filter_lists(a, b):
+    """Merge the items for a Blocklist or Passlist."""
```
mgalka, line 94: Maybe  we can come up with more descriptive names of parameters here?

Please, update the docstring to

kernelci/config/base.py
```diff
@@ -80,6 +80,24 @@ def match(self, **kw):
...
+    for key, value in b.items():
+        if key not in a:
+            a[key] = value
+        else:
+            a[key].extend(value)
```
mgalka, line 99: I think we can make use of `dict.setdefault()` here

```suggestion
        a.setdefault(key, []).extend(value)
```

kernelci/config/base.py
```diff
@@ -156,6 +182,14 @@ def match(self, **kw):
         filter_values = tuple(kw.get(k) for k in self._keys)
         return filter_values in self._values
 
+    def combine(self, items):
+        keys = tuple(items['keys'])
+        if keys != self._keys:
+            return False
+
+        self._values.extend(list(tuple(values) for values in items['values']))
```
mgalka, line 190: I think we can simply use a list comprehension here

```suggestion
        self._values.extend([tuple(values) for values in items['values']])
```

kernelci/config/base.py
```diff
@@ -171,10 +205,25 @@ class FilterFactory(YAMLObject):
...
+                    filter_list.append(filter_instance)
+                else:
+                    for g in filters[filter_type]:
+                        if g.combine(items):
+                            break
+                    else:
+                        filter_cls = cls._classes[filter_type]
+                        filter_instance = filter_cls(items)
+                        filters[filter_type].append(filter_instance)
+                        filter_list.append(filter_instance)
```
mgalka, line 225: This block looks very similar to lines 212-216. Maybe we could extract it to a helper function/method?

mgalka (commented):
I like the idea of having a possibility of specifying extra config. 
Some style/refactoring remarks to consider.

kernelci/config/__init__.py
eds-collabora: I don't believe `validate_entries` is currently used; I've noted that the value is passed unchanged to the underlying load function, and I've added a note to that function that it doesn't currently consume that flag.

kernelci/config/__init__.py
eds-collabora: I don't think this is introduced by these changes, but I've added this change a separate commit onto the MR, if this is a change we want to make here.

mgalka (approved):
LGTM

PR #1568, opened 2022-12-12
Add ChromeOS support for arm64 and one arm64 Chromebook

nuclearcat:
This series of patches add support for arm64 Chromebooks and enable one of them sc7180-trogdor-lazor-limozeen, which have rootfs generated and almost ready for testing.

### DIFF
--- a/kernelci/test.py
+++ b/kernelci/test.py
@@ -58,17 +58,20 @@ def match_configs(configs, meta, lab):
     match = set()
 
     for test_config in configs:
+        dtb_missing = False
         if not test_config.match(arch, flags, filters):
             continue
         dtb = test_config.device_type.dtb
         if dtb and (not dtbs or dtb not in dtbs):
-            continue
+            dtb_missing = True
         for plan_name, plan in test_config.test_plans.items():
             if not plan.match(filters):
                 continue
             filters['plan'] = plan_name
             if lab.match(filters):
-                match.add((test_config.device_type, plan))
+                if not dtb_missing or \
+                   'fixed_kernel' in plan.params:
+                    match.add((test_config.device_type, plan))
 
     return match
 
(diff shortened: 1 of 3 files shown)


kernelci/test.py
```diff
@@ -58,17 +58,20 @@ def match_configs(configs, meta, lab):
...
         if dtb and (not dtbs or dtb not in dtbs):
-            continue
+            dtb_missing = True
         for plan_name, plan in test_config.test_plans.items():
             if not plan.match(filters):
                 continue
             filters['plan'] = plan_name
             if lab.match(filters):
-                match.add((test_config.device_type, plan))
+                if dtb_missing is False or \
```
mgalka, line 72: `is` is an identity operator and shouldn't be used with boolean values, unless you have good reason to do so. In this case I'd probably just use 

```suggestion
                if not dtb_missing or \
```

kernelci/test.py
```diff
@@ -58,17 +58,20 @@ def match_configs(configs, meta, lab):
...
+                   plan.params.__contains__('fixed_kernel'):
```
mgalka, line 73: `__contains__` is a special method, hence not meant to be called directly. __contains__ is what `in` operator uses under the hood.
```suggestion
                   'fixed_kernel' in plan.params:
```

mgalka (commented):
Some Python related remarks

kernelci/test.py
nuclearcat: Updated!

kernelci/test.py
nuclearcat: Updated!

nuclearcat:
Trogdor available in dashboard of tests: https://chromeos.kernelci.org/test/plan/id/6399f23ed70514180924cc5c/

mgalka (approved):
Good to have trogdor running :)

Issue #1463, opened 2022-10-04
Add branch fixes from riscv tree


khilman:
Please provide the information described below in order to add a new branch to
kernelci.org:

- **Which Git branch do you want to add?**

⇨ Git repo URL: https://git.kernel.org/pub/scm/linux/kernel/git/riscv/linux.git

⇨ Git branch name: fixes & for-next

- **How much build coverage do you need for your branch?**

⇨ Build coverage choice:

Build only ARCH=riscv, but for all upstream defconfigs (arch/riscv/defconfig/*) and for both the "fixes" and the "for-next" branch.

- **How often do you expect this branch to be updated?**

⇨ Estimated frequency: ~2 times per week

- **Who should the email reports be sent to?**

⇨ Recipients: kernel-ci-triage@rivosinc.com

PR #1569, opened 2022-12-13
build-configs.yaml: Add 2 riscv trees to monitor

nuclearcat:
As requested by Kevin Hilman, adding two trees for riscv, testing all available riscv configurations.

Fixes https://github.com/kernelci/kernelci-core/issues/1463

### DIFF
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -97,6 +97,9 @@ trees:
   renesas:
     url: "https://git.kernel.org/pub/scm/linux/kernel/git/geert/renesas-devel.git"
 
+  riscv:
+    url: "https://git.kernel.org/pub/scm/linux/kernel/git/riscv/linux.git"
+
   rmk:
     url: "git://git.armlinux.org.uk/~rmk/linux-arm.git"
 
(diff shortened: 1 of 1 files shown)


config/core/build-configs.yaml
```diff
@@ -966,6 +969,25 @@ build_configs:
...
+      gcc-10:
+        build_environment: gcc-10
+        architectures:
+          riscv:
+            base_defconfig: 'defconfig'
+            extra_configs:
+              - 'nommu_k210_defconfig'
+              - 'nommu_k210_sdcard_defconfig'
+              - 'nommu_virt_defconfig'
+              - 'rv32_defconfig'
```
khilman, line 985: Please leave off the `extra_configs` section.  We just want it to build all upstream defconfigs, which it will do by default, so pretty sure this is redundant.

khilman (changes requested):

config/core/build-configs.yaml
nuclearcat: Done

khilman (approved):

config/core/build-configs.yaml
```diff
@@ -966,6 +969,20 @@ build_configs:
     tree: renesas
     branch: 'next'
 
+  riscv-fixes: &riscv-tree
```
gctucker, line 972: The convention is to use `tree_branch-name` so `riscv_fixes` and `riscv_for-next` would have been more correct.

khilman: Fixed in PR #1577

PR #1570, opened 2022-12-13
test-configs: add riscv to arch map

khilman:
Map the kernel arch name "riscv" to debian arch name "riscv64"


khilman:
Tested kselftest LAVA` jobs `with` `kci_test generate` and works in staging.
c.f. https://staging.kernelci.org/test/job/khilman/branch/to-build/kernel/v6.1/

PR #1571, opened 2022-12-15
HACK/DONOTMERGE: update temporary rootfs images to test bootrr properly

nuclearcat:
As we rebased bootrr, we need to check if it is working well with bullseye and buildroot.


nuclearcat:
not required anymore

PR #1576, opened 2022-12-16
gcloud.jinja2: Add gcloud plugin as required by google cloud

nuclearcat:
```
CRITICAL: ACTION REQUIRED: gke-gcloud-auth-plugin, which is needed for continued use of kubectl, was not found or is not executable. Install gke-gcloud-auth-plugin for use with kubectl by following https://cloud.google.com/blog/products/containers-kubernetes/kubectl-auth-changes-in-gke
```
As recommended it can be installed by apt, in our gcloud jinja template fragment

### DIFF
--- a/config/docker/fragment/gcloud.jinja2
+++ b/config/docker/fragment/gcloud.jinja2
@@ -3,6 +3,10 @@
 # https://cloud.google.com/sdk/docs/downloads-apt-get
 #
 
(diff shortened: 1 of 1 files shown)


nuclearcat:
Rebuilt docker image on staging, without this patch kubectl was refusing to run. Now builds are fine: https://bot.staging.kernelci.org/job/kernel-build/

mgalka (approved):

PR #1580, opened 2022-12-17
build-configs: rust: disable IBT/Retpoline/Rethunk for the moment

ojeda:
Until we have the needed pieces in place in `rustc` and the kernel to support IBT/Retpoline/Rethunk (which will take at least a couple months); disable the relevant configs for the moment.

This avoids the 2000+ warnings from `objtool` that we currently generate in the logs.

Link: https://storage.kernelci.org/mainline/master/v6.1-11554-g785d21ba2f44/x86_64/x86_64_defconfig+rust/rustc-1.62/logs/build-warnings.log
Link: https://github.com/Rust-for-Linux/linux/issues/945
Signed-off-by: Miguel Ojeda <ojeda@kernel.org>

### DIFF
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -237,6 +237,9 @@ fragments:
   rust:
     path: "kernel/configs/rust.config"
     configs:
+      - '# CONFIG_X86_KERNEL_IBT is not set'
+      - '# CONFIG_RETPOLINE is not set'
+      - '# CONFIG_RETHUNK is not set'
       - 'CONFIG_RUST=y'
 
   rust-for-linux-samples:


ojeda:
I tested the fragment locally, and I see `kselftest` uses `# CONFIG_DUMMY is not set`, so I hope this is the right way!

ojeda:
@10ne1 Pinging in case you didn't see it. What should we do to proceed?

ojeda:
No rush, Alice may be taking a look too -- Happy New Year! :)

aliceinwire:
yes, I asked @gctucker or @mgalka to enable rustc-1.66 docker image on kernelci for testing. still waiting for a replay.

aliceinwire:
Thanks

aliceinwire:
@ojeda just as a suggestion. I just realized that the link file in the comment is on storage.kernelci.org. files on storage.kernelci.org get cleaned after around one month AFAIR, maybe is better to copy the content to gist.github.com and use that link next time.

ojeda:
Yeah, avoiding ephemeral links sounds better. Thanks for pointing it out! Perhaps I should have just copied a sample inside the Git commit directly.

Should it be done only for the links in the commit messages, or also for the ones sometimes are put here in GitHub from staging? i.e. is it only for things that end up in the Git repository?

aliceinwire:
@ojeda KernelCI doesn't have a rule about this. that's why it was only a suggestion. In my opinion commit messages as they are lasting more time, is maybe better if they link something a bit more permanent.

ojeda:
Understood, thanks!

PR #1582, opened 2022-12-17
Add kci single entry point with validate sub-command

gctucker:
Add a `kci` single entry point executable script with a first sub-command `validate` to validate the YAML configuration.  This should simplify and consolidate the implementation of the various command line tools, especially as things need to be redesigned now that `kci_build`, `kci_test` and `kci_data` seem like very arbitrary names as the concept of nodes with the new API is orthogonal to the use-cases.

Here's a sample command:
```
kci validate yaml --verbose
```


kernelci/cli/__init__.py
```diff
@@ -0,0 +1,42 @@
...
+from . import validate
+
+_COMMANDS = {
+    'validate': validate.main,
+}
+
+
+def list_command_names():
+    """Return a list with the command names"""
+    return _COMMANDS.keys()
```
mgalka, line 30: Either docstring or implementation is wrong. `.keys()` return type is not `list`.

```python
>>> d = {'a': 1}
>>> isinstance(d.keys(), list)
False
>>> type(d.keys())
<class 'dict_keys'>


```

gctucker:
```
keys(...)
    D.keys() -> a set-like object providing a view on D's keys
```
Well it's not strictly a Python list but it's close enough and still kind of true in terms of semantics.  I could just say, "Return the command names".  Or `list()` it.  But what's your opinion?

gctucker: OK I've turned the returned value into an actual `list` object.

gctucker:
@nuclearcat Why did you add `staging-skip` to this PR?

nuclearcat:
Sorry, i was searching PR that breaks staging (it was turned out different one, this one https://github.com/kernelci/kernelci-core/pull/1586), and forgot to remove flag here.

gctucker:
I think I'll send an email to the list about what I have in mind here.  Basically it's to introduce a new set of command line tools which will be using the new API.  That way, we can leave the current ones in place for now and deprecate them while working on the new ones, and also it's a good opportunity for introducing a single `kci` entry point command.

@mgalka Do you have any further comments for this PR?

gctucker:
@mgalka Please take another look now I've addressed things from your initial review.  Also I've sent an email detailing what I had in mind: https://lore.kernel.org/kernelci/4ae8d73f-ed10-c3b5-c6c3-e30ffdf95d7b@collabora.com/T/#u

kernelci/cli/base.py
```diff
@@ -582,7 +570,7 @@ def parse_args_with_parser(parser, glob):
     *parser* is an `ArgumentParser` instance used to parse the command line.
```
mgalka, line 570: `args` parameter description missing in docstring

kernelci/cli/base.py
```diff
@@ -604,7 +592,7 @@ def make_options(args, prog):
     return opts
 
 
-def parse_opts(prog, glob, yaml_config_path=None):
+def parse_opts(prog, glob, yaml_config_path=None, args=None):
```
mgalka, line 595: missing docstring for `args`

mgalka (commented):
I think it's worth to add the `args` parameters description to the docstrings

kernelci/cli/__init__.py
mgalka: Thanks for this change, I guess it's fine to return an actual list here. 
I don't have strong opinions on what kind of object it should be, I just wanted to have the description match the actual object. The list in python is both iterable and subscriptable, while set-like object is only subscriptable and is not meant to keep it's elements order.

kernelci/cli/base.py
gctucker: Fixed

kernelci/cli/base.py
gctucker: Fixed

gctucker:
@mgalka Please take another look.

mgalka (approved):

PR #1584, opened 2022-12-18
config/k8s: add set -e to build jobs

gctucker:
Add set -e to the build job template so that if any command fails the job aborts rather than trying to run subsequent commands as that's destined to keep piling up more failure.

### DIFF
--- a/config/k8s/job-build.jinja2
+++ b/config/k8s/job-build.jinja2
@@ -84,7 +84,7 @@ install: true\n\
 [kci_data]\n\
 output: ${KDIR}/build\n\
 \" > kernelci.conf; \
-set -x; \
+set -xe; \
 \
 kci_build pull_tarball \
   --url ${SRC_TARBALL} \


config/k8s/job-build.jinja2
```diff
@@ -84,7 +84,7 @@ install: true\n\
 [kci_data]\n\
 output: ${KDIR}/build\n\
 \" > kernelci.conf; \
-set -x; \
+set -xe; \
```
gctucker, line 87: Tested OK on staging:
https://bot.staging.kernelci.org/job/kernel-build/58748/consoleFull
```
[kci_data]
      output: ${KDIR}/build
      " > kernelci.conf; set -xe; kci_build pull_tarball --url ${SRC_TARBALL} ...
```

nuclearcat (approved):
Works fine on staging, kernel build statistic looks ok and no errors observed, i think it is good to merge

PR #1585, opened 2022-12-18
Add kernelci.cli.sub_main()

gctucker:
Add `kernelci.cli.sub_main()` and use it with `kernelci.cli.validate`.

Depends on #1582

### DIFF
--- a/kernelci/cli/__init__.py
+++ b/kernelci/cli/__init__.py
@@ -5,7 +5,7 @@
 
 import sys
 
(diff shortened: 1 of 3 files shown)


gctucker:
Adding `staging-skip` to only get #1582 deployed on staging for now to see if that works as expected and get it merged first before enabling this PR on top of it.

gctucker:
#1582 is now merged so removing `staging-skip` here.

mgalka (approved):

PR #1587, opened 2022-12-19
T39494 kernelci_api: Add enum class for node states

JenySadadia:
Add an enum class for holding pending Node states. 
Add instance variable`_pending_node_states` to `KernelCI_API` class store enum for pending node states.
Implement method `get_pending_node_states` to get pending node states to use in pipeline client services.
This is to align the node states with API until we have a mechanism to get the states from API in place.

### DIFF
--- a/kernelci/db/kernelci_api.py
+++ b/kernelci/db/kernelci_api.py
@@ -48,6 +48,12 @@ class KernelCI_API(Database):
             'Content-Type': 'application/json',
         }
         self._filters = {}
+        self._node_states = NodeStates
+
+    @property
+    def node_states(self):
+        """Get node states"""
+        return self._node_states
 
     def _make_url(self, path):
         return urllib.parse.urljoin(self.config.url, path)
--- a/kernelci/db/kernelci_api.py
+++ b/kernelci/db/kernelci_api.py
@@ -16,12 +16,25 @@
 # 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
 
 import json
+import enum
 import requests
 import urllib.parse
 
 from cloudevents.http import from_json
 
 from kernelci.db import Database
+
+
+class NodeStates(enum.Enum):
+    """Enumeration for node states
+    This is a temporary implementation until we have API versioning
+    in place. Once we have it, we can directly use a version specific
+    endpoint to get node states"""
+
+    RUNNING = 'running'
+    AVAILABLE = 'available'
+    CLOSING = 'closing'
+    DONE = 'done'
 
 
 class KernelCI_API(Database):


kernelci/db/kernelci_api.py
```diff
@@ -24,6 +25,14 @@
 from kernelci.db import Database
 
 
+class PendingNodeStates(enum.Enum):
+    """Enumeration for pending node states"""
```
mgalka, line 29: Maybe it's worth mentioning that this is a temporary solution to be replaced with an API call when the versioning is in place.

mgalka (commented):
Small suggestion

kernelci/db/kernelci_api.py
JenySadadia: Updated the docstring.

gctucker (commented):
Why "pending" node states and not just all the states?

JenySadadia:
> Why "pending" node states and not just all the states?

If the method is sending all the states, how would the pipeline know what is the name of the pending state to exclude it from the returned node states?
The node states' names can be different for various API versions.

gctucker:
The concept of "pending states" is internal to the timeout service.  The API doesn't define pending states, just a series of states.  So I think the enum is incomplete if it doesn't have all the states.

By using an enum, we can then have a list using the enum values in the timeout script rather than raw strings.  That I believe is the main advantage, so if we change the status strings to use capitals or just any text changes we won't have to update the code everywhere state names are used.

gctucker:
Think about it this way: having a `get_node_states()` method or just an enum without a method would provide the abstraction and scripts using it would use the enum values.

If we have things like `get_pending_node_states()`, why not also have `get_done_states()`, `get_running_states()`, `get_closing_states()` with various kinds of semantics specific to all the use-cases built on top of the state names?  I think that's the kind of things that is relevant to the applications (e.g. the timeout service) rather than the API.

JenySadadia:
Added `staging-skip` as it is conflicting with https://github.com/kernelci/kernelci-core/pull/1624

JenySadadia:
> The concept of "pending states" is internal to the timeout service. The API doesn't define pending states, just a series of states. So I think the enum is incomplete if it doesn't have all the states.
> 
> By using an enum, we can then have a list using the enum values in the timeout script rather than raw strings. That I believe is the main advantage, so if we change the status strings to use capitals or just any text changes we won't have to update the code everywhere state names are used.

Yes, I am aware of the advantages of using Enum here.
My point is if the timeout is receiving all the node states, it needs to exclude `done` state from the node states received while trying to get pending nodes from the API.

So, the code snippet will be changed to:
```
for state in self._pending_node_states:
            if state.name != 'DONE':
                      node_filters['state'] = state.value
```
What if API renames the state `done` to something else? In that case, the client will still try to exclude the `done` state from the list of node states received.
How will the client be aware of these kinds of changes in node states?

Also, I agree that we can not have methods specific to return done and running state.
But how will we cope with the issue I mentioned?  I guess in that case we need to adjust manually every time the node state schema changes in the API.

gctucker:
There's documentation which developers need to follow, and with the API version checks we can ensure that client code won't be using an incompatible API version.  And of course, state names shouldn't change without an API version change.

gctucker:
(at least once we've started enforcing the API version with v1)

JenySadadia:
> There's documentation which developers need to follow, and with the API version checks we can ensure that client code won't be using an incompatible API version. And of course, state names shouldn't change without an API version change.

OK, in that case, proper documentation will be required.
Anyway, I have updated the code to return all the node states in Enum.

gctucker:
From my earlier comment:

> having a get_node_states() method or just an enum without a method

So how about just using the enum?

JenySadadia:
> From my earlier comment:
> 
> > having a get_node_states() method or just an enum without a method
> 
> So how about just using the enum?

When we have API versioning in place, we need to use a method to get node states from an API endpoint.
At that time I can directly update `get_node_states` method to request the endpoint.
That's why I defined it in this one.

gctucker:
> We need to use a method to get node states from an API endpoint.

Well, I'm not sure we need that.  We can, but what's the reason for doing it?  The code using the API needs to know about all the states anyway just like it needs to know about all the endpoints to be able to use the API.  I think that using versioned API and documentation, we can just have the code in kernelci-core implement bindings that match the documentation.  If we can do things without creating more endpoints and more code then it's better, especially in the lower levels of the stack such as the API and the Python bindings.

JenySadadia:
> > We need to use a method to get node states from an API endpoint.
> 
> Well, I'm not sure we need that. We can, but what's the reason for doing it? The code using the API needs to know about all the states anyway just like it needs to know about all the endpoints to be able to use the API. I think that using versioned API and documentation, we can just have the code in kernelci-core implement bindings that match the documentation. If we can do things without creating more endpoints and more code then it's better, especially in the lower levels of the stack such as the API and the Python bindings.

OK. Fixed it.

gctucker (approved):

PR #1588, opened 2022-12-20
test-configs.yaml: Add more tests to trogdor

nuclearcat:
As we identified problem with networking (slow detection), and have workaround for it, we can enable more advanced networked tests on this type of device.
It can be very useful for comparison with other
arm64 devices, for example Mediatek chipset.


nuclearcat:
Alsa test is ok on stable 5.19: https://staging.kernelci.org/test/plan/id/63a18df2a89fa06667e6d68c/
baseline-nfs is ok too: https://staging.kernelci.org/test/plan/id/63a18db5aeddc0ab04e6d65f/

nuclearcat:
Still, even with all patches it fails to detect GPU and video decoding.
Some additional firmware also required:
qcom-venus aa00000.video-codec: Direct firmware load for qcom/venus-5.4/venus.mdt failed with error -2

nuclearcat:
Ok, so i took config ideas from: https://www.spinics.net/lists/dri-devel/msg363666.html
And voila, it run: https://lava.collabora.dev/scheduler/job/8430857
Now i need to reduce and remove unnecessary options

nuclearcat:
Trogdor serial issue is very problematic.
I added small "fake" test workaround to reduce kernel messages except critical:
```
- test:
    failure_retry: 2
    timeout:
      minutes: 1
    definitions:
    - repository:
        metadata:
          format: Lava-Test Test Definition 1.0
          name: disable-kernel-messages
          description: "Redirect kernel messages to file"
          os:
          - oe
          scope:
          - functional
        run:
          steps:
          - busybox klogd -c 2
      from: inline
      name: redirect-logs
      path: inline/redirect-logs.yaml
      lava-signal: kmsg
```
Here is successfull result with this small workaround:https://lava.collabora.dev/scheduler/job/8432001

nuclearcat:
Even considering racy serial, it is able to pass some msm tests: https://staging.kernelci.org/test/plan/id/63a431dad412a949bc9b92b1/
Some additional work required, to improve tests results and reliability, i will open issue about it later

mgalka (approved):

mgalka:
Let's land it and work on improving the stability, also if it misbehaves we may drop it.

PR #852, opened 2021-10-22
rootfs-configs: Trogdor: Add Adreno 630 and Venus-5.2 firmwares

kholk:
This patchset adds the required firmwares for GPU and hardware enc/decoder, necessary to run certain tests on Qualcomm sc7180-trogdor

PR #1592, opened 2022-12-20
Add trogdor/qualcomm GPU tests

nuclearcat:
To test properly trogdor GPU and PR https://github.com/kernelci/kernelci-core/pull/852 we need to enable several tests referenced in that commit.


nuclearcat:
Will be added in different PR, to not make conflicts during tests on staging

PR #1594, opened 2022-12-22
kernelci.storage.ssh: Replace scp binary calls with paramiko

mgalka:
Storage_ssh class implementation relied on scp binary being installed in the operating system. Implementation based on paramiko and scp Python packages reduces the dependency on underlying system tools and adds better control of the file transfer.


gctucker:
Rebased and replaced with #1819

PR #1597, opened 2022-12-26
Rust 1.66.0

ojeda:
Rust 1.66.0 will be the next version supported by the kernel [1].

Therefore, add `rustc-1.66` as a new build environment alongside `rustc-1.62` (which can be removed later on when the new version hits mainline), and then move `rust-for-linux_rust` to it.

Link: https://github.com/Rust-for-Linux/linux/pull/947 [1]
Signed-off-by: Miguel Ojeda <ojeda@kernel.org>

### DIFF
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -1106,8 +1106,8 @@ build_configs:
     tree: rust-for-linux
     branch: 'rust'
     variants:
-      rustc-1.62:
-        build_environment: rustc-1.62
+      rustc-1.66:
+        build_environment: rustc-1.66
         fragments: [rust, rust-for-linux-samples, kselftest]
         architectures:
           x86_64:
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -499,6 +499,14 @@ build_environments:
         <<: *x86_64_params
         name:
 
+  rustc-1.66:
+    cc: clang
+    cc_version: 15
+    arch_params:
+      x86_64:
+        <<: *x86_64_params
+        name:
+
 # Default config with full build coverage
 build_configs_defaults:
   variants:
--- a/config/docker/rustc-1.66.jinja2
+++ b/config/docker/rustc-1.66.jinja2
@@ -1,0 +1,43 @@
+{% extends 'clang-15.jinja2' %}
+
+{% block packages %}
+{{ super() }}
+
+ARG RUST_VER=1.66.0
+ARG BINDGEN_VER=0.56.0
+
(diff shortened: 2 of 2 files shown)


aliceinwire:
Related: https://github.com/kernelci/kernelci-core/pull/1580

aliceinwire:
currently testing this.

ojeda:
Thanks a lot!

aliceinwire:
@gctucker or @mgalka could you enable rustc-1.66 docker image on KernelCI for testing?

nuclearcat:
I guess -next not supposed to be buildable by 1.66?
https://storage.staging.kernelci.org/kernelci/staging-next/staging-next-20230108.7/x86_64/x86_64_defconfig+rust/rustc-1.66/logs/build-errors.log
You can see here also builds with rust in general: https://staging.kernelci.org/build/kernelci/branch/staging-next/kernel/staging-next-20230108.7/

ojeda:
> I guess -next not supposed to be buildable by 1.66?

Yeah. Right now, no branch is buildable by 1.66. If we merge https://github.com/Rust-for-Linux/linux/pull/947, then `rust-for-linux_rust` will be, but at the same time, it will stop being buildable by 1.62. That is why I haven't merged it yet -- whenever you are ready we can do it. I am not sure which is the best order (or at once) to do so for KernelCI. On our side, we can do it at any time (for that branch).

Then, after that is merged and we check it is OK, my plan is to submit it to the mailing list and soon after put it into `rust-for-linux_rust-next`, and at that point I can submit another PR here to update it, I guess. Finally the same for mainline.

But the above is just what I came up with... If you think KernelCI could handle these "flag day" upgrades in a better way, it would be nice. Ideally, KernelCI could fetch the right environment as given by `scripts/min-tool-version.sh`, but I assume that is non-trivial unless the system is designed to do things like bisections anyway.

nuclearcat:
@aliceinwire maybe we can merge it then? So https://github.com/Rust-for-Linux/linux/pull/947 will be merged too.
Generally rust-1.66 looks ok and working, everything prepared and tested what possible.

gctucker:
I think it's best if KernelCI is always upgrading rust toolchains to the latest used by Rust-for-Linux, to provide CI as early as possible.  As Rust support in the kernel is still in active development, it's expected that things will break in linux-next etc.  It's better if KernelCI shows problems that then get fixed, in the end what really matters is that it helps Rust support in the kernel to get tested and working rather than a quiet sea of green results on linux.kernelci.org.

ojeda:
Yeah, I think it is fine, so let's merge both branches and see if everything goes well. Thanks a lot!

aliceinwire:
@nuclearcat doing last tests before merging

PR #1598, opened 2022-12-27
CHROMEOS test-configs-chromeos.yaml: Disable QEMU

nuclearcat:
For last 2 releases QEMU is broken for ChromeOS, and it doesn't look there is easy solution to fix that.
I am commenting temporary all relevant entries with hope, that in new releases it might work or issue might be fixed. Also commenting is important to not waste lab QEMU resources and docker pull limits.


mgalka:
I think it may be worth putting a line to explain why it was commented out and not removed for example.

mgalka (approved):
Let's disable it for now and check  with the new revision

Issue #670, opened 2021-05-18
Hardware video codecs in Chrome OS


gctucker:
Some embedded SoCs have hardware blocks to accelerate video decoding. This is about enabling some tests in KernelCI for them, at least on one type of hardware.

The Chrome OS tests for this can be found here:
https://chromium.googlesource.com/chromium/src/+/refs/heads/master/docs/media/gpu/video_decoder_test_usage.md

nuclearcat:
I think in ChromeOS R106 we have tests video.PlatformDecoding.ffmpeg_vaapi* which all use acceleration api and we can achieve this issue by using them.

PR #1599, opened 2022-12-28
CHROMEOS test-configs-chromeos.yaml: Add additional hardware offload tests

nuclearcat:
For video encoding and decoding in x86 (amd/intel) we have vaapi and some arm64 have v4l m2m video codec offloading implemented in kernel drivers. ChromeOS uses this API have tast tests for it, so we need to use them, as we have most of Chromebooks x86 - add first VAAPI tast tests.
Fixes https://github.com/kernelci/kernelci-core/issues/670

### DIFF
--- a/config/core/test-configs-chromeos.yaml
+++ b/config/core/test-configs-chromeos.yaml
@@ -124,6 +124,72 @@ test_plans:
       <<: *cros-tast-base-qemu-params
       tests: *cros-tast-kernel-tests
 
+  cros-tast-mm-decode:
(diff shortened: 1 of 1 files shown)


nuclearcat:
Tests can be seen with prefix cros-tast-mm-* in recent tests: https://chromeos.kernelci.org/test/job/kernelci/branch/chromeos-stable/kernel/chromeos-stable-20230102.0/

mgalka (approved):
Thanks for adding these tests

PR #1600, opened 2022-12-29
CHROMEOS cros-boot.jinja2: Add lava-test-raise for modules-install

nuclearcat:
If we fail installing modules there is no point to continue test, as in best case it will fail, and in worst it will give false results


nuclearcat:
Here it helped to expose issue with modules install: https://lava.collabora.dev/scheduler/job/8543729

Issue #1496, opened 2022-11-08
V4L2 decoder conformance tests incorrectly run on net-next tree


laura-nao:
V4L2 decoder conformance tests are configured to run on the `mainline`, `media` and `next` trees (https://github.com/kernelci/kernelci-core/blob/main/config/core/test-configs.yaml#L550), but the dashboard shows that tests on `rk3399-gru-kevin` are run on the `mainline`, `media` and `net-next` trees instead (e.g. H264 test run on net-next: https://linux.kernelci.org/test/job/net-next/branch/master/kernel/v6.1-rc3-796-gbf46390f39c6/plan/v4l2-decoder-conformance-h264/)

PR #1601, opened 2022-12-29
test-configs.yaml: Fix tree match for v4l tests

nuclearcat:
As explained in https://github.com/kernelci/kernelci-core/issues/1496 we need to test only several trees, and current way it match also net-next kernel, not only next. With regex, as verified manually it should solve issue.
Fixes https://github.com/kernelci/kernelci-core/issues/1496


nuclearcat:
Regex tested manually by fetching net-next, next, mainline, media kernels. Now it doesnt run tests on net-next.

mgalka (approved):
LGTM

PR #1605, opened 2023-01-03
Add more crypto options

montjoie:
While working on refresing https://github.com/kernelci/kernelci-core/pull/511, I see that KCI already have now a crypto fragment, but it miss some important options.
So let's enable in-kernel crypto selftest.


broonie:
LGTM and boot on platforms with crypto hardware seems unaffected (eg, https://staging.kernelci.org/test/plan/id/63bd2b33468060ee6cbb05d2/)

nuclearcat (approved):
LGTM2

gctucker:
@montjoie Please follow the convention for the commit message subject next time, it helps a lot when reading the git log as this project has different separate parts.

PR #1606, opened 2023-01-04
test-configs.yaml: Remove kselftest from baseline* blocklist

nuclearcat:
As real use cases show - we need to test kselftest kernels on baseline/baseline-nfs, because for example it looks like some boards unable to boot +kselftest builds, but booting fine with similar build without this fragment.
baseline tests help a lot with testing general "bootability" of kernel.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -69,7 +69,7 @@ test_plans:
     params:
       priority: 90
     filters:
-      - blocklist: *kselftest_defconfig_filter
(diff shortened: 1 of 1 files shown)


nuclearcat:
kselftest are being tested on baseline now.
Also according to discussion on IRC with @broonie it helped to troubleshoot issue with overlapping memory regions between kernel/initrd.

mgalka (approved):
lgtm

PR #1614, opened 2023-01-06
build-configs.yaml: Enable chromeos_acpi support

nuclearcat:
As TPM reset and several other features that require NVRAM offset lookups - we need to enable newly added option for acpi tables exported on Chromebooks. It will allow to lookup for NVRAM offset to manage such things like TPM ownership.

### DIFF
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -482,6 +482,7 @@ fragments:
   x86-chromebook:
     path: "kernel/configs/x86-chromebook.config"
     configs:
+      - 'CONFIG_CHROMEOS_ACPI=y'
       - 'CONFIG_SERIAL_8250=y'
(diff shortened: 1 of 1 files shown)


nuclearcat:
Kernel built with ACPI support: https://storage.staging.kernelci.org/kernelci/staging-next/staging-next-20230209.0/x86_64/x86_64_defconfig+x86-chromebook/gcc-10/config/kernel.config
No problems observed with Chromebooks boot, for example: https://storage.staging.kernelci.org/kernelci/staging-next/staging-next-20230209.0/x86_64/x86_64_defconfig+x86-chromebook/gcc-10/lab-collabora/baseline-nfs-dell-latitude-5400-8665U-sarien.html

mgalka (approved):

PR #1616, opened 2023-01-06
Make the install_path argument work for kci_build too

eds-collabora:
This allows you to pick the directory into which kci_build will install the built kernel and modules, rather than forcing it to be `_install_` within the built kernel's tree. This is convenient for some automated processes, where we want to control what the generated path is, without copying.

### DIFF
--- a/kci_test
+++ b/kci_test
@@ -199,10 +199,8 @@ class cmd_list_jobs(Command):
             print("Either --build-output or --install-path is required")
             return False
 
-        install = (
-            args.install_path if args.install_path else
-            kernelci.build.Step.get_install_path(None, args.build_output)
-        )
+        install = kernelci.build.Step.get_install_path(None, args.build_output,
+                                                       args.install_path)
         meta = kernelci.build.Metadata(install)
 
         if meta.get('bmeta', 'build', 'status') != "PASS":
@@ -288,10 +286,8 @@ class cmd_generate(Command):
             print("Either --build-output or --install-path is required")
             return False
 
-        install = (
-            args.install_path
-            or kernelci.build.Step.get_install_path(None, args.build_output)
-        )
+        install = kernelci.build.Step.get_install_path(None, args.build_output,
+                                                       args.install_path)
         meta = kernelci.build.Metadata(install)
 
         if meta.get('bmeta', 'build', 'status') != "PASS":
--- a/kernelci/build.py
+++ b/kernelci/build.py
@@ -609,7 +609,8 @@ class Step:
 
     KVER_RE = re.compile(r'^v([\d]+)\.([\d]+)')
 
-    def __init__(self, kdir, output_path=None, log=None, reset=False):
+    def __init__(self, kdir, output_path=None, log=None, reset=False,
+                 install_path=None):
         """Each Step deals with a part of the build and its related meta-data
 
         This abstract class handles the common code to run any kernel build
@@ -631,7 +632,8 @@ class Step:
         self._output_path = output_path or self.get_default_output_path(kdir)
         if not os.path.exists(self._output_path):
             os.mkdir(self._output_path)
-        self._install_path = self.get_install_path(kdir, self._output_path)
+        self._install_path = self.get_install_path(kdir, self._output_path,
+                                                   install_path)
         self._create_install_dir(reset)
         self._meta = Metadata(self._output_path, reset)
         self._meta.clear_artifacts(self.name)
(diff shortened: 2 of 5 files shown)


mgalka:
@eds-collabora I guess this needs to be rebase due to the storage abstraction changes landed recently.

gctucker:
We might be able to drop the `--install` argument as a follow-up.  If you specify an install path then do the install steps there, otherwise don't.  I think it's fine to merge this PR as-is (as long as it's working on staging).

Also please note that the builds should soon be reworked as job templates to be treated like any pipeline step.  That shouldn't really make much difference but it'll cause some code churn, so it's better if you land all the changes related to your use-case first.

mgalka:
> We might be able to drop the `--install` argument as a follow-up. If you specify an install path then do the install steps there, otherwise don't. I think it's fine to merge this PR as-is (as long as it's working on staging).

If we want to drop the `--install` argument, I guess it'd be better to rename the option to something like `--install-to` instead of `--install-path` as the latter suggest that it's just a path definition not an action to be taken.

gctucker:
Let's see how this looks:
```
# Build and no files copied to install directory
./kci_build make_kernel

# Opt.1: Build and files copied to specified install directory
./kci_build make_kernel --install-path=some/install/path

# Opt.2: Build and install and and specify install directory
./kci_build make_kernel --install --install-path=some/install/path

# Opt.2: What do we do here?
./kci_build make_kernel --install # stick to some default?
./kci_build make_kernel --install-path=some/install/path # raise error as this is used without --install
```

So that's why I think Opt.1 is more intuitive and easier to use.

gctucker:
Based on today's Open Hours discussion, here's what's now suggested:
```
# To build and install in default path like with --install in the current version, just we drop --install:
./kci_build make_kernel
# To override the install path:
./kci_build make_kernel --install-path=some/path
# To use the default install path like in the current version:
./kci_test generate
# To override the install path (also I believe just like in the current version):
./kci_test generate --install-path=some/path/or/other
```

gctucker:
The suggestion above avoids breaking existing code that relies on the default install path.  I believe the only thing that would need to be done in the Jenkins jobs is to drop the `--install` argument as the code would always be doing the install step.  It's still probably worth keeping separate methods in the code, and maybe we'll need a `--skip-install` argument one day...

kci_test
```diff
@@ -288,10 +286,8 @@ class cmd_generate(Command):
             print("Either --build-output or --install-path is required")
             return False
 
-        install = (
-            args.install_path
-            or kernelci.build.Step.get_install_path(None, args.build_output)
-        )
+        install = kernelci.build.Step.get_install_path(None, args.build_output,
+                                                       args.install_path)
```
gctucker, line 290: Maybe use keyword arguments:
```suggestion
        install = kernelci.build.Step.get_install_path(
            output_path=args.build_output, install_path=args.install_path
        )
```
I guess it's just a bit clearer and less prone to problems if the arguments are changing.

kernelci/build.py
```diff
@@ -666,7 +668,7 @@ def get_default_output_path(cls, kdir):
         return os.path.join(kdir, 'build')
 
     @classmethod
-    def get_install_path(cls, kdir=None, output_path=None):
+    def get_install_path(cls, kdir=None, output_path=None, install_path=None):
         """Get the default build install path
```
gctucker, line 672: This needs to be updated as it's not just about getting the default install path.  Also the new `install_path` argument needs to be documented.

Another way to implement this would be to not add `install_path` and not return straight away when it's `None`, but rename this method to `.get_default_install_path()` and in the constructor do something like this:
```python
self._install_path = install_path or self.get_default_install_path(kdir, self._output_path)
```
and same thing in `kci_build`.  Part of this starts to look a bit clunky though with a `@classmethod` but the whole thing should be redesigned as job templates anyway so I don't know if it's worth making it nicer.  At least it should be consistent with the docstring.

gctucker: It's already there.  Did you mean something slightly different?

mgalka: I meant the changes you suggested above.

gctucker: ...which are just about how to pass the default value around.  But actually I think this is all more of a distraction, whatever works from a functional point of view is good at this stage.

nuclearcat:
It looks like this patch conflict with some other patches on staging, will try to figure out which one.

gctucker:
This is conflicting with other arguments being added to `kernelci.cli.base` so adding the `staging-skip` label for now.

gctucker:
@eds-collabora Please take another look, I think this will need some rework and potentially a new PR.

gctucker:
Same as #1425, leaving it open a bit longer.

nuclearcat:

PR #1617, opened 2023-01-06
Enable rseq kselftests

broonie:
Ideally we'd cover more boards (rseq is supported on x86 too and a big.LITTLE board wouldn't go amiss for arm64) but this gets things going.


nuclearcat (approved):
LGTM

PR #1622, opened 2023-01-09
config/lava/kselftest: Don't install dependencies for kselftest at runtime

broonie:
We build our kselftest rootfs images with all the dependencies for kselftest already installed and therefore do not need to install the incomplete set that the test-defintions tempalte attempts to install at runtime, skip doing so in order to speed up our kselftest jobs and allow them to run on very low memory systems where apt struggles without swap.

### DIFF
--- a/config/lava/kselftest/kselftest.jinja2
+++ b/config/lava/kselftest/kselftest.jinja2
@@ -26,3 +26,4 @@
         TST_CASENAME: {{ kselftest_tests }}
         BOARD: {{ device_type }}
         BRANCH: {{ tree }}
+        SKIP_INSTALL: True


broonie:
@nkbelin @khilman I suspect this might help with the issues you're seeing on RISC-V with apt not being present on the rootfs.

gctucker:
Are there actually any packages listed in the runtime install that aren't already in the rootfs image?

gctucker:
OK, so a very long noop :p Thanks for confirming.

broonie:
Tested OK on staging - https://staging.kernelci.org/test/plan/id/63bd757e61f334db93bb05ba/ doesn't install any deps: https://storage.staging.kernelci.org/kernelci/staging-mainline/staging-mainline-20230110.0/arm64/defconfig/clang-11/lab-collabora/kselftest-arm64-bcm2711-rpi-4-b.html#L1219

broonie:
(the actual tests aren't working but that's a separate issue with clang builds, https://staging.kernelci.org/test/plan/id/63bd5e7c98ed025d00bb05be/ would have been a better pick)

nuclearcat (approved):
LGTM, i was always wondering why those kselftests are calling apt while we have all packages installed. Turned out reason is this option :)

khilman:
Thanks for this @broonie. We have noticed problems with kselftest on riscv also.

gctucker:
Note: We'll just have to keep an eye on the list of packages required and make sure the rootfs image is updated accordingly.  We might have kselftests running on staging with this option turned on in some cases to double check the rootfs really has all the requirements already installed.

PR #1623, opened 2023-01-09
test-configs: Enable arm64 kselftests on qemu

broonie:
In order to get coverage of new architecture extension that are only available in emulation enable the arm64 selftests in qemu, using a ramdisk for the filesystem as we do for the RISC-V kselftests running qemu.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -228,6 +228,14 @@ test_plans:
     filters:
       - passlist: {defconfig: ['kselftest']}
 
+  # Add _a_ into the name so we can keep this before the individual
+  # definitions
+  kselftest_a_qemu: &kselftest_qemu
+    rootfs: debian_bullseye-kselftest_ramdisk
+    pattern: 'kselftest/{category}-{method}-{rootfs}-kselftest-template.jinja2'
+    filters:
+      - passlist: {defconfig: ['kselftest']}
+
   kselftest-alsa:
     <<: *kselftest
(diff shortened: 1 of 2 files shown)


broonie:
Testing OK in staging: https://staging.kernelci.org/test/plan/id/63bd3000394500e08ebb05d6/

The fact that it reports with a different name is a bit annoying but I'm not clear we can do better with the current setup - ideally we'd describe filesystems with multiple formats rather than a specific one, then the one selected would be picked based on the platform boot method.

config/core/test-configs.yaml
```diff
@@ -228,6 +228,12 @@ test_plans:
     filters:
       - passlist: {defconfig: ['kselftest']}
 
+  kselftest_a_qemu: &kselftest_qemu
```
gctucker, line 231: What's the `a` for? Alphabetical order?

broonie: Yes, like the commit log says it was just a bodge so it could stay at the top of the list of kselftests.

gctucker: Ah yes sorry, the PR description only has half of the actual commit message (guess that was an update).  Might be worth a comment in YAML though about that.

broonie: Added a comment.

nuclearcat (approved):
LGTM, results are interesting.
Let's hope it will help us catch regressions before they land in release! :)

PR #1624, opened 2023-01-10
T39784 Add API version to url

JenySadadia:
Depends on https://github.com/kernelci/kernelci-api/pull/209

Implement `KernelCI_API._get_path_with_version_prefix()` to prefix version number to path i.e. `v0/PATH`.
Use the method to create URLs from the provided path in `_make_url` function.


JenySadadia:
After testing this on staging, we can use `pipeline.yaml` to prefix the version to the URL as per our discussion.

JenySadadia:
This was just for testing. Hence closing this one.

PR #1625, opened 2023-01-10
CHROMEOS build_board.sh: Add build improvements

nuclearcat:
We need to address several issues that appear during development: 1)In case of repeating builds, especially if they fail by some reason, we need to keep SDK intact. Introduce "flag file" .keep, which is if present in output directory will prevent script from deleting SDK after build. 2)On repeating builds, if some directories already exist - don't fail


nuclearcat:
Tested by hand, rebuilt trogdor and octopus images

mgalka (approved):

PR #1628, opened 2023-01-11
kernelci.config rework: add get_yaml_attributes() method

gctucker:
Introduce a new method `._get_yaml_attributes()` as an intermediate step towards getting all the configuration objects inherit `yaml.YAMLObject` so they can be parsed and dumped directly by the `yaml` package.  This PR covers all the primitive configuration attributes and leaves out special ones that relate to other specific config objects.  It enables many objects to be serialised back to YAML already.

### DIFF
--- a/kernelci/config/build.py
+++ b/kernelci/config/build.py
@@ -32,16 +32,6 @@ class Tree(YAMLObject):
         self._name = name
         self._url = url
 
-    @classmethod
-    def from_yaml(cls, config, name):
-        kw = {
-            'name': name,
-        }
-        kw.update(cls._kw_from_yaml(config, [
-            'url', 'name',
-        ]))
-        return cls(**kw)
-
     @property
     def name(self):
         return self._name
@@ -49,6 +39,12 @@ class Tree(YAMLObject):
     @property
     def url(self):
         return self._url
+
+    @classmethod
+    def _get_yaml_attributes(cls):
+        attrs = super()._get_yaml_attributes()
+        attrs.update({'url'})
+        return attrs
 
 
 class Reference(YAMLObject):
@@ -103,16 +99,6 @@ class Fragment(YAMLObject):
         self._configs = configs or list()
         self._defconfig = defconfig
 
-    @classmethod
-    def from_yaml(cls, config, name):
-        kw = {
-            'name': name,
-        }
-        kw.update(cls._kw_from_yaml(config, [
-            'name', 'path', 'configs', 'defconfig',
-        ]))
-        return cls(**kw)
-
     @property
     def name(self):
         return self._name
@@ -128,6 +114,12 @@ class Fragment(YAMLObject):
     @property
     def defconfig(self):
         return self._defconfig
+
+    @classmethod
+    def _get_yaml_attributes(cls):
+        attrs = super()._get_yaml_attributes()
+        attrs.update({'path', 'configs', 'defconfig'})
+        return attrs
 
 
 class Architecture(YAMLObject):
@@ -217,16 +209,6 @@ class BuildEnvironment(YAMLObject):
         self._cc_version = str(cc_version)
         self._arch_params = arch_params or dict()
 
-    @classmethod
-    def from_yaml(cls, config, name):
-        kw = {
-            'name': name,
-        }
-        kw.update(cls._kw_from_yaml(config, [
-            'name', 'cc', 'cc_version', 'arch_params',
-        ]))
-        return cls(**kw)
-
     @property
     def name(self):
         return self._name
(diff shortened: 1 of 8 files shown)


gctucker:
OK the issue about `{min,max}_priority` attributes having `None` value is fixed now.

kernelci/config/build.py
```diff
@@ -239,6 +221,16 @@ def cc(self):
     def cc_version(self):
         return self._cc_version
 
+    @property
+    def arch_params(self):
+        return self._arch_params.copy()
```
mgalka, line 226: I don't quite understand a copy of the attribute is returned here? returning a copy of an object with a `.` operator seems a bit counter intuitive.

gctucker: That's to ensure the client code can't change the parameters.  If we return the dictionary object reference then it's not read-only any more.

gctucker: The property is meant to be read-only afaik.  There could be a setter property instead if we wanted this to be mutable.

gctucker:
@mgalka I've rebased this one on top of `main` so it's the next one in the series.

kernelci/config/build.py
mgalka: The property is meant to be read-only in a way that you should be unable to replace the object with another one by a simple assignment. If you decide to return a mutable object from a property it's pretty similar to putting dictionary as an element of a tuple.
While I understand your point I'd probably go with something more explicit like a generator method returning 2-tuples containing param name and value. That'd leave using the values or creating a dictionary out of  them to the code using this value.

Having said that, I guess it's fine to leave it as is, although it makes object's behaviour a bit obscure.

mgalka (approved):
I think we can land it, although returning a copy of an attribute from a property is a bit obscure.
https://github.com/kernelci/kernelci-core/pull/1628/files#r1082717399

gctucker:
> I think we can land it, although returning a copy of an attribute from a property is a bit obscure. https://github.com/kernelci/kernelci-core/pull/1628/files#r1082717399

What do you mean by "obscure" here?  Do you mean you don't really understand the reason for doing it, or that it's obscure for the users of the class what they can do with the attribute?

kernelci/config/build.py
gctucker: We should be adding docstrings at some point and enable pylint on this code.  I'd like to understand why you think there's a problem with returning a dictionary and why a generator with 2-tuples would be better, or any other kinds of concerns so that it can be clarified in the docstring or the implementation can be updated if it's causing some real issues.

I guess I'm seeing this from a C++ point of view and returning the object by value instead of by reference.  I really don't see how it makes a difference from a user point of view whether the attribute returns the reference of the dictionary (which is mutable) or a copy of it as it's exactly the same data structure.

Actually, I've just looked at this again and realised the arch params contain a dictionary for each architecture, so it should be a deep copy to achieve the intended purpose of keeping this read-only.

mgalka:
> What do you mean by "obscure" here? Do you mean you don't really understand the reason for doing it, or that it's obscure for the users of the class what they can do with the attribute?

As I mentioned in the comment I understand the reason of getting a copy of a dictionary which is to protect the class internal dict against modification. What is obscure is the behaviour of the class for the users.
They're using a `.` to get the class attribute, which is a dictionary, so a mutable object. After modifying it the changes will not be reflected in a class. I know it's the expected behaviour from your class point of view, but it's not a standard behaviour of a mutable object. 

Having said that it's fairly easy to see the class source code or use `id()` to figure out that you've been served a copy of an object so I don't find it as something that needs to be changed. Maybe mentioning it in class's docstring could be a good idea , just to point out that this class returns copies of mutable objects from properties.

gctucker:
OK, alternatively we might also just drop this attribute as there are methods for getting particular architecture parameters e.g. `.get_arch_opts()` etc.  So I think I'll rework this part as a follow-up.

There are probably other cases in the code where an attribute is returning a dictionary or some mutable object, so that would need to be reviewed too.

gctucker:
> There are probably other cases in the code where an attribute is returning a dictionary

Namely, `RootFSType.arch_map` in `config/test.py`.  I think that's all.

gctucker:
> it's not a standard behaviour of a mutable object.

@mgalka Actually, the dictionary you get is just mutable so that's not really the issue.  What I think you mean, is you can't do things like `build_env.arch_params.update({...})` as even though it will "work", if you read `build_env.arch_params` again after that you won't get the updated one.

mgalka:
> @mgalka Actually, the dictionary you get is just mutable so that's not really the issue. What I think you mean, is you can't do things like `build_env.arch_params.update({...})` as even though it will "work", if you read `build_env.arch_params` again after that you won't get the updated one.

This is exactly what I meant by non-standard, maybe "non-intuitive" would be more appropriate.

mgalka:
> > There are probably other cases in the code where an attribute is returning a dictionary
> 
> Namely, `RootFSType.arch_map` in `config/test.py`. I think that's all.

Yes, I just thought that it may become a pattern if we need to return dictionaries from YAMLObject. I guess that atm it's just this occurrence.

mgalka:
I guess it's ok to merge it

gctucker:
@mgalka OK I've made #1664 as a follow-up to address this issue in the build config.  I did it on top of #1633 so I could check the unit tests were still passing, and that way it can be reviewed separately from this PR.

gctucker:
BTW, now it's merged I'll update the commit in #1664 with a `Fixes:` trailer.

PR #1630, opened 2023-01-12
config: add tests and use yaml.YAMLObject

gctucker:
Add test cases using `unittest` for loading, dumping and loading again the configuration to check the generated YAML is correct.  Also start using `yaml.YAMLObject` as a base class for the `Tree` and `RootFSType` objects to make the standard code dump them automatically with `yaml.dump()`.

### DIFF
--- a/kernelci/config/base.py
+++ b/kernelci/config/base.py
@@ -24,7 +24,7 @@ import copy
 # Common classes for all config types
 #
 
-class YAMLObject:
+class _YAMLObject:
     """Base class with helper methods to initialise objects from YAML data."""
 
     @classmethod
@@ -32,7 +32,7 @@ class YAMLObject:
         """Load the YAML configuration
 
         Load the YAML configuration passed as a *config* data structure with a
-        given *name*.  This method should return an instance of a YAMLObject
+        given *name*.  This method should return an instance of a _YAMLObject
         subclass.
         """
         yaml_attributes = cls._get_yaml_attributes()
@@ -221,7 +221,7 @@ class Combination(Filter):
         return True
 
 
-class FilterFactory(YAMLObject):
+class FilterFactory(_YAMLObject):
     """Factory to create filters from YAML data."""
 
     _classes = {
--- a/kernelci/config/base.py
+++ b/kernelci/config/base.py
@@ -24,6 +24,55 @@ import copy
 # Common classes for all config types
 #
 
+class YAMLConfigObject(yaml.YAMLObject):
+    """Base class with helper methods to handle configuration YAML data
+
+    This class contains methods to help constructing configuration objects from
+    YAML data.  Then each subclass should implement its standard `to_yaml()`
+    method to be able to dump the whole configuration hierarchy back to YAML.
+    """
+
+    @classmethod
+    def load_from_yaml(cls, config, **kwargs):
+        """Load the YAML configuration
+
+        Load the YAML configuration passed as a *config* data structure with a
+        given *name*.  This method should return an instance of a
+        YAMLConfigObject subclass.
+        """
+        yaml_attributes = cls._get_yaml_attributes()
+        kwargs.update(cls._kw_from_yaml(config, yaml_attributes))
+        return cls(**kwargs)
+
+    @classmethod
+    def _kw_from_yaml(cls, data, attributes):
+        """Create some keyword arguments based on a YAML dictionary
+
+        Return a dictionary suitable to be used as Python keyword arguments in
+        an object constructor using values from some YAML *data*.  The
+        *attributes* are a list of keys to look up from the *data* and convert
+        to a dictionary.  Keys that are not in the YAML data are simply omitted
+        from the returned keywords, relying on default values in object
+        constructors.
+
+        """
+        return {
+            k: v for k, v in ((k, data.get(k))for k in attributes)
+            if v is not None
+        } if data else dict()
+
+    @classmethod
+    def _get_yaml_attributes(cls):
+        """Get a set of YAML attribute names
+
+        Get a set object with all the YAML configuration attribute names for
+        the configuration class.  This can be used to make keyword arguments
+        when creating a configuration object as well as when serialising it
+        back to YAML.
+        """
+        return set()
+
+
 class _YAMLObject:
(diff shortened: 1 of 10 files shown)


gctucker:
@mgalka @nuclearcat This is the next PR in the series, could you please take a look?  The following ones convert all the build config classes to use the new base class and `_YAMLObject` eventually should go away.

kernelci/config/base.py
```diff
@@ -24,15 +24,64 @@
...
+
+        Get a set object with all the YAML configuration attribute names for
+        the configuration class.  This can be used to make keyword arguments
+        when creating a configuration object as well as when serialising it
+        back to YAML.
+        """
+        return set()
+
+
+class _YAMLObject:
```
mgalka, line 76: I don't undertstand the reason behind renaming `YAMLObject` to `_YAMLObject`. 

Usually in Python any name starting with an underscore is meant for internal use. It's clearly specified for method, attribute and function names. I think it's not for class names. In the standard library there is a `_thread` module which is basically provides a low-level API for `threading` module which is supposed to be used in public. 
In this case `_YAMLObject` is used outside of this module which may be a bit confusing.

kernelci/config/base.py
```diff
@@ -24,15 +24,64 @@
...
+    method to be able to dump the whole configuration hierarchy back to YAML.
+    """
+
+    @classmethod
+    def load_from_yaml(cls, config, **kwargs):
+        """Load the YAML configuration
+
+        Load the YAML configuration passed as a *config* data structure with a
+        given *name*.  This method should return an instance of a _YAMLObject
+        subclass.
```
mgalka, line 41: Isn't it returning `YAMLConfigObject` which is a subclass of `yaml.YAMLObject`?

kernelci/config/base.py
gctucker: This is to make it clear `_YAMLObject` is not the standard `yaml.YAMLObject`.

gctucker: We can call it `NotRealYAMLObject` or anything else if you want, but it has to be a different name as the real `yaml.YAMLObject` standard class is now starting to be used.

kernelci/config/base.py
gctucker: Yes, that's a mistake in the docstring.

kernelci/config/test.py
```diff
@@ -255,11 +245,25 @@ def name(self):
     def url(self):
         return self._url
 
-    def _get_attrs(self):
-        attrs = super()._get_attrs()
+    @property
+    def arch_map(self):
+        return self._arch_map.copy()
```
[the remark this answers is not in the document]
gctucker: See #1664

kernelci/config/build.py
```diff
@@ -239,6 +229,16 @@ def cc(self):
     def cc_version(self):
         return self._cc_version
 
+    @property
+    def arch_params(self):
+        return self._arch_params.copy()
```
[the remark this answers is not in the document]
gctucker: See #1664

kernelci/config/test.py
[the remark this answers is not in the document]
gctucker: It's obviously to ensure the property is read-only.  But yes, this should be avoided as discussed before.  I've fixed that for `BuildEnvironment.arch_params`, can do this again here as a follow-up later once we have unit tests in place.

kernelci/config/base.py
mgalka: Ok, I get it, readability matters. I can see couple of ways dealing with it:

- Rename to something like `KciYAMLObject` (maybe a bit too many acronyms here)
- Leave it to namespaces - I think the base approach to make it clearly visible that it's not the `yaml.YAMLObject`

```python
from kernelci.config import base

class MyClass(base.YAMLObject)
```
It can be also left as is. All in all there seem to be no convention defined for `_Class` in PEP8.

tests/test_configs.py
```diff
@@ -55,3 +58,45 @@ def test_architecture_init_name_only():
...
+class ConfigTest(unittest.TestCase):
+
+    def _load_config(self, yaml_file_path):
+        with open(yaml_file_path) as yaml_file:
+            ref_data = yaml.safe_load(yaml_file)
+        config = kernelci.config.load(yaml_file_path)
+        return ref_data, config
+
+    def _reload(self, ref_data, config, name):
+        assert name in config
```
mgalka, line 72: Erm, that's actually not how you write assertions for `unittest.TestCase` based tests. There are special methods to do that https://docs.python.org/3/library/unittest.html#basic-example

mgalka: Maybe this can work if you use pytest as a runner (haven't checked), but even if you write something which is pytest-only then it shouldn't extend `unittests.TestCase`

kernelci/config/base.py
gctucker: The goal is actually to remove `_YAMLObject` by converting all the classes to `yaml.YAMLObject`.  There's a PR for the build configs already, and follow-up ones will cover the rest of them.

tests/test_configs.py
gctucker: Yes, sorry I just stuck to the `pytest` approach even though this is `unittest` (run inside `pytest`).

mgalka: It's still misleading to use `unittest.TestCase` here, especially that the whole `test_configs.py` is written in a pytest convention and is not meant to be run with unittest test runner.
Having said that, I'm all for using pytest, which has a lot of advantages over unittest.

gctucker: Yes I think I was more interested in reusing code from a base class than using `unittest`.  So I'll see if I can rework this with classes and play `pytest`, maybe it's just the case of dropping the `TestCase` inheritance since there's no `setUp()` or `tearDown()`.

gctucker: OK I've reworked this without `unittest`, the base class name just has to start with `Test` as per the naming convention used by `pytest` to discover tests.

kernelci/config/base.py
gctucker: Fixed

gctucker:
@mgalka OK I think this is ready for another review now.

mgalka (approved):
LGTM

PR #1631, opened 2023-01-13
test-configs-riscv.yaml: fix conflict with global default_filters

gctucker:
Rename the default_filters top-level tag with as riscv_default_filters to avoid conflict with the global default_filters defined in test-configs.yaml.  As all the YAML files are loaded together, this was overriding the defaults for all the configs and not just riscv.

Also, as the filters defined in riscv_default_filters are used with anchors, it's just a place to define them within this file so the test_plans tag is not needed.  It is however used with the special default_filters tag as defaults for the Python code that loads the configs and not just YAML anchors.

Fixes: 39d2645e7e4a ("test-configs.yaml: Duplicate the tests to be run on riscv64")

### DIFF
--- a/config/core/test-configs-riscv.yaml
+++ b/config/core/test-configs-riscv.yaml
@@ -1,9 +1,7 @@
-default_filters:
+riscv_default_filters:
 
-  test_plans:
-
-    - blocklist: &kselftest_defconfig_filter
-        defconfig: ['kselftest']
+  - blocklist: &kselftest_defconfig_filter
+      defconfig: ['kselftest']
 
 
 test_plans:


ojeda:
Hmm... not sure why I got the review request, but since I am here, the explanation sounds reasonable.

> Also, as the filters defined in riscv_default_filters are used with anchors, it's just a place to define them within this file so the test_plans tag is not needed.

I only have superficial YAML experience so this is probably obvious but why do we know this? Is it because YAML documents are independent according to the spec (thus anchors cannot be shared) + each `.yaml` is loaded separately in `_iterate_yaml_files` (which is what I would expect, i.e. no hacks concatenating files into a single YAML document)?

> It is however used with the special default_filters tag as defaults for the Python code that loads the configs and not just YAML anchors.

This sentence was a bit confusing due to my lack of context -- is it only saying that the name `default_filters` should not be used for the YAML mapping key to avoid mixing it up with its special meaning for KernelCI (i.e. in the `FilterFactory` logic)? Or is it saying something extra/else?

I hope this all makes some sense, I just glanced through the code...

Also, a general question: could the same problem be automatically detected/prevented in the future somehow?

gctucker:
Sorry @ojeda I think I added you by mistake while I was trying to add @nkbelin as a reviewer :p  Thanks for the comments though!

Yes this issue is not related to how YAML works but `default_filters` is used by the KernelCI code as a particular entry to provide defaults.  Also, all the top-level entries in separate YAML files get merged together by the code which isn't a YAML standard thing either.  So setting `default_filters` in 2 files means they'll get merged together which is not what we want here for riscv.

ojeda:
No worries! I hope the comments (questions, really) are useful nevertheless :)

gctucker:
> Also, a general question: could the same problem be automatically detected/prevented in the future somehow?

I'm actually adding unit tests for the code that handles the YAML config now, and reviewing how the whole thing is set up.  That's how I hit this problem as a side-effect.  I think it's more of a bad design decision to have this special tag as we can just use regular YAML anchors or an explicit attribute to inherit filters with a name.

Also the ability to dump the whole config objects as YAML can help with spotting what actually changes as a result of editing the YAML files.  I'm not sure we can automatically detect when some unwanted changes happen, but at least it should make it easier to debug or verify changes by hand.  I'm reworking the config objects to be able to do that.

ojeda:
Thanks for explaining it to me. Being able to verify/debug them easier sounds good.

nkbelin (approved):
Looks good to me. I did not realise that this would affect the main yaml

gctucker (commented):
Tested OK on staging: https://staging.kernelci.org/build/id/63c1294b5789c867e2bb05f0/

PR #1632, opened 2023-01-13
An example of using the existing --extra-configs

eds-collabora:
This is not buildable in its current form, but it's intended as a baseline for showing how external users can use kernelci to build and test their own custom kernel.


gctucker:
@eds-collabora Would it be possible to add this to the unit tests?  There's a series of tests for the YAML configuration now but it's not covering extra files like this.

nuclearcat:

PR #1636, opened 2023-01-13
Revert "Allow filters of the same type to be combined"

khilman:
This reverts commit e4e70f0bce4954471ebf3bde7ea9444a1c8cc1e3.

I don't fully understand what's going on here, but this patch has some unintended consequences when using `kci_test generate`.

For example, if a single device-type in test-configs has a lab in the blocklist (e.g. beaglebone-black has this):

    filters:
      - blocklist: {lab: ['lab-baylibre']}

then this somehow has the affect of applying that blocklist to *every* device-type, meaning NO jobs are generated for lab-baylibre for *any* device-type.

If you comment out *either* blocklist line from this beaglebone-black filter, then jobs are generated for lab-baylibre.  ?!?!

### DIFF
--- a/kernelci/config/base.py
+++ b/kernelci/config/base.py
@@ -80,31 +80,6 @@ class Filter:
         """Return True if the given *kw* keywords match the filter."""
         raise NotImplementedError("Filter.match() is not implemented")
 
-    def combine(self, items):
-        """Try to avoid making a new filter if we can make a combined
-        filter matching both our existing data and the new items.
-
-        The *items* can be any data used to filter configurations.
-        Return True if we can combine, False otherwise.
-        """
-        return False
-
-
-def _merge_filter_lists(old, update):
-    """Merge the items for a Blocklist or Passlist.
-
-    *old*    is the items from the existing filter list that we
-             have already created.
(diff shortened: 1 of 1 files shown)


khilman:
@mgalka this should be reverted ASAP since it's having the side-effect of having no jobs generated for lab-baylibre.

@eds-collabora could you have a closer look and see why blocklists are being combined for unrelated device-types.

eds-collabora:
@khilman I had a look into this. Essentially what happens is that when we create combined filters, the item dictionaries they contain still share their internal arrays with the default filters. I did not expect that interaction when I wrote the code. There's a candidate fix as #1638, which does not smear the blocklists across random other devices. It might be worth checking whether this solves the problem you saw.

khilman:
Added `staging-skip` so the fix from @eds-collabora can get tested without the revert.

khilman:
Closing this revert since the fix from Ed resolves the issue I was having. 

As to whether we need this feature at all, I admit to being confused by the feature, but TBH I don't fully understand it's utility, so that should probably be brought up in a separate issue.

PR #1638, opened 2023-01-13
kernelci.config.base: Do not corrupt default filter lists

eds-collabora:
This is an unexpected consequence of filter merging: the arrays inside the default filter items we accidentally shared with the filter instances, which meant that, unexpectedly, when merging in new filter items we also corrupted the default filters. This in turn led to filter items being applied inappropriately in other places, in a way that was impossible to predict from the surface YAML.

Signed-off-by: Ed Smith <ed.smith@collabora.com>

### DIFF
--- a/kernelci/config/base.py
+++ b/kernelci/config/base.py
@@ -17,6 +17,7 @@
 
 import re
 import yaml
+import copy
 
 
 # -----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


khilman:
I tried this locally, and it fixes the problem I was seeing.  Thanks @eds-collabora !

Tested-by: Kevin Hilman <khilman@baylibre.com>

khilman (approved):

mgalka (approved):

PR #1639, opened 2023-01-13
kernelci.legacy: Install vmlinux to aid debugging

broonie:
Some debugging, such as the use of scripts/decode_stacktrace.sh and other things involving looking at the disassembly of the built kernel, need the vmlinux to get debug information which doesn't make it into the image files deployed in onto targets for most architectures. Support such uses by including the vmlinux in our shipped artefacts.

Since this is frequently a large file but compresses well use xz to compress as we install. For example, a build I have here is 176M but with xz we can compress it down to 52M. gzip is much less effective at 80M although it is faster, the tradeoff seems worth it given that the binaries are likely to be used infrequently.

### DIFF
--- a/kernelci/build.py
+++ b/kernelci/build.py
@@ -19,6 +19,7 @@ from datetime import datetime
 import fnmatch
 import itertools
 import json
+import lzma as xz
 import os
 import platform
 import re
@@ -869,7 +870,7 @@ class Step:
         """
         raise NotImplementedError("Step.run() needs to be implemented.")
 
-    def install(self, verbose=False, status=True):
+    def install(self, verbose=False, status=True, configs=None):
         """Base method to install the build artifacts.
 
         The default behaviour is to install bmeta.json and steps.json in the
@@ -878,6 +879,7 @@ class Step:
 
         *verbose* is whether to show what is being installed
         *status* is True if install commands succeeded, False otherwise
+        *configs* a BuildConfigs object
         """
         self._add_run_step(status, action='install')
         files = [
@@ -1292,7 +1294,26 @@ class MakeKernel(Step):
             self._add_artifact('kernel', file_name, 'system_map')
             kbmeta['text_offset'] = '0x{:08x}'.format(text_offset)
 
-    def install(self, verbose=False):
+    def _install_vmlinux(self, kbmeta, verbose):
+        file_name = 'vmlinux'
+        vmlinux = os.path.join(self._output_path, file_name)
+        file_name_xz = 'vmlinux.xz'
+        vmlinux_xz = os.path.join(self._output_path, file_name_xz)
+        if os.path.exists(vmlinux):
+            if verbose:
+                print("Creating {}".format(file_name_xz))
+
+            data = open(vmlinux, 'rb').read()
+            zf = xz.open(vmlinux_xz, mode='wb',
+                         format=xz.FORMAT_XZ, preset=9)
+            zf.write(data)
+            zf.close()
+        if os.path.exists(vmlinux_xz):
+            item = self._install_file(vmlinux_xz, 'kernel', file_name_xz,
+                                      verbose)
+            self._add_artifact('kernel', file_name_xz, 'vmlinux.xz')
+
+    def install(self, verbose=False, configs=None):
         """Install the kernel image
 
         Install the Linux kernel image as well as System.map.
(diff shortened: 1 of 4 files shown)


broonie:
File present on staging: https://storage.staging.kernelci.org/kernelci/staging-mainline/staging-mainline-20230115.0/arm64/defconfig/gcc-10/kernel/ and listed in the YAML: https://storage.staging.kernelci.org/kernelci/staging-mainline/staging-mainline-20230115.0/arm64/defconfig/gcc-10/artifacts.json

gctucker:
> kernelci.legacy: Install vmlinux to aid debugging

This change is actually in `kernelci.build`.

gctucker:
> Since this is frequently a large file but compresses well use xz to compress as we install. For example, a build I have here is 176M but with xz we can compress it down to 52M. gzip is much less effective at 80M although it is faster, the tradeoff seems worth it given that the binaries are likely to be used infrequently.

Yes I think this is the actual reason why it's not sent to storage at the moment.  If we do that, we need to extend our storage disk by a significant amount (to be determined) or reduce the garbage collection period.

gctucker:
Maybe a reasonable compromise would be to only store vmlinux on the defconfig builds rather than all the builds?

broonie:
That might work - I'd suggest also including the kselftest builds in there given what it's testing, there are some kselftests that explicitly create oopses. We could also do it per tree, for example focusing on mainline and next. We could also look into something like having a faster expiration policy for vmlinux files - probably if someone hasn't looked at a vmlinux for say a week it's probably safer to delete it for example - or having cold storage in the Azure equivalent of a S3 bucket that we shuffle things off to.

gctucker:
A dedicated service with the new pipeline could be using any separate arbitrary storage for that.

OK so I guess we could add an optional parameter to the build configs e.g.:
```yaml
params:
    upload_vmlinux: true
```
for now?  Then the code in `build.py` would use that to implement its `is_enabled()` method.  It's not perfect design but it would be not too intrusive and a fair compromise I believe.

broonie:
Yes, sounds OK for now - we can remove it later if we get less constrained on storage one way or another.

kernelci/build.py
```diff
@@ -1286,6 +1286,21 @@ def _install_system_map(self, kbmeta, verbose):
...
 
+    def _install_vmlinux(self, kbmeta, verbose):
+        file_name = 'vmlinux'
+        vmlinux = os.path.join(self._output_path, file_name)
+        file_name_xz = 'vmlinux.xz'
+        vmlinux_xz = os.path.join(self._output_path, file_name_xz)
+        if os.path.exists(vmlinux):
+            if verbose:
+                print("Creating {}".format(file_name_xz))
+            shell_cmd("xz -T0 --compress < {src} > {dest}".format(
```
nuclearcat, line 1297: Probably you can use same approach as we have already in code? (i know it ingest whole file in RAM, but i think it is fine with size we have):
https://github.com/kernelci/kernelci-core/blob/4c6c5065c4e9a0b57bbedc148db6edc517876764/kci_build#L373

broonie: I can't tell what you're suggesting here, sorry.

nuclearcat:
This patch causes significant storage usage increase:
root@kernelci-staging:/data/storage.staging.kernelci.org/next# du --max-depth=1 -h
10G	./master
10G	.
root@kernelci-staging:/data/storage.staging.kernelci.org/next# find . -name "vmlinux.xz" -print0 | xargs -0 du -ch | tail -n 1
3.6G	total

So from total 10G it takes 3.6G.

nuclearcat:
Extract from KCI weekly:
We need to do selective builds and build vmlinux only one arm, one for x86 (chromebook+kselftest+etc+some config options for debugging trace), and one for arm64.

gctucker:
@broonie Could you please update this PR to only store `vmlinux` on the small subset of builds as per the comment above?

broonie:
If we are switching to object storage we may want to revisit saving only sometimes, it should vastly reduce costs.

nuclearcat:
This patch likely causing breakage on staging:
```
  File "/usr/local/bin/kci_build", line 4, in <module>
    __import__('pkg_resources').run_script('kernelci==1.1', 'kci_build')
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 651, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 1448, in run_script
    exec(code, namespace, namespace)
  File "/usr/local/lib/python3.9/dist-packages/kernelci-1.1-py3.9.egg/EGG-INFO/scripts/kci_build", line 480, in <module>
  File "/usr/local/lib/python3.9/dist-packages/kernelci-1.1-py3.9.egg/EGG-INFO/scripts/kci_build", line 304, in __call__
    install_res = self._install_step(step, configs, args)
  File "/usr/local/lib/python3.9/dist-packages/kernelci-1.1-py3.9.egg/EGG-INFO/scripts/kci_build", line 317, in _install_step
    return step.install(args.verbose, configs)
TypeError: install() takes from 1 to 2 positional arguments but 3 were given
Build result: 1
```
Still investigating, will edit this post after run completed.

nuclearcat:

PR #1640, opened 2023-01-14
config/docker: Add architecture specific clang templates

broonie:
Since clang provides a single toolchain binary which supports all target architectures we currently provide a single clang container for each version which is used for all architecture builds.  These containers include some very basic cross compilation support required for things like the vDSO but do not include enough to build real userspace applications since those generally also need libgcc, and adding our current kselftest fragments will only pull in the libraries for one architecture.

Rather than installing all userspace libraries for all architectures follow the pattern for GCC and add per architecture clang templates, layerd on top of the existing architecture neutral ones.  These templates add libgcc and can also be used with the kselftest fragment or other fragments that install userspace libraries.  The existing templates are left in place so existing usage is not disrupted during deployment and so that we share as much of the layers for each clang version as possible.

As well as enabling much more succesful kselftest builds with clang this will allow us to remove some of the compiler dependent special casing.

The MIPS and RISC-V templates are currently stub templates which will not actually work for userspace, this at least allows builds to be converted to use the per-architecture packages without causing regressions for them or requring per architecture special casing.  In the case of MIPS the existing GCC template also does not work for userspace builds and it is not clear what should be done, especially given the multiple subarchitectures.  In the case of RISC-V it is an external Debian port with different toolchain packaging that doesn't follow the same pattern so will need handling appropriate to that packaging.

### DIFF
--- a/config/docker/clang-11-arm.jinja2
+++ b/config/docker/clang-11-arm.jinja2
@@ -1,0 +1,13 @@
+{%- set sub_arch = 'armhf' %}
+{% extends 'clang-11.jinja2' %}
+
+{% block packages %}
+{{ super() }}
+
+RUN dpkg --add-architecture armhf
+
+RUN apt-get update && apt-get install --no-install-recommends -y \
+    libgcc-10-dev:armhf
+
+{%- endblock %}
+
--- a/config/docker/clang-11-arm64.jinja2
+++ b/config/docker/clang-11-arm64.jinja2
@@ -1,0 +1,13 @@
+{%- set sub_arch = 'arm64' %}
+{% extends 'clang-11.jinja2' %}
+
+{% block packages %}
+{{ super() }}
+
+RUN dpkg --add-architecture arm64
+
+RUN apt-get update && apt-get install --no-install-recommends -y \
+    libgcc-10-dev:arm64
+
+{%- endblock %}
+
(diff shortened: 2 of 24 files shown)


broonie:
@khilman @nkbelin I could use some help with RISC-V for this - the RISC-V toolchain packaging is slightly different and it wasn't immediately clear what to do. As noted in the commit log I've left a stub container in place so there should be no regression if/when we start using these so it's not super urgent.

broonie:
Pull request to enable builds of containers with these templates: https://github.com/kernelci/kernelci-deploy/pull/48

gctucker:
It might be a better idea to keep the plain clang images for regular kernel builds, then also have a collections for kselftest with variants for each architecture.  So one base image, and a series of arch-specific variants for kselftest.  That way we keep the base image for kernel builds with just what it needs, which helps with interoperability.

broonie:
That's actually how the new images are constructed - the architecture specific stuff for kselftest is layered on top of the existing images which are unmodified, adding the userspace build dependencies for each architecture. We could just keep things that way if that's preferred rater than doing further refactoring of the image construction.

gctucker:
Ah great, sorry I guess I didn't look enough into the details of it.  It's also something we should be documenting a bit more, as a follow-up.

khilman:
> @khilman @nkbelin I could use some help with RISC-V for this - the RISC-V toolchain packaging is slightly different and it wasn't immediately clear what to do. As noted in the commit log I've left a stub container in place so there should be no regression if/when we start using these so it's not super urgent.

Just FYI: risc-v updates submitted here: https://github.com/kernelci/kernelci-core/pull/1653

PR #1641, opened 2023-01-15
Enable libhugetlbfs testsuite

broonie:
The libhugetlbfs testsuite is a small, quick running and reliable testsuite which is really good for finding issues in the kernel's memory management subsystem, let's enable it for use in KernelCI. This involves the addition of a new filesystem image containing the testsuite which we then use with the existing LAVA test definitions.

### DIFF
--- a/config/core/rootfs-configs.yaml
+++ b/config/core/rootfs-configs.yaml
@@ -211,6 +211,21 @@ rootfs_configs:
     script: "scripts/bullseye-libcamera.sh"
     test_overlay: "overlays/libcamera"
 
+  bullseye-libhugetlbfs:
+    rootfs_type: debos
+    debian_release: bullseye
+    arch_list:
+      - amd64
+      - arm64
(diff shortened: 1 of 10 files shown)


gctucker:
Do you have a staging Jenkins account to build the rootfs image?

broonie:
I have access to the admin account, it looks like I can go trigger the rootfss - is it OK to just go do that whenever? There's also all my bootrr changes pending which need rootfs builds too.

gctucker:
Yes, it's needed if we want to have the changes tested on staging.  Then the clunky bit is that the URL needs to be changed to use the staging rootfs image, typically that's done with a draft PR in test-configs.yaml.

The buildroot images can be built just in the same way, there's no md5 check on staging builds.

nuclearcat (approved):
Works great on staging: https://staging.kernelci.org/test/plan/id/63c7af4575303a19a10c530e/

PR #1643, opened 2023-01-15
Enable LTP syscalls tests

broonie:
The LTP syscalls test is a big suite but does cover a lot of ground in the ABI, enable it on BeagleBone Black so we get some coverage from it.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -2805,6 +2805,7 @@ test_configs:
       - ltp-ipc
       # ltp-mm - Runs board out of memory
       - ltp-pty
+      - ltp-syscalls
       - ltp-timers
 
   - device_type: meson-gxm-khadas-vim2


broonie:
Skipping for now since this seems to be interfering with the smaller test jobs in ltp-misc and the BeagleBone Black is just too slow - it's taking over 2 hours, looking at runs for more modern boards I suspect Tritium would be a better fit (possibly Potato too).

broonie:
This is restricted to run on mainline and -next but neither of those is run directly in staging so this won't trigger... any suggestions?

broonie:
The tests run fine but don't appear on the dashboard, I suspect there's scaling issues in the server side.

gctucker:
@broonie Do you have links to LAVA jobs running these tests with staging builds?

broonie:
We'll need to wait for it to run again, it was skipped prior to being closed. Though now I think about it there might also have been issues with the database struggling to ingest things.

broonie:
It's not causing any issues for anything else - removed the label and let's see what's happening.

gctucker (commented):
Tests got run but they're all kind of failing, seems more like a LAVA issue:
https://staging.kernelci.org/test/job/next/branch/master/kernel/next-20230331/plan/ltp-syscalls/
https://storage.staging.kernelci.org/next/master/next-20230331/arm64/defconfig/clang-17/lab-broonie/ltp-syscalls-meson-gxl-s905x-libretech-cc.html

Could this be just a timeout issue?

gctucker:

nuclearcat:

PR #1645, opened 2023-01-15
Enable smaller LTP test sets

broonie:
There are a number of smaller LTP test sets that provide some kernel level coverage which we don't currently run. This pull request provides the basic mapping of these tests into KernelCI.

To make merging a bit easier all the tests are only enabled on Tritium for now. In order to avoid conflicts in staging it also incorporates the commit adding the basic stanza for the larger syscalls test set, that is also in a separate pull request to enable that test set which adds enablement of the test set.

If this gets merged then there should most likely be followups revisiting where we run the tests.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -3170,7 +3170,14 @@ test_configs:
       - kselftest-seccomp
       - kselftest-timers
       - libhugetlbfs
+      - ltp-cpuhotplug
       - ltp-crypto
+      - ltp-dio
+      - ltp-fsx
+      - ltp-hugetlb
+      - ltp-io
+      - ltp-sched
+      - ltp-smoketest
       - preempt-rt
 
   - device_type: sun50i-h5-nanopi-neo-plus2
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -441,6 +441,18 @@ test_plans:
       job_timeout: '15'
       skipfile: skipfile-lkft.yaml
 
+  ltp-cap-bounds:
+    <<: *ltp
+    params:
+      <<: *ltp-params
+      tst_cmdfiles: "cap_bounds"
+
+  ltp-cpuhotplug:
+    <<: *ltp
+    params:
+      <<: *ltp-params
+      tst_cmdfiles: "cpuhotplug"
+
   ltp-crypto:
     <<: *ltp
     params:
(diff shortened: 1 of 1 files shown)


gctucker:
When we looked into the LTP tests we could run with @Lakshmipathi a couple of years ago we came up with a list of about 40 I believe.  Then a subset was picked with a few platforms based on what seemed like the most relevant ones, see this documentation page: https://kernelci.org/docs/tests/ltp/

In theory we should enable all of the tests that cover the kernel, however it's really tedious to manage that in the current YAML configuration due to the fact that each combination of test plan / target platform has to be curated by hand.  So I think that's why there hasn't been a push for expanding coverage.  The idea of having a more dynamic scheduling should simplify this greatly, and then we'll be able to enable pretty much everything to then let the scheduler run the tests on some platforms selected automatically.

In the meantime, I would suggest that only test cases that are really likely to show some kernel problems should be enabled as otherwise it'll be adding some overhead for both maintaining the configs and for the labs to run the extra tests.  It also adds more results on the current web dashboard which won't scale with all the possible tests we could run, the table with test plans doesn't even do pagination.

broonie:
So, this is motivated by me looking at a combination of what we're wanting for the internal KernelCI deployment at work and also by the capacity I have available in my lab - there's some boards I've got a lot of so they're able to keep up with the current test load. For the work stuff it would be helpful to at least get the generic definitions merged even if we don't enable them on any boards since it makes it easier to share the generic "here's how this test fits into KernelCI" bit, as well as the lab capacity the other bit of the board enablement is me knowing that people will want to see running jobs for review purposes.

A better scheduler would definitely help with this stuff, we've got somewhere with the existing use of priorities (in that LAVA gives us some of the scheduling) but there's more that could be done, the other thing which would help would be to coalesce multiple tests into a single LAVA job. I've got a design for that but I've been holding off for the pipleine stuff to get to a point where it makes sense to start looking at. Combining the jobs would be especially good for the smaller test you see with some of LTP or especially in kselftest, there's a lot of things where we spend more time deploying and booting the boards than running the tests so even without doing anything particularly smart we'd be able to get substantially more capacity on our existing board set.

nuclearcat (changes requested):
https://staging.kernelci.org/test/job/kernelci/branch/staging-stable/kernel/staging-stable-20230119.0/
shows dio, filecaps, fsx, and others on dashboard, but i see several issues with them:
https://storage.staging.kernelci.org/kernelci/staging-stable/staging-stable-20230119.0/arm64/defconfig/gcc-10/lab-broonie/ltp-cap-bounds-sun50i-h5-libretech-all-h3-cc.html
```
 1506 05:27:35.103698  cap_bounds_r    1  TCONF  :  cap_bounds_r.c:103: System doesn't have POSIX capabilities.
 ```
 also i can see "Tests which require a big block device are disabled."
 test doesn't really run.
 filecaps - same.
 dio - works
 fsx - works
smoketest - works

i suggest to exclude cap-bounds and filecaps, then we can merge this PR

broonie:
Done

nuclearcat:
Thank you! :)

PR #1649, opened 2023-01-17
kernelci.cli: move Args.storage to kernelci.legacy

gctucker:
The --storage command line argument is only used with legacy commands now so move it to kernelci.legacy accordingly.


mgalka (approved):
LGTM

PR #1650, opened 2023-01-17
Add kernelci.config.api

gctucker:
Add kernelci.config.api to load the YAML configuration required to access a KernelCI API instance.  This is intended to replace the config.db which was about providing an abstraction layer for various API and backend / database implementations.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -52,6 +52,7 @@ jobs:
         run: |
           pylint kci
           pylint kernelci.cli
+          pylint kernelci.config.api
           pylint kernelci.storage
 
       - name: Run YAML config validation
--- a/config/core/api-configs.yaml
+++ b/config/core/api-configs.yaml
@@ -1,0 +1,7 @@
+api_configs:
+
+  docker-host:
+    url: http://172.17.0.1:8001
+
+  staging.kernelci.org:
+    url: https://staging.kernelci.org:9000
--- a/config/runtime/base/python.jinja2
+++ b/config/runtime/base/python.jinja2
@@ -16,13 +16,13 @@ import yaml
 {% endblock %}
 
 {%- block python_local_imports %}
-import kernelci.db
+from kernelci.db import kernelci_api
 import kernelci.config
 {%- endblock %}
 
 {%- block python_globals %}
-DB_CONFIG_YAML = """
-{{ db_config_yaml }}"""
+API_CONFIG_YAML = """
+{{ api_config_yaml }}"""
 NODE_ID = '{{ node_id }}'
 TARBALL_URL = '{{ tarball_url }}'
 WORKSPACE = '/tmp/kci'
@@ -34,18 +34,18 @@ class BaseJob:
     def __init__(self, workspace):
         self._workspace = workspace
 
-    def _get_db(self, db_config_yaml):
-        db_config = kernelci.config.db.DatabaseFactory.from_yaml(
-            'db', yaml.load(db_config_yaml, Loader=yaml.CLoader)
+    def _get_api(self, api_config_yaml):
+        api_config = kernelci.config.api.API.load_from_yaml(
+            yaml.safe_load(api_config_yaml), name='api'
         )
-        if not db_config:
-            return None
+        if not api_config:
+            raise ValueError("API configuration not found")
 
         api_token = os.getenv('API_TOKEN')
         if not api_token:
-            return None
+            raise ValueError("API_TOKEN environment variable not found")
 
-        return kernelci.db.get_db(db_config, api_token)
+        return kernelci_api.KernelCI_API(api_config, api_token)
 
     def _get_source(self, url):
         resp = requests.get(url, stream=True)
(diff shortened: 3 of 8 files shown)


JenySadadia:
Staging is not working with these changes:
```
today at 11:05:2502/07/2023 05:35:25 AM UTC [ERROR] Traceback (most recent call last):
today at 11:05:25  File "/home/kernelci/pipeline/base.py", line 68, in run
today at 11:05:25    status = self._run(context)
today at 11:05:25  File "/home/kernelci/./pipeline/trigger.py", line 96, in _run
today at 11:05:25    self._iterate_build_configs(force, build_configs_list)
today at 11:05:25  File "/home/kernelci/./pipeline/trigger.py", line 74, in _iterate_build_configs
today at 11:05:25    self._run_trigger(config, force)
today at 11:05:25  File "/home/kernelci/./pipeline/trigger.py", line 37, in _run_trigger
today at 11:05:25    node_list = self._db.count_nodes({
today at 11:05:25  File "/usr/local/lib/python3.10/site-packages/kernelci-1.1-py3.10.egg/kernelci/db/kernelci_api.py", line 109, in count_nodes
today at 11:05:25    resp = self._get('count', params=attributes)
today at 11:05:25  File "/usr/local/lib/python3.10/site-packages/kernelci-1.1-py3.10.egg/kernelci/db/kernelci_api.py", line 50, in _get
today at 11:05:25    url = self._make_url(path)
today at 11:05:25  File "/usr/local/lib/python3.10/site-packages/kernelci-1.1-py3.10.egg/kernelci/db/kernelci_api.py", line 40, in _make_url
today at 11:05:25    version_path = '/'.join((self.config.version, path))
today at 11:05:25AttributeError: 'DatabaseAPI' object has no attribute 'version'
today at 11:05:25
today at 11:25:31Container stopped
```

JenySadadia (commented):
Shouldn't the `version` attribute be added to `config/core/api-configs.yaml`?

gctucker:
> Shouldn't the `version` attribute be added to `config/core/api-configs.yaml`?

There's a default value `latest`.  We could decide to drop the default.

JenySadadia:
> > Shouldn't the `version` attribute be added to `config/core/api-configs.yaml`?
> 
> There's a default value `latest`. We could decide to drop the default.

The default should be added to config file instead to modify it easily without touching the code.

gctucker:
> > > Shouldn't the `version` attribute be added to `config/core/api-configs.yaml`?
> > 
> > 
> > There's a default value `latest`. We could decide to drop the default.
> 
> The default should be added to config file instead to modify it easily without touching the code.

I don't understand what you mean.  The default is in the code so it doesn't need to be specified in the YAML config.  If we have to define the version in the YAML config, then it's not a default any more.

JenySadadia:
> > > > Shouldn't the `version` attribute be added to `config/core/api-configs.yaml`?
> > > 
> > > 
> > > There's a default value `latest`. We could decide to drop the default.
> > 
> > 
> > The default should be added to config file instead to modify it easily without touching the code.
> 
> I don't understand what you mean. The default is in the code so it doesn't need to be specified in the YAML config. If we have to define the version in the YAML config, then it's not a default any more.

Do you think is it beneficial to discard the default value from the constructor and add it to the `api-configs.yaml` instead?
I think that way it can be altered more easily if required in the future.

JenySadadia (changes requested):
There is a typo in the second last commit message:

Add a version attribute to the API YAML configuration.  This can be
used in the URLs for making queries and also to create an API object
that **matches** the API version when we start supporting multiple
versions.

gctucker:
> Do you think is it beneficial to discard the default value from the constructor and add it to the `api-configs.yaml` instead?
> I think that way it can be altered more easily if required in the future.

I think `latest` is a special value used in the API and it makes sense as a default here.  So if someone sets up an instance without specifying a particular version in YAML then it will be using the latest.

JenySadadia:
> > Do you think is it beneficial to discard the default value from the constructor and add it to the `api-configs.yaml` instead?
> > I think that way it can be altered more easily if required in the future.
> 
> I think `latest` is a special value used in the API and it makes sense as a default here. So if someone sets up an instance without specifying a particular version in YAML then it will be using the latest.

Makes sense. So is it possible to configure it from the pipeline?

gctucker:
> Makes sense. So is it possible to configure it from the pipeline?

Just like any part of the YAML configuration, yes.

gctucker (commented):
Tested OK on staging:
```
[INFO] Listening for events... 
[INFO] Press Ctrl-C to stop.
[INFO] Time                        Commit        Node Id                   State      Result    Name
[INFO] 2023-02-10 17:35:40.579328  f3fbc982f678  63e6806c2a8c8ff21d312f52  Running    -         checkout
[INFO] 2023-02-10 17:36:37.622352  f3fbc982f678  63e6806c2a8c8ff21d312f52  Available  -         checkout
[INFO] 2023-02-10 17:36:37.718588  f3fbc982f678  63e680a52a8c8ff21d312f53  Running    -         kver
[INFO] 2023-02-10 17:36:37.836213  f3fbc982f678  63e680a52a8c8ff21d312f54  Running    -         kunit
[INFO] 2023-02-10 17:37:01.810400  f3fbc982f678  63e680a52a8c8ff21d312f53  Done       Pass      kver
[INFO] 2023-02-10 17:44:12.781973  2a71288b233e  63e67fb12a8c8ff21d312f4f  Closing    -         checkout
```
The `kunit` job seems to have hit an issue but I think that's another kind of problem.  I'll be investigating that next.

Runner log:
```
02/10/2023 05:35:10 PM UTC [INFO] Listening for available checkout events
02/10/2023 05:35:10 PM UTC [INFO] Press Ctrl-C to stop.
Getting kernel source tree...
Source directory: /tmp/kci/linux-kernelci-staging-mainline-staging-mainline-20230210.8
Running job...
Checking kernel revision
Result: pass
```

PR #1656, opened 2023-01-19
test-configs-cip: update isar-cip-core rootfs

aliceinwire:

### DIFF
--- a/config/core/build-configs-cip.yaml
+++ b/config/core/build-configs-cip.yaml
@@ -237,6 +237,40 @@ build_configs:
               - 'allnoconfig'
               - 'cip://5.10.y-cip/riscv/qemu_riscv64_defconfig'
 
+  cip_6.1:
+    tree: cip
+    branch: 'linux-6.1.y-cip'
+    variants:
+      gcc-10:
+        build_environment: gcc-10
+        architectures:
+          arm:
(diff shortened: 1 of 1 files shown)


aliceinwire:
@nuclearcat can you check and merge this? thanks

gctucker:
Yes, latest is this one: https://storage.kernelci.org/images/rootfs/cip/20230321/

gctucker:
@aliceinwire Could you please update this PR with the latest CIP rootfs images?

config/core/build-configs-cip.yaml
```diff
@@ -237,6 +237,40 @@ build_configs:
               - 'allnoconfig'
               - 'cip://5.10.y-cip/riscv/qemu_riscv64_defconfig'
 
+  cip_6.1:
```
patersonc, line 240: Hello @aliceinwire 

I believe we need to move this below `cip_5.10-rt` to maintain alpha order.

Thanks, Chris

aliceinwire: this is just the pr for updating the isar-cip-core not 6.1. I used same branch for another pr and the commit come also here. by the way for replaying to your question, there is no alpha order, the order is checked (I think alphabetically) by githubCI and that's the only order that doesn't give an error on the PR

PR #1658, opened 2023-01-19
vDSO tests bodge

broonie:


broonie:
All the stable trees and possibly some others are more active than usual so are DOSing the labs for the time being.

config/core/rootfs-images.yaml
```diff
@@ -115,6 +115,40 @@ file_systems:
     ramdisk: bullseye/20230127.0/{arch}/rootfs.cpio.gz
     root_type: ramdisk
     type: debian
+  debian_bullseye-vdso_nfs:
+    boot_protocol: tftp
+    nfs: bullseye-vdso/20230116.1/{arch}/full.rootfs.tar.xz
```
mgalka, line 120: It looks that the `20230116.1` version of all `bullseye-vdso` images is not on present at [staging storage](https://storage.staging.kernelci.org/images/rootfs/debian/bullseye-vdso). There is one for `20230119.0` which has images for `armhf` and `arm64`. I also built all images yesterday. @broonie Can you update the paths?

```suggestion
    nfs: bullseye-vdso/20230202.0/{arch}/full.rootfs.tar.xz
```

PR #1660, opened 2023-01-19
Enable kvm-unit-tests

broonie:
The kvm-unit-tests provide low level coverage of KVM functionality, as the name suggests being very simple low level guests which exercise functionality in a much more focused fashion than running a full guest OS would.

This series depends on the very latest test definitions upstream, specifically https://github.com/Linaro/test-definitions/commit/755466ad638e89c4ddd95a93f3479b24f6e23ad3 - otherwise the only option is to build the testsuite at runtime.

### DIFF
--- a/config/core/rootfs-configs.yaml
+++ b/config/core/rootfs-configs.yaml
@@ -43,6 +43,22 @@ rootfs_configs:
       - xz-utils
     script: "scripts/bookworm-kselftest.sh"
 
+  bookworm-kvm-unit-tests:
+    rootfs_type: debos
+    debian_release: bookworm
+    arch_list:
+      - amd64
+      - arm64
+    extra_packages:
+      - python3-minimal
+      - qemu-system
+    extra_packages_remove:
+      - bash
+      - e2fslibs
+      - e2fsprogs
+    debos_scratchsize: "10G"
+    script: "scripts/bookworm-kvm-unit-tests.sh"
+
   buildroot-baseline:
     rootfs_type: buildroot
     git_url: https://github.com/kernelci/buildroot
--- a/config/core/rootfs-images.yaml
+++ b/config/core/rootfs-images.yaml
@@ -80,6 +80,23 @@ file_systems:
     ramdisk: bullseye-kselftest/20230623.0/{arch}/rootfs.cpio.gz
     root_type: ramdisk
     type: debian
+  debian_bookworm-kvm-unit-tests_nfs:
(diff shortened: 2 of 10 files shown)


broonie:
This doesn't seem to be taking on staging - I'm assuming it's conflicting with something but any ideas what or how likely that is to progress?

config/core/rootfs-configs.yaml
```diff
@@ -195,21 +195,32 @@ rootfs_configs:
...
+
   bullseye-libcamera:
     rootfs_type: debos
     debian_release: bullseye
     arch_list:
       - amd64
       - arm64
       - armhf
-    extra_packages:
-      - libevent-dev
```
nfraprado, line 206: Accidental changes? This is a different rootfs. Same below.

broonie: Ugh, this is junk due to having to rebase this for so long - git must've got confused at some point.

nuclearcat (commented):
Please check https://github.com/kernelci/kernelci-core/pull/1753 , i am not sure if this particular PR have issue or PR 1753.

broonie:
That's merged now, we just need a sync with upstream.

gctucker (commented):
No results on staging:
https://staging.kernelci.org/test/job/next/branch/master/kernel/next-20230331/

I've no idea why that is.  @broonie Could you please take a look?  It seems something is broken somewhere.  I guess Jenkins logs on staging could provide additional information, and checking locally with `kci_test` that the jobs are getting generated.

broonie:
This is the thing I keep on raising where the rootfs builds have started constantly failing due to what appears to be some sort of infra issue with running out of space in the target image during debos processing of Debian packages.

We're using this in production internally FWIW.

gctucker:
Ah OK so the problem is that the rootfs image is missing.

@mgalka Could you please take a look?  I think this might be related to the `debos` configuration thing with the rootfs image size.

gctucker:
I believe it's not an infrastructure error but a limitation of how debos creates images, it has a fixed maximum size for the rootfs which can be configured so we might just need to add a parameter in `rootfs.yaml` to extend it.

broonie:
From a user point of view it's infra - it's something I can't change (and something which started failing after I pushed the PR). @nuclearcat did mention making this configurable but I've no idea what the status is on that.

gctucker:

nuclearcat:

PR #1661, opened 2023-01-20
CHROMEOS build-configs-chromeos.yaml: Add Kernel 6.1 builds

nuclearcat:
Update and add to build variants Kernel 6.1 as it is new LTS, and possible target for future uprevs.

### DIFF
--- a/config/core/build-configs-chromeos.yaml
+++ b/config/core/build-configs-chromeos.yaml
@@ -77,13 +77,43 @@ chromeos_5.15_variants: &chromeos_5_15_variants
   clang-14: *clang-14
 
 
+chromeos_6.1_variants: &chromeos_6_1_variants
+  chromeos-clang-14:
+    build_environment: clang-14
+    architectures:
+      arm:
+        base_defconfig: 'cros://chromeos-6.1/armel/chromiumos-arm.flavour.config'
+        extra_configs:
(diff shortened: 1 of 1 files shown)


nuclearcat:
Many Chromebooks just failing to boot, need to investigate reasons: https://chromeos.kernelci.org/test/job/kernelci/branch/chromeos-stable/kernel/chromeos-stable-20230121.1/

nuclearcat:
Which is actually very strange, because single stage boot works fine:
https://lava.collabora.dev/scheduler/job/8812476
And dual stage with same kernel fails to boot second kernel:
https://lava.collabora.dev/scheduler/job/8812783

mgalka:
@nuclearcat Is my impression right that this is still work in progress? If it's the case, please convert the PR to draft.

nuclearcat:
It was supposed to go test phase, but turned out kernel 6.1 compiled with clang doesnt boot on many devices.
https://github.com/kernelci/kernelci-project/issues/166

nuclearcat:
Turned out regressions are real kernel bug.
Otherwise 6.1 works fine: https://chromeos.kernelci.org/build/kernelci/branch/chromeos-stable/kernel/chromeos-stable-20230126.0/

mgalka (approved):

Issue #1619, opened 2023-01-09
Building bullseye-v4l2 rootfs image fails


mgalka:
Building `bullseye-v4l2` rootfs image files on `add firmware files` action in debos.

```
2023/01/09 10:13:17 ==== add firmware files ====
2023/01/09 10:13:18 venus.mdt | hint: Using 'master' as the name for the initial branch. This default branch name
2023/01/09 10:13:18 venus.mdt | hint: is subject to change. To configure the initial branch name to use in all
2023/01/09 10:13:18 venus.mdt | hint: of your new repositories, which will suppress this warning, call:
2023/01/09 10:13:18 venus.mdt | hint: 
2023/01/09 10:13:18 venus.mdt | hint: 	git config --global init.defaultBranch <name>
2023/01/09 10:13:18 venus.mdt | hint: 
2023/01/09 10:13:18 venus.mdt | hint: Names commonly chosen instead of 'master' are 'main', 'trunk' and
2023/01/09 10:13:18 venus.mdt | hint: 'development'. The just-created branch can be renamed via this command:
2023/01/09 10:13:18 venus.mdt | hint: 
2023/01/09 10:13:18 venus.mdt | hint: 	git branch -m <name>
2023/01/09 10:13:18 venus.mdt | Initialized empty Git repository in /tmp/linux-firmware/.git/
2023/01/09 10:13:38 venus.mdt | From https://git.kernel.org/pub/scm/linux/kernel/git/firmware/linux-firmware
2023/01/09 10:13:38 venus.mdt |  * branch            HEAD       -> FETCH_HEAD
2023/01/09 10:13:44 venus.mdt | Note: switching to 'FETCH_HEAD'.
2023/01/09 10:13:44 venus.mdt | 
2023/01/09 10:13:44 venus.mdt | You are in 'detached HEAD' state. You can look around, make experimental
2023/01/09 10:13:44 venus.mdt | changes and commit them, and you can discard any commits you make in this
2023/01/09 10:13:44 venus.mdt | state without impacting any branches by switching back to a branch.
2023/01/09 10:13:44 venus.mdt | 
2023/01/09 10:13:44 venus.mdt | If you want to create a new branch to retain commits you create, you may
2023/01/09 10:13:44 venus.mdt | do so (now or later) by using -c with the switch command. Example:
2023/01/09 10:13:44 venus.mdt | 
2023/01/09 10:13:44 venus.mdt |   git switch -c <new-branch-name>
[...]

PR #1667, opened 2023-01-21
rootfs-configs.yaml: Fix qcom/venus-5.4/venus firmware files issues

nuclearcat:
Recently firmware structure was changed in commit: https://git.kernel.org/pub/scm/linux/kernel/git/firmware/linux-firmware.git/commit/?id=ad9fdba
So now we have just two files.
But one of them missing due mistake.
I've sent fix upstream also: https://lore.kernel.org/linux-firmware/5dc7081a8b793baf08f20df6b641a86a80c9e15c.camel@collabora.com/T/#u
Fixes: https://github.com/kernelci/kernelci-core/issues/1619

### DIFF
--- a/config/core/rootfs-configs.yaml
+++ b/config/core/rootfs-configs.yaml
@@ -268,12 +268,8 @@ rootfs_configs:
       - e2fslibs
       - e2fsprogs
     extra_firmware:
-        - qcom/venus-5.4/venus.b00
-        - qcom/venus-5.4/venus.b01
-        - qcom/venus-5.4/venus.b02
-        - qcom/venus-5.4/venus.b03
-        - qcom/venus-5.4/venus.b04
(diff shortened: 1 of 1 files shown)


gctucker:
Please use links to permanent locations such as a mailing list thread for the patch you sent upstream and the linux-firmware upstream git repository rather than your fork.

mgalka:
@nuclearcat Is this fix still valid as it looks like a workaround which is solved by the `copy-firmware.sh`. Maybe it's better to have a proper solution in place instead of landing this one?

nuclearcat:
https://bot.staging.kernelci.org/job/rootfs-builder/412/console
bullseye-v4l2 build ok, built together with https://github.com/kernelci/kernelci-core/pull/1672
verified also rootfs, both firmware files present

mgalka (approved):

PR #1668, opened 2023-01-23
lab-configs: lab-baylibre: use passlist

khilman:
lab-baylibre is having large queues of jobs that don't finish in time for (useful) KernelCI results.  Let's attempt to reduce the load on teh lab by limiting the number of trees that are tested in lab-baylibre.  To be expanded later after reevaluating lab load.

### DIFF
--- a/config/core/lab-configs.yaml
+++ b/config/core/lab-configs.yaml
@@ -7,11 +7,17 @@ labs:
     queue_timeout:
       days: 1
     filters:
-      - blocklist:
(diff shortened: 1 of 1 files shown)


nuclearcat:
I think if we skip staging-next, staging-... - baylibre cannot participate in staging tests. Is it supposed to be like this?
Maybe better to leave at least one staging tree.

nuclearcat (approved):
LGTM

PR #1669, opened 2023-01-23
build-configs.yaml: fix clang arch names

gctucker:
Now that the Clang images are built on a per-arch basis, update the build_environment definitions in build-configs.yaml to use the same arch name mapping as for GCC.

Fixes: 2fbb59c2c768 ("config/docker: Add architecture specific clang templates")

### DIFF
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -569,21 +569,11 @@ build_environments:
     cc: clang
     cc_version: 11
     arch_params: &clang_11_arch_params
-      arm:
-        <<: *arm_params
-        name:
-      arm64:
-        <<: *arm64_params
-        name:
-      i386:
-        <<: *i386_params
-        name:
-      mips:
-        <<: *mips_params
-        name:
-      x86_64:
-        <<: *x86_64_params
-        name:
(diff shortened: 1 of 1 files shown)


gctucker:
Looking good on staging with today's linux-next:

https://staging.kernelci.org/build/kernelci/branch/staging-next/kernel/staging-next-20230124.0/

Only `arm`, `arm64` and `x86_64` are built with `clang-16` on staging.  They're all there, and `x86_64` uses the `x86` Docker image.  So it's not testing everything but I think that's enough to show the arch name translation works as expected.

Well, apart from an actual kernel build error but that's why we have KernelCI:
```
../io_uring/io_uring.c:1240:3: error: variable 'count' is uninitialized when used here [-Werror,-Wuninitialized]
                count += handle_tw_list(node, &ctx, &uring_locked, &fake);
                ^~~~~
../io_uring/io_uring.c:1230:20: note: initialize the variable 'count' to silence this warning
        unsigned int count;
                          ^
                           = 0
1 error generated.
```

nuclearcat (approved):
mie bel mie!

Issue #1579, opened 2022-12-16
test-configs: conditional QEMU command-line args based on Kconfig


khilman:
When generating tests, would like to add QEMU command-line arguments that are dependent on kernel Kconfig options.

For example, add the `-smp` argument to QEMU only if the kernel has `CONFIG_SMP=y` in the defconfig. 

Or maybe easier would be to use existing defconfig filters and only add `-smp` to the command-line if `CONFIG_SMP=n` is *not* in the defconfig filter.

khilman:
This isn't actually resolved.  The PR linked simply adds an SMP variant.  It doesn't yet do so based on checking Kconfig values.

PR #1670, opened 2023-01-24
test-configs: qemu-riscv: add a SMP variant

montjoie:
Current RISCV qemu have only one cpu, so let's add a variant with more CPU.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -1635,6 +1635,23 @@ device_types:
       - blocklist: *device_config_filter
       - regex: {defconfig: '^defconfig'}
 
+  qemu_smp8_riscv64:
+    base_name: qemu
+    mach: qemu
+    arch: riscv
+    boot_method: qemu
+    context:
+      arch: riscv64
+      cpu: 'rv64'
+      guestfs_interface: 'virtio'
+      machine: 'virt'
+      no_kvm: True
+      memory: 1024
+      extra_options: ['-bios default -device virtio-net,netdev=main -netdev user,id=main -smp 8']
+    filters:
+      - blocklist: *device_config_filter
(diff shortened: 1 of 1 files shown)


config/core/test-configs.yaml
```diff
@@ -1635,6 +1635,23 @@ device_types:
       - blocklist: *device_config_filter
       - regex: {defconfig: '^defconfig'}
 
+  qemu_smp8_riscv64:
```
gctucker, line 1638: How about using a YAML anchor to avoid duplicating all the common parts?  The only thing that differs afaict is the `extra_options`.

khilman: Instead of creating a new device_type for smp=8, please just make the current device type default to smp=8, which is what we want for all the existing tests

As follow-on work, we need to explore how to update/modify QEMU cmdline args based on kernel options (e.g. defconfig.)  For example if the defconfig has `CONFIG_SMP=n` then we should not have any `-smp` on the cmdline.

khilman: After discussion with @montjoie we decided that having the existing no-smp device and smp8 would both be useful.

khilman: @gctucker FYI: an issue was created for this back in Dec: https://github.com/kernelci/kernelci-core/issues/1579

gctucker: Tip: Adding "Fixes #1579" in the PR description helps in this respect.  The only gotcha is that the issue will now be closed automatically when this PR is merged, which is desirable as long as it really solves the issue entirely.

nuclearcat (approved):
LGTM, tested and works on dashboard: https://staging.kernelci.org/test/plan/id/63df7065544f358b166711b6/

PR #1672, opened 2023-01-25
Fix firmware handling

nuclearcat:
Two patches to fix firmware handling using copy-firmware.sh, as without it we doesn't have symlinked firmware files.
First to address firmware extraction for rootfs, second to address firmware extraction for building in-kernel.

### DIFF
--- a/config/rootfs/debos/scripts/install-firmware.sh
+++ b/config/rootfs/debos/scripts/install-firmware.sh
@@ -40,11 +40,19 @@ mkdir -p $DEST
 # git init hack to fetch a single shallow commit to minimize time,
 # space and network bandwidth
 TMP_REPO=/tmp/linux-firmware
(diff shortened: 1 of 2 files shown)


nuclearcat:
Tested by hand building kernel and rootfs with firmware files, where one of them symlink:
```
        - qcom/venus-5.4/venus.mdt
        - qcom/venus-5.4/venus.mbn
```
Next step is merging some patches for updated firmware and testing in staging jenkins.

nuclearcat:
Tested rootfs with fixed firmware revision, works fine

nuclearcat:
https://bot.staging.kernelci.org/job/rootfs-builder/412/console
bullseye-v4l2 build ok, built together with https://github.com/kernelci/kernelci-core/pull/1667
verified also rootfs, both firmware files present

PR #1676, opened 2023-01-25
kernelci.lab.lava: Fix priority scaling

broonie:
Since 96d636627f ("kernelci.lab.lava: rework priority scale") jobs submitted to at least my lab have been ignoring the configured priority range and using the full range of available priorities.  The commit has the logic to scale plan priorities but it doesn't seem to be taking effect for reasons that are not obvious to me, possibly due to confusion over the treatment of a 0 priority value as a logical value.

This is severely disrupting my own use of my lab since the baseline jobs are submitted with vastly higher priority than any of my normal jobs so juts revert the commit for now.

### DIFF
--- a/kernelci/lab/lava/__init__.py
+++ b/kernelci/lab/lava/__init__.py
@@ -37,17 +37,22 @@ class LavaAPI(LabAPI):
 
         # Scale the job priority (from 0-100) within the available levels
         # for the lab, or use the lowest by default.
-        plan_priority = plan_config.params.get('priority', 20)
-        if all((self.config.priority_max, self.config.priority_min)):
-            prio_range = self.config.priority_max - self.config.priority_min
-            prio_min = self.config.priority_min
-            plan_priority = int((plan_priority * prio_range / 100) + prio_min)
+        if 'priority' in plan_config.params:
+            priority = plan_config.params['priority']
+            if priority > 100:
+                priority = 100
+        else:
+            priority = 20
+
+        prio_range = self.config._priority_max - self.config._priority_min
+        priority = int(((priority * prio_range) / 100) +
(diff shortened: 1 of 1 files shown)


gctucker:
I think that's not the best way to fix the issue, the Python syntax is not what's causing the problem.  I did test it locally by generating job definitions and it seemed to be working as expected.  So yes it maybe something to do with integers vs strings.

Could you please give some examples of the expected behaviour?  Ideally we should have unit tests for things like this.

broonie:
I agree that it's not the best way to fix the issue, this is just the quickest way to fix the severe issues being caused in production.

The issue can be seen by looking at the priority of a baseline job generated for my lab, it should be 36 but it is in fact 90 with current main. It appears that the raw priority for the baseline job is not being scaled into the configured range 0..40.

broonie:
Note that I suspect it's not integers vs strings, I suspect it's the fact that a valid priority for my lab is in the range 0..40 but there's a test been added in there for if both min and max priority are defined for the lab and there is some confusion possible with the treatment of 0 as an integer in a logical context vs None in a logical context.

broonie:
This is working on staging, eg this baseline job has the expected priority of 36:

https://lava.sirena.org.uk/scheduler/job/182651

gctucker:
OK thanks, then I think we can merge this to fix the regression.

Meanwhile I'm adding unit tests to ensure the functionality doesn't change when reworking this part of the code.

nuclearcat (approved):
LGTM

gctucker:
For the record, #1681 and #1682 are follow-ups to add unit tests and rework the code slightly.

nuclearcat:
Thanks, i will review them in next hour or two, so they will go together

Issue #1677, opened 2023-01-25
Commit 46976d4df cause k8s to restart pods with bugs indefinitely


nuclearcat:
Commit: https://github.com/kernelci/kernelci-core/commit/46976d4df272d2d16edfda80fdf57e44f72b9e52

While originally commit was intended to prevent jobs continuing in case of error, it caused unexpected, opposite result, when kubernetes pod program (shell script) return non-zero code, pod will consider it as crash and will try to restart.
Another negative result, that job in jenkins will not show error, since job is still trying to run, but rather return timeout, making troubleshooting of bugs difficult.

For example:
```
build-j61810-x86-64-gcc-10-x86-64-defconfig-crypto-vmds9          0/1     CrashLoopBackOff   19 (2m28s ago)   92m
pod log:
```
Upload path: kernelci/staging-stable/staging-stable-20230125.0/x86_64/x86_64_defconfig+crypto/gcc-10
400 Client Error: Provided JSON is not valid for url: https://api.staging.kernelci.org/build
'kernel' is a required property
 
Failed validating 'required' in schema['properties']['artifacts']:
    {'patternProperties': {'.*': {'items': {'properties': {'contents': {'items': {'type': 'string'},
                                                                        'type': 'array'},
                                                           'key': {'type': 'string'},
                                                           'path': {'type': 'string'},
                                                           'type': {'enum': ['file',
                                                                             'directory',
                                                                             'tarball'],
                                                                    'type': 'string'}},
                                            'required': ['type',
                                                         'path'],
                                            'type': 'object'},
                                  'type': 'array'}},
     'required': ['kernel'],
     'type': 'object'}
 
On instance['artifacts']:
    {u'config': [{u'key': u'config',
                  u'path': u'config/kernel.config',
                  u'type': u'file'},
                 {u'key': u'fragment',
                  u'path': u'config/kernelci.config',
                  u'type': u'file'},
                 {u'key': u'log',
                  u'path': u'logs/config.log',
                  u'type': u'file'}]}
```

nuclearcat:
After discussion we have two approaches to solve this:
1)Revert commit
2)Change Kubernetes job policy to not restart pod on failure

@broonie maybe you have any ideas? 
I might make temporary patch to revert PR, so staging jobs with errors will not case indefinite loops

gctucker:
The root cause for the kernel build error is that cross-compile strings are populated with `null` in `bmeta.json`:
[...]

PR #1678, opened 2023-01-25
Revert "config/k8s: add set -e to build jobs"

nuclearcat:
Being discussed in https://github.com/kernelci/kernelci-core/issues/1677 
This reverts commit 46976d4df272d2d16edfda80fdf57e44f72b9e52.


nuclearcat:
As discussed during Weekly meeting, we need also to add "|| exit 0" to commands to exit early in case of failures, so original goal of quitting early in case of failure will be achieved.

gctucker (approved):
I think it's fine to merge this for now and then we can improve error reporting as a follow-up.

PR #1679, opened 2023-01-26
build-configs.yaml: Add trogdor configuration bits from chromeos

nuclearcat:
As we are still experiencing problem with enabling MSM GPU we will try to enable configuration bits to match cros:// fragments.


nuclearcat:
Turned out 6.1 trogdor kernel issues not related to config

PR #1681, opened 2023-01-26
Add tests for `runtime` starting with LAVA job priorities

gctucker:
Add a new file `test_runtime.py` to cover things related to the runtime environments which are currently implemented in `kernelci.lab`.  Start with testing the LAVA job priority logic since there was a regression recently and some further code changes are likely to be made in this area to accommodate generating jobs for Kubernetes etc. with the new API & Pipeline.

This also fixes some corner-cases such as when no min/max priorities are provided.


kernelci/lab/lava/__init__.py
```diff
@@ -62,16 +62,14 @@ def job_file_name(self, params):
...
-                priority = 100
-        else:
-            priority = 20
-
-        prio_range = self.config._priority_max - self.config._priority_min
-        priority = int(((priority * prio_range) / 100) +
-                       self.config._priority_min)
+        priority = plan_config.params.get('priority', 20)
+        if self.config.priority:
+            priority = int(priority * self.config.priority / 100)
```
broonie, line 67: The intent of the old code was that if a single priority was set then we'd use exactly that priority rather than scaling within it (min and max would end up being set to the same value IIRC). I can't see anyone objecting to us lowering the priority though. The code was written the way it was just to ensure there were no changes for any labs that didn't explicitly configure a range.

gctucker: OK, I guess that's related to some changes in how the YAML config is handled as I've reworked it to be able to dump the input config back into YAML.  That meant removing some mangling, basically it's better if the config objects stay true to the original YAML data and the code using that adds any extra logic on top.

So I'll tweak this again to not scale when there's no range as that's a separate thing, and just include the fix for when the min/max values are not provided as in that case they now resolve as `None` from the config object.

broonie (approved):

gctucker:
This was merged with #1682 (cc @nuclearcat)

nuclearcat (commented):
LGTM

nuclearcat:
Oh just noticed it is closed :)

gctucker:
> Oh just noticed it is closed :)

Yes you've already merged it, by accident I gather ;)

Issue #1652, opened 2023-01-18
v4l compliance missing firmware for mt8173-elm-hana


nuclearcat:
As reported in https://storage.staging.kernelci.org/kernelci/staging-mainline/staging-mainline-20230118.0/arm64/defconfig+arm64-chromebook/gcc-10/lab-collabora/v4l2-compliance-uvc-mt8173-elm-hana.html:
```
 1661 03:18:57.241218  <LAVA<3>[   23.468844] mtk_vpu 10020000.vpu: Failed to load vpu_p.bin, -2
 1662 03:18:57.249448  _SIGNAL_STARTRUN<3>[   23.476197] mtk_vpu 10020000.vpu: Failed to request vpu_p.bin, -2
 1663 03:18:57.256657   0_v4l2-complian<3>[   23.483966] mtk-mdp 14001000.rdma: vpu_load_firmware failed -2
 1664 03:18:57.268091  ce-uvc 8765680_1<4>[   23.491635] mtk_vpu 10020000.vpu: Direct firmware load for mediatek/mt8173/vpu_p.bin failed with error -2
 1665 03:18:57.268880  .5.2.3.1>
 1666 03:18:57.277661  + /us<6>[   23.502606] mtk_vpu 10020000.vpu: Failed to load mediatek/mt8173/vpu_p.bin, -2, retry
 1667 03:18:57.278448  
 1668 03:18:57.287595  r/bin/v4l2-parse<4>[   23.512218] mtk_vpu 10020000.vpu: Direct firmware load for vpu_p.bin failed with error -2
 1669 03:18:57.293750  <3>[   23.522052] mtk_vpu 10020000.vpu: Failed to load vpu_p.bin, -2
 1670 03:18:57.301089  r.sh -d uvcvideo<3>[   23.528234] mtk_vpu 10020000.vpu: Failed to request vpu_p.bin, -2
 1671 03:18:57.301669  
 1672 03:18:57.309540  <3>[   23.535998] fops_vcodec_open(),205: [MTK_V4L2][ERROR] failed to load firmware!
 1673 03:18:57.319074  <4>[   23.543703] mtk_vpu 10020000.vpu: Direct firmware load for mediatek/mt8173/vpu_p.bin failed with error -2
 1674 03:18:57.326899  <6>[   23.553557] mtk_vpu 10020000.vpu: Failed to load mediatek/mt8173/vpu_p.bin, -2, retry
[...]

PR #1694, opened 2023-01-27
build-configs.yaml: Add missing firmware for Mediatek and Rockchip

nuclearcat:
Mediatek and Rockchip emit a lot of error messages about missing firmware even on baseline job, which might affect tests not only by fact of missing firmware, but also repeating messages disrupting LAVA console. As firmware is relative small we can add it directly to Chromebook kernels.
Fixes: https://github.com/kernelci/kernelci-core/issues/1652

### DIFF
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -332,6 +332,11 @@ fragments:
       - 'CONFIG_VIDEO_OV02A10=m'
       - 'CONFIG_VIDEO_OV5695=m'
       - 'CONFIG_VIDEO_OV8856=m'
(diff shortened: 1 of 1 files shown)


nuclearcat:
Firmware loaded successfully: https://storage.staging.kernelci.org/kernelci/staging-next/staging-next-20230130.0/arm64/defconfig+arm64-chromebook/gcc-10/lab-collabora/v4l2-compliance-uvc-mt8173-elm-hana.html
Can be compared with previous result in: https://github.com/kernelci/kernelci-core/issues/1652

mgalka (approved):

PR #1699, opened 2023-01-30
Clean up unit tests

gctucker:
Fix all pycodestyle and pylint errors in unit tests and enable coverage accordingly in GitHub actions.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -55,7 +55,8 @@ jobs:
             kci \
             kernelci.cli \
             kernelci.config.api \
-            kernelci.storage
+            kernelci.storage \
+            tests
 
       - name: Run YAML config validation
         run: |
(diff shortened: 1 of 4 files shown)


nuclearcat (approved):
LGTM
(first i was confused about using 172.17.0.1 then i realize it is just test case for config, nothing else)

nuclearcat:
I will wait until someone can help me to review few things in https://github.com/kernelci/kernelci-core/pull/1650 at Monday, as it depends on it

gctucker:
> LGTM (first i was confused about using 172.17.0.1 then i realize it is just test case for config, nothing else)

It's the standard host IP address when using docker-compose.yaml so it's actually functional.  But it's true it's not used in these unit tests, just the config gets loaded but no code is actually run.

gctucker:
@mgalka @nuclearcat @aliceinwire I've enabled the Merge Queue feature on this repository as an experiment.  It's nice that it's removing the branches automatically, and maybe it can help when merging a number of PRs in a row e.g. ahead of production updates.

PR #1707, opened 2023-02-02
test-configs.yaml: Prune tests targetted at Tritum

broonie:
With the current level of activity in the covered trees in my lab and our inability to coalesce LAVA jobs we're never getting anywhere near clearing the backlog of test jobs scheduled on the Tritiums, tune down the set of jobs scheduled with the goal of ensuring that tests run reliably.


nuclearcat (approved):
LGTM

PR #1709, opened 2023-02-02
test-configs.yaml: Use config fragment for KVM selftests

broonie:
The KVM selftests do actually have a config fragment which is needed on architectures other than arm64 to get KVM actually built, and will be needed on architectures other than S/390 to get the page_fault_test test to run once a fix is merged upstream.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -326,7 +326,6 @@ test_plans:
     params:
       job_timeout: '15'
       kselftest_collections: "kvm"
-    filters: *kselftest_no_fragment
 
   kselftest-landlock:
     <<: *kselftest


gctucker (dismissed):
Tested on staging, it's failing because the tests are using `kselftest-slim`:
https://staging.kernelci.org/test/job/next/branch/master/kernel/next-20230331/plan/kselftest-vm/
https://storage.staging.kernelci.org/next/master/next-20230331/x86_64/x86_64_defconfig+x86-chromebook+kselftest-slim/gcc-10/lab-collabora/kselftest-vm-hp-11A-G6-EE-grunt.html#L2953
```
No such collection 'vm'
```

broonie:
...which I can't actually identify in the commit, but however we're deciding that we should be looking for a collection "vm" that's not going to work since there's no vm kselftest suite, it should be "kvm".

gctucker:
Ah yes sorry I picked the wrong test plan results on the frontend.

gctucker:

nuclearcat:

Issue #1696, opened 2023-01-30
arm64 qcom/trogdor fail to enable GPU,  "disp_cc_mdss_ahb_clk status stuck"


nuclearcat:
After trying to enable relevant kernel entries from ChromiumOS kernel tree, i'm still unable to run GPU of Trogdor.
It is likely either bug or/and combination of kernel options.
Probably related to issue: https://lore.kernel.org/all/YZSfSgz%2FALWTfLQ5@builder.lan/raw

Relevant kernel message:
```

 1520 01:00:24.196844  <3>[   17.935071] disp_cc_mdss_mdp_clk_src: RCG did not turn on
 1521 01:00:24.197778  <3>[   17.935366] Failed to enable clk 'core': -110
 1522 01:00:24.200011  <3>[   17.935394] msm-mdss ae00000.mdss: clock enable failed, ret:-110
 1523 01:00:24.206931  <3>[   17.936319] OF: graph: no port node found in /soc@0/geniqup@ac0000/spi@a80000/ec@0/typec/connector@0
 1524 01:00:24.217512  <3>[   17.936332] OF: graph: no port node found in /soc@0/geniqup@ac0000/spi@a80000/ec@0/typec/connector@0
 1525 01:00:24.228005  <3>[   17.936337] OF: graph: no port node found in /soc@0/geniqup@ac0000/spi@a80000/ec@0/typec/connector@0
 1526 01:00:24.237348  <3>[   17.936381] OF: graph: no port node found in /soc@0/geniqup@ac0000/spi@a80000/ec@0/typec/connector@0
 1527 01:00:24.247434  <3>[   17.936387] OF: graph: no port node found in /soc@0/geniqup@ac0000/spi@a80000/ec@0/typec/connector@0
 1528 01:00:24.257244  <3>[   17.936392] OF: graph: no port node found in /soc@0/geniqup@ac0000/spi@a80000/ec@0/typec/connector@0
 1529 01:00:24.266993  <3>[   17.936398] OF: graph: no port node found in /soc@0/geniqup@ac0000/spi@a80000/ec@0/typec/connector@0
 1530 01:00:24.276797  <3>[   17.936405] OF: graph: no port node found in /soc@0/geniqup@ac0000/spi@a80000/ec@0/typec/connector@0
 1531 01:00:24.283879  <3>[   17.936420] OF: graph: no port node found in /soc@0/geniqup@ac0000/spi@a80000/ec@0/typec/connector@0
[...]

PR #1713, opened 2023-02-05
build-configs.yaml: Add QCOM_QFPROM option for QCOM GPU

nuclearcat:
As discovered in https://github.com/kernelci/kernelci-core/issues/1696 Qualcomm GPU need this option to retrieve some data relevant to GPU.


nuclearcat:
Unforuntately this is relevant only for 6.0 kernels

PR #1717, opened 2023-02-08
build-configs.yaml: Add NFS options to arm64-chromebook

nuclearcat:
As we have some configs with this fragment without defconfig, we need to add NFS options, so kernel configuration can pass baseline-nfs tests.

### DIFF
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -157,6 +157,12 @@ fragments:
   arm64-chromebook:
     path: "kernel/configs/arm64-chromebook.config"
     configs:
+      - 'CONFIG_NFS_FS=y'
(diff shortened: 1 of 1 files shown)


nuclearcat:
Kernel with arm64 fragment built, no issues observed: https://storage.staging.kernelci.org/kernelci/staging-mainline/staging-mainline-20230209.0/arm64/defconfig+arm64-chromebook/gcc-10/config/kernel.config
https://staging.kernelci.org/test/job/kernelci/branch/staging-mainline/kernel/staging-mainline-20230209.0/plan/baseline-nfs/
Positive results can be observed only after merge in ChromeOS branch after rebase.

mgalka (approved):

PR #1721, opened 2023-02-09
CHROMEOS test-configs-chromeos.yaml: Add new Mediatek Chromebook jacuzzi

nuclearcat:
As Mediatek upstream efforts significantly improved kernel support of relevant platforms, it will be great to test upstream kernels with ChromeOS rootfs.
This patch add mt8183 based kukui-jacuzzi-juniper Chromebook.


mgalka (approved):
Tests visible on chromeos.kernelci.org

PR #1730, opened 2023-02-10
test-configs.yaml: Only run ALSA selftests where they exist

broonie:
The ALSA selftests were merged in v5.17 so there is no point in running them on older kernels, add a filter. We block all v5.x since there are no longer any active v5.x kernels based on anything newer than v5.15 so it's not worth the effort of writing the filter.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -255,6 +255,7 @@ test_plans:
     filters:
       - passlist: {defconfig: ['kselftest']}
 
+  # Supported from v5.17 on but the latest v5.x stable is v5.15
   kselftest-alsa:
     <<: *kselftest
     rootfs: debian_bookworm-kselftest_nfs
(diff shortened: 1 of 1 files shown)


config/core/test-configs.yaml
```diff
@@ -236,13 +236,15 @@ test_plans:
...
 
+  # Supported from v5.17 on but the latest v5.x stable is v5.15
   kselftest-alsa:
     <<: *kselftest
     params:
       job_timeout: '10'
       kselftest_collections: "alsa"
     filters: &kselftest_no_fragment
       - combination: *arch_defconfig_filter
+      - blocklist: {kernel: ['v3.', '4.', '5.']}
```
gctucker, line 247:
```suggestion
      - blocklist: {kernel: ['v3.', 'v4.', 'v5.']}
```

gctucker:

PR #1731, opened 2023-02-10
test-configs.yaml: Tune down number of tests run on i.MX UDOO boards

broonie:
Due to an increase in the number of commits tested, due to increasing numbers of trees and activity in some of the larger ones like stable, we are not managing to get through all the scheduled jobs on the UDOO boards.  Trim the list of tests scheduled on these boards to just those that exercise hardware and features specific to them.


nuclearcat (approved):
LGTM

PR #1733, opened 2023-02-10
test-configs.yaml: Tune down number of tests run on Juno

broonie:
Due to increased numbers of trees and activity in some of those trees we are not managing to get through the tests we currently schedule on Juno. Tune down the number of tests to try to ensure we get through things promptly and reliably.


nuclearcat (approved):
LGTM

PR #1739, opened 2023-02-13
workflows use checkout v3 for older checks

gctucker:
Update the checkout action to v3 to fix this warning:

  Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions/checkout@v2.


nuclearcat (approved):
LGTM

PR #1740, opened 2023-02-14
Add and use bookworm rootfs for kselftest-alsa

nfraprado:
Add a rootfs to run kselftest on Debian bookworm, and update the kselftest-alsa test plan to use it instead. Bookworm is needed since it includes more recent versions of the alsa-ucm-conf, libasound2 and alsa-utils packages than the ones available in bullseye, allowing us to access UCM files for more recent platforms. In the end this allows more thorough testing of audio on these machines and better scores.

I created [a bug report](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1029272) on libasound2 to see if a backport to stable was possible, but since I haven't heard back from the maintainer in a while, let's get this going with debian testing. If the packages are ever backported we can switch to those then.

### DIFF
--- a/config/core/rootfs-configs.yaml
+++ b/config/core/rootfs-configs.yaml
@@ -1,4 +1,44 @@
 rootfs_configs:
+  bookworm-kselftest:
+    rootfs_type: debos
+    debian_release: bookworm
+    arch_list:
+      - amd64
+      - arm64
+      - armhf
+    extra_packages:
+      - alsa-ucm-conf
+      - alsa-utils
+      - bc
+      - ca-certificates
+      - iproute2
+      - jdim
+      - libasound2
+      - libatm1
+      - libcap2-bin
+      - libelf1
+      - libgdbm-compat4
+      - libgdbm6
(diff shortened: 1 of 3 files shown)


nfraprado:
Let me know when this rootfs is generated so I can update the URL and we can compare the test results to make sure none of the boards regress.

broonie:
I've triggered builds with version 20230224.0 - they should be on staging by the time you see this.

broonie:
Didn't see an update to the rootfs images so I made my own: https://github.com/kernelci/kernelci-core/pull/1767

broonie:
Successful runs on arm and arm64 :

https://staging.kernelci.org/test/plan/id/63fd58faa54190f51f9d58bc/
https://staging.kernelci.org/test/plan/id/63fd607c41a941721b9d58d7/
https://staging.kernelci.org/test/plan/id/63fd57e42ba491e02b9d590e/

nuclearcat (approved):
LGTM

PR #1741, opened 2023-02-14
Enable kselftest-alsa (bookworm) on mt8173-elm-hana

nfraprado:
Based on #1740.

Add and enable the kselftest-alsa-bookworm test plan on mt8173-elm-hana. The UCM
files for mt8173-elm-hana are under ucm2/MediaTek/mtk-rt5650, which is
only present starting with v1.2.6.3 of alsa-ucm-conf. Bullseye provides
alsa-ucm-conf v1.2.4, and therefore the bookworm rootfs is needed to
completely initialize audio on this machine.

Without the UCM, errors like the following show up during the pcm-test
in alsa-kselftest:

```
rt5650 Playback: ASoC: no backend DAIs enabled for rt5650 Playback
rt5650 Playback: ASoC: error at dpcm_fe_dai_prepare on rt5650 Playback: -22
```

And with the UCM applied, results from pcm-test go from

```
Totals: pass:2 fail:2 xfail:0 xpass:0 skip:17 error:0
```

to

```
Totals: pass:12 fail:1 xfail:0 xpass:0 skip:8 error:0
```

The default selftest timeout is too short to run pcm-test on this machine, so I had to apply [[PATCH v2] kselftest/alsa: Increase kselftest timeout](https://lore.kernel.org/all/20221214130353.1989075-1-nfraprado@collabora.com/#r) to be able to run the test through the end. But it wasn't accepted upstream, so a different approach might be needed (or more data to show that the current timeout isn't enough). Related to this, there has also been `69218b59be20 kselftest/alsa: Run PCM tests for multiple cards in parallel` merged recently, but that doesn't help here, since this machine has a single sound card.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -2846,6 +2846,7 @@ test_configs:
       - baseline-nfs
       - cros-ec
       - igt-kms-mediatek
+      - kselftest-alsa
       - kselftest-arm64
       - kselftest-cpufreq
       - kselftest-filesystems


broonie:
Should we perhaps just use bookworm for all ALSA kselftest runs rather than just for specific boards? It seems more trouble than it's worth to handle the separate rootfs images and worry about moving boards between them. We're probably going to be upgrading to bookworm whenever it gets released this year anyway.

nuclearcat:
Adding skip-staging as it doesn't pass CI checks and break staging:
```
./kci_test list_labs
  File "/home/nuclearcat/Documents/kernelci-core/./kci_test", line 400, in <module>
    configs = kernelci.config.load_all(opts.get_yaml_configs())
  File "/home/nuclearcat/Documents/kernelci-core/kernelci/config/__init__.py", line 192, in load_all
    return from_data(data)
  File "/home/nuclearcat/Documents/kernelci-core/kernelci/config/__init__.py", line 159, in from_data
    config.update(kernelci.config.test.from_yaml(data, filters))
  File "/home/nuclearcat/Documents/kernelci-core/kernelci/config/test.py", line 559, in from_yaml
    test_plans = {
  File "/home/nuclearcat/Documents/kernelci-core/kernelci/config/test.py", line 560, in <dictcomp>
    name: TestPlan.from_yaml(name, test_plan, file_systems, plan_filters)
  File "/home/nuclearcat/Documents/kernelci-core/kernelci/config/test.py", line 434, in from_yaml
    'rootfs': file_systems[rootfs_name] if rootfs_name else None,
KeyError: 'debian_bookworm-kselftest_nfs'
```
Let me know when it is fixed.

nfraprado:
@broonie yes, makes sense. I'll change alsa kselftest to use a bookworm rootfs.

@nuclearcat sorry, since the rootfs-images file is automatically generated I thought it would get fixed up automatically after the rootfs got built. Do I need to manually add an entry then? And what date should I use?

broonie:
Take a look at https://github.com/kernelci/kernelci-core/pull/1660 and https://github.com/kernelci/kernelci-core/pull/1753 - you need to manually add the template (the date doesn't really matter there since it'll get overwritten after being merged) then add a dummy PR on top which points at staging images with a date/tag that's determined by when a manual build of the images has been done via the web dashboard.

Unfortunately we don't automatically build any rootfs images for staging like we do for production which tends to cause a bunch of pain.

broonie:
One other thought here: might it be better to just grab a more current git and overlay it onto the filesystem rather than relying on things going through the distro? Debian releases slowly and this is likely to be an ongoing problem.

nfraprado:
That was what I was originally doing in #1483. So let's discuss it more thoroughly to avoid going in circles.

Using the packages from stable is clearly too slow, as they're 2 years old at this point.

On debian testing, the [latest alsa-ucm-conf was released in 30 November 2022](https://metadata.ftp-master.debian.org/changelogs//main/a/alsa-ucm-conf/alsa-ucm-conf_1.2.8-1_changelog), about one month after [the release was tagged in the git repository](https://github.com/alsa-project/alsa-ucm-conf/releases/tag/v1.2.8). So if we were to get it from the latest release in git, we would be sparing that one month. We should keep in mind though that UCMs aren't completely backward compatible, for example the syntax version of an already used UCM might change, and so we would also need to fetch libasound2 and alsa-utils from git to avoid regressions.

Another option is to get from the master branch in git (rather than the latest release). Since it has been 4 months since the latest release, we would spare that much waiting time. The UCM for mt8192-asurada-spherion in particular was merged shortly after that release, and would be already available to use in KernelCI if we followed that route. While I think doing this for the UCMs would be straightforward, I'd expect issues once in a while building libasound2 and alsa-utils directly from the master branch.

Besides that, when I originally did this, I believe the libasound2 package couldn't be easily removed from the system since others depended on it, and due to that I ended up overriding the ld load order to prioritize the libasound2 built from source. I might have missed something there so this might not really be needed, but it's something to keep in mind.

This all to say, the biggest source of delay right now comes from the sparsely tagged releases on the alsa repositories, rather than debian packaging, so building everything directly from master would be needed to avoid that, but I'm worried that would require more maintenance than it's worth it.

broonie:
I'm not sure how much of an issue backwards incompatible changes are here - if things are relying on newer features presumably they just need those newer features. Though I guess at some point someone will want such a board...

That said if the intention is to keep tracking testing rather than use bookworm specifically in order to ensure we get updated ALSA and everything as well then probably the rootfs should be called testing rather than bookworm? That's probably what we'd need ongoing.

nfraprado:
What I meant by backward incompatible is that newer UCM files might not work with older versions of libasound, due to [changes like this](https://github.com/alsa-project/alsa-ucm-conf/commit/e2cfda5b53f4718f1128fb4df75ec8fb7274d36b): the syntax version of a UCM gets bumped, and now it needs a more recent libasound to work. So we should keep the versions for alsa-ucm-conf, libasound and alsa-utils all in sync.

And yes, agreed, it would make more sense to base the rootfs on testing instead of bookworm, so we don't need to do this again in the future after bookworm gets released. Although I wonder if when testing starts pointing to the next release there won't be big changes at first, which could cause regressions. Or if packages with version embedded on the name like `libperl5.36` will disappear in the newer version. I'm basically wondering if there will be any manual intervention required when testing starts tracking the next release anyway, because if so, then we might be better off keeping the rootfs on bookworm and just do the manual update to the next testing once we have sorted those differences out, that way at least there wouldn't be big breaks in the test results.

nfraprado:
We can further discuss whether it's better to use debian testing or the current testing debian release (maybe in a separate issue?), but for running alsa kselftest on mt8173-elm-hana specifically, bookworm is all it needs, and that's already in place, so marking PR as ready.

gctucker:
Removing `staging-skip` so this can actually be tested on staging...  Well check again for results next week.

nuclearcat:

PR #1745, opened 2023-02-16
build-configs.yaml: Add MediaTek thermal driver modules

kholk:
Add new configuration options for both the legacy soc-thermal, for older SoCs like mt6795, mt8173, mt8183 and others, and the new LVTS thermal for newer SoCs like MT8195.

This solves the following test regression:
https://groups.io/g/kernelci-results/message/38269


nuclearcat:
According to https://staging.kernelci.org/test/plan/id/63ef3646f5b79db55c9d598c/ it fixes issue with thermal driver

PR #1746, opened 2023-02-16
test-configs.yaml: Expand coverage for QDF2400

broonie:
Now we have curtailed the set of trees we cover with the QDF2400 we are getting through all the currently scheduled tests with room to spare, expand the coverage a bit and also cover the arm64 selftests (this is a unique microarchitecture so it's good to cover them) and since we're relying on it for all the other tests include baseline-nfs too.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -2931,6 +2931,8 @@ test_configs:
   - device_type: qcom-qdf2400
     test_plans:
       - baseline
+      - baseline-nfs
+      - kselftest-arm64
       - kselftest-filesystems
       - kselftest-futex
       - kselftest-kvm


broonie:
We don't seem to be scheduling kselftest-arm64 on *anything* in staging at the minute...

nuclearcat:
I have to check that, but i think i did scheduled at least on qualcomm and probably mediatek.

broonie:
I suspect the issue is that we only run mainline and -next on qdf2400 but those trees aren't built directly in staging (they're branches of a "kernelci" tree as far as the UI is concerned) so nothing is matching on staging any more. Not sure how to deal with that sensibly.

nuclearcat:
It doesnt look device appeared in results: https://staging.kernelci.org/test/job/mainline/branch/master/kernel/v6.3-rc2/plan/baseline/

broonie:
Right, there's something broken with the board - it's not had any jobs submitted since the 10th (Friday), even from production.

broonie:
Successful runs: https://staging.kernelci.org/test/plan/id/6429c00207223a10140b9bd2/ https://staging.kernelci.org/test/plan/id/6429b001c37089a50d0b9bd1/

gctucker:

PR #1747, opened 2023-02-17
CHROMEOS build_board.sh: Extract and move CR50 firmware during build

nuclearcat:
This commit introduces a patch that allows for the extraction of CR50 firmware, enabling it to be flashed separately. Prior to this patch, ChromeOS would attempt to reflash the CR50 during rootfs installation, which required manual intervention to powercycle the device. By extracting the firmware and making it a separate step, we can now plan for proper testing by Labs personnel.


nuclearcat:
trogdor rootfs rebuilt in jenkins and CR50 firmware extracted
https://storage.chromeos.kernelci.org/images/rootfs/chromeos/chromiumos-trogdor/20230221.0/arm64/

mgalka (approved):

PR #1750, opened 2023-02-17
build-configs.yaml: Compile-in TPM related modules

nuclearcat:
ChromeOS require TPM to be compiled in, otherwise it might fail to pass Tast tests.


nuclearcat:
Tast tests passing now on trogdor, for example: https://chromeos.kernelci.org/test/plan/id/63fc46817f62f0fdd5a9d867/

mgalka (approved):

PR #1698, opened 2023-01-30
test-configs.yaml: Cover NFS boots on Raspberry Pi 1B

broonie:
The ethernet works fine on these boards so we can cover NFS.

Signed-off-by: Mark Brown <broonie@kernel.org>

PR #1752, opened 2023-02-20
test-configs.yaml: Enable NFS baseline tests on bcm2385-rpi-b-rev2

broonie:
These boards have everything needed for NFS boot so let's cover it.


broonie:
Dupicate of https://github.com/kernelci/kernelci-core/pull/1698

PR #1755, opened 2023-02-21
Rename kernelci.lab as kernelci.runtime

gctucker:
Rename the `kernelci.lab` package as `kernelci.runtime` and update all areas of the code accordingly.  Add an extra unit test to check all the runtime types can be initialised.


gctucker:
Actually I should first rename the classes in `kernelci.lab` as that's less invasive, and then rename the package to `kernelci.runtime` once the `kernelci.config.runtime` changes have been merged.  So this will be rebased later.

PR #1737, opened 2023-02-13
Runtime config

gctucker:
Rename `kernelci.config.lab` as `kernelci.config.runtime`.  Update all the code and unit tests accordingly.

Fixes #1726

PR #1756, opened 2023-02-21
Rename kernelci.lab.LabAPI to kernelci.lab.Runtime

gctucker:
Rename the `kernelci.lab.LabAPI` base class to `Runtime` as a first step to make it more generic.  Update unit tests accordingly.

Depends on #1737


gctucker:
Tested OK on staging and with the new pipeline using #1755.

PR #1758, opened 2023-02-21
CHROMEOS build-configs-chromeos.yaml: Add collabora-chromeos-kernel

nfraprado:
Add build using the for-kernelci branch of Collabora's chromeos-kernel tree. This branch is used to integrate patches for MediaTek based Chromebooks.

### DIFF
--- a/config/core/build-configs-chromeos.yaml
+++ b/config/core/build-configs-chromeos.yaml
@@ -1,5 +1,8 @@
 trees:
 
+  collabora-chromeos-kernel:
+    url: "https://gitlab.collabora.com/google/chromeos-kernel.git"
+
   kernelci:
     url: "https://github.com/kernelci/linux.git"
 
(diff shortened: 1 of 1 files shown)


nuclearcat:
Build tested: https://chromeos.kernelci.org/job/collabora-chromeos-kernel/branch/for-kernelci/
At moment you are doing only one fragment, so it be tested on real devices.
To be tested on Mediatek chromebooks you need to add something like:
```
        fragments: [arm64-chromebook]
+        extra_configs:
+          - 'cros://chromeos-6.1/arm64/chromiumos-mediatek.flavour.config+arm64-chromebook+CONFIG_MODULE_COMPRESS_GZIP=n'
```
Maybe you can do it in follow up PR, if it is required.

nfraprado:
@nuclearcat shouldn't what I added here be enough to get the mainline tests (like baseline and kselftests) running on the chromebooks? I'd like to see test results without the ChromeOS fragments applied.

nuclearcat:
For baseline and kselftest it is enough

PR #1759, opened 2023-02-21
Mitigate size issues with kselftest kernels

broonie:
We struggle to run kselftest on a lot of u-boot platforms since the kselftest builds produce a kernel which is much larger than usual, and some of the emulated platforms can struggle at runtime since they are also slower than usual. The increase in size is mostly due to only two of the config fragments, those for cpufreq and LKDTM. These add options that instrument the entire kernel, inflating the size of the resulting image substantially.

While we can address the configuration of individual boards this is a constant burden and we still have the overhead of downloading the larger images and slower execution caused by the instrumentation to deal with. Instead let's define a second kselftest fragment which excludes the problematic fragments and use that unless we need the extra fragments.

This does mean we end up doing an extra build whenever we build kselftest but the improved test coverage and runtime performance seems worth it.

### DIFF
--- a/kernelci/build.py
+++ b/kernelci/build.py
@@ -241,11 +241,12 @@ def make_tarball(kdir, tarball_name):
     os.chdir(cwd)
 
 
-def generate_kselftest_fragment(frag, kdir):
+def generate_kselftest_fragment(frag, kdir, skip=[]):
     """Create a config fragment file for kselftest
 
     *frag* is a Fragment object
     *kdir* is the path to a kernel source directory
+    *skip* is a list of kselftests to skip
     """
     shell_cmd(r"""
 set -e
@@ -259,6 +260,8 @@ find \
 """.format(kdir=kdir, frag_path=frag.path))
     with open(os.path.join(kdir, frag.path), 'a') as f:
         for kernel_config in frag.configs:
+            if kernel_config in enumerate(skip):
+                continue
             f.write(kernel_config + '\n')
 
 
@@ -271,6 +274,12 @@ def generate_config_fragment(frag, kdir):
     with open(os.path.join(kdir, frag.path), 'w') as f:
         for kernel_config in frag.configs:
             f.write(kernel_config + '\n')
+
+
+slim_skip_frags = [
+    "tools/testing/selftests/lkdtm/config",
+    "tools/testing/selftests/cpufreq/config"
+]
 
 
 def generate_fragments(config, kdir):
@@ -288,6 +297,8 @@ def generate_fragments(config, kdir):
         print(frag.path)
         if frag.name == 'kselftest':
             generate_kselftest_fragment(frag, kdir)
+        elif frag.name == 'kselftest-slim':
+            generate_kselftest_fragment(frag, kdir, skip=slim_skip_frags)
         elif frag.configs:
             generate_config_fragment(frag, kdir)
 
--- a/config/core/build-configs-cip.yaml
+++ b/config/core/build-configs-cip.yaml
@@ -59,19 +59,21 @@ cip_variants_kselftest: &cip_variants_kselftest
       <<: *cip_architectures
       arm:
         base_defconfig: 'multi_v7_defconfig'
-        fragments: [kselftest]
+        fragments: [kselftest, kselftest-slim]
         extra_configs: ['allnoconfig']
       arm64:
         fragments: [kselftest, arm64-chromebook]
         extra_configs:
           - 'allnoconfig'
           - 'defconfig+arm64-chromebook+kselftest'
+          - 'defconfig+arm64-chromebook+kselftest-slim'
       x86_64:
         base_defconfig: 'x86_64_defconfig'
         fragments: [kselftest, x86-chromebook]
         extra_configs:
           - 'allnoconfig'
           - 'x86_64_defconfig+x86-chromebook+kselftest'
+          - 'x86_64_defconfig+x86-chromebook+kselftest-slim'
 
 
 cip_variants_preempt_rt: &cip_variants_preempt_rt
--- a/config/core/build-configs-stable.yaml
+++ b/config/core/build-configs-stable.yaml
@@ -70,18 +70,20 @@ stable_variants_kselftest: &stable_variants_kselftest
       arm:
         base_defconfig: 'multi_v7_defconfig'
         extra_configs: ['allnoconfig']
-        fragments: [kselftest]
+        fragments: [kselftest, kselftest-slim]
       arm64:
         extra_configs:
           - 'allnoconfig'
           - 'defconfig+arm64-chromebook+kselftest'
+          - 'defconfig+arm64-chromebook+kselftest-slim'
         fragments: [arm64-chromebook, kselftest]
       x86_64:
         base_defconfig: 'x86_64_defconfig'
         extra_configs:
           - 'allnoconfig'
           - 'x86_64_defconfig+x86-chromebook+kselftest'
-        fragments: [x86-chromebook, kselftest]
+          - 'x86_64_defconfig+x86-chromebook+kselftest-slim'
+        fragments: [x86-chromebook, kselftest, kselftest-slim]
 
 
 build_configs:
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -418,7 +418,7 @@ fragments:
       - 'CONFIG_IMA=y'
       - 'CONFIG_IMA_READ_POLICY=y'
 
-  kselftest:
+  kselftest: &kselftest-fragment
     path: "kernel/configs/kselftest.config"
     configs:
       - '# CONFIG_DUMMY is not set'
@@ -426,11 +426,8 @@ fragments:
       - 'CONFIG_NET_IPGRE_DEMUX=m'
 
   kselftest-slim:
+    <<: *kselftest-fragment
     path: "kernel/configs/kselftest-slim.config"
-    configs:
-      - '# CONFIG_DUMMY is not set'
-      - 'CONFIG_NET_IPGRE=m'
-      - 'CONFIG_NET_IPGRE_DEMUX=m'
 
   preempt_rt:
     path: "kernel/configs/preempt_rt.config"
@@ -723,7 +720,6 @@ build_configs_defaults:
 
       fragments: &default_fragments
         - 'debug'
-        - 'kselftest'
         - 'kselftest-slim'
         - 'tinyconfig'
 
@@ -761,7 +757,6 @@ build_configs_defaults:
             - 'allnoconfig'
             - 'defconfig+CONFIG_CPU_BIG_ENDIAN=y'
             - 'defconfig+CONFIG_RANDOMIZE_BASE=y'
-            - 'defconfig+arm64-chromebook+kselftest'
             - 'defconfig+arm64-chromebook+kselftest-slim'
             - 'defconfig+arm64-chromebook+videodec'
           fragments: [arm64-chromebook, crypto, ima, videodec]
@@ -791,7 +786,6 @@ build_configs_defaults:
           extra_configs:
             - 'allmodconfig'
             - 'allnoconfig'
-            - 'x86_64_defconfig+x86-chromebook+kselftest'
             - 'x86_64_defconfig+x86-chromebook+kselftest-slim'
             - 'x86_64_defconfig+x86-chromebook+amdgpu'
           fragments: [amdgpu, crypto, ima, x86_kvm_guest, x86-chromebook]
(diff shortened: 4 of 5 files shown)


broonie:
Successful builds of the kselftest-slim fragment: https://staging.kernelci.org/build/kernelci/branch/staging-mainline/kernel/staging-mainline-20230227.1/
Full fat kselftest config still running fine: https://staging.kernelci.org/test/job/kernelci/branch/staging-mainline/kernel/staging-mainline-20230227.1/plan/kselftest-lkdtm/

broonie:
Some selftest-slim configs running:

https://staging.kernelci.org/test/plan/id/63fd530c6f12a6da709d58cc/
https://staging.kernelci.org/test/plan/id/63fd509bf999869d549d5912/

and one example that didn't finish booting due to an underlying regression in mainline (fix on the way) but which you can see from the logs manages to download and boot the kernel:

https://staging.kernelci.org/test/plan/id/63fd56b469ca1c7df59d59a2/
https://storage.staging.kernelci.org/kernelci/staging-next/staging-next-20230228.0/arm64/defconfig+kselftest-slim/clang-17/lab-broonie/kselftest-landlock-meson-gxl-s905x-libretech-cc.html

Run that actually runs but skips all the tests due to an issue in test-definitions:

https://staging.kernelci.org/test/plan/id/63fd548a130886541b9d58df/

which was the issue here. uImage size for arm64 goes from 53.2Mb to 43.1Mb.

kernelci/build.py
```diff
@@ -273,6 +276,12 @@ def generate_config_fragment(frag, kdir):
             f.write(kernel_config + '\n')
 
 
+slim_skip_frags = [
+    "tools/testing/selftests/lkdtm/config",
+    "tools/testing/selftests/cpufreq/config"
+]
+
+
```
gctucker, line 284: How about defining fragments in YAML which disable some configs and then have e.g. `kselftest+no-lkdtm+no-cpufreq`?  That way we don't need to pile up more logic on this kselftest corner-case fragment.  All the other fragments are entirely defined in YAML and I think it would be better if we also tried to remove the kselftest special case.

Note: A more longer-term solution I think would be to use the generic way of making the kselftest config with `make kselftest-merge` and some way of skipping fragments in the upstream Makefile.

gctucker (commented):
To minimise the extra build load, maybe we could just replace all the `kselftest` builds with `kselftest-slim` and also enable the full `kselftest` builds with the main branches e.g. `next/master`, `mainline/master`, `kselftest/fixes`?

Another, simpler approach is to just disable kselftest on the platforms that can't handle the kernel image size...  At least we know that won't have unexpected side-effects.

broonie:
> To minimise the extra build load, maybe we could just replace all the `kselftest` builds with `kselftest-slim` and also enable the full `kselftest` builds with the main branches e.g. `next/master`, `mainline/master`, `kselftest/fixes`?

Sure, though I'm not sure how exactly to express that in the filters and since we don't actually run branches called those things in staging it makes it very difficult get any further changes in. I'm not sure how loaded we are for builds at the minute.

> Another, simpler approach is to just disable kselftest on the platforms that can't handle the kernel image size... At least we know that won't have unexpected side-effects.

That severely cuts down the set of platforms we've got available to us, and/or increases the admin cost for playing with image size options for those platforms. The whole goal here is to avoid having to do that and get more tests run. The resulting kernels running faster is also a win under emulation.

gctucker:
> since we don't actually run branches called those things in staging

We can run any build config on staging, just some specially tailored ones are run periodically.

> I'm not sure how loaded we are for builds at the minute.

We occasionally hit some congestion.  And yes, not knowing precisely how loaded the system is means we may be overloading it by adding more builds.

> That severely cuts down the set of platforms we've got available to us

In a way, yes of course.  But if such platforms can't run with the full kselftest build then we're not losing any coverage by disabling them.  Then we can start introducing new builds with smaller images and more incrementally extend build load and test coverage.

Whichever path we take, it seems unavoidable to try and get some metrics about the additional build costs, whether we can avoid unnecessary builds, and any change in the number of tests run in the labs.

kernelci/build.py
broonie: That would be nice but it's much more work than I'm likely to be able to do in the medium term - we don't have the infrastructure to negatively apply config fragments like this. Getting something done upstream is probably interesting too, it's not clear that it'd even be recommended to put all the config fragments in one kernel as standard (for the reasons we're seeing here).

broonie:
> > since we don't actually run branches called those things in staging
> 
> We can run any build config on staging, just some specially tailored ones are run periodically.

Right, but doing anything non-standard is really painful - at a minimum you need to send a pull request to kernelci-deploy then wait some indeterminate time to get it deployed which adds a lot of latency and stop energy.

> We occasionally hit some congestion. And yes, not knowing precisely how loaded the system is means we may be overloading it by adding more builds.

It would probably help to get some progress on using the Kubernetes clusters better but that's all stalled out... I suspect we have idle capacity at the minute.

> > That severely cuts down the set of platforms we've got available to us
> 
> In a way, yes of course. But if such platforms can't run with the full kselftest build then we're not losing any coverage by disabling them. Then we can start introducing new builds with smaller images and more incrementally extend build load and test coverage.

It's hurting in cases where things add a fragment (like KVM has done, it's running on some of the affected u-boot platforms) if nothing else, and it is generally a pain when it comes to taking advantage of the labs/boards we have. 

> Whichever path we take, it seems unavoidable to try and get some metrics about the additional build costs, whether we can avoid unnecessary builds, and any change in the number of tests run in the labs.

What would you suggest here?

gctucker:
> Right, but doing anything non-standard is really painful - at a minimum you need to send a pull request to kernelci-deploy then wait some indeterminate time to get it deployed which adds a lot of latency and stop energy.

No you just start a monitor job on bot.staging.kernelci.org with the build config you want to cover.

> What would you suggest here?

Run linux-next on staging, in fact it happens every weekend already so maybe check results on Monday and compare with the production ones to see the increase in build and tests.  Getting some metrics from the build clusters and lab usage would be neat but I don't think we have enough tools in place for that right now.

broonie:
> > Right, but doing anything non-standard is really painful - at a minimum you need to send a pull request to kernelci-deploy then wait some indeterminate time to get it deployed which adds a lot of latency and stop energy.
> 
> No you just start a monitor job on bot.staging.kernelci.org with the build config you want to cover.

Which monitor job are you thinking of here? kernel-tree-monitor-next doesn't seem to have done anything for example.
 
> > What would you suggest here?
> 
> Run linux-next on staging, in fact it happens every weekend already so maybe check results on Monday and compare with the production ones to see the increase in build and tests. Getting some metrics from the build clusters and lab usage would be neat but I don't think we have enough tools in place for that right now.

It does? I'm not seeing the results showing up in the UI.

gctucker:
> > > Right, but doing anything non-standard is really painful - at a minimum you need to send a pull request to kernelci-deploy then wait some indeterminate time to get it deployed which adds a lot of latency and stop energy.
> > 
> > 
> > No you just start a monitor job on bot.staging.kernelci.org with the build config you want to cover.
> 
> Which monitor job are you thinking of here? kernel-tree-monitor-next doesn't seem to have done anything for example.

https://bot.staging.kernelci.org/job/kernel-tree-monitor/build?delay=0sec

Enter the build configs in the `CONFIG_LIST` parameter.

> > > What would you suggest here?
> > 
> > 
> > Run linux-next on staging, in fact it happens every weekend already so maybe check results on Monday and compare with the production ones to see the increase in build and tests. Getting some metrics from the build clusters and lab usage would be neat but I don't think we have enough tools in place for that right now.
> 
> It does? I'm not seeing the results showing up in the UI.

OK it hasn't run since 13th January but that's because something is broken:
https://staging.kernelci.org/job/next/

@mgalka @VinceHillier Could you please take a look why the linux-next jobs aren't running?

gctucker:
This is the starting point, on `main` branch:
```
$ ./kci_build generate_fragments --build-config=next --kdir=/home/gtucker/src/linux
kernel/configs/debug.config
kernel/configs/kselftest.config
kernel/configs/tiny.config
kernel/configs/amdgpu.config
kernel/configs/crypto.config
kernel/configs/ima.config
kernel/configs/kvm_guest.config
kernel/configs/arm64-chromebook.config
kernel/configs/crypto.config
kernel/configs/ima.config
kernel/configs/videodec.config
kernel/configs/rust.config
kernel/configs/rust-samples.config
kernel/configs/kselftest.config
$ ./kci_build list_kernel_configs --build-config=mainline --kdir=~/src/linux | grep kselftest
x86_64 x86_64_defconfig+x86-chromebook+kselftest gcc-10
arm64 defconfig+arm64-chromebook+kselftest gcc-10
$ ./kci_build list_kernel_configs --build-config=next --kdir=~/src/linux | grep kselftest
arm64 defconfig+arm64-chromebook+kselftest gcc-10
x86_64 x86_64_defconfig+x86-chromebook+kselftest gcc-10
```

gctucker:
Now with this change in YAML:
```diff
diff --git a/config/core/build-configs.yaml b/config/core/build-configs.yaml
index f4a2422fdd6a..759ce711596d 100644
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -418,13 +418,17 @@ fragments:
       - 'CONFIG_IMA=y'
       - 'CONFIG_IMA_READ_POLICY=y'
 
-  kselftest:
+  kselftest: &kselftest-fragment
     path: "kernel/configs/kselftest.config"
     configs:
       - '# CONFIG_DUMMY is not set'
       - 'CONFIG_NET_IPGRE=m'
       - 'CONFIG_NET_IPGRE_DEMUX=m'
 
+  kselftest-slim:
+    <<: *kselftest-fragment
+    path: "kernel/configs/kselftest-slim.config"
+
   preempt_rt:
     path: "kernel/configs/preempt_rt.config"
     configs:
@@ -711,7 +715,7 @@ build_configs_defaults:
 
       fragments: &default_fragments
         - 'debug'
-        - 'kselftest'
+        - 'kselftest-slim'
         - 'tinyconfig'
 
       architectures: &default_architectures
@@ -748,7 +752,7 @@ build_configs_defaults:
             - 'allnoconfig'
             - 'defconfig+CONFIG_CPU_BIG_ENDIAN=y'
             - 'defconfig+CONFIG_RANDOMIZE_BASE=y'
-            - 'defconfig+arm64-chromebook+kselftest'
+            - 'defconfig+arm64-chromebook+kselftest-slim'
             - 'defconfig+arm64-chromebook+videodec'
           fragments: [arm64-chromebook, crypto, ima, videodec]
 
@@ -777,7 +781,7 @@ build_configs_defaults:
           extra_configs:
             - 'allmodconfig'
             - 'allnoconfig'
-            - 'x86_64_defconfig+x86-chromebook+kselftest'
+            - 'x86_64_defconfig+x86-chromebook+kselftest-slim'
             - 'x86_64_defconfig+x86-chromebook+amdgpu'
           fragments: [amdgpu, crypto, ima, x86_kvm_guest, x86-chromebook]
 
@@ -1097,15 +1101,46 @@ build_configs:
     tree: mainline
     branch: 'master'
     variants:
-      gcc-10: *default_gcc-10
+      gcc-10:
+        <<: *default_gcc-10
+        fragments: &fragments-kselftest
+          - 'debug'
+          - 'kselftest'
+          - 'kselftest-slim'
+          - 'tinyconfig'
+
+        architectures:
+          <<: *default_architectures
+          arm64: &arm64_arch-kselftest
+            <<: *arm64_arch
+            extra_configs:
+              - 'allmodconfig'
+              - 'allnoconfig'
+              - 'defconfig+CONFIG_CPU_BIG_ENDIAN=y'
+              - 'defconfig+CONFIG_RANDOMIZE_BASE=y'
+              - 'defconfig+arm64-chromebook+kselftest'
+              - 'defconfig+arm64-chromebook+kselftest-slim'
+              - 'defconfig+arm64-chromebook+videodec'
+
+          x86_64: &x86_64_arch-kselftest
+            <<: *x86_64_arch
+            extra_configs:
+              - 'allmodconfig'
+              - 'allnoconfig'
+              - 'x86_64_defconfig+x86-chromebook+kselftest'
+              - 'x86_64_defconfig+x86-chromebook+kselftest-slim'
+              - 'x86_64_defconfig+x86-chromebook+amdgpu'
+
       # Minimum version
       clang-11:
         build_environment: clang-11
         architectures: *arch_clang_configs
+
       # Latest stable release
       clang-16:
         build_environment: clang-16
         architectures: *arch_clang_configs
+
       rustc-1.62:
         build_environment: rustc-1.62
         fragments: [rust, rust-samples, kselftest]
@@ -1150,10 +1185,10 @@ build_configs:
     variants:
       gcc-10:
         build_environment: gcc-10
-        fragments: *default_fragments
+        fragments: *fragments-kselftest
         architectures:
           i386: *i386_arch
-          x86_64: *x86_64_arch
+          x86_64: *x86_64_arch-kselftest
           mips: *mips_arch
           riscv: *riscv_arch
           sparc: *sparc_arch
@@ -1168,6 +1203,7 @@ build_configs:
               - 'defconfig+CONFIG_CPU_BIG_ENDIAN=y'
               - 'defconfig+CONFIG_RANDOMIZE_BASE=y'
               - 'defconfig+arm64-chromebook+kselftest'
+              - 'defconfig+arm64-chromebook+kselftest-slim'
           arm:
             base_defconfig: 'multi_v7_defconfig'
             extra_configs:

```

```
$ ./kci_build generate_fragments --build-config=mainline --kdir=/home/gtucker/src/linux
kernel/configs/debug.config
kernel/configs/kselftest-slim.config
kernel/configs/tiny.config
kernel/configs/crypto.config
kernel/configs/ima.config
kernel/configs/arm64-chromebook.config
kernel/configs/crypto.config
kernel/configs/ima.config
kernel/configs/videodec.config
kernel/configs/amdgpu.config
kernel/configs/crypto.config
kernel/configs/ima.config
kernel/configs/kvm_guest.config
```

```
$ ./kci_build list_kernel_configs --build-config=mainline --kdir=~/src/linux | grep kselftest
x86_64 x86_64_defconfig+x86-chromebook+kselftest gcc-10
x86_64 x86_64_defconfig+x86-chromebook+kselftest-slim gcc-10
arm64 defconfig+arm64-chromebook+kselftest-slim gcc-10
arm64 defconfig+arm64-chromebook+kselftest gcc-10
```

gctucker (commented):
See the comment above, using `kselftest-slim` by default and only enabling the full `kselftest` builds in mainline and next.

gctucker:
Well I think the comments above show how to do that, right?

broonie:
That doesn't seem to be working properly, the fragements for GCC 10 seem to be being ignored.

gctucker:
Guess if you don't want to fix it then we'll just close it and stop running kselftest on boards that can't load the kernel, and create an issue for someone else to have a look?

gctucker:
Of course, it could be that you can't spend more time on it because it's not user-friendly and too time-consuming.  Whatever the reason, we need to get PRs resolved and either someone has to fix these issues they get closed.  Finding simpler alternatives like disabling tests that don't run and creating issues to keep track of what needs to be improved is also a good way of making progress and reducing the PR backlog.  We can't wait another few months until we have the new YAML configuration format and associated CLI tools e.g. `kci config` to merge these things (hopefully they'll be easier to use).

broonie:
Some more guidance would have really helped here, for example a pointer that the config language can't really cope here would have helped a lot here (or just generally some sort of pointer when I said I couldn't see how to express this in the filter language).

Please also consider that there's been a massive change in pace recently with how quickly things are looked at, that does have knock on effects on expectations as a submitter.

gctucker:
Adding `staging-skip` to avoid conflict with #1926 and unblock progress with Rust coverage.

gctucker:
There was just no verification of the results on staging. I can review it again next week.

gctucker:
Finally got round to discussing this with the Collabora lab folks, will come back with a summary soon.  Thanks for your patience :)

nuclearcat:
@broonie 
Do you think it is still will be useful to make such rootfs? As rootfs still used in Maestro, i think PR still relevant.

PR #1761, opened 2023-02-22
Support arm64 virtconfig

broonie:
Upstream has just introduced a new configuration virtconfig for arm64, intended to build more quickly but test mach-virt only. Since it's implemented with a fragment we need to manually trigger builds.


gctucker:

nuclearcat:

PR #1763, opened 2023-02-23
CHROMEOS test-configs-chromeos.yaml: Add kingoftown Chromebook

nuclearcat:
We have one more Trogdor Chromebook which is a bit different from Limozeen. Add standard tast tests and start testing it.


nuclearcat:
kingoftown on dashboard and passing tast tests: https://chromeos.kernelci.org/test/plan/id/63fc471a2fcd2db80fa9d85d/

mgalka (approved):

PR #1764, opened 2023-02-24
Copy `runtime.lava` to `legacy.lava`

gctucker:
Copy all the code in `runtime.lava` to `legacy.lava` to keep it for backwards compatibility while the new runtime implementations are evolving for the new API.


gctucker:
Tested OK on staging with manual build:
https://bot.staging.kernelci.org/job/gtucker/job/gtucker-kernel-build-trigger/14/console

PR #1771, opened 2023-02-27
Consolidate YAML config file arguments

gctucker:
Allow `yaml_config` and `extra_configs` to be defined in the settings file and consolidate the code to have only one path for loading the configuration.

### DIFF
--- a/kci_bisect
+++ b/kci_bisect
@@ -62,6 +62,6 @@ class cmd_get_mbox(Command):
 
 if __name__ == '__main__':
     opts = parse_opts("kci_bisect", globals())
-    configs = kernelci.config.load_all(opts.get_yaml_configs())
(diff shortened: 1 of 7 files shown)


gctucker:
@eds-collabora Hello, could you please confirm whether this has any side-effects on your use-cases with the extra config files?

eds-collabora:
> @eds-collabora Hello, could you please confirm whether this has any side-effects on your use-cases with the extra config files?

I don't see any structural issues here. It seems like a reasonable clean-up, and the existing functionality looks to be preserved. Unless there are subtleties I've missed, I don't think there's any conflict.

gctucker:
@eds-collabora OK thanks.  There's no unit tests or anything on staging.kernelci.org to check whether your use-cases aren't broken, so let me know if you spot anything.

PR #1772, opened 2023-02-27
test-configs.yaml: arm64 ftrace selftsts coverage

broonie:
The ftrace kselftests are very useful but currently we don't try to run
it, add the config fragment and enable it on Juno.  Since a config
fragment is required this will be a lot easier to deploy with the
kselftest-slim PR (https://github.com/kernelci/kernelci-core/pull/1759)
merged.

We will also need https://github.com/Linaro/test-definitions/pull/431
for this to be useful, at the minute the test is unconditionally skipped
everywhere by the Linaro skipfile due to what appear to be no longer
relevant reasons.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -333,6 +333,12 @@ test_plans:
       kselftest_collections: "filesystems"
     filters: *kselftest_no_fragment
 
+  kselftest-ftrace:
+    <<: *kselftest
+    params:
+      job_timeout: '20'
+      kselftest_collections: "ftrace"
+
   kselftest-futex:
     <<: *kselftest
     params:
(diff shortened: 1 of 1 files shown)


broonie:
This is running happily according to LAVA but no sign in the dashboard for some reason https://lava.sirena.org.uk/scheduler/job/196095 - probably my patch at https://lore.kernel.org/linux-kselftest/20230319232438.786c46feaf9bb7ddcb78a731@kernel.org/ will improve that? We should at least have the single pass/fail for the suite as a whole but it's vanished into the ether.

broonie:
Running OK on staging now https://staging.kernelci.org/test/plan/id/647c3953fee8acc1eac5926a/ (there's issues with the tests failing but the bit where we run things works OK).

gctucker:

broonie:
Once more running happily on staging: https://staging.kernelci.org/test/plan/id/659bbac514a1801bdd62ee68/

nuclearcat:
Thanks, merging

PR #1774, opened 2023-03-02
Add `kci job` command

### LINKED ISSUE #1728: Enable generating jobs using `kci job generate`
Once the `lab` stuff has been reworked as `runtime`, the next step is to implement the `kci job` commands with the matching parts in `kernelci.runtime` and potentially `kernelci.job`.  It should take a checkout node as input and generate a job for `kver` or `kunit`.  Then a `kci job submit` command should be available to run the job via the `runtime` implementation.

- [x] tested locally with `shell` runtime (`kver` and `kunit` tests)
- [x] tested manually with `k8s` runtime (`kver`)
- [x] tested on staging with both `shell` ~~and `k8s` runtimes~~ kind of, but KUnit builds are broken in mainline kernel

gctucker:
Add a `kci job` command to run jobs with the new API.  Sample commands:

```
kci job register --input-node-id=<checkout node id> --plan=kver --id-only
kci job generate --runtime-config=shell --platform-config=shell --node-id=<job node id>
kci job submit --runtime-config=shell job/kver
```

Fixes #1728 
Depends on #1773

### DIFF
--- a/kernelci/cli/__init__.py
+++ b/kernelci/cli/__init__.py
@@ -9,12 +9,14 @@ import sys
 
 from .base import Args, Command, parse_opts, sub_main
 from . import (
+    job,
     node,
     user,
     validate,
 )
 
 _COMMANDS = {
+    'job': job.main,
     'node': node.main,
     'user': user.main,
     'validate': validate.main,
--- a/kernelci/cli/base.py
+++ b/kernelci/cli/base.py
@@ -377,8 +377,8 @@ class Command(abc.ABC):
     """
 
     help = None
-    args = None
-    opt_args = None
+    args = []
+    opt_args = []
 
     def __init__(self, sub_parser, name):
         """This class is to facilitate creating command line utilities
@@ -445,10 +445,9 @@ class Command(abc.ABC):
 
 class APICommand(Command):  # pylint: disable=too-few-public-methods
     """Base command class for interacting with the KernelCI API"""
-    args = [
+    args = Command.args + [
         Args.api_config, Args.api_token,
     ]
-    opt_args = []
 
     @classmethod
     def _get_api(cls, configs, args):
--- a/kernelci/cli/base.py
+++ b/kernelci/cli/base.py
@@ -459,6 +459,11 @@ class APICommand(Command):  # pylint: disable=too-few-public-methods
         n_indent = 0 if indent is None else int(indent)
         print(json.dumps(data, indent=n_indent))
 
+    @classmethod
+    def _load_json(cls, json_path, encoding='utf-8'):
+        with open(json_path, encoding=encoding) as json_file:
+            return json.load(json_file)
+
 
 class Options:
     """Options based on user settings with CLI override."""
--- a/kernelci/cli/job.py
+++ b/kernelci/cli/job.py
@@ -1,0 +1,13 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+#
+# Copyright (C) 2022-2023 Collabora Limited
+# Author: Guillaume Tucker <guillaume.tucker@collabora.com>
+
+"""Tool to generate and run KernelCI jobs"""
+
+from .base import sub_main
+
+
+def main(args=None):
+    """Entry point for the command line tool"""
+    sub_main("job", globals(), args)
--- a/kernelci/cli/job.py
+++ b/kernelci/cli/job.py
@@ -8,7 +8,7 @@
 import yaml
 
 import kernelci.runtime
-from .base import APICommand, Args, sub_main
+from .base import APICommand, Args, Command, sub_main
 
 
 class cmd_register(APICommand):  # pylint: disable=invalid-name
@@ -125,6 +125,25 @@ Invalid arguments.  Either --node-id or --node-json is required.")
             print(job)
 
         return True
+
+
+class cmd_submit(Command):  # pylint: disable=invalid-name
+    """Submit a job definition from a file"""
+    args = Command.args + [
+        {
+            'name': '--runtime-config',
+            'help': "Name of the runtime config",
+        },
+        {
+            'name': 'job_path',
+            'help': "Path of the job file to submit",
+        },
+    ]
+
+    def __call__(self, configs, args):
+        runtime_config = configs['runtimes'][args.runtime_config]
+        runtime = kernelci.runtime.get_runtime(runtime_config)
+        runtime.submit(args.job_path)
 
 
 def main(args=None):
--- a/kernelci/cli/job.py
+++ b/kernelci/cli/job.py
@@ -5,6 +5,9 @@
 
 """Tool to generate and run KernelCI jobs"""
 
+import yaml
+
+import kernelci.runtime
 from .base import APICommand, Args, sub_main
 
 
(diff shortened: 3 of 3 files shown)


JenySadadia (commented):
I tried to run the `job register` command locally.
```
$ ./kci job register --input-node-id=63c549319fb3b62c7626e7f9 --plan=kver --id-only --api-config=docker-host --api-token=
  File "/home/jeny/kernelCI/kernelci-core-env/src/gctucker/kernelci-core/./kci", line 29, in <module>
  File "/home/jeny/kernelCI/kernelci-core-env/src/gctucker/kernelci-core/kernelci/cli/__init__.py", line 40, in call
    _COMMANDS[name](args)
  File "/home/jeny/kernelCI/kernelci-core-env/src/gctucker/kernelci-core/kernelci/cli/job.py", line 151, in main
    sub_main("job", globals(), args)
  File "/home/jeny/kernelCI/kernelci-core-env/src/gctucker/kernelci-core/kernelci/cli/base.py", line 693, in sub_main
  File "/home/jeny/kernelCI/kernelci-core-env/src/gctucker/kernelci-core/kernelci/cli/job.py", line 58, in __call__
    plan_config = configs['test_plans'][args.plan]
KeyError: 'kver'
```
Do I need any additional configuration?

gctucker:
> Do I need any additional configuration?

Yes you need to use the `pipeline.yaml` file from `kernelci-pipeline`.

JenySadadia:
> > Do I need any additional configuration?
> 
> Yes you need to use the `pipeline.yaml` file from `kernelci-pipeline`.

OK. Now, it is not able to find job template.
```
$ ./kci job --yaml-config /home/jeny/kernelCI/kernelci-pipeline-env/src/kernelci-pipeline/config/pipeline.yaml generate --runtime-config=shell --platform=shell --node-id=6401db0f1d5105667a52c6f1 --api-token=x --api-config=docker-host --runtime-config=shell
  File "/home/jeny/kernelCI/kernelci-core-env/src/gctucker/kernelci-core/./kci", line 29, in <module>
  File "/home/jeny/kernelCI/kernelci-core-env/src/gctucker/kernelci-core/kernelci/cli/__init__.py", line 40, in call
    _COMMANDS[name](args)
  File "/home/jeny/kernelCI/kernelci-core-env/src/gctucker/kernelci-core/kernelci/cli/job.py", line 151, in main
    sub_main("job", globals(), args)
  File "/home/jeny/kernelCI/kernelci-core-env/src/gctucker/kernelci-core/kernelci/cli/base.py", line 693, in sub_main
  File "/home/jeny/kernelCI/kernelci-core-env/src/gctucker/kernelci-core/kernelci/cli/job.py", line 120, in __call__
    job = runtime.generate(params, platform_config, plan_config)
  File "/home/jeny/kernelCI/kernelci-core-env/src/gctucker/kernelci-core/kernelci/runtime/shell.py", line 21, in generate
    template = jinja2_env.get_template(template_path)
  File "/home/jeny/kernelCI/kernelci-core-env/lib/python3.10/site-packages/jinja2/environment.py", line 1000, in get_template
    return self._load_template(name, globals)
  File "/home/jeny/kernelCI/kernelci-core-env/lib/python3.10/site-packages/jinja2/environment.py", line 959, in _load_template
    template = self.loader.load(self, name, self.make_globals(globals))
  File "/home/jeny/kernelCI/kernelci-core-env/lib/python3.10/site-packages/jinja2/loaders.py", line 126, in load
    source, filename, uptodate = self.get_source(environment, name)
  File "/home/jeny/kernelCI/kernelci-core-env/lib/python3.10/site-packages/jinja2/loaders.py", line 218, in get_source
    raise TemplateNotFound(template)
jinja2.exceptions.TemplateNotFound: kver.jinja2
```

JenySadadia (commented):
It would be good to have some documentation for these newly added commands.

gctucker:
Yes I think documentation will come after, I want to get some initial feedback on them first.

gctucker:
@JenySadadia I guess the easiest way to try the commands is to start an interactive shell in a `kernleci-pipeline` container as the tools should be in the `kernelci/staging-kernelci` image.  That way you get the YAML config and job templates etc. from the pipeline already in the container too.

gctucker:
A few things to consider, for example:

* Is `register` a good name in `kci job register` to create the node before running a job?
* Likewise, how about `generate` and `submit`?
* Does it seem confusing to have to provide `--runtime-config` both for `generate` and `submit`?
* Is it a good idea to both have a way to provide the path to JSON files and just the raw node ID?
* Is the `--indent` argument with a number intuitive enough?

Generally speaking I'd like to know if this makes sense to someone else.  I wanted to have something functional and that's solved as far as I can tell, but the names of the commands and arguments as well as the general CLI workflow / UX matters from a usability point of view.

JenySadadia:
> A few things to consider, for example:
> 
> * Is `register` a good name in `kci job register` to create the node before running a job?

This is about creating a job node rather than registering the node with some job. How about naming it something with `create` in it?
> * Likewise, how about `generate` and `submit`?

Looks OK to me.
> * Does it seem confusing to have to provide `--runtime-config` both for `generate` and `submit`?

I don't think so.
> * Is it a good idea to both have a way to provide the path to JSON files and just the raw node ID?

Yes, it is a good idea to provide multiple ways for the node ID. But the user would prefer to provide ID rather than creating JSON and providing the path. At least I would prefer that.
> * Is the `--indent` argument with a number intuitive enough?
> 

How about `--output-indent`?
> Generally speaking I'd like to know if this makes sense to someone else. I wanted to have something functional and that's solved as far as I can tell, but the names of the commands and arguments as well as the general CLI workflow / UX matters from a usability point of view.

gctucker:
> > A few things to consider, for example:
> > 
> > * Is `register` a good name in `kci job register` to create the node before running a job?
> 
> This is about creating a job node rather than registering the node with some job. How about naming it something with `create` in it?

It's not actually just about creating a node, even though that's what it does under the hood.  It's from a logic point of view, to tell the API that a job is going to be run.  I first called this command `create_node` but then found if very confusing, because it's not clear whether this should be done before generating a job and it also seems too similar to some `kci node` commands.

> > * Likewise, how about `generate` and `submit`?
> 
> Looks OK to me.
> 
> > * Does it seem confusing to have to provide `--runtime-config` both for `generate` and `submit`?
> 
> I don't think so.

OK.  Somehow I was wondering if there could be a way to deduce the runtime from the generated job to avoid having to provide it again in `submit`.  Also I'm wondering if there's any valid use-case where we would want to generate a job for one runtime and submit it with another.  But we'll see how things evolve.

> > * Is it a good idea to both have a way to provide the path to JSON files and just the raw node ID?
> 
> Yes, it is a good idea to provide multiple ways for the node ID. But the user would prefer to provide ID rather than creating JSON and providing the path. At least I would prefer that.

I think that depends on the use case.

> > * Is the `--indent` argument with a number intuitive enough?
> 
> How about `--output-indent`?

Well that's longer, did you find any ambiguity with `--indent`?

> > Generally speaking I'd like to know if this makes sense to someone else. I wanted to have something functional and that's solved as far as I can tell, but the names of the commands and arguments as well as the general CLI workflow / UX matters from a usability point of view.

Have you managed to run e.g. `kver` with these commands?

gctucker:
OK so I think the summary is that maybe `register` could be changed but there's no obvious alternative name now, and if `--indent` is ambiguous we could tweak the name too. Otherwise, I'll keep considering how to include the runtime name or config in the generated job to avoid having to supply it again with `submit`.

gctucker:
Maybe `kci job init` would be better than `kci job register`.  But we can rework that later, just a thought.

JenySadadia:
> Maybe `kci job init` would be better than `kci job register`. But we can rework that later, just a thought.

Yes, seems OK to me.

PR #1775, opened 2023-03-02
Add `kernelci.api` to load the existing API module

gctucker:
As an intermediate step before having a new `API` class with API bindings, add the `kernelci.api` package with a loader to dynamically import the matching module.  Right now the module is hard-coded but it will become useful when supporting different API versions.

This also decouples the Python dependencies needed for the API with the command line tools e.g. `cloudevents` is now only imported when an API object is needed (this particular Python package is not in Debian so it's particularly inconvenient).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,7 @@ test: \
 pylint:
 	pylint --reports=y \
 		kci \
+		kernelci.api \
 		kernelci.cli \
 		kernelci.config.api \
 		kernelci.config.runtime \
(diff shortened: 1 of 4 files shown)


gctucker:
@mgalka @nuclearcat @aliceinwire Hello, could you please review and merge if you're OK with this?  I think it would be good to have it in today's release so it propagates to the production Docker images.

mgalka (approved):
LGTM

PR #1777, opened 2023-03-02
Runtime rework

gctucker:
Rework the `Runtime` class to make it better aligned with the new API & Pipeline.

### DIFF
--- a/kernelci/runtime/__init__.py
+++ b/kernelci/runtime/__init__.py
@@ -87,8 +87,8 @@ class Runtime(abc.ABC):
         return params
 
     @abc.abstractmethod
-    def generate(self, params, device_config, plan_config, callback_opts=None,
-                 templates_paths=None, runtime_config=None):
+    def generate(self, params, device_config, plan_config,
+                 templates_paths=None):
         """Generate a test job definition.
 
         *params* is a dictionary with the test parameters which can be used
@@ -100,14 +100,10 @@ class Runtime(abc.ABC):
         *plan_config* is a TestPlan configuration object for the target test
              plan
 
-        *callback_opts* is a dictionary with extra options used for callbacks
-
         *templates_paths* is an optional argument to specify the path(s) where
             the template files should be found, when not in the standard
             location.  This could be either a string or a list of strings to
             provide several paths.
-
-        *runtime_config* is a configuration object for the runtime environment
         """
 
     def save_file(self, job, output_path, params):
--- a/kernelci/runtime/__init__.py
+++ b/kernelci/runtime/__init__.py
@@ -9,6 +9,8 @@ import json
 import os
 import yaml
 
+from jinja2 import Environment, FileSystemLoader
+
 
 def add_kci_raise(env):
     def template_exception(msg):
@@ -50,6 +52,11 @@ class Runtime(abc.ABC):
 
     def _connect(self, *args, **kwargs):
         return None
+
+    def _get_template(self, plan_config):
+        jinja2_env = Environment(loader=FileSystemLoader(self.templates))
+        template_path = plan_config.get_template_path(None)
+        return jinja2_env.get_template(template_path)
 
     def import_devices(self, data):
         """Import devices information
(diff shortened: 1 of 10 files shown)


gctucker:
Tested OK on staging with job submitted by hand:
https://bot.staging.kernelci.org/job/gtucker/job/gtucker-kernel-build-trigger/16/console
https://staging.kernelci.org/test/job/gtucker/branch/kernelci-staging/kernel/kernelci-staging-20230306.0/

gctucker:
Tested OK on staging with the new API & Pipeline:
```
03/06/2023 01:43:35 PM UTC [INFO] 2023-03-06 13:43:35.785458  396702e0748e  6405edf0d32374a94665efdd  Done       Pass      kver
```

kernelci/runtime/__init__.py
```diff
@@ -3,44 +3,68 @@
...
+def add_kci_raise(jinja2_env):
+    """Add a kci_raise function to use in templates
 
-def add_kci_raise(env):
+    This adds a `kci_raise` function to a given Jinja2 environment `jinja2_env`
+    so it can be used to raise exceptions from template files when for example
+    some template parameters are not valid.
+    """
     def template_exception(msg):
         raise Exception(msg)
```
mgalka, line 25: It's not part of this PR, but maybe it'd be good to use specialized exception class here like `TemplateError` or `TemplateParameterError` instead of a generic `Exception`. This could improve further exceptions handling.

kernelci/runtime/kubernetes.py
```diff
@@ -3,28 +3,28 @@
...
-        template_path = plan_config.get_template_path(None)
-        template = jinja2_env.get_template(template_path)
-        job_name = '-'.join([
-            'kci', params['node_id'], params['name'][:24]
-        ])
+    def generate(self, params, plan_config):
+        template = self._get_template(plan_config)
+        job_name = '-'.join(['kci', params['node_id'], params['name'][:24]])
         safe_name = re.sub(r'[\:/_+=]', '-', job_name).lower()
         rand_sx = ''.join(random.sample(self.RANDOM_CHARACTERS, 8))
```
mgalka, line 29: Again, this change hasn't been introduced here, but maybe it's a good idea to change `RANDOM_CHARACTERS` name to something more descriptive - this characters are not really random. Maybe  simply use `(ascii_lowercase+digits)` here.

mgalka (commented):
small suggestions

kernelci/runtime/kubernetes.py
gctucker: I think `RANDOM_CHARACTERS` is actually better as the set is an implementation detail.  We could call it `RANDOM_K8S_JOB_NAME_CHARACTERS` but that's longer and not very useful for a class that's not used anywhere else.

kernelci/runtime/__init__.py
gctucker: Yes, error handling needs to be revisited throughout the code base in general.  In this case, we might need to define several handlers or add a variable with the type of exception to raise.  Also there might be something more suitable in recent versions of Jinja2.

mgalka: Maybe we could make it a follow-up?

kernelci/runtime/kubernetes.py
gctucker: OK so something like `K8S_JOB_NAME_CHARACTERS` would be more descriptive, even though I think that's still not the best name we can come up with.

gctucker: Alright I've renamed it.

kernelci/runtime/__init__.py
gctucker: #1797

gctucker:
@mgalka I think I've addressed all the points, unless there's anything else I guess this could be merged later today after the production update.

mgalka:
> @mgalka I think I've addressed all the points, unless there's anything else I guess this could be merged later today after the production update.

All looks good.

mgalka (approved):

gctucker:
Or before :p There was no rush as this isn't required in production, but as long as it's not breaking anything then fine.

PR #1781, opened 2023-03-04
test-configs.yaml: Raise memory allocation for arm64 qemu

broonie:
arm64 debug configurations are starting to get rather large and run into
problems booting with the default memory allocation of 512m, especially in
UEFI configurations. Increase the amount of memory guests are given to a
gigabyte to avoid these issues.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -1655,6 +1655,7 @@ device_types:
       cpu: 'max,pauth-impdef=on'
       guestfs_interface: 'virtio'
       machine: 'virt,gic-version=2,mte=on,accel=tcg'
+      memory: 1g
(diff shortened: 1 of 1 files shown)


nuclearcat (approved):
Can see successful jobs:
  242 01:03:37.875296  <6>[    0.000000] Memory: 947360K/1048576K available (16192K kernel code, 4182K rwdata, 9856K rodata, 8256K init, 607K bss, 68448K reserved, 32768K cma-reserved)
https://storage.staging.kernelci.org/kernelci/staging-next/staging-next-20230313.0/arm64/defconfig/gcc-10/lab-broonie/baseline-qemu_arm64-virt-gicv2.html

PR #1782, opened 2023-03-06
bullseye-gst-fluster.sh: Disable caching if proxy is unreachable

nuclearcat:
As caching service available on production only, we need to find way to disable it dynamically on staging.
One of ways to test if it works by using curl.
If port is not reachable, then disable it.


nuclearcat:
Built job with this patch on staging, it worked well as no proxy on staging available.
https://bot.staging.kernelci.org/job/rootfs-builder/445/console

config/rootfs/debos/scripts/bullseye-gst-fluster.sh
```diff
@@ -29,12 +29,24 @@ CACHING_SERVICE="http://kernelci1.eastus.cloudapp.azure.com:8888/cache?uri="
...
 pip3 install meson
 
+# Verify if CACHING_SERVICE is reachable by curl, if not, unset it
+if [ ! -z ${CACHING_SERVICE} ]; then
+  # disable set -e
+  set +e
+  curl -s --head --request GET ${CACHING_SERVICE}
+  # if exit code is not 0, unset CACHING_SERVICE
+  if [ $? -ne 0 ]; then
+	unset CACHING_SERVICE
```
mgalka, line 45: It'd be good to put a message here to say explicitly that we're disabling the use of caching service. Just to make sure we're not falling back silently.

nuclearcat: Done, i did that actually, just forgot to push updates :)

PR #1784, opened 2023-03-06
Add and run igt-kms-mediatek on MediaTek Chromebooks

nfraprado:
Add IGT KMS test plan for MediaTek and enable on MediaTek Chromebooks: Hana, Juniper and Spherion. Even though all three use the same DRM driver (mediatek), each one has a different display pipeline, so it's worth testing on all of them.

Hana (MT8173 Elm based) uses a PS8640 DSI-eDP bridge. Spherion (MT8192 Asurada based) and Juniper (MT8183 Kukui Jacuzzi based) both use the ANX7625 DSI-DP bridge, but they use different eDP panels: Spherion has IVO 0x075d; Juniper has AUO 0x145c.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -204,6 +204,12 @@ test_plans:
       <<: *igt-kms-test-list
       drm_driver: "exynos"
 
+  igt-kms-mediatek:
+    <<: *igt-kms
+    params:
+      <<: *igt-kms-test-list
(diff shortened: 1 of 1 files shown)


nuclearcat (approved):
It works and producing useful test results:
https://staging.kernelci.org/test/plan/id/640e73cbd13d595d4f9d58cf/

PR #1790, opened 2023-03-09
`kci config` command

gctucker:
Rename `kci validate` as `kci config` with commands to validate the YAML config and list the YAML source files.


gctucker:
This depends on #1789 only to avoid conflicts in `kernelci.cli` as both `api` and `config` commands are being added.  As such, also mark it as a draft PR with the `staging-skip` label for now.

PR #1798, opened 2023-03-13
Add API implementation for node requests

gctucker:
Add implementation for the requests to get / count / create / update nodes.  Update the `kci node` tool accordingly.


gctucker:
Tested OK on staging in combination with all the other related PRs.

PR #1457, opened 2022-10-02
CVE-2007-4559 Patch

TrellixVulnTeam:
# Patching CVE-2007-4559

Hi, we are security researchers from the Advanced Research Center at [Trellix](https://www.trellix.com). We have began a campaign to patch a widespread bug named CVE-2007-4559. CVE-2007-4559 is a 15 year old bug in the Python tarfile package. By using extract() or extractall() on a tarfile object without sanitizing input, a maliciously crafted .tar file could perform a directory path traversal attack. We found at least one unsantized extractall() in your codebase and are providing a patch for you via pull request. The patch essentially checks to see if all tarfile members will be extracted safely and throws an exception otherwise. We encourage you to use this patch or your own solution to secure against CVE-2007-4559. Further technical information about the vulnerability can be found in this [blog](https://www.trellix.com/en-us/about/newsroom/stories/research/tarfile-exploiting-the-world.html).

If you have further questions you may contact us through this projects lead researcher [Kasimir Schulz](mailto:kasimir.schulz@trellix.com).

PR #1799, opened 2023-03-13
CVE-2007-4559 Patch by Trellix

nuclearcat:
This is just copy of PR https://github.com/kernelci/kernelci-core/pull/1457

Cherry-picked their patches and added DSO and copyright/kudos to Trellix team.


nuclearcat:
These patches doesn't pass pycodestyle checks, its better to rewrite them

PR #1801, opened 2023-03-14
CHROMEOS rootfs-configs-chromeos.yaml: Add arm64 to bullseye-cros-flash

nuclearcat:
As we have ChromiumOS ARM64 Chromebooks, we need "flasher" image for them too.


nuclearcat:
Done

PR #1802, opened 2023-03-14
Add `APIHelper` class

gctucker:
Add a class with middleware ported from `kernelci.db.kernelci_api` on top of the actual `API` class.

Depends on #1793


gctucker:
Merging this now as it has been tested on staging.  It won't be enabled by default yet until other PRs are merged to use `kernelci.api` in the pipeline and job templates.

Issue #1770, opened 2023-02-27
Convert kci_docker into a `kci docker` command


gctucker:
The current `kci_docker` in `config/docker` should be convered to a command line too `kci docker` to align it with the other new commands.  Also, some minor tweaks and improvements might be needed so it's a good opportunity to review it and fix `pylint` errors etc.

In particular, the `--fragments` option currently takes a number of fragments but the order matters.  This is rather implicit as the order optional arguments does not usually matter with command line tools.  So one way to deal with this would be to make it accept a single string with a list of fragments e.g. `--fragments kselftest kernelci`, or another way would be to drop the `--fragments` option and take any positional arguments after the base image name as fragments.  I believe the latter might be a better approach as it would directly map to the image name / tag.  For example:

```
kci docker build --arch=x86 gcc-10 kselftest kernelci
```
would build the `kernelci/gcc-10:x86-kselftest-kernelci` image.

- [ ] `kci_docker` reworked as `kci` sub-command with implementation in `kernelci.cli.docker`
- [ ] `pylint` errors in `kernelci.cli.docker` fixed
- [ ] improvements implemented e.g. `--fragments` optional argument reworked
- [ ] `staging.kernelci.org` and `kernelci.org` scripts in `kernelci-deploy` updated accordingly
- [ ] tested OK on staging

PR #1805, opened 2023-03-15
T41052 Convert kci_docker into a kci docker command

JenySadadia:
Fixes https://github.com/kernelci/kernelci-core/issues/1770

Add a command line utility `kci docker` from `config.docker.kci_docker` script.
The tool will be used in the same way as `kci_docker` script to generate Dockerfiles from Jinja2 templates and build images. It has `name`, `generate`, and `build` as sub-commands.
The script has been reworked to add base command classes to incorporate common arguments and methods for sub-commands.
Also, fixed some `pylint` errors by adding module and function docstrings.
Improved `fragment` option.

### DIFF
--- a/kernelci/cli/docker.py
+++ b/kernelci/cli/docker.py
@@ -1,0 +1,203 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+#
+# Copyright (C) 2022, 2023 Collabora Limited
+# Author: Guillaume Tucker <guillaume.tucker@collabora.com>
+# Author: Jeny Sadadia <jeny.sadadia@collabora.com>
+
+"""Tool to generate Dockerfiles from Jinja2 templates and build images"""
+
+
+import io
+import json
+
+from jinja2 import Environment, FileSystemLoader
+import docker
+
+from .base import Command, sub_main
+
+
+class DockerCommand(Command):
+    """Base command class for docker sub-commands"""
+
+    args = Command.args + [
+        {
+            'name': 'image',
+            'help': "Docker image name, e.g. gcc-10:x86",
+        },
+        {
+            'name': '--prefix',
+            'help': "Docker image tag prefix",
+        },
+    ]
+    opt_args = Command.opt_args + [
+        {
+            'name': '--arch',
+            'help': "CPU architecture, e.g. x86",
+        },
+        {
+            'name': '--fragment',
+            'action': 'append',
+            'help': "Extra fragments, e.g. kernelci",
+        },
+        {
+            'name': '--image-version',
+            'help': "Docker image version tag e.g. 20221011.0",
+        },
+    ]
+
+    @classmethod
+    def _gen_image_name(cls, args):
+        base_name = args.prefix + args.image
+        tag_strings = ([args.arch] if args.arch else []) + (
+            args.fragment or [])
+        if args.image_version:
+            tag_strings.append(args.image_version)
+        tag_name = '-'.join(tag_strings)
+        name = ':'.join((base_name, tag_name)) if tag_name else base_name
+        return base_name, tag_name, name
+
+
+class DockerBuildGenerateCommand(DockerCommand):
+    """Base command class for docker generate and build sub-commands"""
+
+    TEMPLATE_PATH = 'config/docker'
+
+    @classmethod
+    def _get_template_params(cls, prefix, fragment):
+        params = {
+                'prefix': prefix,
+                'fragments': [
+                    f'fragment/{fragment}.jinja2'
+                    for fragment in fragment
+                ] if fragment else []
+            }
+        return params
+
+    def _gen_dockerfile(self, img_name, params):
+        jinja2_env = Environment(loader=FileSystemLoader(self.TEMPLATE_PATH))
+        template = jinja2_env.get_template(f"{img_name}.jinja2")
+        return template.render(params)
+
+    def _get_dockerfile(self, args):
+        params = self._get_template_params(args.prefix, args.fragment)
+        template = (
+            '-'.join((args.image, args.arch)) if args.arch else args.image
+        )
+        return self._gen_dockerfile(template, params)
+
+
+class cmd_build(DockerBuildGenerateCommand):  # pylint: disable=invalid-name
+    """Build image"""
+
+    opt_args = DockerCommand.opt_args + [
+        {
+            'name': '--build-arg',
+            'action': 'append',
+            'help': "Build argument to pass to docker build",
+        },
+        {
+            'name': '--no-cache',
+            'action': 'store_true',
+            'help': "Disable Docker cache when building the image",
+        },
+        {
+            'name': '--push',
+            'action': 'store_true',
+            'help': "Push the image to Docker Hub after building",
+        },
+        {
+            'name': '--verbose',
+            'action': 'store_true',
+            'help': "Verbose output",
+        },
+    ]
+
+    @classmethod
+    def _build_image(cls, dockerfile, tag, buildargs, nocache):
+        client = docker.from_env()
+        dockerfile_obj = io.BytesIO(dockerfile.encode())
+        return client.images.build(
+            fileobj=dockerfile_obj, tag=tag, buildargs=buildargs,
+            nocache=nocache,
+        )
+
+    @classmethod
+    def _dump_dockerfile(cls, dockerfile):
+        sep = '--------------------------------------------------' * 2
+        print()
+        print(sep)
+        print(dockerfile)
+        print(sep)
+        print()
+
+    @classmethod
+    def _dump_log(cls, build_log):
+        for chunk in build_log:
+            stream = chunk.get('stream') or ""
+            for line in stream.splitlines():
+                print(line)
+
+    @classmethod
+    def _push_image(cls, base_name, tag_name):
+        client = docker.from_env()
+        push_log_json = client.images.push(base_name, tag_name)
+        return list(
+            json.loads(json_line)
+            for json_line in push_log_json.splitlines()
+        )
+
+    @classmethod
+    def _dump_push_log(cls, log):
+        for line in log:
+            if 'status' in line and 'progressDetail' not in line:
+                print(line['status'])
+
+    def __call__(self, configs, args):
+        base_name, tag_name, name = self._gen_image_name(args)
+
+        dockerfile = self._get_dockerfile(args)
+        if args.verbose:
+            self._dump_dockerfile(dockerfile)
+
+        print(f"Building {name}")
+        buildargs = dict(
+            barg.split('=') for barg in args.build_arg
+        ) if args.build_arg else {}
+        _, build_log = self._build_image(
+            dockerfile, name, buildargs, args.no_cache
+        )
+        if args.verbose:
+            self._dump_log(build_log)
+        if args.push:
+            print(f"Pushing {name} to Docker Hub")
+            push_log = self._push_image(base_name, tag_name)
+            if args.verbose:
+                self._dump_push_log(push_log)
+            for line in push_log:
+                error = line.get('errorDetail')
+                if error:
+                    print(error['message'])
+                    return False
+        return True
+
+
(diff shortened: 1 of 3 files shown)


JenySadadia:
Fixed `pycodestyle` errors.

JenySadadia:
Rebased it.

kernelci/cli/docker.py
```diff
@@ -0,0 +1,203 @@
...
+            'action': 'append',
+            'help': "Extra fragments, e.g. kernelci",
+        },
+        {
+            'name': '--image-version',
+            'help': "Docker image version tag e.g. 20221011.0",
+        },
+        {
+            'name': '--prefix',
+            'default': 'kernelci/',
```
gctucker, line 45: Please remove the default, otherwise we can't load it from the `kernelci.conf` settings as this always takes priority.  It's a small weakness of how the command line arguments and settings are handled, at the moment the only way to provide a last-resort default is in the code.  This wasn't a problem for `kci_docker` as it's not using the settings file.

kernelci/cli/docker.py
```diff
@@ -0,0 +1,203 @@
...
+            'help': "Docker image tag prefix",
+        },
+    ]
+
+    @classmethod
+    def _gen_image_name(cls, args):
+        base_name = args.prefix + args.image
+        tag_strings = ([args.arch] if args.arch else []) + (
+            args.fragment or [])
```
gctucker, line 54: I know it was written this way in `kci_docker` but now I think the `tag_strings` should be rewritten this way for clarity as it can't really be all defined on one line with 3 conditionals:
```python
        tag_strings = [args.arch] if args.arch else []
        if args.fragment:
            tag_strings.extend(args.fragment)
        if args.image_version:
            tag_strings.append(args.image_version)
        tag_name = '-'.join(tag_strings)
```
(also `fragment` should probably become `fragments` but that's another thing).

kernelci/cli/docker.py
```diff
@@ -0,0 +1,203 @@
...
+            tag_strings.append(args.image_version)
+        tag_name = '-'.join(tag_strings)
+        name = ':'.join((base_name, tag_name)) if tag_name else base_name
+        return base_name, tag_name, name
+
+
+class DockerBuildGenerateCommand(DockerCommand):
+    """Base command class for docker generate and build sub-commands"""
+
+    TEMPLATE_PATH = 'config/docker'
```
gctucker, line 65: I think we'll also need to look in `/etc/kernelci/config/docker` as a standard installation path.  The `kernelci` Docker images already have all the config files installed there.  We might also add a command line argument to specify other paths for user-supplied templates but that's not strictly required right now.

kernelci/cli/docker.py
```diff
@@ -0,0 +1,203 @@
...
+
+    @classmethod
+    def _get_template_params(cls, prefix, fragment):
+        params = {
+                'prefix': prefix,
+                'fragments': [
+                    f'fragment/{fragment}.jinja2'
+                    for fragment in fragment
+                ] if fragment else []
```
gctucker, line 74: Using `fragment` for 2 different things is going to cause some problems.  This wasn't a problem in `kci_docker` as it had `for fragment in args.fragment`.  In this case, I think we can just rename the command line arg to `--fragments` and then we'll have `for fragment in fragments`.

kernelci/cli/docker.py
```diff
@@ -0,0 +1,203 @@
...
+            }
+        return params
+
+    def _gen_dockerfile(self, img_name, params):
+        jinja2_env = Environment(loader=FileSystemLoader(self.TEMPLATE_PATH))
```
gctucker, line 79: I think we can directly pass a list of paths to the loader, so having `TEMPLATE_PATHS` with both a local directory and the one in `/etc` should just work here.

kernelci/cli/docker.py
```diff
@@ -0,0 +1,203 @@
...
+        template = jinja2_env.get_template(f"{img_name}.jinja2")
+        return template.render(params)
+
+    def _get_dockerfile(self, args):
+        params = self._get_template_params(args.prefix, args.fragment)
+        template = '-'.join(
+            (args.image, args.arch)) if args.arch else args.image
```
gctucker, line 86: Maybe it's just me but I find it confusing to have an open bracket and the closing one in the middle of the next line.  This would be easier to read I think:
```python
        template = (
            '-'.join((args.image, args.arch)) if args.arch else args.image
        )
```

kernelci/cli/docker.py
```diff
@@ -0,0 +1,203 @@
...
+        if args.push:
+            print(f"Pushing {name} to Docker Hub")
+            push_log = self._push_image(base_name, tag_name)
+            if args.verbose:
+                self._dump_push_log(push_log)
+            for line in push_log:
+                error = line.get('errorDetail')
+                if error:
+                    print(error['message'])
+                    sys.exit(1)
```
gctucker, line 180:
```python
return False
```
is the regular way to report errors with the `kci` command line tools.

kernelci/cli/docker.py
```diff
@@ -0,0 +1,203 @@
...
+    def __call__(self, configs, args):
+        print(self._get_dockerfile(args))
+        return True
+
+
+class cmd_name(DockerCommand):  # pylint: disable=invalid-name
+    """Get the full name of a Docker image without building it"""
+
+    def __call__(self, configs, args):
+        _, _, name = self._gen_image_name(args)
```
gctucker, line 196: Or maybe this instead:
```python3
name = self._gen_image_name(args)[2]
```

kernelci/cli/docker.py
```diff
@@ -0,0 +1,203 @@
...
+        dockerfile = self._get_dockerfile(args)
+        if args.verbose:
+            self._dump_dockerfile(dockerfile)
+
+        print(f"Building {name}")
+        buildargs = dict(
+            barg.split('=') for barg in args.build_arg
+        ) if args.build_arg else {}
+        _, build_log = self._build_image(dockerfile, name, buildargs,
+                                         args.no_cache)
```
gctucker, line 168: Maybe having less indentation and not breaking the line with the arguments would be easier to read?
```python3
        _, build_log = self._build_image(
            dockerfile, name, buildargs, args.no_cache
        )        
```

kernelci/cli/docker.py
```diff
@@ -0,0 +1,203 @@
...
+    opt_args = Command.opt_args + [
+        {
+            'name': '--arch',
+            'help': "CPU architecture, e.g. x86",
+        },
+        {
+            'name': '--fragment',
+            'action': 'append',
+            'help': "Extra fragments, e.g. kernelci",
+        },
```
gctucker: This should be renamed to `--fragments`, even if having `--fragments` several times on the command line looks a bit weird.  That's being reworked with the 2nd commit anyway and then it makes more sense to have `--fragments="foo bar"`

gctucker (commented):
Mostly coding style things, but there's a problem with `fragment` variables conflicting and renaming the command line argument to `--fragments` would solve that and also seem like a more appropriate name anyway.

gctucker:
Also, please rebase (should be trivial, now `kci config` has been merged)

kernelci/cli/docker.py
JenySadadia: That's true. Thanks for pointing this out.

kernelci/cli/docker.py
JenySadadia: Renamed the option.

kernelci/cli/docker.py
JenySadadia: Yes, this is another way of doing it.
However, the downside can be in case the function signature changes of `get_image_name` and if it returns a different number of parameters, then this syntax could lead to a problem.

kernelci/cli/docker.py
JenySadadia: Fixed it.

kernelci/cli/docker.py
gctucker: I think there would be the same problem with the `_, _, name =` syntax.

JenySadadia: However, we will receive at least some error with the `_` syntax.
Something like `ValueError: not enough values to unpack (expected 4, got 3)`

kernelci/cli/docker.py
JenySadadia: Removed it.

kernelci/cli/docker.py
gctucker: So how about changing this to a list?

```suggestion
    TEMPLATE_PATHS = ['config/docker', '/etc/kernelci/config/docker']
```

kernelci/cli/docker.py
gctucker: Well either way are both technically and stylishly correct so it's probably more a question of how this is done elsewhere in the code base for consistency.  I don't think it's a very common pattern so I guess we can just leave it as-is.

gctucker (commented):
About the command line arguments for fragments etc. I think we can compare the following examples:

```
# with --fragments as '*'
kci docker build gcc-10 --arch=x86 --fragments kunit kernelci
kci docker build --fragments kunit kernelci --arch x86 gcc-10

# however it means the image name can't be placed just after --fragments
kci docker build --arch x86 --fragments kunit kernelci gcc-10

# with --fragments as 'append' we could also have this:
kci docker build gcc-10 --fragments kunit --arch x86 --fragments kernelci

# with fragments as positional argument '*' the order would be fixed:
# e.g. kci docker build <OPTS> IMAGE FRAGMENT1 FRAGMENT2...
kci docker build --arch=x86 gcc-10 kunit kernelci
```
The last option is what I'm proposing as there's only one way to provide the image name and fragments so it removes any ambiguity or corner-cases.

kernelci/cli/docker.py
JenySadadia: Updated it.

JenySadadia:
> About the command line arguments for fragments etc. I think we can compare the following examples:
> 
> ```
> # with --fragments as '*'
> kci docker build gcc-10 --arch=x86 --fragments kunit kernelci
> kci docker build --fragments kunit kernelci --arch x86 gcc-10
> 
> # however it means the image name can't be placed just after --fragments
> kci docker build --arch x86 --fragments kunit kernelci gcc-10
> 
> # with --fragments as 'append' we could also have this:
> kci docker build gcc-10 --fragments kunit --arch x86 --fragments kernelci
> 
> # with fragments as positional argument '*' the order would be fixed:
> # e.g. kci docker build <OPTS> IMAGE FRAGMENT1 FRAGMENT2...
> kci docker build --arch=x86 gcc-10 kunit kernelci
> ```
> 
> The last option is what I'm proposing as there's only one way to provide the image name and fragments so it removes any ambiguity or corner-cases.

I understand your point.

I dig a little dipper and found that the issue you are mentioning has already been reported here  https://bugs.python.org/issue9338. It mentions a fix that we can separate optional and positional arguments by `--`.

```
$ ./kci docker name --arch=x86 --prefix kernelci/ --fragments kernelci -- gcc-10
kernelci/gcc-10:x86-kernelci

$ ./kci docker name --arch=x86 --prefix kernelci/ --fragments kernelci kselftest -- gcc-10
kernelci/gcc-10:x86-kernelci-kselftest
```
If we decide to go with this, we will need to update the usage information as well.
Otherwise we can proceed with your proposed solution.

mgalka:
Provided that single call of `kci docker` is going to  be used to build one docker image for one architecture I'd go with the solution provided by @gctucker. This will simplify the code and give a clear syntax which will be some sort of a  "recipe" for the image which will be a base imgae and subsequents on top.

JenySadadia:
> Provided that single call of `kci docker` is going to be used to build one docker image for one architecture I'd go with the solution provided by @gctucker. This will simplify the code and give a clear syntax which will be some sort of a "recipe" for the image which will be a base imgae and subsequents on top.

Thanks for the feedback.
Converted it to the optional position argument.

gctucker (approved):
Tested OK on staging with https://github.com/kernelci/kernelci-deploy/pull/69

PR #1810, opened 2023-03-17
kernelci.cli.show: add command to show results

gctucker:
Add "kci show results" to show results from a node and recursively all its child nodes hierarchy in a human-readable way.

### DIFF
--- a/kernelci/cli/__init__.py
+++ b/kernelci/cli/__init__.py
@@ -14,6 +14,7 @@ from . import (
     job,
     node,
     pubsub,
+    show,
     user,
 )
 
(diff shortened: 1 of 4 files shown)


gctucker:
Tested locally against the staging API:
kci-show-001

gctucker:
It's not very efficient as it's doing one query for each node, it could be doing a search for all node with a same kernel revision and result group name but this is just a first implementation.  Still it seems to be slower than expected with the staging API so there might be some sysadmin work to tweak network configuration, or we'll see how that goes with a new deployment in Kubernetes or on a new dedicated VM anyway.

PR #1813, opened 2023-03-20
Print job id in `kci job submit`

gctucker:
Add `Runtime.get_job_id()` method and use it in `kci job submit` to print the job id.


kernelci/runtime/lava/__init__.py
```diff
@@ -49,6 +49,9 @@ def submit(self, job_path):
     def job_file_name(self, params):
         return '.'.join([params['name'], 'yaml'])
 
+    def get_job_id(self, job_object):
+        return job_object
```
mgalka, line 53: Is the `job_object` just an id in this case?

gctucker: Yes

PR #1815, opened 2023-03-20
tests: add API bindings unit tests

gctucker:
Add an initial basic unit test for the API bindings.  This simply creates API and APIHelper objects for each configuration in the sample api-configs.yaml to check the objects can be initialised correctly.

### DIFF
--- a/tests/test_api.py
+++ b/tests/test_api.py
@@ -1,0 +1,22 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+#
+# Copyright (C) 2023 Collabora Limited
+# Author: Guillaume Tucker <guillaume.tucker@collabora.com>
+
+"""Unit tests for KernelCI API bindings"""
+
+import kernelci.api
+import kernelci.api.helper
+import kernelci.config
+
+
+def test_api_init():
+    """Test that all the API configurations can be initialised (offline)"""
+    config = kernelci.config.load('tests/configs/api-configs.yaml')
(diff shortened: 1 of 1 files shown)


tests/test_api.py
```diff
@@ -0,0 +1,21 @@
...
+
+def test_api_init():
+    """Test that all the API configurations can be initialised (offline)"""
+    config = kernelci.config.load('tests/configs/api-configs.yaml')
+    api_configs = config['api_configs']
+    for api_name, api_config in api_configs.items():
+        print(f"API config name: {api_name}")
+        api = kernelci.api.get_api(api_config)
+        assert api is not None
+        kernelci.api.helper.APIHelper(api)
```
mgalka, line 21: Shouldn't we have an assertion for `APIHelper` instance being created?

gctucker: If it fails then an exception is raised which would cause the test to fail.  We can assert that the object is not `None` I guess...

gctucker: After discussing this, I'll add an assert that the created object is of the type `APIHelper`.  We'll still have an exception if the contructor fails, but then we'll also have an explicit test to check the object is of the expected base type.

gctucker: Updated with asserts using `isinstance()`.

gctucker:
@mgalka Could you please take another look?

mgalka (approved):

PR #1816, opened 2023-03-20
cr50_updater.sh: Add script to update CR50 firmware on Chromebooks

nuclearcat:
During rootfs update we need also update CR50 firmware. ChromeOS designed to update it during update process, but it does require additional operations to reboot CR50 after update, which are not acceptable for labs.
As it require additional steps, testing labs design requre to reparate process of CR50 update, so we can do it from debos image separately.


config/rootfs/debos/overlays/cros-flash/opt/chromeos/cr50_updater
```diff
@@ -0,0 +1,127 @@
...
+if [ -z "$1" ]; then
+    echo "Error: no URL for firmware"
+    echo "Usage: $0 <URL>"
+    echo "Where <URL> is base URL for both prod and prepvt, e.g. https://storage..../dir"
+    echo "Both <URL>/cr50.bin.prod and cr50.bin.prepvt must be available"
+    exit 1
+fi
+
+expr "x$1" : 'x.*[^/]$' >/dev/null || {
+    echo "Error: URL must not end with slash"
```
mgalka, line 15: Maybe it's worth putting error messages to sterr instead of stdout? It applies to all error messages in the script.

config/rootfs/debos/overlays/cros-flash/opt/chromeos/cr50_updater
```diff
@@ -0,0 +1,127 @@
+#!/bin/sh
+# SPDX-License-Identifier: LGPL-2.1-or-later
+# Copyright (C) 2023 Collabora Ltd.
+# Author: Denys Fedoryshchenko <denys.f@collabora.com>
+
+if [ -z "$1" ]; then
+    echo "Error: no URL for firmware"
+    echo "Usage: $0 <URL>"
+    echo "Where <URL> is base URL for both prod and prepvt, e.g. https://storage..../dir"
```
mgalka, line 9: I don't think "where" is something to be used here.

```suggestion
    echo "The <URL> is base URL for both prod and prepvt images, e.g. https://storage..../dir"
```

nuclearcat:
Tested on octopus and zork on staging LAVA, but unfortunately logs are gone.

mgalka (approved):

PR #1823, opened 2023-03-20
build-configs.yaml: Enable AMD audio co-processor

nfraprado:
Enable configs for the AMD audio co-processor on the x86-chromebook fragment. This allows the audio card to probe on grunt, so alsa-kselftest can actually run and produce useful results.

That said, pcm-test results aren't stable: some of the tests are only passing sometimes. Moreover, the current kselftest timeout for alsa is too short which cuts the run short.


nfraprado:
I tested this change on other grunt chromebooks and the result was the same on all of them. So, I think it would make sense to move this test to `acer-R721T-grunt`, of which there are more available in the lab and currently only runs baseline tests, while `hp-11A-G6-EE-grunt` is already completely overloaded. Should I add a commit for that in this PR or in a separate one?

gctucker:
> hp-11A-G6-EE-grunt is already completely overloaded

Sounds like another PR would be needed to move more test plans over to `acer-R721T-grunt` then.

nuclearcat:
Works:
https://storage.staging.kernelci.org/kernelci/staging-mainline/staging-mainline-20230323.1/x86_64/x86_64_defconfig+x86-chromebook/clang-11/lab-collabora/baseline-acer-R721T-grunt.html
```
<6>[   11.411860] cz-da7219-max98357a AMD7219:00: ASoC: CPU DAI designware-i2s.1.auto not registered

<6>[   11.584482]  amd-da7219-play: codec dai name = da7219-hifi

<6>[   11.594001] input: acpd7219m98357 Headset Jack as /devices/platform/AMD7219:00/sound/card0/input5
```
Options present in kernel config: https://storage.staging.kernelci.org/kernelci/staging-mainline/staging-mainline-20230323.1/x86_64/x86_64_defconfig+x86-chromebook/clang-11/config/kernel.config

nuclearcat (approved):
LGTM

PR #1824, opened 2023-03-21
config/runtime: set imagePullPolicy to Always

gctucker:
Set the imagePullPolicy to Always in kubernetes.jinja2 to always pull the latest tagged image.  This still relies on local caching, so if an image with the same digest has already been pulled then nothing gets done.

It's important as KernelCI Docker images are currently "abusing" tags since they get updated regularly.  This will become less of an issue when image tags have a version or date and become 1-to-1 mappings with digests.  It will still be useful for development to avoid having to manually create many tags when iterating.


nuclearcat (approved):
LGTM

PR #1825, opened 2023-03-21
config/runtime: set restartPolicy to Never

gctucker:
As these jobs are "batch" jobs, disable automatic restarts and add options to automatically delete the jobs when they're done after a short period of time.  This is still useful to manually inspect logs when some jobs fail.


nuclearcat (approved):
LGTM, but need to be battle tested :)

PR #1829, opened 2023-03-21
config/lava/kselftest: Don't use a skipfile

broonie:
Currently we are using the LKFT skipfile for kselftest however this is
not ideal since it contains a lot of bitrotten entries for tests that
now run correctly (eg, the recently removed skip for the futex tests)
and masks any changes in state for the listed tests.  Since the
test-defintions integration will try to insert defaults if we don't
supply anything at all use /dev/null as the default test list, providing
an empty skiplist.

This also means we don't need to worry about getting skipgen installed
on the targets and running it which speeds things up ever so slightly.

If we need to skip tests we should probably maintain our own lists.

### DIFF
--- a/config/lava/kselftest/kselftest.jinja2
+++ b/config/lava/kselftest/kselftest.jinja2
@@ -21,7 +21,7 @@
       name: {{ plan }}
       parameters:
         TESTPROG_URL: {{ kselftests_url }}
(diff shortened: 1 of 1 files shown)


nuclearcat:
Several jobs with this PR works: https://lava.collabora.dev/scheduler/job/9894241 , https://lava.collabora.dev/scheduler/job/9894237

PR #1844, opened 2023-03-27
test-configs: add mt8365-evk board

khilman:
Minimal support for mt8365-evk board has recently landed upstream.  Add to KernelCI.


khilman:
Example working LAVA job in lab-baylibre: https://lava.baylibre.com:51443/scheduler/job/3474731

PR #1845, opened 2023-03-28
generate_cros_manifest.sh: Add tool to generate ChromiumOS xml manifest

nuclearcat:
To create new version manifest we need to fetch it a bit unusual way using ChromiumOS SDK.
We had it as comments in build_board.sh, but this create more complexity, so it is better to make proper script to automate this process.

### DIFF
--- a/config/rootfs/chromiumos/scripts/generate_cros_manifest.sh
+++ b/config/rootfs/chromiumos/scripts/generate_cros_manifest.sh
@@ -1,0 +1,58 @@
+#!/bin/bash
+set -e
+BRANCH=$1
+DATA_DIR=$(pwd)
+USERNAME=$(/usr/bin/id -run)
+
+function cleanup()
+{
+  rc=$?
+  # During development we might need to keep SDK for debugging
+  # then please "touch .keep" in --output directory
+  if [ -f "$DATA_DIR/../.keep" ]; then
+    echo "Keeping SDK as required by .keep file flag"
+    exit $rc
+  fi
+  echo Cleanup on exit
+  # Delete old SDK directory to not waste space
+  [ -d "${DATA_DIR}/chromiumos-sdk" ] && sudo rm -rf ${DATA_DIR}/chromiumos-sdk && echo Old SDK deleted
+  exit $rc
+}
+
+trap cleanup EXIT
+
+# Check if branch is set
+# You can check list of existing at: https://chromium.googlesource.com/chromiumos/manifest/+refs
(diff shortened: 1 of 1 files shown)


nuclearcat:
Way how this script tested:
```
git clone https://github.com/kernelci/kernelci-core/

docker run \                                                                                                                                                                                                       
    --name kernelci-build-cros \                                                                                                                                                                                   
    -itd \                                                                                                                                                                                                         
    -v $(pwd):/kernelci-core \                                                                                                                                                                                     
    --device /dev/kvm \                                                                                                                                                                                            
    -v /dev:/dev \                                                                                                                                                                                                 
    -v $(pwd)/config:/etc/kernelci \                                                                                                                                                                               
    --privileged \                                                                                                                                                                                                 
    --user cros \                                                                                                                                                                                                  
    kernelci/cros-sdk:kernelci                                                                                                                                                                                     
                                                                                                                                                                                                                   
# Execute shell
docker exec \                                                                                                                                                                                                      
    -it kernelci-build-cros \                                                                                                                                                                                      
    /bin/bash            

# then manually run the script inside docker shell
/etc/kernelci/rootfs/chromiumos/scripts/generate_cros_manifest.sh

Preparing depot tools
... lot of text will be generated...
repo sync has finished successfully.
Generating manifest
Saved manifest to cros-snapshot.xml
Copying manifest
```
After that file cros-snapshot.xml can be found in kernelci-core directory

JenySadadia:
Why can't we un-comment lines from `build_board.sh`  to generate manifest?
What are the complexities will it introduce?

nuclearcat:
> Why can't we un-comment lines from `build_board.sh` to generate manifest? What are the complexities will it introduce?

It was just a hack to have them there as commented, i prefer to introduce separate script and remove commented code from build_board.sh, to not mix each tool purpose.
Especially as build_board.sh becoming too long/complicated and might need better structuring to understand what it does.

mgalka (approved):
A follow-up should be a clean-up and rework of the build_board.sh script

JenySadadia:
> > Why can't we un-comment lines from `build_board.sh` to generate manifest? What are the complexities will it introduce?
> 
> It was just a hack to have them there as commented, i prefer to introduce separate script and remove commented code from build_board.sh, to not mix each tool purpose. Especially as build_board.sh becoming too long/complicated and might need better structuring to understand what it does.

OK. Maybe we should think of migrating common functions to a script and using them in `build_board` and `generate_cros_manifest` files to avoid re-writing the same code.

PR #1850, opened 2023-03-30
rustc-1.6*-x86.jinja2: Add arch specific templates

nuclearcat:
As we have now docker image template with arch, such as: compiler:arch-fragments...
We need to add also template for rustc arch, x86 at moment.

### DIFF
--- a/config/docker/rustc-1.66-x86.jinja2
+++ b/config/docker/rustc-1.66-x86.jinja2
@@ -1,0 +1,9 @@
+{%- set sub_arch = 'amd64' %}
+{% extends 'rustc-1.66.jinja2' %}
+
+{% block packages %}
+{{ super() }}
+
+RUN dpkg --add-architecture amd64
+
+{%- endblock %}
--- a/config/docker/rustc-1.62-x86.jinja2
+++ b/config/docker/rustc-1.62-x86.jinja2
@@ -1,0 +1,9 @@
+{%- set sub_arch = 'amd64' %}
+{% extends 'rustc-1.62.jinja2' %}
+
+{% block packages %}
+{{ super() }}
+
+RUN dpkg --add-architecture amd64
+
+{%- endblock %}


gctucker:
A better way to do this would be to rework the templates so we don't need this kind of hack.  But I guess that would require a bit more typing.

gctucker:
OK, could you please give an example of an image name currently being pull but does not exist in Docker Hub?  This would help understand why we need these empty templates to have new image names and resolve the problem.

nuclearcat:
For example: kernelci/staging-rustc-1.62:x86-kselftest-kernelci

```
  Normal  BackOff  24s (x1461 over 5h35m)  kubelet  Back-off pulling image "kernelci/staging-rustc-1.62:x86-kselftest-kernelci"
```

nuclearcat:
It turned out it is not really no-op templates, but kselftest fragment really need "RUN dpkg --add-architecture amd64"

nuclearcat:
Tested manually by deploy on staging:
```
     for rustc in rustc-1.62 rustc-1.66; do
-    ./kci docker $args $rustc kselftest kernelci
+    ./kci docker $args $rustc kselftest kernelci --arch x86
     done
```

gctucker:
> kernelci/staging-rustc-1.62:x86-kselftest-kernelci

OK so right now we only have:
```
kernelci/staging-rustc-1.62:kselftest-kernelci
```
but with this PR and some changes in `kernelci-deploy` we'll start having:
```
kernelci/staging-rustc-1.62:x86-kselftest-kernelci
```
and there's a requirement somewhere that the images need to have the CPU arch name in them.

Is that right?

gctucker:
> Tested manually by deploy on staging:
> 
> ```
>      for rustc in rustc-1.62 rustc-1.66; do
> -    ./kci docker $args $rustc kselftest kernelci
> +    ./kci docker $args $rustc kselftest kernelci --arch x86
>      done
> ```

Guess we need a PR for that too?  Or just push a fix to `main`.

Note: Testing this shouldn't need to be run on staging, I believe anyone can run `kci docker` locally and achieve the same result without `--push` or with a different `--prefix`.

nuclearcat:
Yes, i will do PR on kernelci-deploy as soon as this PR considered as ok/reviewed.
As we are adding support for x86 arch, this PR alone wont break anything.
I tested on staging with dual purpose - as we had pending k8s jobs that need this image, it helped to create image in docker registry with all required tags, so jobs can complete.

config/docker/rustc-1.66-x86.jinja2
```diff
@@ -0,0 +1,9 @@
+{%- set sub_arch = 'amd64' %}
+{% extends 'rustc-1.62.jinja2' %}
```
gctucker, line 2:
```suggestion
{% extends 'rustc-1.66.jinja2' %}
```
I guess

gctucker:
LGTM, but please provide a bit more context about how it was tested and which issues it's fixing.

nuclearcat:
I run manually deploy script: https://github.com/kernelci/kernelci-core/pull/1850#issuecomment-1490058774
Otherwise there is jobs created that need images such as "kernelci/staging-rustc-1.62:**x86**-kselftest-kernelci"

PR #1851, opened 2023-03-30
T41052 config/docker: remove `kci_docker` script

JenySadadia:
The `kci_script` has been converted to `kci docker` command. Since the script is not in use anymore,
remove it.


gctucker (commented):
As discussed before, it would seem worth updating `README.md` as it's still mentioning `kci_docker`.  Maybe updating it with `kci docker` and the fact that there's no default for `--prefix` any more would be fine for now, then this could be turned into a dedicated documentation page under the `kci` tool section when we have one.

JenySadadia:
> As discussed before, it would seem worth updating `README.md` as it's still mentioning `kci_docker`. Maybe updating it with `kci docker` and the fact that there's no default for `--prefix` any more would be fine for now, then this could be turned into a dedicated documentation page under the `kci` tool section when we have one.

I am sending a separate PR for that one.

JenySadadia:
> > As discussed before, it would seem worth updating `README.md` as it's still mentioning `kci_docker`. Maybe updating it with `kci docker` and the fact that there's no default for `--prefix` any more would be fine for now, then this could be turned into a dedicated documentation page under the `kci` tool section when we have one.
> 
> I am sending a separate PR for that one.

Please see https://github.com/kernelci/kernelci-core/pull/1853

mgalka:
I'd consider adding deprecation warning at first before actually removing the script as there may be some users having their flow based on the kci_* tools. It'd be good to have a deprecation warning maybe with a date when the tool will be removed. The warning could also have a link to the documentation of `kci docker`

gctucker:
> I'd consider adding deprecation warning at first before actually removing the script as there may be some users having their flow based on the kci_* tools. It'd be good to have a deprecation warning maybe with a date when the tool will be removed. The warning could also have a link to the documentation of `kci docker`

That's generally a good idea and we'll definitely need to do that for `kci_build` etc, but this tool has only been there for a few months and `kci docker` provides the same functionality so I don't think it's worth the complications related to having a deprecation period.

mgalka:
I think we can merge it, but I think we need to establish some clear rules on how the compatibility breaking changes are going to be introduced to the project especially that there are a lot of changes coming in due to the development of API and pipeline.

mgalka (approved):

PR #1852, opened 2023-03-30
rootfs-configs.yaml: Fetch ALSA UCM files from git

nfraprado:
The releases for alsa-ucm-conf are sparse. Instead of relying on the alsa-ucm-conf debian package, fetch the UCM directly from the main branch in the git repository to avoid having to wait for releases. This will allow us to start testing audio on machines using alsa-kselftest as soon as the machine's UCM is merged upstream.

It's worth pointing out that this is fetching the latest UCM files from git while still using the latest tagged release for alsa-lib and alsa-utils. It is possible that a UCM could be updated to rely on functionality from the latest alsa-lib, causing a regression for the machines that use that UCM, which would only be fixed when a new alsa-lib version gets released and packaged. A way around this would be to also build alsa-lib and alsa-utils from source. But since I'm not sure this will even happen, I'm starting with the simplest approach, so we can add this complexity once it becomes necessary.

Also, for now it seems unnecessary, but if we start building everything from source, might make sense to test both using the latest tagged release and the latest commit, so we can have a more stable reference in case there's a regression in alsa-lib.

### DIFF
--- a/config/core/rootfs-configs.yaml
+++ b/config/core/rootfs-configs.yaml
@@ -7,7 +7,6 @@ rootfs_configs:
       - arm64
       - armhf
     extra_packages:
-      - alsa-ucm-conf
       - alsa-utils
       - bc
       - ca-certificates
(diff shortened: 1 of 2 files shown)


nuclearcat (approved):
Builds success!

PR #1855, opened 2023-03-31
T41021 build_board: disable mali-G57 workaround firmware

JenySadadia:
Disable `builtin_fw_mali_g57` flag as the firmware is not required when `panfrost` is enabled.


nuclearcat:
Which board affected by this change? To make sure it builds successfully and works

JenySadadia:
> Which board affected by this change? To make sure it builds successfully and works

All the boards will be affected.

nuclearcat (approved):
It is not just remove firmware, it is actually fixing opensource kernel build, so it's quite important patch.
Tested on R111, thanks!

PR #1857, opened 2023-03-31
Add initial `Job` config class and rework `Runtime` accordingly

gctucker:
This is a first step to move away from build and test configurations and start using a new common job definition.  Update `kci job` accordingly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,7 @@ pylint:
 		kernelci.api \
 		kernelci.cli \
 		kernelci.config.api \
+		kernelci.config.job \
 		kernelci.config.runtime \
 		kernelci.config.storage \
 		kernelci.storage \
--- a/kernelci/cli/config.py
+++ b/kernelci/cli/config.py
@@ -26,7 +26,7 @@ class cmd_validate(Command):  # pylint: disable=invalid-name
     def __call__(self, configs, args):
         entries = [
             'device_types',
-            'test_plans',
+            'jobs',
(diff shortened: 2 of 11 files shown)


mgalka:
The changes look rather cosmetic. I think that the unit tests may need some adjustments, too.

mgalka:
> > The changes look rather cosmetic. I think that the unit tests may need some adjustments, too.
> 
> Could you please expand? Did you forget to approve this PR or post some review comments?

I think that the  [runtime unit test](https://github.com/kernelci/kernelci-core/blob/51c365bfb30f297f7ae1e83c08b754a73b8b99d2/tests/test_runtime.py#L32) still uses 'plans' instead of 'jobs' or is it something different?

gctucker:
> > > The changes look rather cosmetic. I think that the unit tests may need some adjustments, too.
> > 
> > 
> > Could you please expand? Did you forget to approve this PR or post some review comments?
> 
> I think that the [runtime unit test](https://github.com/kernelci/kernelci-core/blob/51c365bfb30f297f7ae1e83c08b754a73b8b99d2/tests/test_runtime.py#L32) still uses 'plans' instead of 'jobs' or is it something different?

Yes that's separate, it's for runtime tests.  I guess the name is misleading.  It's not related to this PR.  I guess this will also be entirely rewritten anyway when we finally drop test plans and only have jobs.

mgalka (approved):

PR #1860, opened 2023-04-03
Clean up the `Runtime` base class

gctucker:
Remove unused code from the base `kernelci.runtime.Runtime` now that we have `shell`, `docker` and `kubernetes` runtimes implemented.

### DIFF
--- a/kernelci/runtime/kubernetes.py
+++ b/kernelci/runtime/kubernetes.py
@@ -22,6 +22,10 @@ class Kubernetes(Runtime):
 
     JOB_NAME_CHARACTERS = string.ascii_lowercase + string.digits
 
+    @classmethod
+    def _get_job_file_name(cls, params):
+        return '.'.join([params['k8s_job_name'], 'yaml'])
+
     def generate(self, params, job_config):
         template = self._get_template(job_config)
         job_name = '-'.join(['kci', params['node_id'], params['name'][:24]])
@@ -30,9 +34,6 @@ class Kubernetes(Runtime):
         k8s_job_name = '-'.join([safe_name[:(62 - len(rand_sx))], rand_sx])
         params['k8s_job_name'] = k8s_job_name
         return template.render(params)
-
-    def job_file_name(self, params):
-        return '.'.join([params['k8s_job_name'], 'yaml'])
 
     def submit(self, job_path):
         kubernetes.config.load_kube_config(context=self.config.context)
--- a/kci_test
+++ b/kci_test
@@ -209,8 +209,10 @@ class cmd_list_jobs(Command):
 
         runtime_config = configs['runtimes'][args.runtime_config]
         runtime = kernelci.runtime.get_runtime(
-            runtime_config, args.user, args.runtime_token, args.runtime_json
+            runtime_config, args.user, args.runtime_token
         )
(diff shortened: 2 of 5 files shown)


kernelci/runtime/kubernetes.py
```diff
@@ -22,18 +22,19 @@ class Kubernetes(Runtime):
 
     JOB_NAME_CHARACTERS = string.ascii_lowercase + string.digits
 
-    def generate(self, params, plan_config):
-        template = self._get_template(plan_config)
+    @classmethod
+    def _get_job_file_name(cls, params):
+        return '.'.join([params['k8s_job_name'], 'yaml'])
```
mgalka, line 27: isn't it the same as `params['k8s_job_name'] + '.yaml'`?

kernelci/runtime/kubernetes.py
```diff
@@ -22,18 +22,19 @@ class Kubernetes(Runtime):
...
+
+    def generate(self, params, job_config):
+        template = self._get_template(job_config)
         job_name = '-'.join(['kci', params['node_id'], params['name'][:24]])
         safe_name = re.sub(r'[\:/_+=]', '-', job_name).lower()
         rand_sx = ''.join(random.sample(self.JOB_NAME_CHARACTERS, 8))
         k8s_job_name = '-'.join([safe_name[:(62 - len(rand_sx))], rand_sx])
```
mgalka, line 34: I'd consider extracting it to a separate function.

kernelci/runtime/kubernetes.py
gctucker: I've always thought that join was better and faster than + for joining strings so I'm always using this approach. I think there's an article somewhere comparing these things but I read it several years ago.

kernelci/runtime/kubernetes.py
gctucker: That's unrelated to this PR. Feel free to send a change to rework this if you want.

kernelci/runtime/kubernetes.py
gctucker: Also it's not a change related to this PR either, it's only reworking the method as a private class one.

mgalka: > I've always thought that join was better and faster than + for joining strings so I'm always using this approach. I think there's an article somewhere comparing these things but I read it several years ago.

Yes, in general it is, I think still the fastest method of string concatenation and + is the slowest. I put '+' expression there as an example of what I have in mind. Maybe not clearly enough. `.join` is the fastest, slower, but still fast is `f-string`, the slowest is `+`. However, in this case we're joining 2 strings and the speed is not the priority, but "readability counts` principle is still in use. This is just for the sake of clarification.

If it's not part of your PR, feel free to ignore it. I'll possible convert it to an f-string at some other occasion.

kernelci/runtime/kubernetes.py
mgalka: ack

kernelci/runtime/kubernetes.py
gctucker: Right, well once we have the whole pipeline actually functional then we can review the whole code and tweak tiny details like this I guess.

PR #1862, opened 2023-04-03
cros-snapshot-release-R111-15329.B.xml: Add R111 manifest

nuclearcat:
Add manifest for new ChromiumOS R111 build with our modifications and custom trees.


nuclearcat:
Firmware built by hand, flashed to devices ,verified by tast tests. Example, asurada, arm64, tast tests: https://lava.collabora.dev/scheduler/job/9906424

nuclearcat:
Will rebase now and edit commits

nuclearcat:
Done, please let me know if i missed anything

gctucker:
> I'll defer to @mgalka about whether this PR should target the main branch or chromeos (I'm ok either way).

It seems like it would be good to have some documented guidelines about what is eligible for the `main` branch so we don't end up flooding it with code that only applies to special instances.

mgalka:
> It seems like it would be good to have some documented guidelines about what is eligible for the `main` branch so we don't end up flooding it with code that only applies to special instances.

I'd be good to establish those I guess.  Fwiw, the manifest files are part of the Chromium OS rootfs images used by the `build_board.sh` script [quite explicitly](https://github.com/kernelci/kernelci-core/pull/1881#issuecomment-1504991942). So if we want to make them a part of the `chromeos` branch only we need to re-work the way the script works.

gctucker:
Yes I guess the Chromium OS rootfs config was initially all contained in one YAML file but now it's a lot more data with extra scripts and XML manifests.  It's convenient for the `chromeos` branch to have these things in `main` to reduce the rebase burden, but on the other hand it's not necessarily helpful to have it in `main`.  A real issue I think would be if the git history or files specific to chromeos.kernelci.org were making things significantly harder to maintain in the `main` branch.  Also it's setting a precedent for any future specialised instance, so the upstream `main` guidelines have to be scalable.

nuclearcat:
Tested over many images in jenkins: https://bot.staging.kernelci.org/job/chromeos/job/rootfs-builder/

PR #1865, opened 2023-04-05
Handle file collision better in case of very long test names

nuclearcat:
Right now some test names might exceed LAVA name limit and as result overwrite each other definition filenames.
We need to handle it properly:
1)Better to fail on generation, than silently overwrite
2)Try to shorten some keywords, when it is possible

### DIFF
--- a/kci_test
+++ b/kci_test
@@ -344,6 +344,9 @@ class cmd_generate(Command):
                 return False
             if args.output:
                 output_file = runtime.save_file(job, args.output, params)
+                if output_file is None:
+                    print("Failed to save the job definition, file collision")
+                    return False
                 print(output_file)
             else:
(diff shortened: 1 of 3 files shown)


nuclearcat:
Right now this patches fixing issue with ChromiumOS/tast tests on ChromeOS instance, and looks like no negative impact on staging instance.
For example: https://chromeos.kernelci.org/test/job/kernelci/branch/chromeos-stable/kernel/chromeos-stable-20230410.0/

mgalka (approved):
Let's have it merged as it adds code only to the `legacy` module. However, we should think of a solution that would fit in the current `kci` tool model

gctucker:
There's no support for LAVA jobs yet with the new Runtime, but yes this will need to be handled as part of the implementation.

PR #1868, opened 2023-04-06
CHROMEOS docker/data add tast_test_list script

alexandrasp:
Copy the script responsible for getting the list of tast tests available on the current build to the docker image used by cros.

After discussing with @mgalka we decided to make this script available on the public repository of the Kernel CI project because we had a problem sharing the file with the LAVA, once it was available in a Gitlab repository with limited access by the members. Other pros involve installing a dependency on the image and making it available; it will also avoid blocking a LAVA device when installing pip and requests library, which is a requirement for the script. 
I did not apply the changes on the `old-docker`. I assumed it is something deprecated on some level, but let me know if I am wrong.

### DIFF
--- a/config/docker/cros-tast.jinja2
+++ b/config/docker/cros-tast.jinja2
@@ -14,6 +14,7 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
     inetutils-ping \
     python3 \
     python-pkg-resources \
+    python3-pip \
     sudo \
     ssh \
     wget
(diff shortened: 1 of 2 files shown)


nuclearcat:
Can you rebase please?
Also change from chromeos.kernelci.org to chromeos

nuclearcat:
ping

alexandrasp:
> ping

done @nuclearcat

nuclearcat (approved):
LGTM

Issue #1854, opened 2023-03-31
Implement unit tests for `APIHelper`


gctucker:
The `APIHelper` class adds some middleware features on top of the API bindings to simplify application code.  There's currently a basic unit test in `tests/test_api.py` to instantiate `API` and `APIHelper` objects, now we need to add at least a unit test for each method of `APIHelper`.  This would involve a mockup `API` implementation as the unit tests should be self-contained and run without an actual API service.

Note: Writing unit tests for the `API` class itself is a bit more tricky since it's just meant to provide Python bindings to the API and there's no real logic in it.  This might still be done later as a follow-up, maybe using a lower-level mockup at the HTTP level.  It's outside the scope of this current issue.

PR #1869, opened 2023-04-07
T41723 Implement unit tests for `APIHelper`

JenySadadia:
Fixes https://github.com/kernelci/kernelci-core/issues/1854

Added `tests/api` directory for API specific tests.
Moved `tests/test_api.py` file to `tests/api` directory.
Added unit tests to the file. Also, added `conftest.py` file for pytest fixtures.


gctucker (commented):
I think you'll need at least an empty `tests/api/__init__.py` file to make it a package, right?

JenySadadia:
> I think you'll need at least an empty `tests/api/__init__.py` file to make it a package, right?

Yes

JenySadadia:
Added `APIHelperTestData` class to generate and get test data for fixtures and test definitions.

mgalka (approved):

mgalka:
@JenySadadia I approved this PR. Thank you for valuable input. Do you think it'd be worth to implement e.g. a test for `APIHelper._prepare_results()` as a follow-up?

JenySadadia:
> @JenySadadia I approved this PR. Thank you for valuable input. Do you think it'd be worth to implement e.g. a test for `APIHelper._prepare_results()` as a follow-up?

Thank you @mgalka for reviewing this.
Here `APIHelper._prepare_results()` is being used by `APIHelper.submit_results()` method. As we don't have any public wrapper for the method `_prepare_results` and it is an internal logic, I'd say we should include it in e2e tests rather than unit tests.

JenySadadia:
Bdw I have added a small fixup as a follow-up here https://github.com/kernelci/kernelci-core/pull/1888

gctucker:
I think unit tests should cover individual methods, in this case I think the `submit_results()` method is very close to it but we could also have a test with just `_prepare_results()`.  I don't believe it's really necessary as if `_prepare_results()` is broken the test using `submit_results()` will fail and it should be obvious enough to debug it.  If you look at the unit tests for the YAML config, you'll see some are using the `._filters` private field with `pylint: disable=protected-access` at the top of the file explaining why it's OK to do this with unit tests.

End-to-end tests are about trying the system as a whole, so the internal methods being used aren't relevant to them.

JenySadadia:
Please see https://github.com/kernelci/kernelci-core/pull/1891

PR #1870, opened 2023-04-07
kernelci.cli: add data section in "kci show results"

gctucker:
If the root node has some data, also dump it in a "Data" section.

### DIFF
--- a/kernelci/cli/show.py
+++ b/kernelci/cli/show.py
@@ -56,6 +56,13 @@ class cmd_results(APICommand):  # pylint: disable=invalid-name
             print(f"  {self._color(name, 'blue')}")
             print(f"    {url}")
 
+    def _dump_data(self, data):
+        max_len = max(len(key) for key in data)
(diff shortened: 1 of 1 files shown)


gctucker:
kci-show-004

kernelci/cli/show.py
```diff
@@ -56,6 +56,13 @@ def _dump_artifacts(self, artifacts):
             print(f"  {self._color(name, 'blue')}")
             print(f"    {url}")
 
+    def _dump_data(self, data):
+        max_len = max(len(key) for key in data)
```
mgalka, line 60: If data is iterable this can be a better use of `max`

```suggestion
        max_len = max(data, key=len)
```

gctucker: This actually returns the longest key, so I would need to do `len(max(data, key=len))` which I think seems more cryptic than just a comprehension with `len(key)`.

mgalka (approved):

PR #1875, opened 2023-04-10
Merge Docker images related commits from `chromeos` branch to `main`

mgalka:
This portion of commits migrate functionality related to docker images creation and configuration from `chromeos` branch to `main` branch of kernelci-core

This should also remove the docker-old files.
Depends: #1876

### DIFF
--- a/config/docker-old/cros-baseline/Dockerfile
+++ b/config/docker-old/cros-baseline/Dockerfile
@@ -1,0 +1,41 @@
+FROM debian:bullseye-slim
+MAINTAINER "KernelCI TSC" <kernelci-tsc@groups.io>
+
+ARG DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update && apt-get install --no-install-recommends -y \
+    apt-transport-https \
+    ca-certificates 
+
+RUN apt-get update && apt-get install -y \
+    git \
+    inetutils-ping \
+    ssh
+
+RUN \
+  mkdir -p /home/cros && \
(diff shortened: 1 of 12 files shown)


config/docker-old/cros-baseline/Dockerfile
gctucker, line 1: I don't think anything in `docker-old` should be merged in `main`, this directory needs to be deleted.

gctucker (changes requested):

config/docker-old/cros-baseline/Dockerfile
mgalka: I included the commits for `docker-old` just the sake of keeping history and description of some scripts which were originally added in the `docker-old` directory e.g. `tast_parser.py` or `ssh_retry.sh`. 
I was planning to add here a commit that removes `docker-old` before merging this PR (see #1876).

If we don't care about the git I can add remove `docker-old` related commits and add separate PRs to remove `docker-old` from `main` and `chromeos` branches. Do you think in such cases it's worth to modify commit messages so the information about `ssh_retry.sh` , `tast_parser.py` and possibly some others is not lost?

gctucker: Yes I don't think history is useful here, it would just make it harder to use with many more commits. If you know of a good reason to keep the history on the `main` branch then fine.

gctucker:
Replaced with #1964 which also removes the `docker-old` directory.

PR #1876, opened 2023-04-10
Remove docker-old

mgalka:
Files in config/docker-old are leftovers of the old mechanism for docker images creation.
These were ported to new tools and moved to config/docker. These are not used and not needed anymore, hence remove the files


nuclearcat:
Please rebase, otherwise LGTM

PR #1877, opened 2023-04-10
Enable decoder conformance tests on mt8192-asurada-spherion-r0

laura-nao:
Enable V4L2 decoder conformance tests on mt8192-asurada-spherion-r0 for H264, VP8 and VP9 formats.
Set a 90s timeout for each decoding.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -3102,6 +3102,15 @@ test_configs:
 # Disabling to reduce load
 #      - v4l2-compliance-uvc
 
+  - device_type: mt8192-asurada-spherion-r0
+    test_plans:
+      - v4l2-decoder-conformance-h264
+      - v4l2-decoder-conformance-vp8
+      - v4l2-decoder-conformance-vp9
+    filters: &chromebook-decoders-filter
+      - passlist: {defconfig: ['videodec']}
+      - regex: {'tree': '(next|mainline|media)'}
+
   - device_type: mt8365-evk
     test_plans:
       - baseline
@@ -3360,9 +3369,7 @@ test_configs:
       - v4l2-decoder-conformance-h264
       - v4l2-decoder-conformance-vp8
       - v4l2-decoder-conformance-vp9
-    filters: &chromebook-decoders-filter
-      - passlist: {defconfig: ['videodec']}
-      - regex: {'tree': '(next|mainline|media)'}
+    filters: *chromebook-decoders-filter
 
   - device_type: rk3399-puma-haikou
     test_plans:


laura-nao:
Waiting for the decoder support to land upstream.

gctucker:

gctucker (commented):
I think this one will just need to be rebased once everything else has been merged.  Is there any particular reason why it's a draft PR?  Are the tests supposed to be working?

gctucker:
Ah I see the `scp` firmware is still being added to the rootfs image here rather than the kernel build.

laura-nao:
I updated the PR according to the latest changes merged in the `main` branch. I checked that all dependencies were satisfied by running a manual VP8 test in LAVA on recent `mainline` and `next` builds:
* With `mainline-master-v6.5-rc2-arm64-defconfig+arm64-chromebook+videodec` (from staging): https://lava.collabora.dev/scheduler/job/11123696
* With `next-20230720-defconfig+arm64-chromebook`: https://lava.collabora.dev/scheduler/job/11123700

The `media` tree has not been rebuilt since [this PR](https://github.com/kernelci/kernelci-core/pull/1987) with the SCP FW was merged, so the decoder tests on this tree will fail until the next build.

The initial score for the fluster tests is:
* H264: 98/135
* VP8:  59/61
* VP9: 250/305

config/core/test-configs.yaml
```diff
@@ -3360,9 +3369,7 @@ test_configs:
       - v4l2-decoder-conformance-h264
       - v4l2-decoder-conformance-vp8
       - v4l2-decoder-conformance-vp9
-    filters: &chromebook-decoders-filter
-      - passlist: {defconfig: ['videodec']}
-      - regex: {'tree': '(next|mainline|media)'}
+    filters: *chromebook-decoders-filter
```
gctucker, line 3372: This is not related to asurada but to kevin, so maybe it should be split out into a separate commit.

gctucker (commented):
OK sounds good, just one cosmetic comment.  I guess we can have this merged on Monday before the cut-off point for the production update and have everything in place then.

config/core/test-configs.yaml
laura-nao: I just moved the filter definition up since `asurada` comes before `kevin` alphabetically, there's no functional change for `kevin`. I'm happy to split the changes into two commits anyway if it makes things cleaner!

gctucker: Ah sorry I see that now, it's fine then.  OK I think this is ready to be merged :+1:

config/core/test-configs.yaml
```diff
@@ -3102,6 +3102,15 @@ test_configs:
...
 #      - v4l2-compliance-uvc
 
+  - device_type: mt8192-asurada-spherion-r0
+    test_plans:
+      - v4l2-decoder-conformance-h264
+      - v4l2-decoder-conformance-vp8
+      - v4l2-decoder-conformance-vp9
+    filters: &chromebook-decoders-filter
+      - passlist: {defconfig: ['videodec']}
+      - regex: {'tree': '(next|mainline|media)'}
```
gctucker, line 3112: As previously discussed, a potential improvement might be to narrow down the regex strings to avoid matching things like `mainlineXYZ`:
```suggestion
      - regex: {'tree': '^(next|mainline|media)$'}
```
There are however other filters with the same issue so maybe a follow-up PR to solve them all in one go would be good.  It's not currently causing any practical issues as there aren't any other tree names currently that would match those listed there (`net-next` doesn't match this regex).

gctucker (commented):
Got the tests scheduled on staging as well with the same mainline kernel build with `videodec` fragment I used last week:
https://lava.collabora.dev/scheduler/job/11127649

Will wait for them to complete and then I think this should be all ready to go.

gctucker:
@laura-nao I see kernel warnings by the way:
```
+ python3 /usr/bin/fluster_parser.py -ts VP9-TEST-VECTORS -d GStreamer-VP9-V4L2SL-Gst1.0 -t 90 -j 1
<4>[   31.196502] ------------[ cut here ]------------
<4>[   31.201384] WARNING: CPU: 2 PID: 396 at kernel/workqueue.c:3379 __flush_work.isra.0+0x2dc/0x308
```
https://lava.collabora.dev/scheduler/job/11127649

The job is running anyway, but this might be something to investigate.

config/core/test-configs.yaml
laura-nao: Yes, I plan on submitting a separate PR to address all regex strings in the file.

gctucker:
OK got some results: https://staging.kernelci.org/test/plan/id/64be3cb604f93d803db8750a/

laura-nao:
@gctucker thanks for the heads up on the warnings - this is already being discussed with the community:
https://lore.kernel.org/all/5d15905d-2a7b-4eb6-9f52-38303eaf5292@notapiano/

PR #1879, opened 2023-04-11
Add kind field to job node

gctucker:
Add a `kind` field to the pipeline job configuration and use it when creating a node for a new job.

### DIFF
--- a/tests/configs/jobs.yaml
+++ b/tests/configs/jobs.yaml
@@ -12,6 +12,7 @@ jobs:
     template: 'kbuild.jinja2'
     image: 'gcc-10:{arch}{fragments}'
     conditions: *checkout
+    kind: kbuild
     params:
       config: x86_64_defconfig
 
@@ -19,6 +20,7 @@ jobs:
     template: 'kunit.jinja2'
     image: 'gcc-10:x86-kunit-kernelci'
     conditions: *checkout
+    kind: node
     params: {}
 
   kunit-x86_64:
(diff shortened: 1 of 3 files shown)


tests/configs/jobs.yaml
```diff
@@ -12,13 +12,15 @@ jobs:
...
     conditions: *checkout
+    kind: kbuild
     params:
       config: x86_64_defconfig
 
   kunit: &kunit-job
     template: 'kunit.jinja2'
     image: 'gcc-10:x86-kunit-kernelci'
     conditions: *checkout
+    kind: node
```
JenySadadia, line 23: Shouldn't the `kind` here be changed to `test` or `kunit`?

gctucker: Maybe at some point it should be something like `test` or something else, but probably not `kunit` as that's the test name.  For now there's only `node` as the generic kind and `kbuild` for kernel builds.  I'm working on adding LAVA tests now so we might have a new kind for that, depending on how things go with the `kbuild` thing and `Node.data` sub-schema.

JenySadadia: OK

gctucker:
Dropping that for now since we're keeping the job-specific data outside the scope of the API.  Semantics around `Node.kind` need to be better defined I think, to stay safe this should just be about the models used on the API side so either `node` or `regression` at the moment.  There's no plan to have pipeline jobs with another kind than `node` for the time being, we can revisit the situation later once we have the main jobs running and sending their data to the API e.g. `checkout`, `kunit`, `kbuild` and `baseline` in LAVA.

PR #1881, opened 2023-04-12
Migrate manifest changes from chromeos

mgalka:
This PR moves to the main branch the Chromium OS manifest changes that were added only in `chromeos` branch.

It'll make possible to merge #1862 to `main` instead of `chromeos`


nuclearcat (approved):
LGTM

gctucker:
Is it really a good idea to keep the manifest files on the `main` branch?

mgalka:
> Is it really a good idea to keep the manifest files on the `main` branch?

They are a part of Chromium OS rootfs build process. They are used by the `build_board.sh` script. 

https://github.com/kernelci/kernelci-core/blob/ea21f488a6b0b20e01f4c48d56c824727e2a23e2/config/rootfs/chromiumos/scripts/build_board.sh#L52

PR #1885, opened 2023-04-13
test-configs.yaml: Update for vm selftests rename to mm

broonie:
The selftests previously known as vm have been renamed upstream to mm.
Update our configurations to reflect this, leaving the old vm name in place
so that runs on older kernels from before the rename continue to happen and
using filters to avoid scheduling each on the kernels where they don't
exist.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -366,6 +366,15 @@ test_plans:
       kselftest_collections: "mincore"
     filters: *kselftest_no_fragment
 
+  kselftest-mm:
+    <<: *kselftest
+    params:
+      job_timeout: '10'
+      kselftest_collections: "mm"
+    filters:
+      - passlist: {defconfig: ['kselftest']}
+      - blocklist: {kernel: ['v3.', 'v4.', 'v5.', 'v6.0', 'v6,1', 'v6.2']}
+
   kselftest-net:
     <<: *kselftest
     params:
@@ -425,6 +434,10 @@ test_plans:
     params:
       job_timeout: '10'
       kselftest_collections: "vm"
+    filters:
+      - passlist:
+          defconfig: ['kselftest']
+          kernel: ['v3.', 'v4.', 'v5.', 'v6.0', 'v6,1', 'v6.2']
 
   lc-compliance:
     rootfs: debian_bullseye-libcamera_nfs
@@ -2319,6 +2332,7 @@ test_configs:
       - cros-ec
       - kselftest-alsa
       - kselftest-lib
+      - kselftest-mm
       - kselftest-vm
       - ltp-ipc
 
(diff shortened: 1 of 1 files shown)


nuclearcat:
At moment this PR cause staging to crash:
```
  File "/home/nuclearcat/Documents/kernelci-core/./kci_test", line 406, in <module>
  File "/home/nuclearcat/Documents/kernelci-core/./kci_test", line 315, in __call__
    test_configs = kernelci.test.match_configs(
  File "/home/nuclearcat/Documents/kernelci-core/kernelci/test.py", line 83, in match_configs
    if not plan.match(filters):
  File "/home/nuclearcat/Documents/kernelci-core/kernelci/config/test.py", line 489, in match
    return all(f.match(**config) for f in self._filters)
  File "/home/nuclearcat/Documents/kernelci-core/kernelci/config/test.py", line 489, in <genexpr>
    return all(f.match(**config) for f in self._filters)
  File "/home/nuclearcat/Documents/kernelci-core/kernelci/config/base.py", line 238, in match
    for k, wl in self._items.items():
AttributeError: 'list' object has no attribute 'items'
```
I'm investigating reasons

config/core/test-configs.yaml
```diff
@@ -425,6 +434,10 @@ test_plans:
     params:
       job_timeout: '10'
       kselftest_collections: "vm"
+    filters:
+      - passlist:
+        - defconfig: ['kselftest']
+        - kernel: ['v3.', 'v4.', 'v5', 'v6.0', 'v6,1', 'v6.2']
```
nuclearcat, line 441: Changes required:
```
-        - defconfig: ['kselftest']
-        - kernel: ['v3.', 'v4.', 'v5', 'v6.0', 'v6,1', 'v6.2']
+          defconfig: ['kselftest']
+          kernel: ['v3.', 'v4.', 'v5', 'v6.0', 'v6,1', 'v6.2']
```
So this fragment will look like:
```
  kselftest-vm:
    <<: *kselftest
    params:
      job_timeout: '10'
      kselftest_collections: "vm"
    filters:
      - passlist:
          defconfig: ['kselftest']
          kernel: ['v3.', 'v4.', 'v5', 'v6.0', 'v6,1', 'v6.2']
```

nuclearcat (changes requested):

config/core/test-configs.yaml
gctucker: Also it should be `v5.`, or we might match kernel v50 one day...

gctucker: And there's a typo in `v6.1` (`v6,1`)

gctucker: The benevolent dictator in charge of the Linux versioning scheme might decide to increase the major version at every release starting with v10 or something arbitrary, a bit like "hey we're now at v2.6.39 so let's jump to v3.0"...

nuclearcat (approved):
https://staging.kernelci.org/test/plan/id/6440c1d7180cccfbbb134ce4/
Passing on staging dashboard

PR #1886, opened 2023-04-14
CHROMEOS test-configs-chromeos.yaml: Update tast test list for R111

nuclearcat:
Update ChromeOS tast tests, as with version update some got removed, and some, similar, got added.


nuclearcat:
Validated on dashboard

PR #1888, opened 2023-04-18
T41723 tests/api: fix child node `path` field

JenySadadia:
Append child node name to `Node.path` field in `APIHelperTestData.update_kunit_child_node` just like `APIHelper._prepare_results` method that generates various fields for child nodes.

### DIFF
--- a/tests/api/conftest.py
+++ b/tests/api/conftest.py
@@ -210,7 +210,9 @@ class APIHelperTestData:
             "group": self._kunit_node["group"],
             "holdoff": None,
             "kind": self._kunit_node["kind"],
-            "path": self._kunit_node["path"],
+            "path": (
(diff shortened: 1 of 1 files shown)


gctucker:
It's usually a good idea to add a `Fixes:` trailer in fix patches, see the git history for examples.  That's not a hard requirement though, it just helps keeping track of fixes esp for forks and downstream branches.

JenySadadia:
> 

Oh yes, I missed it.

JenySadadia:
> It's usually a good idea to add a `Fixes:` trailer in fix patches, see the git history for examples. That's not a hard requirement though, it just helps keeping track of fixes esp for forks and downstream branches.

Fixed it.

mgalka (approved):

PR #1891, opened 2023-04-19
T41723 tests/api/test_api: implement `test_prepare_results`

JenySadadia:
Add test for `APIHelper._prepare_results` method that used to prepare hierarchy of test results.

### DIFF
--- a/tests/api/test_api.py
+++ b/tests/api/test_api.py
@@ -225,3 +225,41 @@ def test_receive_event_node(get_api_config, mock_receive_event,
             'timeout',
             'updated',
         }
+
+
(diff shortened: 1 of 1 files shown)


mgalka (approved):
LGTM

mgalka:
@JenySadadia It looks good in general. I think that there may be a typo in the commit message

`that used` should be  `that is used`, I guess.

JenySadadia:
> @JenySadadia It looks good in general. I think that there may be a typo in the commit message
> 
> `that used` should be `that is used`, I guess.

Thanks for pointing it out. Fixed it.

PR #1892, opened 2023-04-19
CHROMEOS cros-tast.jinja2: Don't fail if gsutil already latest

nuclearcat:
If gsutil already latest version it will return on exit(1), and will break Docker building recipe.
We add || true to not fail, if it is already latest version, or even if update failed, as it is not critical procedure.

### DIFF
--- a/config/docker/cros-tast.jinja2
+++ b/config/docker/cros-tast.jinja2
@@ -19,9 +19,6 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
     wget
 
 {% include 'fragment/cros-lava.jinja2' %}
-# Update gsutil to latest version, so it doesnt download it in LAVA
-# docker each time
-RUN gsutil update -n
 
 {{ super() }}
 


nuclearcat:
Fixes:
```
Building kernelci/cros-tast
  File "/data/kernelci-deploy-checkout/kernelci-core/./kci", line 29, in <module>
  File "/data/kernelci-deploy-checkout/kernelci-core/kernelci/cli/__init__.py", line 51, in call
    _COMMANDS[name](args)
  File "/data/kernelci-deploy-checkout/kernelci-core/kernelci/cli/docker.py", line 207, in main
    sub_main("docker", globals(), args)
  File "/data/kernelci-deploy-checkout/kernelci-core/kernelci/cli/base.py", line 702, in sub_main
  File "/data/kernelci-deploy-checkout/kernelci-core/kernelci/cli/docker.py", line 170, in __call__
    _, build_log = self._build_image(
  File "/data/kernelci-deploy-checkout/kernelci-core/kernelci/cli/docker.py", line 123, in _build_image
    return client.images.build(
  File "/usr/lib/python3/dist-packages/docker/models/images.py", line 287, in build
    raise BuildError(chunk['error'], result_stream)
docker.errors.BuildError: The command '/bin/sh -c gsutil update -n' returned a non-zero code: 1
```
Tested on kernelci-deploy scripts and manually

config/docker/cros-tast.jinja2
```diff
@@ -20,8 +20,8 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
 
 {% include 'fragment/cros-lava.jinja2' %}
 # Update gsutil to latest version, so it doesnt download it in LAVA
-# docker each time
-RUN gsutil update -n
+# docker each time. But don't fail if we are already latest version
+RUN gsutil update -n || true
```
mgalka, line 24: This will in general silent any error which can occur during gsutil update. Do you think it is ok or maybe we should use `gsutil update -f` instead?

nuclearcat: It looks like we might even need to remove it completely, i just tested it locally and it shows:
```
It looks like you are trying to run "/usr/bin/../lib/google-cloud-sdk/bin/bootstrapping/gsutil.py update".
The "update" command is no longer needed with the Cloud SDK.
To update, run: gcloud components update
Really run this command? (y/N) 
```

nuclearcat: It seems with all these changes there is no point to do that. I will drop this line

nuclearcat: Is it ok like this?

mgalka (approved):

PR #1893, opened 2023-04-19
tests-config: Add Aaeon UP squared 6000 with X6425RE

sjoerdsimons:
The UP squared 6000 boards comes in various different SKUs this device-type is specifically for the variant with an industrial X6425RE CPU which has manufactorer SKU UPN-EHLX4RE-A10-0864

This board is available in both the collabora staging and production labs. Starting with just baseline tests for now.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -692,6 +692,10 @@ test_plans:
       job_timeout: '20'
 
 device_types:
+  aaeon-UPN-EHLX4RE-A10-0864:
+    mach: x86
+    arch: x86_64
+    boot_method: grub
 
   acer-R721T-grunt:
     mach: x86
@@ -2181,6 +2185,10 @@ device_types:
 
 
 test_configs:
+  - device_type: aaeon-UPN-EHLX4RE-A10-0864
+    test_plans:
+      - baseline
+      - baseline-nfs
 
   - device_type: acer-R721T-grunt
     test_plans:


nuclearcat:
It seems board doesnt boot: https://storage.staging.kernelci.org/mainline/master/v6.3-rc7-89-gcb0856346a60/x86_64/x86_64_defconfig/gcc-10/lab-collabora/baseline-aaeon-UPN-EHLX4RE-A10-0864.html

sjoerdsimons:
> It seems board doesnt boot: https://storage.staging.kernelci.org/mainline/master/v6.3-rc7-89-gcb0856346a60/x86_64/x86_64_defconfig/gcc-10/lab-collabora/baseline-aaeon-UPN-EHLX4RE-A10-0864.html

That's surprising as it works in our health checks with a very similar job definition (older kernel ofcourse)  as well as the apertis system tests.. I'll see if i can figure out what in  that particular job is making it unhappy

sjoerdsimons:
Fwiw with a "chromebook" kernel in a test job we can see the device successfully booting, but just failing wrt. networking as CONFIG_IGB is not enabled:
  https://lava.collabora.dev/scheduler/job/10059478

gctucker:
> Ah right for serial on top of defconfg we need CONFIG_MFD_INTEL_LPSS_PCI CONFIG_SERIAL_8250_DW configuration; and for networking CONFIG_IGB on top of the mainline defconfig -- It looks like the first two are configure in the x86-chromebook fragments, but not not generally for x86 (even though a lot of those options are generic, not just for chromebooks).
> 
> Any suggestions on how to best handle that (e.g. would it make sense to add some of these options into all x86 builds?)

One way is to have another inline fragment e.g. `x86-serial`, but that would mean an extra kernel build so it should only be enabled on a few trees e.g. mainline, next and stable.  Another way to look at it would be to always build x86 with this new fragment, and also have the plain defconfig with trees that have the full set of builds.  That might be a bit more tricky to do in the YAML config, but it's feasible.  The simplest thing of course is to use the builds with `x86-chromebook` for this platform but that's a bit of a hack.  It could be tried as a quick experiment to confirm all the tests are running on staging I guess.

So I think the main question here is, do you need to have this platform tested with all the trees covered by KernelCI or is it OK to have only the main ones such as mainline, linux-next and stable?

gctucker:
> Fwiw with a "chromebook" kernel in a test job we can see the device successfully booting, but just failing wrt. networking as CONFIG_IGB is not enabled: https://lava.collabora.dev/scheduler/job/10059478

Ah sorry missed this bit in my previous comment.  So it's not just about the serial console, I guess that's a typical thing with `x86_64_defconfig` to only have basic things turned on unlike the arm64 one.  Assuming we can't just have this turned on in `x86_64_defconfig` upstream, we'll need to have a fragment for this platform or maybe a family of similar platforms.  Not sure about the name but `x86-serial` doesn't seem appropriate in this case.  Maybe `x86-aaeon`?  With `CONFIG_IGB=y` and the serial ones added there too.  Then I think that also means the extra kernel build with this fragment `x86_64_defconfig+x86-aaeon` should only be available with the main trees and maybe one maintainer tree relevant to this platform if there's such a thing.  How does that sound?

nuclearcat:
I suggest if we can rename x86-chromebook to x86-boards fragment, similar as it is done in distribution kernels, and build with it as much as possible variations (that will be tested on real boards),
It should work with all x86 hardware, and if not - in most cases it is considered as a bug. It will provide also more testing coverage for drivers in x86 kernels.

sjoerdsimons:
There is nothing specific to the aaeon here; All this config options are for relatively generic and common Intel parts; E.g. the serial is simply how serial works on these intel chips, which is why there present in the chromebook config. I bet if modern chromebooks still had ethernet the IGB would also already be enabled there :).

So in that that sense i like @nuclearcat suggestion to just make it an x86-boards fragment that has more complete coverage

gctucker:
There are issues about renaming `x86-chromebook` to something more general, for example if we need some changes for x86 Chromebooks that break some non-Chromebook platforms then the Chromebooks will take precedence.  The fragment name isn't necessarily about the configs inside it but it could also be about a certain range of platforms using it.  That's partly to avoid having lots of tiny config fragments and also to keep the number of builds down.  Another issue with having a big and generic fragment for "all things x86" is that it's making the kernel less like the plain defconfig and there's value in testing the plain defconfig, for example to cover patches that make changes to it.  Also it's less likely to break, basically the defconfig is a special case that we need to keep testing.  Builds with a `x86-boards` fragment might become the standard ones instead of the plain defconfig even when it could be used as-is.

So I think it's less invasive and has fewer consequences to add a fragment just for this board.  If we're all fully aware of the implications of reshaping the `x86-chromebook` fragment into a big catch-all one for random x86 platforms and we agree it's the best way forward then fine - but it doesn't seem as clear-cut to me.

nuclearcat:
I am wondering what kind of changes we might need in Chromebook config that might break non-Chromebook platforms?
Sure we can add then Chromebook fragment config back, but i think with proper upstreaming efforts, at least x86 Chromebooks should work with "default" configs, otherwise it is a bug.
We can just also add this device to be tested with chromebook fragment, without renaming anything and just add IGB driver in chromebook fragment. And if at some moment it will require separate kernel fragment, also we will do it.
The main goal now is to add device with minimal efforts, of course without harming anything else. And if, theoretically, problems may arise in the future, we will solve them as they arise.

gctucker:
> I am wondering what kind of changes we might need in Chromebook config that might break non-Chromebook platforms?

Ideally it shouldn't be a problem, but it's a possibility in practice.  Say if we enable some config that cause non-Chromebook platforms to fail to boot, of course that's a kernel bug but it might not be easily fixed.  I'm sure there are tons of kernel configs right now that would break some platforms if you turned them on.  Conversely, if we need a config for a non-Chromebook platform and that breaks a Chromebook, we'll have to disable it.  Or someone will have to fix these issues, that's the point of KernelCI after all - my only concern is if it takes 3 months to get a fix and we lose Chromebook coverage in the meantime then it's hard to justify the costs.  If the fragment is only for Chromebooks then we decouple these things.  Maybe a key to all this is to keep the `x86-chromebook` fragment only for the `chromeos.kernelci.org` instance and only have generic fragments on the main instance `linux.kernelci.org`.  We would just need to also enable some builds with linux-next and mainline on `chromeos.kernelci.org` with `x86-chromebook` to compensate.

nuclearcat:
If there is any issue in ANY driver, doesn't matter if it is Chromebook or not chromebook, we don't wait 3 month until it is fixed, i think we should just exclude this driver from fragment until it is fixed. Because it will block other devices coverage.
Also we need x86-chromebook fragment in main instance, to have proper coverage in testing GPU, EC, and other subsystems.

gctucker:
> If there is any issue in ANY driver, doesn't matter if it is Chromebook or not chromebook, we don't wait 3 month until it is fixed, i think we should just exclude this driver from fragment until it is fixed. Because it will block other devices coverage. Also we need x86-chromebook fragment in main instance, to have proper coverage in testing GPU, EC, and other subsystems.

Yes but my point is in the scenario where we need to exclude it to work around a problem on a non-Chromebook platform but this same config is needed and works on a Chromebook.

How about keeping Chromebook-specific fragments on the ChromeOS instance though?  I think that's always kind of been a grey area anyway, the fragments existed before the instance iirc.

sjoerdsimons:
The same argument goes for configs that breaks some chromebooks but not others right?

Lots of distribution also enable the actual chromebook specific options on their general kernels; So whatever way you cut it,  the chromebook configs options should really also be tested on non-chromebooks on the main kernelci in my opinion :)

gctucker:
Yes, basically what I'm suggesting is to fork the fragment and keep `x86-chromebook` for `chromeos.kernelci.org` builds and call it `x86-boards` for `linux.kernelci.org`. Then we can decide which options to put in them and if there's a conflict we can just have different options to address that naturally.

However I still think we should keep the plain `x86_64_defconfig` in the standard build configurations and only add an extra build with the `x86-boards` fragment to larger ones such as mainline, next and stable.

sjoerdsimons:
@gctucker can you give me some pointers on how to do that; I see two x86-chromebook setups; one in `config/core/build-configs.yaml` and one in `tests/configs/fragments.yaml`  which are not in sync.  It seems typically only `build-configs.yaml` gets updated? is that hte one to focus on ?

For your other point, just keeping plain defconfig, i really i'm not versed enough to understand the impacts but testing the plain defconfig is obviously a good thing :).. I do suspect that limiting x86-boards only to the bigger trees will mean that most others won't be tested on real modernish intel hardware though (but i guess that's not a regression as the same goes for the current chromebook config?)

gctucker:
To "fork" the fragment, I think that's basically the case of renaming `x86-chromebooks` as `x86-boards` in `build-configs.yaml` and making a copy as `x86-chromebooks` in `build-configs-chromeos.yaml` for the ChromeOS KernelCI instance.  Ideally the same thing should be done with `arm64-chromebooks` and `arm64-boards` but that can be done in a separate PR.  Probably it's best to first do that without any changes to the configs listed in them, and check there's no difference in the results apart from the rename, then have a PR like this one on top to start making changes in `x86-boards` and review that on its own.

I think doing this doesn't change the number of builds at all, at least not for the main KernelCI instance.  Maybe some filters will need to be changed in `test-configs.yaml` to get Chromebooks to run with builds that have either the `-chromebooks` or the `-boards` fragment but that's only relevant to the ChromeOS instance.

gctucker:

nuclearcat:
If it is still required, we can add this board to Maestro. Please ping me if you want me to do it.

PR #1894, opened 2023-04-19
Add `Runtime.wait()` method and use it with `kci job submit --wait`

gctucker:
Add a `Runtime.wait()` abstract method with implementations for `shell`, `docker` and `kubernetes` to wait for a given job to complete and get the exit status code.  Use it with `kci job submit --wait`.


gctucker:
Tested OK locally with the 3 runtimes.

PR #1895, opened 2023-04-19
kernelci.cli: pubsub: add --json to parse JSON input

gctucker:
Add a `kci pubsub send_event --json` argument to parse the event input as JSON from stdin.  Otherwise it's simply read as a plain string.


gctucker:
Tested OK locally by submitting CloudEvent data from a stored JSON file.  This is actually quite useful to debug things by simulating a real event sent normally by the API.

PR #1898, opened 2023-04-19
kernelci.config: add Job.runtimes attribute

gctucker:
Add Job.runtimes as a list of dictionaries containing parameters to choose which runtimes the pipeline job should be run in.

Update unit tests accordingly.

This is an initial step to remove the hard-coded logic in runner.py, it will need to be redesigned to take into full consideration all the parameters and platform attributes too.  Finally the YAML configuration should be validated against a schema (e.g. Pydantic models) to specify the required fields with their types etc. Currently the runtimes parameters here are arbitrary and not validated when loading the configuration since this is still being worked on.

### DIFF
--- a/kernelci/config/job.py
+++ b/kernelci/config/job.py
@@ -15,12 +15,13 @@ class Job(YAMLConfigObject):
 
     # pylint: disable=too-many-arguments
     def __init__(self, name, template, image=None,
-                 conditions=None, params=None):
+                 conditions=None, params=None, runtimes=None):
         self._name = name
         self._template = template
         self._image = image
         self._conditions = conditions or []
         self._params = params or {}
+        self._runtimes = runtimes or []
 
     @property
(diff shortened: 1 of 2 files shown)


gctucker:
I'm actually not 100% sure we'll merge this, I'm preparing an email and GitHub discussion about YAML config so let's see how this goes first.

gctucker:
See the discussion related to the new YAML pipeline configuration:
https://github.com/kernelci/kernelci-core/discussions/1907

gctucker:
So the plan is now to have a `scheduler` entry in the YAML config to tie jobs with runtimes and platforms.  We might merge this in the meantime to get things going but it's likely we'll have to rework it later to match the proposal.

gctucker:
OK I think it's better to directly go for the longer-term proposed solution using `scheduler`.

PR #1889, opened 2023-04-18
Run v4l2-compliance on mt8192-asurada-spherion's encoder

nfraprado:
This PR enables v4l2-compliance on the encoder video node of mt8192-asurada-spherion.

The firmwares are added for other MediaTek SoCs as well to make it easier to enable the same test on them.

Using the latest v4l2-compliance, there are two issues on mt8192-asurada-spherion's encoder:

```
[..]
Control ioctls:
[..]
		fail: v4l2-test-controls.cpp(1167): node->codec_mask & STATEFUL_ENCODER
	test VIDIOC_(UN)SUBSCRIBE_EVENT/DQEVENT: FAIL
[..]
Format ioctls:
[..]
		fail: v4l2-test-formats.cpp(924): sel.r.width != fmt.g_width()
	test VIDIOC_S_FMT: FAIL

[..]
Total for mtk-vcodec-enc device /dev/video1: 45, Succeeded: 43, Failed: 2, Warnings: 0
```

I'll be sending a patch to fix the second issue, and asking where the first should be fixed, shortly.

PR #1901, opened 2023-04-20
bullseye-v4l2.sh: Move from autoconf to meson for v4l-utils

nfraprado:
Starting with commit c252c887c4e3 ("Drop autoconf/automake support"), the v4l-utils project has dropped support for autoconf in favor of the meson build system. Update the script to use meson.


nuclearcat:
rootfs was tested at PR https://github.com/kernelci/kernelci-core/pull/1889

PR #1912, opened 2023-04-27
kernelci.config: fix get_config_paths() when single file

gctucker:
Fix get_config_paths() when called with a single file as a string. The returned value is always expected to be a list, so when returning a string the caller would then iterate over the characters in the string.

Fixes: 0f4d9da028a4 ("kernelci.config: add .get_config_path() with defaults")

### DIFF
--- a/kernelci/config/__init__.py
+++ b/kernelci/config/__init__.py
@@ -49,6 +49,8 @@ def get_config_paths(config_paths):
             if os.path.isdir(config_path):
                 config_paths.append(config_path)
                 break
+    elif isinstance(config_paths, str):
+        config_paths = [config_paths]
     return config_paths
 
 


kernelci/config/__init__.py
```diff
@@ -49,6 +49,8 @@ def get_config_paths(config_paths):
             if os.path.isdir(config_path):
                 config_paths.append(config_path)
                 break
+    elif isinstance(config_paths, str):
```
mgalka, line 52: If `configs_paths` was an empty string it'd still get handled by the first  `if` block (line 46) is it an expected behaviour?

gctucker: I guess it would break.  It would also break if it was a floating point number though.

PR #1918, opened 2023-05-04
Rework custom Jinja2 functions

gctucker:
Rework the code that registers Jinja2 custom functions directly into the Runtime class.  This is in preparation for handling LAVA job templates in particular.


gctucker:

PR #1919, opened 2023-05-04
Runtime params rework

gctucker:
Rework the template parameters to directly pass objects rather than YAML dumps or just the node ID.  This relies on the `yaml_dump()` custom Jinja2 function added by #1918.  Update the `python.jinja2` base template accordingly to use the objects and reduce the number of parameters.


gctucker:

PR #1920, opened 2023-05-05
kernelci.cli: job: add --runtime-token

gctucker:
Add the `kci job {generate,submit} --runtime-token` argument for runtimes that require a token to be passed on the command line (e.g. LAVA).  Some runtimes such as Kubernetes load their credentials from the environment instead.


gctucker:

PR #1921, opened 2023-05-05
kernelci.runtime: add ._user and ._token

gctucker:
Store the provided user and token as private class attributes so that derived classes can make use of it.


mgalka (approved):
LGTM

PR #1922, opened 2023-05-05
Initial new LAVA runtime

gctucker:
Add new initial implementation of a LAVA runtime.  This has a minimal set of features to generate and run jobs with LAVA REST API v0.2.

The legacy modules remain unchanged for backwards compatibility with `kci_test` and the Jenkins pipeline.


gctucker:
Tested OK by hand: https://lava.collabora.dev/scheduler/job/10204533

This requires also some changes to define the runtime and baseline job in the YAML config, but that's for `kernelci-pipeline`.

PR #1925, opened 2023-05-08
CHROMEOS test-configs-chromeos.yaml: Update ChromiumOS test devices

nuclearcat:
This commit series fix some entries, update R111 update and add new devices discovered during new firmware update automation tests.


nuclearcat:
Tested and validated on dashboard

PR #1927, opened 2023-05-09
CHROMEOS build_board.sh: Remove hack for corrupted git cache

nuclearcat:
As tested this hack not required anymore as Google fixed corrupted git cache issue.


nuclearcat:
Tested during rootfs builds manually

PR #1931, opened 2023-05-17
kernelci.build: apply a patch to test regression

JenySadadia:
Apply a patch for `linux/drivers/base/test/property-entry-test.c` to test regression tracker script from pipeline.

### DIFF
--- a/kernelci/build.py
+++ b/kernelci/build.py
@@ -165,7 +165,11 @@ cd {path}
 git reset --hard
 git clean -fd
 git checkout --detach {remote}/{branch}
-""".format(path=path, remote=config.tree.name, branch=config.branch))
+sed -i s,'{line1}','{line2}', {test_file}
+""".format(path=path, remote=config.tree.name, branch=config.branch,
+           line1='PROPERTY_ENTRY_U16("prop-u16"\\, 16)',
+           line2='PROPERTY_ENTRY_U16("prop-u16"\\, 17)',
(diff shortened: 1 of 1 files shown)


gctucker:
This could also be done via `kernelci-deploy` by adding the patch file directly there.

JenySadadia:
> This could also be done via `kernelci-deploy` by adding the patch file directly there.

`kernelci.build.update_repo` is being called from pipeline service. The patch can not be applied before starting the containers. Hence, I added it here.

gctucker:
The patches in `kernelci-deploy` are used by the staging script to create the `staging-{next,mainline,stable}` branches and these are monitored by the new pipeline too.

JenySadadia:
> The patches in `kernelci-deploy` are used by the staging script to create the `staging-{next,mainline,stable}` branches and these are monitored by the new pipeline too.

As per my understanding from the code, `pending.py` script is applying all the patches. The update repository will be done by pipeline at the runtime. That will overwrite all the changes in the `linux` source if found. So, we can not apply the patch beforehand.
Please correct me if I am wrong.

JenySadadia:
I had a discussion with @mgalka about this and I understood what you meant @gctucker.
So we can add a patch to be run from `kernel.py` directly.

JenySadadia:
Tested regressions by failing test results instead. Closing this one.

Issue #1913, opened 2023-04-27
Convert kernelci.conf to TOML


gctucker:
The current settings file e.g. `kernelci.conf` uses the old `configparser` module.  It would be good to transition to TOML with the `toml` module as that's a better defined format for handling settings and should be fairly backwards-compatible.

It would also fix some issues with argument types as the `kernelci.conf` format assumes all values are strings, so arguments that are meant to be integers don't work as expected.  With TOML, strings have to be in quotes like in JSON so it's clearer.

The only issue with this is for users who rely on a `kernelci.conf` settings file, I guess we would need to add support for TOML and keep the `configparser` implementation with a deprecation period.  This could probably be done easily just by checking the file name e.g. `.config` or `.toml`.

gctucker:
@JenySadadia Maybe we could use dotted keys in the section names rather than `:`?  That would require a bit more rework but it could have some advantages, relying more on the TOML standard than special implementation in the KernelCI code.

Documentation: https://toml.io/en/v1.0.0#keys

JenySadadia:
> @JenySadadia Maybe we could use dotted keys in the section names rather than `:`? That would require a bit more rework but it could have some advantages, relying more on the TOML standard than special implementation in the KernelCI code.
> 
> Documentation: https://toml.io/en/v1.0.0#keys

Dotted keys will be confusing when we have dotted values.
e.g. we have `db:staging.kernelci.org`.

The parser will assume it is a nested dictionary and parse it accordingly.
```
[db.staging.kernelci.org]
test = 2
> 'db': {'staging': {'kernelci': {'org': {'test': 2}}}}
```
In this case, we will still need to use the double quotes around the value:
```
[db."staging.kernelci.org"]
test = 2
> 'db': {'staging.kernelci.org': {'test': 2}}
```
I'd suggest keeping keys as is with `:` or else it will be more confusing for a user to decide when to use double quotes.

gctucker:
OK, I think we should be sending an email to the list about the added support for TOML and maybe mention this point to see what users think.  TOML is a widely used language and clearly documented so I think users might already be used to this kind of thing.  It's all about the tradeoff between user-friendliness and code sustainability, using built-in TOML features such as dotted keys is better from a design point-of-view.

JenySadadia:
[...]

PR #1936, opened 2023-05-22
Convert `.conf` to TOML

JenySadadia:
Fixes https://github.com/kernelci/kernelci-core/issues/1913

Convert user settings file format from `.conf` to `.toml`:
- Add sample settings file
- Add parsing of TOML file
- Return deprecation warning with `.conf` and keep backwards compatibility with the older format


JenySadadia:
Pipeline is receiving deprecation warning:
```
today at 15:47:34Warning: user settings file format '.conf' will soon be deprecated. Please use '.toml' file format and provide 'kernelci.toml' file instead.
today at 15:47:3405/23/2023 10:17:34 AM UTC [INFO] Delay: 3s
today at 15:47:3705/23/2023 10:17:37 AM UTC [INFO] Resubmitting existing revision   kernelci_staging-mainline        52b79c04229450b0bc28ac570b285b0bc04b9787
```

gctucker:
There also appears to be an issue with default values as the same revisions are being triggered again even though the `force` flag is not set in the config file:
```
Warning: user settings file format '.conf' will soon be deprecated. Please use '.toml' file format and provide 'kernelci.toml' file instead.
05/23/2023 02:18:06 PM UTC [INFO] Delay: 3s
05/23/2023 02:18:10 PM UTC [INFO] New revision                     kernelci_staging-mainline        de3f3b63180c6c1416722da140303c5715121586
05/23/2023 02:18:10 PM UTC [INFO] Resubmitting existing revision   kernelci_staging-next            83d3bf28cab8d6c9f4d563659cca1325e194a8f1
05/23/2023 02:18:11 PM UTC [INFO] Resubmitting existing revision   kernelci_staging-stable          b56850d464e68aa39403954185f74cd870937e79
05/23/2023 02:18:12 PM UTC [INFO] Resubmitting existing revision   mainline                         ae8373a5add4ea39f032563cf12a02946d1e3546
05/23/2023 02:18:12 PM UTC [INFO] Sleeping for 3600s
```
Adding skip label to try again and check it's related to the changes in this PR.

kernelci/cli/base.py
```diff
@@ -510,8 +511,12 @@ def __init__(self, path, command, cli_args, section=None):
         *section* is a section name to use in the settings file, to provide a
                   way to have default values for each CLI tool
         """
+        self._deprecated_settings = False
```
gctucker, line 514: A nicer way to do this I think would be to have a derived class with the `configparser` code and a factory to create objects either of the main Options class which would be using TOML or the .conf one.  That way you wouldn't have to test `if self._deprecated_settings` in multiple places but have overridden methods instead.  And when the time has come to drop support for `configparser`, we can just delete the derived class.

I think it's fine to have the current approach too, the first objective is to have something functional.  Then maybe this could be reworked a bit if the deprecated code is meant to be supported for a long time to make it easier to maintain.

gctucker:
> There also appears to be an issue with default values as the same revisions are being triggered again even though the `force` flag is not set in the config file:

Revisions are still being resubmitted so I don't know what's going on, I've double-checked and the trigger force flag is not in the config file.  Will investigate further tomorrow.

JenySadadia:
> > There also appears to be an issue with default values as the same revisions are being triggered again even though the `force` flag is not set in the config file:
> 
> Revisions are still being resubmitted so I don't know what's going on, I've double-checked and the trigger force flag is not in the config file. Will investigate further tomorrow.

It is not related to this PR. I added the flag manually to `trigger` service on the staging instance for the regression tracker. Just removed it.

gctucker:
OK, to force a new set of staging jobs you can just create a new kernel revision with `staging.kernelci.org kernel mainline` then the trigger will pick it up. Another trick is to restart just the trigger service with docker-compose to avoid waiting for the next timer tick.

kernelci/cli/base.py
JenySadadia: I don't think we should put more effort to rework it atm as `configparser` would be removed eventually.
I agree that let's have this functional first and we can rework it later if required.

JenySadadia:
> OK, to force a new set of staging jobs you can just create a new kernel revision with `staging.kernelci.org kernel mainline` then the trigger will pick it up. Another trick is to restart just the trigger service with docker-compose to avoid waiting for the next timer tick.

Thank you for the tip :)
I don't think restarting the trigger service works in this case.

gctucker:
> I don't think restarting the trigger service works in this case.

Well if there's a new git revision, restarting the trigger service will make it detect it sooner than waiting until the next time it wakes up on its own.

In fact I think it might be better to have the trigger service as a "one shot" service and start it with an external timer e.g. systemd or Kubernetes config.  But that's slightly off-topic for this discussion.

kernelci/cli/base.py
```diff
@@ -520,9 +525,24 @@ def __init__(self, path, command, cli_args, section=None):
...
                     break
-        self._settings = configparser.ConfigParser()
+
+        if path and path.endswith('.conf'):
+            self._deprecated_settings = True
+
+        if self._deprecated_settings:
+            print("Warning: user settings file format '.conf' will soon be \
+deprecated. Please use '.toml' file format and provide 'kernelci.toml' file \
+instead.")
```
gctucker, line 535:
```suggestion
instead.", file=sys.stderr)
```
The stdout data is used as input for other commands, so printing arbitrary messages on stdout can break things.

JenySadadia: Fixed it. Thanks.

gctucker:
Could you please rebase?  #1944 depends on this PR and it's causing some conflicts with other open PRs

JenySadadia:
> Could you please rebase? #1944 depends on this PR and it's causing some conflicts with other open PRs

Rebased it.

JenySadadia:
Fixed a typo in a comment in `kernelci.toml.sample` file.

kernelci/cli/base.py
```diff
@@ -560,11 +580,18 @@ def get(self, option, as_list=False):
...
+                return None
+            value = self._settings.get(section, option).split()
+            if not as_list and len(value) == 1:
+                value = value[0]
+        else:
+            value = None
+            if self._settings.get('DEFAULT'):
+                value = self._settings.get('DEFAULT').get(option)
+            if self._settings.get(section):
+                value = self._settings.get(section).get(option) or value
```
gctucker, line 594: BTW I think this can be optimised a bit by first looking for the value in the designated `section` as in this case there's no need to look up the `DEFAULT` value.  Also we can keep a shortcut to the default section in the object.  Something like this:

```python
# in __init__():
        if path and os.path.exists(path):
            if self._deprecated_settings:
                self._settings.read(path)
            else:
                self._settings = toml.load(path)
                self._default_section = self._settings.get('DEFAULT')

# [...]

# in get_from_section()
            value = None
            section_data = self._settings.get(section)
            if section_data:
                value = section_data.get(option)
            if value is None and self._default_section:
                value = self._default_section.get(option)
```

JenySadadia: So, at first, I added code to look up values from the designated section and then the default one. But then I ended up in so many `if` conditions somehow and `pycodestyle` was not happy about it.
OK. this seems fair and I'll update it.

JenySadadia: Oh, I remember now.
If we go with checking the specific section first and then the default one, the value from default would override.
That was the reason.

JenySadadia: If we define `default_section`, `self._default_section.get(option)` apparently calls `Options.get` method and it ends up in a recursive call. Hence, discarded it.

```
  File "/home/runner/work/kernelci-core/kernelci-core/kernelci/cli/base.py", line 552, in __getattr__
    return self.get(name)
  File "/home/runner/work/kernelci-core/kernelci-core/kernelci/cli/base.py", line 595, in get
    if value is None and self._default_section:
  File "/home/runner/work/kernelci-core/kernelci-core/kernelci/cli/base.py", line 552, in __getattr__
    return self.get(name)
  File "/home/runner/work/kernelci-core/kernelci-core/kernelci/cli/base.py", line 574, in get
    opt_data = self._command.get_arg_data(option)
  File "/home/runner/work/kernelci-core/kernelci-core/kernelci/cli/base.py", line 470, in get_arg_data
    return self._args_dict.get(arg_name)
RecursionError: maximum recursion depth exceeded while calling a Python object
```

gctucker: OK it would be good to get to the bottom of this and understand why it's a problem in the constructor but not in the method.  I'll take a look, it's not a blocker for this PR.

gctucker: This works for me:
```diff
diff --git a/kernelci/cli/base.py b/kernelci/cli/base.py
index 49017c8051e8..3128ff74354a 100644
--- a/kernelci/cli/base.py
+++ b/kernelci/cli/base.py
@@ -542,6 +542,7 @@ instead.", file=sys.stderr)
                 self._settings.read(path)
             else:
                 self._settings = toml.load(path)
+                self._default_section = self._settings.get('DEFAULT')
 
         self._command = command
         self._cli_args = cli_args
@@ -591,8 +592,8 @@ instead.", file=sys.stderr)
             section_data = self._settings.get(section)
             if section_data:
                 value = section_data.get(option)
-            if value is None and self._settings.get('DEFAULT'):
-                value = self._settings.get('DEFAULT').get(option)
+            if value is None and self._default_section:
+                value = self._default_section.get(option)
         return value
 
     def get_missing_args(self):
```
I guess we should also add some unit tests to cover the settings / `Options` logic at some point.

JenySadadia: > This works for me:
> 
> ```diff
> diff --git a/kernelci/cli/base.py b/kernelci/cli/base.py
> index 49017c8051e8..3128ff74354a 100644
> --- a/kernelci/cli/base.py
> +++ b/kernelci/cli/base.py
> @@ -542,6 +542,7 @@ instead.", file=sys.stderr)
>                  self._settings.read(path)
>              else:
>                  self._settings = toml.load(path)
> +                self._default_section = self._settings.get('DEFAULT')
>  
>          self._command = command
>          self._cli_args = cli_args
> @@ -591,8 +592,8 @@ instead.", file=sys.stderr)
>              section_data = self._settings.get(section)
>              if section_data:
>                  value = section_data.get(option)
> -            if value is None and self._settings.get('DEFAULT'):
> -                value = self._settings.get('DEFAULT').get(option)
> +            if value is None and self._default_section:
> +                value = self._default_section.get(option)
>          return value
>  
>      def get_missing_args(self):
> ```
> 
> I guess we should also add some unit tests to cover the settings / `Options` logic at some point.

Please see https://github.com/kernelci/kernelci-core/actions/runs/5140867870/jobs/9252740078

gctucker: OK this can be fixed somewhat by relying on the convention that settings don't start with `_`, that way private attributes will be found in the object while other attributes will return the settings value.

If the attribute way isn't actually used in the code then we might simply drop it and request all the code to use `.get()`, but since it's been there for a while I'm not entirely sure of this.  Maybe it's something we could consider deprecating when transitioning to TOML only, if it's actually a problem to override `__getattr__` and not that useful in practice.

JenySadadia: This seems like a hack to me than a fix.
Or we can use `self._deprecated_settings` flag to bifurcate both the logics.

gctucker: Found the actual problem, the `._default_section` wasn't created when the settings file was not present.  So a fix for this would be:
```diff
diff --git a/kernelci/cli/base.py b/kernelci/cli/base.py
index 49017c8051e8..28d6aa18a1c3 100644
--- a/kernelci/cli/base.py
+++ b/kernelci/cli/base.py
@@ -542,6 +542,7 @@ instead.", file=sys.stderr)
                 self._settings.read(path)
             else:
                 self._settings = toml.load(path)
+        self._default_section = self._settings.get('DEFAULT')
 
         self._command = command
         self._cli_args = cli_args
```
Then we can use `self._default_section` without causing an infinite recursion.

gctucker (approved):
OK we can now merge this as-is and optimise it as a follow-up with `self._default_section` if we want.

kernelci/cli/base.py
JenySadadia: > This seems like a hack to me than a fix. Or we can use `self._deprecated_settings` flag to bifurcate both the logics.

I was referring to this change by this comment:
```
+        if name.startswith('_'):
+            return super().__getattr__(name)
```

JenySadadia: > Found the actual problem, the `._default_section` wasn't created when the settings file was not present. So a fix for this would be:
> 
> ```diff
> diff --git a/kernelci/cli/base.py b/kernelci/cli/base.py
> index 49017c8051e8..28d6aa18a1c3 100644
> --- a/kernelci/cli/base.py
> +++ b/kernelci/cli/base.py
> @@ -542,6 +542,7 @@ instead.", file=sys.stderr)
>                  self._settings.read(path)
>              else:
>                  self._settings = toml.load(path)
> +        self._default_section = self._settings.get('DEFAULT')
>  
>          self._command = command
>          self._cli_args = cli_args
> ```
> 
> Then we can use `self._default_section` without causing an infinite recursion.

Oh, okay.
We should have a follow-up PR for this.

JenySadadia:
> OK we can now merge this as-is and optimise it as a follow-up with `self._default_section` if we want.

Please see https://github.com/kernelci/kernelci-core/pull/1947

PR #1938, opened 2023-05-22
build-configs-stable.yaml: Add Linux 6.3 stable trees

patersonc:
Even though it's not an LTS, I still think it's worth testing the latest stable releases and release candidates.

Note we missed adding v6.2.y testing.

### DIFF
--- a/config/core/build-configs-stable.yaml
+++ b/config/core/build-configs-stable.yaml
@@ -186,6 +186,11 @@ build_configs:
     branch: 'linux-6.1.y'
     variants: *stable_variants
 
+  stable_6.3:
+    tree: stable
+    branch: 'linux-6.3.y'
+    variants: *stable_variants
+
   stable-rc_3.16:
     tree: stable-rc
     branch: 'linux-3.16.y'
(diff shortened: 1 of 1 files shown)


gctucker:
Yes it's always been the policy to test all the stable branches and then drop them when they reach EOL.  It's a shame stable maintainers aren't updating the YAML config for this any more.  I think the issue was the manual work required every 3 months or so.  With the new API & Pipeline, we could have a trigger service dedicated to stable branches to do this more automatically without having everything hard-coded in YAML I think.

Thanks!

nuclearcat:
Started monitor job for all added trees: https://bot.staging.kernelci.org/job/kernel-tree-monitor/1740/console

gctucker:
> Started monitor job for all added trees: https://bot.staging.kernelci.org/job/kernel-tree-monitor/1740/console

Tested OK manually:
```
$ ./kci_build --yaml-config=config/core list_kernel_configs --build-config=stable_6.3 --kdir=/home/gtucker/src/linux | wc
    198     611    5916
$ ./kci_build --yaml-config=config/core list_kernel_configs --build-config=stable-rc_6.3 --kdir=/home/gtucker/src/linux | wc
    198     611    5916
$ ./kci_build --yaml-config=config/core list_kernel_configs --build-config=stable-rc_queue-6.3 --kdir=/home/gtucker/src/linux | wc
    198     611    5916
```

Whether the builds pass or fail doesn't really matter from a YAML configuration point of view, this is enough to show it's valid (comparing the list of configs with another branch e.g. stable-6.1 also helps).

gctucker (approved):

Issue #1941, opened 2023-05-24
Add support for arrays in dtb parameter


nuclearcat:
Currently kernelci allows to set only one dtb file per device, such as:
```
  sc7180-trogdor-kingoftown:
    mach: qcom
    class: arm64-dtb
    boot_method: depthcharge
    dtb: 'qcom/sc7180-trogdor-kingoftown-r1.dtb'
    filters: *arm64-chromebook-filters
```
But this particular device we got situation, in kernels 6.1, for example, dtb name is as set. But after this patch https://lore.kernel.org/lkml/20230302131031.v2.2.I68cbe5d5d45074428469da8c52f1d6a78bdc62fc@changeid/ name changes to sc7180-trogdor-kingoftown.dtb and as tests verify presence of this particular file, tests will not run anymore.
We can set it to new name, but then tests for old kernels wont run too.

I suggest to add support of arrays, so config will support dtb as before and:
```
  sc7180-trogdor-kingoftown:
    mach: qcom
    class: arm64-dtb
    boot_method: depthcharge
    dtb: ['qcom/sc7180-trogdor-kingoftown-r1.dtb','qcom/sc7180-trogdor-kingoftown.dtb']
    filters: *arm64-chromebook-filters
```

jluebbe:
With https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=724ba6751532055db75992fc6ae21c3e322e94a7 (part of v6.5-rc1) all ARM-32 dts files have moved. We probably need something like this PR to keep testing both current and stable kernels.

nfraprado:
There's always a chance issues like this will arise given that the DT filenames aren't stable... The stable way of doing it would be to have the compatible strings of the device in the device definition in the YAML instead of the dtb filename, and search for the best-matching DTB based on those compatibles and use that one.

gctucker:
There's already a simple solution for this in the YAML config for other devices, by creating 2 entries in YAML with one for each DTB name.  For example:
```yaml
  sc7180-trogdor-kingoftown: &sc7180-trogdor-kingoftown
    mach: qcom
    class: arm64-dtb
    boot_method: depthcharge
    dtb: 'qcom/sc7180-trogdor-kingoftown.dtb'
    filters: *arm64-chromebook-filters

  sc7180-trogdor-kingoftown-r1:
    <<: *sc7180-trogdor-kingoftown
    base_name: sc7180-trogdor-kingoftown
[...]

PR #1942, opened 2023-05-24
test-configs.yaml: Add one more device entry for different dtb file

nuclearcat:
As kingoftown got different dtb name in new kernels, as mentioned in https://github.com/kernelci/kernelci-core/issues/1941 , one of ways to solve this issue to add one more device entry with base_name pointing to initial device name.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -1928,12 +1928,17 @@ device_types:
     class: arm64-dtb
     boot_method: uboot
 
-  sc7180-trogdor-kingoftown:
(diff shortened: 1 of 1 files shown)


nuclearcat:
https://storage.staging.kernelci.org/kernelci/staging-mainline/staging-mainline-20230525.0/arm64/defconfig+arm64-chromebook+kselftest/gcc-10/lab-collabora/baseline-sc7180-trogdor-kingoftown.html
Works for new dtb, but waiting one of stable kernels tested on -r1, to make sure both entries working.

PR #1943, opened 2023-05-25
Build qcom-venus firmware into kernel image

laura-nao:
The qcom-venus driver is loaded quite early on in boot and requires the venus firmware. As the firmware binary is relatively small, build it into the Chromebooks kernel image to allow the video codecs driver to probe correctly on `trogdor` devices.

Remove the firmware binary from the `bullseye-v4l2` ramdisk/rootfs config.

### DIFF
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -348,6 +348,8 @@ fragments:
       mediatek/mt8173/vpu_d.bin
       mediatek/mt8173/vpu_p.bin
       rockchip/dptx.bin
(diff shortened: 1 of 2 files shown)


nuclearcat:
Firmware works: https://storage.staging.kernelci.org/mainline/master/v6.4-rc6/arm64/defconfig+arm64-chromebook/gcc-10/lab-collabora/v4l2-decoder-conformance-vp8-sc7180-trogdor-kingoftown.html

PR #1946, opened 2023-06-01
kernelci.cli: fix __getattr__() for private attributes

gctucker:
When accessing "private" attributes which by convention have a name that starts with _, rely on the standard method as settings don't normally start with '_'.  This will avoid obscure issues when trying to access a private attribute and the code looks for an option with the attribute's name instead.

### DIFF
--- a/kernelci/cli/base.py
+++ b/kernelci/cli/base.py
@@ -528,6 +528,8 @@ class Options:
         self._section = section
 
     def __getattr__(self, name):
+        if name.startswith('_'):
+            return super().__getattr__(name)
         return self.get(name)
 
     @property


gctucker:
Ah, found the actual issue.  With the previous implementation using `Options._default_section`, the attribute wasn't created when there was no file.  So this is actually the fix we need:
```diff
diff --git a/kernelci/cli/base.py b/kernelci/cli/base.py
index 49017c8051e8..28d6aa18a1c3 100644
--- a/kernelci/cli/base.py
+++ b/kernelci/cli/base.py
@@ -542,6 +542,7 @@ instead.", file=sys.stderr)
                 self._settings.read(path)
             else:
                 self._settings = toml.load(path)
+        self._default_section = self._settings.get('DEFAULT')
 
         self._command = command
         self._cli_args = cli_args
```

gctucker:
Closing this PR as `__getattr__()` is actually behaving as expected and as per the documentation.

PR #1947, opened 2023-06-02
kernelci.cli.base: optimize TOML file parsing

JenySadadia:
Add `Options._default_section` variable to read `DEFAULT` section from `.toml` file and update `Options.get` method accordingly.

### DIFF
--- a/kernelci/cli/base.py
+++ b/kernelci/cli/base.py
@@ -543,6 +543,9 @@ instead.", file=sys.stderr)
             else:
                 self._settings = toml.load(path)
 
+        if not self._deprecated_settings:
+            self._default_section = self._settings.get('DEFAULT')
+
         self._command = command
         self._cli_args = cli_args
         self._section = section
@@ -591,8 +594,8 @@ instead.", file=sys.stderr)
             section_data = self._settings.get(section)
             if section_data:
                 value = section_data.get(option)
-            if value is None and self._settings.get('DEFAULT'):
-                value = self._settings.get('DEFAULT').get(option)
+            if value is None and self._default_section:
+                value = self._default_section.get(option)
         return value
 
     def get_missing_args(self):


JenySadadia:
Needed to check for the `_deprecated_settings` flag.
```
if not self._deprecated_settings:
    self._default_section = self._settings.get('DEFAULT')
```

Otherwise, it will fail when the `.conf` file will be sent.
```
./kci config validate
Warning: user settings file format '.conf' will soon be deprecated. Please use '.toml' file format and provide 'kernelci.toml' file instead.
  File "/home/jeny/kernelCI/kernelci-core_env/src/kernelci-core/./kci", line 29, in <module>
  File "/home/jeny/kernelCI/kernelci-core_env/src/kernelci-core/kernelci/cli/__init__.py", line 53, in call
    _COMMANDS[name](args)
  File "/home/jeny/kernelCI/kernelci-core_env/src/kernelci-core/kernelci/cli/config.py", line 74, in main
    sub_main("config", globals(), args)
  File "/home/jeny/kernelCI/kernelci-core_env/src/kernelci-core/kernelci/cli/base.py", line 733, in sub_main
    opts = parse_opts(name, glob, args=args)
  File "/home/jeny/kernelCI/kernelci-core_env/src/kernelci-core/kernelci/cli/base.py", line 719, in parse_opts
    return make_options(args, prog)
  File "/home/jeny/kernelCI/kernelci-core_env/src/kernelci-core/kernelci/cli/base.py", line 693, in make_options
    opts = Options(args.settings, args.func, args, prog)
  File "/home/jeny/kernelCI/kernelci-core_env/src/kernelci-core/kernelci/cli/base.py", line 547, in __init__
    self._default_section = self._settings.get('DEFAULT')
TypeError: RawConfigParser.get() missing 1 required positional argument: 'option'
make: *** [Makefile:49: validate-yaml] Error 1
```

gctucker (commented):
LGTM - let's just wait for the staging jobs to confirm it's all working as expected.

gctucker:
> `./kci config validate`

BTW This command is to validate the YAML config, the user settings just happen to be loaded as a side-effect like with any command line tool.  I guess it's OK to run this as a quick check but having unit tests would be great to validate the logic in the code and maybe we could also have a way to validate any user's settings file but I'm not sure what's the best way to do that.  Maybe a dedicated tool in `scripts` or `kci settings` which would not be using the standard `sub_main` function from `kernelci.cli`?  The `kci` tool doesn't load settings itself afaict, it's at the discretion of each sub-command to implement their main function.

JenySadadia:
> > `./kci config validate`
> 
> BTW This command is to validate the YAML config, the user settings just happen to be loaded as a side-effect like with any command line tool. I guess it's OK to run this as a quick check but having unit tests would be great to validate the logic in the code and maybe we could also have a way to validate any user's settings file but I'm not sure what's the best way to do that. Maybe a dedicated tool in `scripts` or `kci settings` which would not be using the standard `sub_main` function from `kernelci.cli`? The `kci` tool doesn't load settings itself afaict, it's at the discretion of each sub-command to implement their main function.

Maybe we can use https://pypi.org/project/toml-validator/ to validate the settings file. I haven't tried it though.
As you suggested, we can use a separate `kci` tool for it and add it to `Actions`.

gctucker:
> Maybe we can use https://pypi.org/project/toml-validator/

Yes, I guess.  That's another great thing about using a standard settings format with specifications, we can use standard validators for it.

gctucker (approved):

PR #1949, opened 2023-06-07
CHROMEOS test-configs-chromeos.yaml: Update download links for trogdor/grunt

nuclearcat:
As images got fixed and rebuilt, we need to update download links


nuclearcat:
Grunt: https://chromeos.kernelci.org/test/plan/id/6486e28d2911523338302659/
Trogdor: https://chromeos.kernelci.org/test/plan/id/6486d75e81150541c630266e/

PR #1950, opened 2023-06-07
kernelci.cli: job: drop submit --api-token

gctucker:
Drop "kci job submit --api-token" argument as it's not used.  That way we can submit jobs without having to provide an API token.

Fixes: 3f9b582c6676 ("kernelci.cli: drop Args.api_token for anonymous requests")


gctucker:
Tested OK on staging.

PR #1952, opened 2023-06-07
config/runtime: lava: add storage_config_name metadata

gctucker:
Add a metadata entry with the name of the storage config so that the callback handler can upload artifacts to storage (e.g. job logs).

Fixes: https://github.com/kernelci/kernelci-pipeline/issues/214


gctucker:
Tested OK on staging.

PR #1954, opened 2023-06-08
CHROMEOS Enable fluster tests on asurada with collabora-chromeos-kernel

laura-nao:
Run the fluster decoder conformance tests on the `collabora-chromeos-kernel` tree.

### DIFF
--- a/config/core/test-configs-chromeos.yaml
+++ b/config/core/test-configs-chromeos.yaml
@@ -522,6 +522,52 @@ test_plans:
       <<: *cros-tast-base-qemu-params
       tests: *cros-tast-video-tests
 
+  v4l2-decoder-conformance-av1_chromeos: &v4l2-decoder-conformance_chromeos
+    rootfs: debian_bullseye-gst-fluster_nfs
+    pattern: 'v4l2-decoder-conformance/{category}-{method}-{protocol}-{rootfs}-v4l2-decoder-conformance-template.jinja2'
+    params: &v4l2-decoder-conformance-params_chromeos
+      job_timeout: '30'
+      testsuite: 'AV1-TEST-VECTORS'
+      decoders:
+        - 'GStreamer-AV1-V4L2SL-Gst1.0'
+      videodec_parallel_jobs: '1'
+      videodec_timeout: 90
+
+  v4l2-decoder-conformance-h264_chromeos:
+    <<: *v4l2-decoder-conformance_chromeos
+    params:
+      <<: *v4l2-decoder-conformance-params_chromeos
+      testsuite: 'JVT-AVC_V1'
+      decoders:
+        - 'GStreamer-H.264-V4L2-Gst1.0'
+        - 'GStreamer-H.264-V4L2SL-Gst1.0'
+
+  v4l2-decoder-conformance-h265_chromeos:
+    <<: *v4l2-decoder-conformance_chromeos
+    params:
+      <<: *v4l2-decoder-conformance-params_chromeos
+      testsuite: 'JCT-VC-HEVC_V1'
+      decoders:
+        - 'GStreamer-H.265-V4L2-Gst1.0'
+        - 'GStreamer-H.265-V4L2SL-Gst1.0'
+
+  v4l2-decoder-conformance-vp8_chromeos:
+    <<: *v4l2-decoder-conformance_chromeos
+    params:
+      <<: *v4l2-decoder-conformance-params_chromeos
+      testsuite: 'VP8-TEST-VECTORS'
+      decoders:
+        - 'GStreamer-VP8-V4L2-Gst1.0'
+        - 'GStreamer-VP8-V4L2SL-Gst1.0'
+
+  v4l2-decoder-conformance-vp9_chromeos:
+    <<: *v4l2-decoder-conformance_chromeos
+    params:
+      <<: *v4l2-decoder-conformance-params_chromeos
+      testsuite: 'VP9-TEST-VECTORS'
+      decoders:
+        - 'GStreamer-VP9-V4L2SL-Gst1.0'
+
 device_types:
 
   acer-R721T-grunt_chromeos:
--- a/config/core/test-configs-chromeos.yaml
+++ b/config/core/test-configs-chromeos.yaml
@@ -1096,6 +1096,13 @@ test_configs:
     filters: *collabora-chromeos-kernel-filter
     test_plans: *chromebook-arm64-test-plans
 
+  - device_type: mt8192-asurada-spherion-r0_chromeos
+    filters: *collabora-chromeos-kernel-filter
+    test_plans:
+      - v4l2-decoder-conformance-h264_chromeos
+      - v4l2-decoder-conformance-vp8_chromeos
+      - v4l2-decoder-conformance-vp9_chromeos
+
 # Test plans temporarily disabled as QEMU build doesnt work
 #  - device_type: qemu_x86_64-uefi-chromeos
 #    test_plans:


gctucker:
This kernel branch is only tested by the chromeos.kernelci.org instance so I wonder if the YAML config should be contained in the ChromeOS files instead.

laura-nao:
You mean the changes in `test-configs.yaml` should be in `test-configs-chromeos.yaml` instead? 

I noticed there are some tests with binaries from both https://storage.chromeos.kernelci.org and http://storage.kernelci.org (e.g. `baseline-nfs`: https://lava.collabora.dev/scheduler/job/9933145/definition) and assumed it was not required, but now I'm second guessing this assumption.

nfraprado:
@laura-nao That test is submitted by the ChromeOS instance of KernelCI.

I don't see anything in this PR that would cause ChromeOS specific tests or builds to happen in the main instance, so it's just about cleanly separating the two.

@gctucker Do you want to even avoid 'collabora-chromeos-kernel' from being part of the regex/filters of tests in the main YAML? Because if we want to avoid that, then the test configs will need to be duplicated in the chromeos YAML just to have that.

For the `test_configs` entries, I agree there's no current reason for them to be in the main YAML, but these tests on these machines should be enabled on linux-next very soon, at which point we'd be moving them back to the main YAML (assuming we can have the collabora regex in the main YAML).

gctucker:
Yes it's more from a maintenance point of view, it's useful to separate the configs for the main linux.kernelci.org instance and chromeos.kernelci.org so that making changes in some files doesn't cause side-effects in the other instance.

The `collabora-chromeos-kernel` branch is only tested on chromeos.kernelci.org and I don't think it's really eligible to be added to linux.kernelci.org as it's a "downstream" development branch.  So it makes sense to keep filters that refer to it only in the YAML config files for chromeos.kernelci.org.

gctucker:
@laura-nao I'm not sure about `baseline-nfs`, I'll take a closer look.  Generally speaking, I think it's OK if chromeos.kernelci.org uses binaries from storage.kernelci.org although it's not ideal.  Maybe it just makes sense in practice rather than rebuilding exactly the same rootfs images.  The other way round would not be OK though, the main instance shouldn't depend on the ChromeOS one.

laura-nao:
@gctucker  I reworked this PR to target the `test-configs-chromeos.yaml` file on the `chromeos` branch. In order to keep the filter for the `collabora-chromeos-kernel` tree out of the main instance YAML, I opened https://github.com/kernelci/kernelci-core/pull/1986 . I also opened https://github.com/kernelci/kernelci-core/pull/1987 to add the required scp firmware binaries to the gst+fluster bullseye rootfs.

Both these PRs target the `main` branch and the changes are required for this PR to work correctly. I can wait for all the required changes to land in the `main` branch first and on the `chromeos` branch after it is rebased. Otherwise I can push all the relevant commits to the branch in this PR too, so we can merge it right away (the commits will then be dropped in the chromeos-to-main rebase). Any thoughts on the best approach?

gctucker:

laura-nao:
I incorporated the changes from https://github.com/kernelci/kernelci-core/pull/1986 here, so that this patch can be picked without having to necessarily wait for the chromeos branch to be rebased on top of main. 

Waiting for https://github.com/kernelci/kernelci-core/pull/1989 to be merged before marking this as ready (adds required MT8192 SCP FW binary to the arm64-chromebook fragment).

laura-nao:
All the required changes were merged to the main branch and are now included in the chromeos branch too. Marked as ready for review.

gctucker:
Sorry I added the skip label to this PR a while ago when reducing coverage due to the Azure resource limitations but didn't realise it was a `chromeos` PR.  So I've removed the skip label again now.

nuclearcat:
I did run collabora kernel as specified in filter, but tests didnt got scheduled.
I think because conformance tests have their own filter:
```
      - passlist: {defconfig: ['videodec']}
```
https://github.com/kernelci/kernelci-core/blob/7755ba945b95f10917b9a6a276bcadef34344a03/config/core/test-configs.yaml#L695

nuclearcat:
If we still need this PR, can you rebase it please and fix filter issue?

laura-nao:
I originally thought it was possible to define `passlist: {defconfig: ['videodec']}` on a per-device basis; apparently this does not work though, and the defconfig passlist was defined in the test plan definition instead: https://github.com/kernelci/kernelci-core/pull/2012/

As a consequence, the tests on the `collabora-chromeos-kernel` tree don't get scheduled (as there's no `videodec` fragment). Defining an empty passlist [here](https://github.com/kernelci/kernelci-core/blob/chromeos/config/core/test-configs-chromeos.yaml#L828)) doesn't seem to help either, the videodec fragment requirement remains.

Any idea on the proper way to handle this corner case?

nuclearcat:
I think it might work like this:
```
diff --git a/config/core/test-configs.yaml b/config/core/test-configs.yaml
index 9dabbf09f..c06d35462 100644
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -692,7 +692,11 @@ test_plans:
       videodec_parallel_jobs: '1'
       videodec_timeout: 90
     filters:
-      - passlist: {defconfig: ['videodec']}
+      - passlist:
+          defconfig: ['videodec']
+      - passlist:
+          tree: ['collabora-chromeos-kernel']
+          branch: ['for-kernelci']
 
   v4l2-decoder-conformance-h264:
     <<: *v4l2-decoder-conformance
```
Can you try this and rebase again please? i had to rebase chromeos to include fix for docker images builds

laura-nao:
I tried running `kci_test list_jobs` with the change suggested above, using bmeta.json and artifacts.json from the `collabora-chromeos-kernel` and from `mainline` build folders.

With the latest `collabora-chromeos-kernel` build I get this list of tests for `spherion`, which doesn't include the `v4l2-decoder-conformance-*` tests:
```
./kci_test list_jobs --runtime-config=lab-collabora  --runtime-token <token> --install-path data/collabora
mt8192-asurada-spherion-r0 baseline
mt8192-asurada-spherion-r0 baseline-nfs
mt8192-asurada-spherion-r0 cros-ec
mt8192-asurada-spherion-r0 sleep
mt8192-asurada-spherion-r0 v4l2-compliance-mtk-vcodec-enc
mt8192-asurada-spherion-r0_chromeos cros-baseline
mt8192-asurada-spherion-r0_chromeos cros-tast-hardware
mt8192-asurada-spherion-r0_chromeos cros-tast-kernel
mt8192-asurada-spherion-r0_chromeos cros-tast-mm-decode
mt8192-asurada-spherion-r0_chromeos cros-tast-mm-decode-v4l2-stateless-h264
mt8192-asurada-spherion-r0_chromeos cros-tast-mm-decode-v4l2-stateless-vp8
mt8192-asurada-spherion-r0_chromeos cros-tast-mm-decode-v4l2-stateless-vp9-group1
mt8192-asurada-spherion-r0_chromeos cros-tast-mm-decode-v4l2-stateless-vp9-group2
mt8192-asurada-spherion-r0_chromeos cros-tast-mm-decode-v4l2-stateless-vp9-group3
mt8192-asurada-spherion-r0_chromeos cros-tast-mm-decode-v4l2-stateless-vp9-group4
mt8192-asurada-spherion-r0_chromeos cros-tast-mm-misc
mt8192-asurada-spherion-r0_chromeos cros-tast-perf
mt8192-asurada-spherion-r0_chromeos cros-tast-platform
mt8192-asurada-spherion-r0_chromeos cros-tast-power
mt8192-asurada-spherion-r0_chromeos cros-tast-sound
mt8192-asurada-spherion-r0_chromeos cros-tast-ui
mt8192-asurada-spherion-r0_chromeos cros-tast-video
```

The change above seems to also prevent jobs on `mainline` from being scheduled:
```
./kci_test list_jobs --runtime-config=lab-collabora  --runtime-token <token> --install-path data/mainline
mt8192-asurada-spherion-r0 baseline
mt8192-asurada-spherion-r0 baseline-nfs
```

I believe with the changes above the regex filter [here](https://github.com/kernelci/kernelci-core/blob/main/config/core/test-configs.yaml#L3178) isn't applied anymore, and as a consequence it only matches `collabora-chromeos-kernel` trees with `videodec` fragment (which does not exist).

The kernel tree filters seems to work well as-is (i.e. `mainline`, `next`, `media` and `collabora-chromeos-kernel` are all matched), it's just the defconfig passlist that cannot be overridden for the test plans on the `collabora-chromeos-kernel` tree.

laura-nao:
@nuclearcat as the filter solution above didn't work as expected, I've reworked this PR to duplicate the fluster test plan definitions in the `test-configs-chromeos.yaml` file in order to be able to run these tests on kernel trees from the ChromeOS instance too. In particular, this PR enables the tests on `spherion` for the `collabora-chromeos-kernel` tree. The resulting LAVA jobs use the gst+fluster ramdisk/rootfs archives from the main KernelCI instance, and the kernel artifacts from the chromeos instance.

I was not able to proof test the jobs manually as there's no build for the `collabora-chromeos-kernel` tree in the storage at the moment, so I just visually checked the job definitions generated by `kci_test generate`.

nuclearcat (approved):
Results are on dashboard

PR #1955, opened 2023-06-08
test-configs.ymal: Add sigaltstack kselftest coverage

broonie:
The sigaltstack test is *tiny* so the time taken to boot a test platform is
going to dwarf the time taken to actually run the tests but it's relevant
to some development work I'm doing so let's add some test coverage for it.
Add the fragment for it and run it on two platforms in my lab.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -408,6 +408,13 @@ test_plans:
       job_timeout: '10'
       kselftest_collections: "seccomp"
 
+  kselftest-sigaltstack:
+    <<: *kselftest
+    params:
+      job_timeout: '10'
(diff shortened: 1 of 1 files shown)


gctucker:
Sounds like a good candidate for trying multi-test jobs when we get there in the future.

broonie:
My theory with the multi test jobs is that we should be able to just opportunistically combine jobs that use the same rootfs unless they're explicitly flagged as disruptive. I've got a design for it, been holding off due to a combination of being very busy and the pipeline rework.

nuclearcat (approved):
Results available on dashboard: https://staging.kernelci.org/test/plan/id/64899da9cf5e4598eec5925c/

PR #1956, opened 2023-06-08
config: add `rustc-1.70` build environment

ojeda:
Rust 1.70.0 will be the next version supported by the kernel.

Therefore, add it as a new build environment.


ojeda:
I also have to upgrade `bindgen` (for all build environments), and that one is more urgent since we would like to make it to the merge window since LLVM 16 breaks the build.

So I will send another one for `bindgen`, and I will update the `bindgen` version here too.

ojeda:
We decided to put the `bindgen` upgrade in after the merge window, so no need for that.

We are moving all our branches to Rust 1.70.0, by the way, so we could, in principle, remove the old build environments too.

gctucker:
OK catching up with all this now, will take a look to see what can / needs to be merged before Monday's production update.

ojeda:
Thanks @gctucker!

I have added here the commits that move all the branches too (please let me know if you need this split into different PRs), as well as the removal of the old build environments, in case that is useful. Though I am not sure if they are usually kept for some reason (e.g. for testing of stable branches).

gctucker:
Sounds good.  I think it's probably a bit early for testing stable branches with Rust so for now it seems like the best approach to remove the older versions.  Again I'll take a look at all this tomorrow as I should finally have the time to do that.

gctucker:

gctucker:
Reviving this _again_ now that the Azure issue has settled.  We can't just re-enable all the build and test coverage yet but should be able to address the Rust support updates at least.

ojeda:
Thanks Guillaume!

So Rust 1.71 got released in the meantime, and I was considering submitting that one to mainline directly for this cycle if all works on my side. If you are planning to land an update on Monday, should I update this PR or send another? What is best for you? (i.e. in case it avoids you some work).

gctucker:
I think it's probably best to leave this one as-is and then have a follow-up for rustc-1.71.  Thanks for the update :)

gctucker:
Waiting for this build to complete on staging... will check again on Monday.
https://bot.staging.kernelci.org/job/kernel-build/82737/console

gctucker:
..that failed for some reason, probably Jenkins as usual.  I've restarted it with a bigger builder to get it to complete quicker, let's see how that goes.  At least the kernel build jobs with `rust-1.70` appear to be getting scheduled as expected.

gctucker:
Alright there's some congestion in the Kubernetes cluster doing kernel builds.  That'll get resolved later today, in the meantime I'm double-checking this is fine manually.  I think it's ready to be merged now.

@ojeda Next step, `rust-1.71` ;)

gctucker:
We knew it wasn't going to be straightforward, I guess we asked for it a few months ago ;)

Also I've replied to the email thread about the missing results, I think we need to add `LLVM=1`:
https://lore.kernel.org/kernelci/2f443a56-3a84-dead-2c79-a25781780357@collabora.com/T/#m3d2277a6d27cacffb4930bc030ad1ad1464ca842

ojeda:
> @ojeda Next step, `rust-1.71` ;)

Done! https://github.com/kernelci/kernelci-core/pull/2039

Thanks!

PR #1957, opened 2023-06-09
test-configs.yaml: Add Cherry Tomato MT8195 Chromebook

nuclearcat:
Add new Chromebook for testing, for now just baseline.


gctucker:

nuclearcat:
Booting, as confirmed in : https://github.com/kernelci/kernelci-core/pull/1998

PR #1961, opened 2023-06-13
test-configs: Add Renesas R-car Starter Kit H3e-2G

sjoerdsimons:
This board is currently available in the Collabora lava staging and should soon also be in the collabora production lab


nuclearcat:
Results available on staging dashboard: https://staging.kernelci.org/test/plan/id/64894781f1c93d76a3c59267/

nuclearcat (approved):

PR #1962, opened 2023-06-16
CHROMEOS cros-tast.jinja2: Reduce chance of filesystem corruption

nuclearcat:
As some of DUT might have batteries removed, abrupt poweroff might cause issues with storage. Better to make sure all buffers are sync and try to poweroff gracefully.
Not sure yet, if we will give them enough time, probably worth to add small additional delay later.

### DIFF
--- a/config/lava/chromeos/cros-tast.jinja2
+++ b/config/lava/chromeos/cros-tast.jinja2
@@ -38,6 +38,13 @@
             - >-
               ./tast_parser.py
               {{ tests|wordwrap(1, False, "\n              ") }}
+            - >-
+              ./ssh_retry.sh
+              -o StrictHostKeyChecking=no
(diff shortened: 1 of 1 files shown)


gctucker:
Maybe add the `chromeos` label so this will also be rolled out for the ChromeOS jobs?

nuclearcat:
> Maybe add the `chromeos` label so this will also be rolled out for the ChromeOS jobs?

oops my mistake, sorry

gctucker:
Ah yes this is for the `chromeos` branch so `CHROMEOS` in the commit subject is good.

I first thought this was for the `main` branch in which case setting the `chromeos` GitHub label would have made this PR to get deployed for ChromeOS jobs with the script in `kernelci-deploy`.

gctucker:
@nuclearcat Please rebase

PR #1966, opened 2023-06-20
Enable V4L2 decoder tast tests on trogdor Chromebooks

laura-nao:
Enable V4L2 stateful decoder tast tests on `sc7180-trogdor-kingoftown` and `sc7180-trogdor-lazor-limozeen`.


nuclearcat:
Success! For example results on dashboard: 
https://chromeos.kernelci.org/test/plan/id/649bf9117af9027ac530265d/

Issue #1968, opened 2023-06-22
BUG:  bullseye-cros-flash, time stamp in the future


nuclearcat:
During rootfs build, in CR50 unpack step we are getting following errors:
```
2023/06/21 20:56:10 bullseye-cros-flash.sh | tar: .checkpatch.conf: time stamp 2023-06-21 20:56:10.825 is 0.354338957 s in the future

2023/06/21 20:56:10 bullseye-cros-flash.sh | tar: .clang-format: time stamp 2023-06-21 20:56:10.825 is 0.354137976 s in the future

2023/06/21 20:56:10 bullseye-cros-flash.sh | tar: .dir-locals.el: time stamp 2023-06-21 20:56:10.827 is 0.356021579 s in the future

2023/06/21 20:56:10 bullseye-cros-flash.sh | tar: .gitignore: time stamp 2023-06-21 20:56:10.827 is 0.355895947 s in the future

2023/06/21 20:56:10 bullseye-cros-flash.sh | tar: .pylintrc: time stamp 2023-06-21 20:56:10.827 is 0.355774235 s in the future
```
Most likely it is a glitch or small desync in timing of servers and we should just ignore this error.

nuclearcat:
Issue was solved

PR #1969, opened 2023-06-22
bullseye-cros-flash.sh: Update repository for CR50 flashing tools

nuclearcat:
As we faced error while extracting CR50 archive, https://github.com/kernelci/kernelci-core/issues/1968, it is better to ignore modified time in tar archive.


nuclearcat:
This PR not useful, reason is that tools was moved to another repository

nuclearcat:
Updated commit, changed repository to new one, tested rootfs build on staging.
https://storage.chromeos.kernelci.org/images/rootfs/debian/bullseye-cros-flash/20230623.0/
Tested flashing with it

PR #1970, opened 2023-06-22
config/docker: use kernelci.org branch for data files

gctucker:
Use the kernelci.org production branch when downloading Docker data files at build time rather than the fixed git sha1 revisions.  This was needed initially to bootstrap the templates before they got merged, now we can use the kernelci.org branch to get updates automatically in sync with production.

### DIFF
--- a/config/docker/base/host-tools.jinja2
+++ b/config/docker/base/host-tools.jinja2
@@ -19,15 +19,15 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     cdbs \
     devscripts \
     equivs \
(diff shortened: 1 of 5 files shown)


gctucker:
Adding `staging-skip` for now as this will only actually work after the next production update since some files aren't yet on the `kernelci.org` branch.  So it can be tested and merged next week in principle.

gctucker (commented):
Tested OK on staging.  All the Docker images built OK and the jobs were run fine as well.  We can also double check things are looking fine with the linux-next job results on Monday on staging.

PR #1972, opened 2023-06-22
`kci user` command for user groups

JenySadadia:
Added commands to `kci user` for getting user groups by ID and attributes.
Newly added sample commands:
```
kci user get_group <group-id>
kci user find_groups
kci user find_groups name=admin
```

### DIFF
--- a/kernelci/api/latest.py
+++ b/kernelci/api/latest.py
@@ -124,29 +124,8 @@ class LatestAPI(API):
         offset: Optional[int] = None, limit: Optional[int] = None
     ) -> Sequence[dict]:
         params = attributes.copy() if attributes else {}
-        nodes = []
-
-        if any((offset, limit)):
-            params.update({
-                'offset': offset or None,
-                'limit': limit or None,
-            })
-            resp = self._get('nodes', params=params)
-            nodes = resp.json()['items']
-        else:
-            offset = 0
-            limit = 100
-            params['limit'] = limit
-            while True:
-                params['offset'] = offset
-                resp = self._get('nodes', params=params)
-                items = resp.json()['items']
-                nodes.extend(items)
-                if len(items) < limit:
-                    break
-                offset += limit
-
-        return nodes
+        return self._get_api_objs(params=params, path='nodes',
+                                  limit=limit, offset=offset)
 
     def count_nodes(self, attributes: dict) -> int:
         return self._get('count', params=attributes).json()
--- a/kernelci/api/latest.py
+++ b/kernelci/api/latest.py
@@ -89,6 +89,33 @@ class LatestAPI(API):
                 continue
             return event
 
+    def _get_api_objs(self, params: dict, path: str,
+                      limit: Optional[int] = None,
+                      offset: Optional[int] = None) -> list:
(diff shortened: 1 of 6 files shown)


kernelci/api/latest.py
```diff
@@ -89,37 +89,42 @@ def receive_event(self, sub_id: int):
...
-        return self._get(f'node/{node_id}').json()
-
-    def get_nodes(
-        self, attributes: dict,
-        offset: Optional[int] = None, limit: Optional[int] = None
-    ) -> Sequence[dict]:
-        params = attributes.copy() if attributes else {}
-        nodes = []
-
+    def get_api_objs(self, params: dict, path: str, limit: int = None,
```
gctucker, line 92: This seems like it should be a "private" method e.g. `_get_api_objs()`.

kernelci/cli/base.py
```diff
@@ -178,6 +178,11 @@ class Args:  # pylint: disable=too-few-public-methods
         'help': "Path to the dtbs.json file",
     }
 
+    group_id = {
+        'name': '--group-id',
```
gctucker, line 182: How about just using it as a positional argument `group_id` like for `kci node get`?

kernelci/cli/user.py
```diff
@@ -11,6 +11,23 @@
 from .base_api import APICommand
 
 
+class UserAttributesCommand(APICommand):
```
gctucker, line 14: Is this different to how nodes are being searched using arbitrary attributes?  Maybe something more generic for any kind of search (find) query could be used instead.

JenySadadia: `pylint` was also complaining about it. Added a generic class in `base_api.py`. Bdw I am thinking of a more suitable name for the class than `AttributesCommand`.

kernelci/api/latest.py
JenySadadia: Oh, yes. I'll fix it.

kernelci/cli/base.py
JenySadadia: Yes, updated it.

JenySadadia:
Tested OK on staging:
```
$ ./kci user find_groups --api-config=staging.kernelci.org
[
{
"id": "649190b01432ccfdb575a250",
"name": "admin"
}
]
```
```
$ ./kci user get_group 649190b01432ccfdb575a250 --api-config=staging.kernelci.org
{
"id": "649190b01432ccfdb575a250",
"name": "admin"
}

```

kernelci/cli/base.py
```diff
@@ -178,6 +178,11 @@ class Args:  # pylint: disable=too-few-public-methods
         'help': "Path to the dtbs.json file",
     }
 
+    group_id = {
+        'name': 'group_id',
+        'help': "User group id",
+    }
+
```
gctucker, line 185: I'm not sure it's useful to have it in the shared class, but that's a grey area anyway.  All the other arguments there are optional ones with `--something`.  This is not a blocking issue but rather something that would need to be revisited once we remove all the deprecated command line tools (or move the code to the `kernelci.legacy` package).

gctucker (approved):

PR #1973, opened 2023-06-23
CHROMEOS cros-flash: Add support for flashing over ssh and image as URL

nuclearcat:
As serial proven to be unreliable, this is attempt to improve reliability of flashing process.


gctucker:
I need to do the `chromeos` branch rework today so I guess you'll just rebase this afterwards?

nuclearcat:
yes, sure

nuclearcat:
Tested on LAVA manually, works well and more reliable, might be used as foundation to convert tests images to ssh

gctucker (commented):
Out of interest, have you tried the regular way of providing images in LAVA job definitions?  See how some Apertis jobs do it.

The issue with directly downloading images is that it doesn't use the cache, which is why it was always embedded into the rootfs images as they are cached with LAVA downloads.

nuclearcat:
I found way to use cache during direct downloads, in my templates in flashing (i should integrate them eventually with kernelci) it is done like this:
```
            ./flash http://${NFS_SERVER_IP}/cache/?uri={{ cros_flash_chromiumos_bin }} {{ cros_flash_blockdevice }}
```
On my experience it is most fast and storage efficient way, as any other methods due LAVA inefficiency might exceed storage limits (if it is in tmpfs, as some dispatchers have). Unfortunately when LAVA download any kind of archive and unpack it, it often use double space (leaving original tar for example), and parallel flashing was often causing OOM.

PR #1975, opened 2023-06-23
config/rootfs: add R111 and R114 Chromium OS manifests

gctucker:
Add manifests for Chromium OS R111 and R114 from the `chromeos` branch.


gctucker:
The other snapshots were already in `main` so that seemed like an easy one to just merge as well.  I've rebased the `chromeos` branch on top of this PR and there's no diff with the resulting code.

PR #1976, opened 2023-06-23
config/rootfs: build_board.sh: Remove hack for corrupted git cache

gctucker:
As tested this hack not required anymore as Google fixed corrupted git cache issue.

Fixes: e8c2b38d4ff3 ("build_board.sh: Add workaround for chrome-icu git cache corruption")


nuclearcat:
This patch was tested before and OK in staging and ChromeOS

PR #1977, opened 2023-06-23
Chromium OS: use R114 rootfs images

gctucker:
Also merge the 2 commits to update the Chromium OS rootfs images to R111 and then R114 following the added snapshots.


nuclearcat:
Works well on staging (well they are not used, but either not breaking anything)

PR #1985, opened 2023-06-27
Smcgannon/add lava dev

Smcgannon:


config/core/test-configs.yaml
```diff
@@ -2986,6 +2986,13 @@ test_configs:
   - device_type: mt8365-evk
```
broonie, line 2986: This device type already exists, this is expanding the test set for the existing device rather than adding a new device.

PR #1986, opened 2023-06-28
test-configs.yaml: Move videodec_timeout param and kernel tree filter for decoder tests

laura-nao:
Move the kernel tree filter used for the V4L2 decoder conformance tests out of the test plan definitions and specify the reference trees on a per-device basis. Use the same `videodec_timeout` value for all the devices running these decoder tests.


gctucker:
Adding skip label because of #1994 l

gctucker (approved):
Actually this isn't conflicting with the other changes done to reduce test coverage, so this might still just work.  In principle, every test plan should still be run but on a reduced number of platforms and with fewer kernel trees.  So hopefully we'll get some decent amount of results for this, but let's double check all that next week once it's in production.

PR #1990, opened 2023-07-02
test-configs.yaml: Expand coverage for i.MX8MP-EVK

broonie:
The i.MX8MP has networking so we can at least cover the baseline-nfs
tests, and the main reason it's in my lab is that it has interesting
audio hardware so run the ALSA selftests too (the audio isn't in the
upstream DT yet but the tests handle this gracefully).


nuclearcat (approved):
LGTM

PR #1993, opened 2023-07-03
install-modules: Enforce root ownership

nuclearcat:
We need to ensure proper ownership, otherwise it might trigger self_repair which might delete persistent partition


gctucker:
@nuclearcat Looks like this also needs the `chromeos` label?

nuclearcat:
I will try to build image on main instance (after production update), if it will work - then chromeos label is not required

gctucker:
OK, I thought ChromeOS staging jobs were run with the `chromeos.kernelci.org` branch - including rootfs builds.  But I guess things might be more complicated than that as there's still be the tech debt issue with the ChromeOS instance being half-way between production and staging.

nuclearcat:
It seems reason of partition corruption is different from ownership issue, closing for now as this is not necessary

PR #1994, opened 2023-07-05
Reduce build and test coverage to cope with Azure limits

gctucker:
As we're moving to the new Azure subscription we're currently limited in terms of build and network bandwidth capacity.  As such, reduce builds by disabling allmodconfig and reducing standard tree coverage to the minimal variants.  Also reduce test coverage to only have a couple of devices running each test plan to minimise the downloads from storage.

### DIFF
--- a/config/core/build-configs-android.yaml
+++ b/config/core/build-configs-android.yaml
@@ -11,7 +11,8 @@ android_variants: &android_variants
         arm: &arm_arch
           base_defconfig: 'multi_v7_defconfig'
           extra_configs:
-            - 'allmodconfig'
+# Disabling to reduce load
+#            - 'allmodconfig'
             - 'allnoconfig'
             - 'multi_v7_defconfig+CONFIG_CPU_BIG_ENDIAN=y'
             - 'multi_v7_defconfig+CONFIG_SMP=n'
@@ -20,7 +21,8 @@ android_variants: &android_variants
 
         arm64: &arm64_arch
           extra_configs:
-            - 'allmodconfig'
+# Disabling to reduce load
+#            - 'allmodconfig'
             - 'allnoconfig'
             - 'defconfig+CONFIG_CPU_BIG_ENDIAN=y'
             - 'defconfig+CONFIG_RANDOMIZE_BASE=y'
@@ -38,7 +40,8 @@ android_variants: &android_variants
         x86_64: &x86_64_arch
           base_defconfig: 'x86_64_defconfig'
           extra_configs:
-            - 'allmodconfig'
+# Disabling to reduce load
+#            - 'allmodconfig'
             - 'allnoconfig'
 
 clang_android_variants: &clang_android_variants
--- a/config/core/build-configs-stable.yaml
+++ b/config/core/build-configs-stable.yaml
@@ -26,13 +26,21 @@ stable_variants: &stable_variants
         extra_configs: ['allnoconfig']
         filters:
           # remove any non-ARCv2 defconfigs since we only have ARCv2 toolchain
-          - blocklist:
+# Disabling to reduce load
+#          - blocklist:
+#              defconfig:
+#                - 'axs101_defconfig'
+#                - 'nps_defconfig'
+#                - 'nsim_700_defconfig'
+#                - 'nsimosci_defconfig'
+#                - 'tb10x_defconfig'
+          - passlist:
               defconfig:
-                - 'axs101_defconfig'
-                - 'nps_defconfig'
-                - 'nsim_700_defconfig'
-                - 'nsimosci_defconfig'
-                - 'tb10x_defconfig'
+                - 'imx_v6_v7_defconfig'
+                - 'multi_v5_defconfig'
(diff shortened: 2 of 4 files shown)


broonie:
On Wed, Jul 05, 2023 at 12:10:00PM -0700, Guillaume Tucker wrote:
> As we&#39;re moving to the new Azure subscription we&#39;re currently limited in terms of build and network bandwidth capacity.  As such, reduce builds by disabling allmodconfig and reducing standard tree coverage to the minimal variants.  Also reduce test coverage to only have a couple of devices running each test plan to minimise the downloads from storage.

Can we put a CDN in front of storage to mitigate the issue with
downloads from storage, Azure has:

   https://azure.microsoft.com/en-gb/products/cdn/

(probably a good idea in general anyway, both from a load on storage and
cost perspectives.)

nuclearcat:
Azure CDN prices are almost same as egress, they have just better geographical proximity to users:
https://azure.microsoft.com/en-us/pricing/details/bandwidth/
Azure -> N.America/Europe  $0.087 per GB 
https://azure.microsoft.com/en-us/pricing/details/cdn/
$0.081 per GB and much higher, depends...

On my opinion to reduce costs we have following options:
1(best)) Hosting storage.*kernelci.org on hetzner server:
https://www.hetzner.com/dedicated-rootserver/sx64
96.75 Eur, unmetered 1Gbit. They have add-on 10Gbit and charging $1/TB($0.001/GB) for overusage of allocated quota.
This might be complex due frontend/backend configuration, but maybe we can sort this out

2)Caching proxy on this hetzner server, so any file from storage.* served only once. Still might reduce bandwidth a lot.
3)Cloudflare pro + Cloudflare R2 https://www.cloudflare.com/products/r2/

Maybe there is more options, i am not sure...

gctucker:
I really think this is an interim solution to avoid hitting some limits inadvertently, and tbh the kernel builds are probably costing us much more than the binary downloads.  But we first need to bring the costs to the bare minimal to ensure continuity of the main services and there's no doubt we'll find a solution in the coming weeks to bring things back to normal.  Making efficient use of the available resources is important in any case, aside from how much things cost it will lead to better performance overall.

Let's see how things go with the changes in this PR with the full linux-next build on staging this weekend and probably we'll already be able to find a balance in between the full set we had and this next week even without changing how the infrastructure is setup.  Then if things go well we'll end up with a more efficient config as well as some long-term sustainable amount of resources.

nuclearcat:
Actually Azure(and other cloud services) is not feasible (unless you are very rich corporation or startup with lot of funding) even for transfer volumes we have, i believe optimizing storage and egress required for the long-term.

broonie:
Ah, that's a shame with the cache service - for AWS CloudFront in front of a S3 buckets actually _super_ cost effective - I'm not paying any bandwidth costs for the binaries I serve up to my lab for my CI (uploads to S3 are free, transfers from S3 to CloudFront are free and even when I get out of the free tier on the CDN it's their cheapest bandwidth IIRC).

nuclearcat:
I am not sure on exact numbers on our egress, because for example we have transfers of sources to GKS nodes (classified as egress), serving kernels to labs and etc, but looking to just number on production instance eth0 - it might be quite significant traffic, 5-8TB/week.
Even AWS Cloudfront it is $0.080/GB.
It is really hard to beat good dedicated servers/colo offers, we are in that zone when having such server might be justified, as we are not earning on distributing this files.
Unfortunately i dont have any kind of logs to estimate numbers, but if our bandwidth costs on Azure are over $2000, and we can save half, we can get dedicated server with 10G port on something like fdcservers(10G unmetered is below $1k) or voxility($6.26/TB).
Even if less we can try to get away with hetzner, they have unmetered 1Gbit/s offers (which might be enough for us), or 1TB/$1 10G offers.
If we really insist on staying on S3 compatible managed services, then maybe we should look at all options, including linode, DO, vultr.
Linode:  
```
Egress transfer costs is free up to 1TB then $.01/GB, AWS egress transfer $.09/GB with first 1GB free, GCP egress transfer $.12/GB.
```
Vultr:
```
 $10 per additional TB transferred  (thats $0.01/GB)
```
(And they have pretty significant free tier)

But even with external server generating content things might be complicated and egress charges will be significant. For example our build K8S clusters will upload binaries elsewhere, this is still egress.

gctucker:
This PR is now producing the kind of discussions we need for https://github.com/kernelci/kernelci-api/issues/9 ;)

broonie:
Right, a CDN is definitely not an ideal solution once you get too far over the free tier - I was just thinking that they're really simple and non-invasive to enable so if the pricing worked out with Azure it might've helped mitigate things with little effort. It seems like there's not enough of a free/cheap tier with them to be relevant for us sadly :(

gctucker:
There was still a couple of allmodconfig left as an oversight, fixed that now.  Otherwise the linux-next results from the staging weekend run are available here:
https://staging.kernelci.org/build/next/branch/master/kernel/next-20230707/

PR #1995, opened 2023-07-06
cros-snapshot-release-R114-15437.B.xml: Update 5.10 kernel tree

nuclearcat:
New boards require up to date kernel tree, let's use Google branch.


nuclearcat:
Tested on Guybrush, fixes build issues

PR #1996, opened 2023-07-06
build_board.sh: Add libsandbox bug workaround

nuclearcat:
During build of ChromiumOS on certain builders it fail during ChromiumOS build with multiple malloc failure messages. Similar but reported for Gentoo and libsandbox, as workaround they just disable build sandbox feature, and we do the same.


nuclearcat:
Tested on different builders, workaround works.

PR #1998, opened 2023-07-10
build-configs.yaml: Kernel configs for MT8195 Tomato

nfraprado:
Kernel config changes to get MT8195 booting and with full platform support (once missing firmware and kernel series for the display land).

This should land after #1957, as that's the PR enabling baseline tests on MT8195 Tomato in the first place.

### DIFF
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -344,6 +344,7 @@ fragments:
       - 'CONFIG_VIDEO_OV02A10=m'
       - 'CONFIG_VIDEO_OV5695=m'
       - 'CONFIG_VIDEO_OV8856=m'
+      - '# CONFIG_ARM_DSU_PMU is not set'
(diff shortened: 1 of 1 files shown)


nfraprado:
I've sent a patch upstream so that disabling `CONFIG_ARM_DSU_PMU` is not necessary: https://lore.kernel.org/all/20230720200753.322133-1-nfraprado@collabora.com

Although given that no other platform needs `CONFIG_ARM_DSU_PMU`, I think it makes more sense to keep it in the fragment, so we can boot Tomato on previous kernel releases.

nuclearcat:
Tomato is booting in baseline: https://storage.staging.kernelci.org/kernelci/staging-next/staging-next-20230729.0/arm64/defconfig+arm64-chromebook/gcc-10/lab-collabora/baseline-mt8195-cherry-tomato-r2.html
https://staging.kernelci.org/test/plan/id/64c54b108064c38582b8750f/

PR #2005, opened 2023-07-12
rootfs-images.yaml: Test new fluster builds

nuclearcat:
We need to test new builds of fluster, i built rootfs images manually, as we have many PR pending for weeks.


gctucker:
Oh, was this a production image?  Well in any case we need to recreate it as there's been some further changes.  Production images should be built with the `main` branch.

PR #2006, opened 2023-07-12
config/docker: fix indentation in GitHub URLs

gctucker:
Fix indentation in GitHub file URLs wrapped over multiple lines.  The indentation was being added in the middle of the URLs, causing a failure to build the image.

Fixes: 6ea6fcbdce1a ("config/docker: use kernelci.org branch for data files")


gctucker:
Tested OK with production image rebuilds (was required as they were out-of-date due to the docker build issues caused by the indentation).

gctucker (commented):
Merging now so it can be used in the production updates to build new Docker images.

PR #2010, opened 2023-07-13
CHROMEOS flash: Add retry to detect p3

nuclearcat:
```
++ losetup -P -f --show chromiumos_test_image.bin
+ dev=/dev/loop0
+ mount /dev/loop0p3 chromeos -o ro
mount: /root/chromeos: mount(2) system call failed: No such file or directory.
```
Workaround is to wait up to 10 sec until partition table get detected. Sometimes it takes a bit of time.


nuclearcat:
Tested during guybrush flashing

PR #2014, opened 2023-07-13
rootfs-images.yaml: use new bullseye-gst-fluster arm64 image

gctucker:
The bullseye-gst-fluster rootfs image for arm64 has been rebuilt to extend coverage.  Update the URL accordingly and by hand since rootfs automated updates are currently disabled.

### DIFF
--- a/config/core/rootfs-images.yaml
+++ b/config/core/rootfs-images.yaml
@@ -33,10 +33,10 @@ file_systems:
     type: debian
   debian_bullseye-gst-fluster_nfs:
     boot_protocol: tftp
-    nfs: bullseye-gst-fluster/20230623.0/{arch}/full.rootfs.tar.xz
+    nfs: bullseye-gst-fluster/20230717.0/{arch}/full.rootfs.tar.xz
     params: {}
     prompt: '/ #'
-    ramdisk: bullseye-gst-fluster/20230623.0/{arch}/initrd.cpio.gz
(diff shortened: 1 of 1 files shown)


gctucker:
There was a glitch in this rootfs image build apparently and the changes in `fluster_parser.py` from https://github.com/kernelci/kernelci-core/commit/beca7af875b64893b38a23e4460df922aa1a9841 didn't make it into the image:

```python
def _run_fluster(test_suite=None, timeout=None, jobs=None, decoders=None):
    cmd = ['python3', 'fluster.py', '-ne', 'run',
           '-f', 'junitxml', '-so', RESULTS_FILE]

    if test_suite:
        cmd.extend(['-ts', test_suite])
    if timeout:
        cmd.extend(['-t', timeout])
    if jobs:
        cmd.extend(['-j', jobs])
    if decoders:
        cmd.extend(['-d', decoders])

    subprocess.run(cmd, cwd=FLUSTER_PATH, check=False)
```

As such, this will need to wait until the next production update.

gctucker:
As it turns out, rootfs images are normally built every Friday with `kernelci/debos:kernelci` Docker images which are updated every Monday.  So of course, they don't include the latest changes.  I'm making a change in the script to update the Docker images for `debos` and `buildroot` with the `main` branch before doing the rootfs image builds on Fridays.  Also I'll use that to get a new build for gst-fluster on arm64 when that's done and update this PR accordingly.

PR #2015, opened 2023-07-14
Sandbox workaround + vanilla manifest

nuclearcat:
We add new features that can be enabled by environment variables for easier development and ChromiumOS uprev.


nuclearcat:
Tested manually

Issue #2008, opened 2023-07-12
Lot of job failure with kselftest on qemu-riscv on lab-baylibre


montjoie:
There is lot of qemu-riscv kselftest job failing to boot on lab-baylibre (ubuntu 20.04 running lava-docker).
The same job boot fine on lab-broonie (buster running lava-docker).

For debugging I ran the same job on lab-clabbe (gentoo running LAVA native) and it fail.
The same job on my LAVA test lab (gentoo running lava-docker) ran with success.

Adding smp:8 lead to a successful boot.
So I suspected some timing issue, and labs working have good cpu (ryzen, i3) and the ones that fail have less good ones.

Removing debug from kernel command line lead to a successful boot on lab-baylibre
On lab-clabbe (the one with the worst CPU) removing debug and verbose lead to success.

I think removing debug from kernel command line is the fast solution.
I think the problem live in systemd but this will be hard to debug.

PR #2017, opened 2023-07-17
qemu: kselftest: remove debug parameter

montjoie:
On kselftest jobs for riscV, the debug parameter lead to lot of systemd output and so jobs fail.
Anyway having systemd debug output is not necessry at all and perhaps should be removed for all jobs some day.

Closes: #2008

### DIFF
--- a/config/lava/kselftest/generic-qemu-ramdisk-kselftest-template.jinja2
+++ b/config/lava/kselftest/generic-qemu-ramdisk-kselftest-template.jinja2
@@ -7,5 +7,5 @@
 {% endblock %}
 
 {%- block image_arg %}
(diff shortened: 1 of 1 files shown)


khilman (approved):

gctucker (approved):
Tested OK on staging:
https://staging.kernelci.org/test/plan/id/64b73fba19ba5c7593b8750a/
https://storage.staging.kernelci.org/kernelci/staging-mainline/staging-mainline-20230719.0/arm64/defconfig/gcc-10/lab-broonie/kselftest-arm64_qemu-qemu_arm64-virt-gicv3.html#L246

PR #2018, opened 2023-07-17
CHROMEOS test-configs-chromeos.yaml: Update bullseye-cros-flash rootfs

nuclearcat:
As we have integrated several bugfixes, especially partition detection delay that fixes unreliable flashing issue, we need to switch to latest rootfs.


nuclearcat:
Tested by PR https://github.com/kernelci/kernelci-core/pull/2019

PR #2019, opened 2023-07-17
CHROMEOS flash: Improve mount reliability

nuclearcat:
Previous attempt to fix partition propagation delay were not fully reliable, still some flashing jobs, such as https://lava.collabora.dev/scheduler/job/11100799 had issue when /dev/loop0p3 available, but fails to mount. 
I tested new approach, where i literally retry to mount partition several times (with delay) until it succeed. 
Tested: https://lava.collabora.dev/scheduler/job/11101029


nuclearcat:
As written - tested in LAVA.

PR #2020, opened 2023-07-17
build-configs.yaml: add videodec to mainline, next, media

gctucker:
Add the videodec arm64 fragment again to mainline, next and media build configs so that the video decoder tests can be run.


gctucker (commented):
Tested OK on staging with mainline:
https://staging.kernelci.org/test/job/mainline/branch/master/kernel/v6.5-rc2/

PR #2022, opened 2023-07-18
config/lava: Add missing extra_kernel_args to NFS variants

nfraprado:
The depthcharge and uboot templates add extra_kernel_args to the context only on the ramdisk templates, but not in the NFS ones. Update the NFS templates so they also set the kernel arguments.

Thanks to @crazoes for noticing this issue.

Note that this only syncs the ramdisk and NFS so that they're the same (except for the `nfs/ramdisk` command) for all boot methods. The barebox, grub and ipxe templates don't set the kernel arguments on neither the ramdisk nor the nfs templates, so they're not touched.

### DIFF
--- a/config/lava/boot-nfs/generic-depthcharge-tftp-nfs-template.jinja2
+++ b/config/lava/boot-nfs/generic-depthcharge-tftp-nfs-template.jinja2
@@ -3,6 +3,9 @@
 {{ super() }}
 {% endblock %}
 {% block main %}
+{%- if extra_kernel_args %}
+{% do context.update({"extra_kernel_args": extra_kernel_args}) %}
+{% endif %}
 {{ super() }}
 {% endblock %}
 {% block actions %}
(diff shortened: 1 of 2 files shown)


gctucker (approved):
Tested OK on staging:
* U-Boot
  * https://staging.kernelci.org/test/plan/id/64b73aa7663ac0dcf7b875fd/
  * https://storage.staging.kernelci.org/kernelci/staging-mainline/staging-mainline-20230719.0/arm64/defconfig/gcc-10/lab-broonie/baseline-nfs-meson-gxl-s905x-libretech-cc.html#L780
* Depthcharge
  * https://staging.kernelci.org/test/plan/id/64b736601cfddc4ca5b87516/
  * https://storage.staging.kernelci.org/kernelci/staging-mainline/staging-mainline-20230719.0/x86_64/x86_64_defconfig+x86-chromebook/gcc-10/lab-collabora/baseline-nfs-acer-R721T-grunt.html#L1901

PR #2026, opened 2023-07-22
Storage fixups

gctucker:
Fix and improve a few things in `kernelci.config.storage` and `kernelci.storage`:
* fix docstrings
* fix YAML tags to match the class names
* convert Storage class to ABC
* add `._connect()` method
* allow `._upload()` method to return the URL mapping to override the default implementation

### DIFF
--- a/kernelci/config/storage.py
+++ b/kernelci/config/storage.py
@@ -11,6 +11,8 @@ from kernelci.config.base import YAMLConfigObject
 class Storage(YAMLConfigObject):
     """Base configuration class for Storage implementations"""
 
+    yaml_tag = '!Storage'
+
     def __init__(self, name, storage_type, base_url):
         """Storage configuration class
 
@@ -50,7 +52,7 @@ class BackendStorage(Storage):
     *api_url* is the URL to access the kernelci-backend API
     """
 
-    yaml_tag = '!Storage_backend'
+    yaml_tag = '!BackendStorage'
 
     def __init__(self, api_url, *args, **kwargs):
         super().__init__(*args, **kwargs)
@@ -77,7 +79,7 @@ class SSHStorage(Storage):
     *path* is the base destination path on the SSH server
     """
 
-    yaml_tag = '!Storage_ssh'
+    yaml_tag = '!SSHStorage'
 
     def __init__(self, host, *args, port=22, user='kernelci', path='~/data',
                  **kwargs):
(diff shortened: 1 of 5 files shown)


JenySadadia (changes requested):
Maybe you could fix the module docstring here as well https://github.com/kernelci/kernelci-core/blob/main/kernelci/cli/storage.py#L6

kernelci/storage/__init__.py
```diff
@@ -33,13 +34,23 @@ def credentials(self):
         """Credentials data"""
         return self._credentials
 
+    def _connect(self):
```
JenySadadia, line 37: This should also be declared as an abstract method.
```suggestion
   @abc.abstractmethod
    def _connect(self):
```

kernelci/config/storage.py
```diff
@@ -11,6 +11,8 @@
 class Storage(YAMLConfigObject):
     """Base configuration class for Storage implementations"""
 
+    yaml_tag = '!Storage'
```
JenySadadia, line 14: Where is `yaml_tag` being used?

gctucker: Every time a base class `Storage` config object get deserialised from YAML.  All the current storage types defined in the file have inherited classes but it's entirely possible to have a storage type that just relies on the base Storage config.

gctucker: See also: https://pyyaml.org/wiki/PyYAMLDocumentation

kernelci/storage/__init__.py
gctucker: No because it doesn't need to be implemented.

gctucker:
> Maybe you could fix the module docstring here as well https://github.com/kernelci/kernelci-core/blob/main/kernelci/cli/storage.py#L6

Right, that's a CLI tool issue rather than storage.  But yes as this PR is a bit of a catch-all one for any storage-related fixes so I'll add a commit for it.

kernelci/storage/__init__.py
JenySadadia: Yes, I saw that it is not implemented in `StorageBackend`.

gctucker:
@JenySadadia Please take another look, I believe I've addressed all your feedback.

JenySadadia (approved):
LGTM

PR #2027, opened 2023-07-22
Add Azure Files storage

gctucker:
Add Storage implementation for Azure Files.

Depends on #2026 
Fixes https://github.com/kernelci/kernelci-api/issues/308


gctucker:
This is quite slow in comparison with the plain Nginx downloads from the staging VM:
```
$ wget http://storage.staging.kernelci.org/api/linux-kernelci-staging-mainline-staging-mainline-20230727.0.tar.gz
# ...
linux-kernelci-staging-mai 100%[======================================>] 217.37M  9.17MB/s    in 41s     
```

gctucker:
Alright this works and it has been tested with the staging API instance.  However, this is creating a new share for each node rather than a directory inside a common share.  I'll update the PR to address that tomorrow.  This should also resolve the issue of not having a destination directory path which should be OK with a share (what it's been doing until now, basically).

Each share can have its own quota, so we can set up one share per user account accordingly.  Then we would have a "staging" share for the staging instance etc.

I've also manually added a rule to remove storage blobs 7 days after they were created, and the default quota is 5TB for the whole account.  We'll see how much this costs with the cumulative egress for builds and LAVA tests after about a week.

gctucker:
> Each share can have its own quota, so we can set up one share per user account accordingly. Then we would have a "staging" share for the staging instance etc.

That's been fixed now, all the files are uploaded under a `staging` share with folders like on a regular file system storage.  Here's a few snapshots from the Azure portal:
azure-files-001
azure-files-002

kernelci/storage/azure.py
```diff
@@ -0,0 +1,67 @@
...
+            if item['name'] == self.config.share:
+                share_found = item
+                break
+        if share_found is None:
+            share.create_share()
+        return share
+
+    def _get_directory(self, share, path):
+        directory = share.get_directory_client(directory_path=path)
+        if not directory.exists():
```
nuclearcat, line 46: I think it is a little bit confusing logic, in _get_directory, if directory doesn't exist - create it.
IMO, it is better to handle that in _upload, if _get_directory returned None, create directory, then use it.
Otherwise, if for example in future this class will be extended, for example to check existence of particular results (directory), it wont be possible to use _get_directory for such purpose.

nuclearcat (approved):

kernelci/storage/azure.py
gctucker: This method is just an implementation detail for AzureFiles storage, the bass class method is just `upload()` and that's expected to create the destination path.  If we want to add methods for exploring storage we can do that later, at the moment it's not a supported feature.

PR #2029, opened 2023-07-24
build-configs-chromeos.yaml: Remove ChromiumOS build configs from main

nuclearcat:
We dont need ChromiumOS config fragments in main branch, otherwise they are causing unnecessary builds and complexities in filters.


gctucker (approved):
Small thing, "config fragment" is usually understood as a kernel config fragment.  Here I think you're referring to the whole KernelCI YAML build configuration relating to ChromeOS.    Maybe just saying that the YAML file is being removed would avoid this ambiguity.

nuclearcat:
Adjusted commit message

gctucker:
OK I'll check again and merge this tomorrow morning once the prod update is done and this has been tested on staging.

gctucker:
Tested OK on staging and with `kci_build list_kernel_configs` to check there's no visible side-effects outside ChromeOS build configs.

gctucker:
https://staging.kernelci.org/build/kernelci/branch/staging-mainline/kernel/staging-mainline-20230727.0/

PR #2030, opened 2023-07-26
CHROMEOS Cleanups, fixes for build_board.sh

nuclearcat:
We have various issues accumulated over time, which this series will address.


config/rootfs/chromiumos/scripts/build_board.sh
```diff
@@ -115,13 +116,6 @@ case ${BOARD} in
     ;;
 esac
 
-# if environment variable SANDBOX_BUG is set, then we need to use sandbox workaround
```
a-wai, line 118: Isn't it still needed? In my experience the R116 fail without the sandbox workaround.

nuclearcat: As i know this sandbox issue and following ones was caused by lack of vm/max_map_count tuning. I got trogdor built without any sandbox workarounds after adjusting this value.
I updated my tool to adjust this value as well (as SDK cannot do it inside docker image).
Can you test that on your side too? (using latest version of kci-cros-builder.py tool)

config/rootfs/chromiumos/scripts/build_board.sh
```diff
@@ -25,6 +25,12 @@ function cleanup()
 
 trap cleanup EXIT
 
+if [ "$2" != "release-R114-15437.B" ] || [ -n "${VANILLA_MANIFEST}" ]; then
```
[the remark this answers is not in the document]
nuclearcat: It is actually good point, i am cooking better idea how to implement this, and something that @mgalka had in mind - to reduce complexity of this shell file, we can split it, and make version specific workarounds as separate shell script, and just verify if it does exist.
This will solve even more issues - we can keep old versions buildable by keeping old workaround files, and if file doesn't exist for new version, it will serve as reminder to verify and probably "port" existing workarounds.

config/rootfs/chromiumos/scripts/build_board.sh
a-wai: Ah right, I tested without the workaround and builds run fine now. So yes, this can go away.

gctucker:
Hello, could you please keep using `CHROMEOS` as a tag in the PR subjects (and git commits)?  That makes it easier to triage the PRs to review.  Of course it's possible to filter PRs based on the [base branch](https://github.com/kernelci/kernelci-core/pulls?q=is%3Apr+is%3Aopen+base%3Achromeos) but not in the GitHub notifications.

nuclearcat:
Tested by building octopus as old build, then vanilla on new manifest

PR #2033, opened 2023-07-26
TOML settings: use dotted keys

JenySadadia:
The older settings file `.conf` is using `":"` to define sub-section of components such as databases and runtimes e.g. `db:docker-host`. Use dotted keys for these components as it is a more standard way to define child sections in TOML. The syntax is `section_name.section_config` e.g. `db.docker-host`. In case the section configuration has `"."` (dot)
in the name, it needs to be surrounded by `""` (double-quotes) to parse it properly with TOML decoder e.g. `db."staging.kernelci.org"`

### DIFF
--- a/kernelci.toml.sample
+++ b/kernelci.toml.sample
@@ -5,8 +5,9 @@
 # are typical ones that shouldn't change in every command line call when used
 # manually for local development purposes.  Any command line argument can have
 # a default value stored in this file.  Components such as databases and labs
-# have their own section too, with a naming convention of the form ["db:db-name"]
-# or ["lab:lab-name"].
+# have their own section too, with a naming convention of the form [db.db-name]
+# or [lab.lab-name]. Make sure to surround sub-section name with "" (double-quotes)
+# if it contains "." (dot) in the name e.g [db."staging.kernelci.org"].
 
 [DEFAULT]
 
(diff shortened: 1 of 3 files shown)


kernelci.toml.sample
```diff
@@ -17,20 +18,23 @@
...
 #
-["db:localhost"]
+[db.localhost]
 #
 # db_token = "XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX"
 # api = "http://192.168.122.1:5001"
 # callback_id = "kernelci-callback-local"
 # callback_url = "http://localhost:5001"
 
+# [db."staging.kernelci.org"]
```
gctucker, line 35: I think we can also just drop the convention of having config entry names match the FQDN.  We could just have `staging` in the YAML config or any arbitrary string as long as there's no ambiguity.

JenySadadia: Yes, that works too. I'll keep this line in the sample though for the reference: `Make sure to surround sub-section name with "" (double-quotes) if it contains "." (dot) in the name e.g [db."staging.kernelci.org"].`

JenySadadia: I'll send a PR for the patch file to update staging settings file with dotted keys.

kernelci/cli/base.py
```diff
@@ -620,14 +620,20 @@ def get_from_section(self, section, option, as_list=False):
...
                 return None
             value = self._settings.get(section, option).split()
             if not as_list and len(value) == 1:
                 value = value[0]
         else:
             value = None
-            section_data = self._settings.get(section)
+            if isinstance(section, tuple):
+                section_data = self._settings.get(
+                    section[0], {}).get(section[1])
```
gctucker, line 634: I think we should be able to just stop using `get_from_section()` when the old settings format is dropped.  The code could just look up the sub-section itself, for example with `--api-config=something` it could get the `api` section from the settings and then look for the `something`  key in the application as the information is already there.  Until now, this used to be "serialised" as `section:subsection` but now we don't need that any more.

Until the old settings format is deprecated, I think it's going to be difficult to manage that.  But maybe the new code based on the `kci` command line tool and the new pipeline services could already require the settings file to be TOML and make use of the dotted-key sections natively.

gctucker (approved):
One comment about the fact that we should be able to simplify the code logic when the old settings file support is dropped.  Otherwise, LGTM.

kernelci/cli/base.py
JenySadadia: Yes, we would be able to drop the `get_from_section` method and move its whole logic to `Options.get` once the old settings are dropped.

gctucker (approved):

PR #2034, opened 2023-07-26
rootfs-configs-chromeos: add brya & puff

a-wai:
Add new rootfs type for `brya` & `puff` boards. Tested to build without any additional change.


nuclearcat:
Built by hand

PR #2035, opened 2023-07-26
build_board.sh: allow working from a kernelci-core fork

a-wai:
During development, it can be useful to run builds based on a fork of `kernelci-core`. This patch relies on the following environment variables:
* KCICORE_URL: the URL to the forked `kernelci-core` repo
* KCICORE_BRANCH: the name of the branch to fetch the manifest from

If those variables aren't set, it will use the `origin` remote URL and the name of the currently checked out branch. If either of those isn't available (no `origin` remote configured, detached state, not a git repo...) it falls back to using the upstream repo and `chromeos.kernelci.org` branch, as was previously the case.


nuclearcat:
Excellent idea, thank you! I will add you to staging users list now and will test it soon.

nuclearcat (approved):
Tested manually

PR #2036, opened 2023-07-27
build-configs.yaml: Drop broken riscv clang-11 config

gctucker:
The clang-11 toolchain doesn't support riscv but the mainline build config for riscv relies on the standard arch list which is also used for clang-16 and clang-17 which do support riscv.  As a result, some builds were being scheduled with clang-11 for riscv and since the clang-11 build environment config doesn't list riscv there's no translation to the Debian arch name riscv64 causing the build jobs to fail to pull the Docker image (and causing some congestion in Kubernetes as a side-effect).

Fix this by removing riscv from the clang-11 mainline build config.

Fixes: ddb1c84f466f ("build-configs: update riscv_clang_configs")

### DIFF
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -1151,7 +1151,41 @@ build_configs:
       # Minimum version
       clang-11:
         build_environment: clang-11
-        architectures: *arch_clang_configs
+        architectures:
+          arm64:
+            extra_configs:
+# Disabling to reduce load
+#            - 'allmodconfig'
+            - 'allnoconfig'
+            - 'defconfig+CONFIG_ARM64_64K_PAGES=y'
+          arm:
+            base_defconfig: 'multi_v7_defconfig'
+            filters:
+              - regex: {
+
+# Disabling to reduce load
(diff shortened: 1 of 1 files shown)


gctucker:
Tested OK by comparing the main branch and this PR:

```
$ ./kci_build list_kernel_configs --build-config=mainline --kdir=~/src/linux | sort
# ...
$ diff -u builds-1.txt builds-2.txt
--- builds-1.txt	2023-07-27 10:32:49.892581928 +0200
+++ builds-2.txt	2023-07-27 10:32:35.480678001 +0200
@@ -33,10 +33,8 @@
 i386 i386_defconfig gcc-10
 mips 32r2el_defconfig gcc-10
 mips allnoconfig gcc-10
-riscv allnoconfig clang-11
 riscv allnoconfig clang-16
 riscv allnoconfig gcc-10
-riscv defconfig clang-11
 riscv defconfig clang-16
 riscv defconfig gcc-10
 sparc allnoconfig gcc-10
```

nuclearcat (approved):
LGTM, this is going to resolve very big issue that blocks k8s clusters by creating jobs referencing non-existing image

PR #2037, opened 2023-07-27
config/runtime: kubernetes: use Azure Files SAS instead of SSH

gctucker:
Switch the storage credentials to the upload SAS token for Azure Files temporarily.  Ideally this should be done dynamically in the template.


gctucker:
As this is still hard-coded, switch to AzureFiles for all Kubernetes jobs for now.  The pipeline job which is run outside of the Kubernetes template can switch to SSH storage on staging to keep testing both methods.

PR #2038, opened 2023-07-28
CHROMEOS test-configs-chromeos.yaml: Update dtb filename in reference

nuclearcat:
Reference kernel in R114 doesn't use -r1 suffix as in newest kernels in upstream.


nuclearcat:
tested during flashing

PR #2039, opened 2023-07-29
Rust 1.71.1

ojeda:
Rust 1.71.0 will be the next version supported by the kernel [1].

Therefore, add it as a new build environment, update each config and finally remove the older build environment.

[1] https://lore.kernel.org/rust-for-linux/20230729220317.416771-1-ojeda@kernel.org/


gctucker:
I think we still need to resolve the missing `LLVM=1` issue causing Rust builds to fail before we can get valid results again, see the email thread about that. Hopefully I'll be able to prepare a fix this week then we could have Rust 1.71 in production next Monday.

ojeda:
Definitely -- I just replied to the list, sorry for the delay!

ojeda:
I have moved this PR from 1.71.0 to Rust 1.71.1 (i.e. patch version increase) since that is what we have ended up using in `rust-next`.

I also moved bindgen to 0.65.1 since we also upgraded that one, and it is required to fix an issue with LLVM 16.

nuclearcat:
https://staging.kernelci.org/build/id/64e89b696bbd9d9c35b8750c/
Tested on staging

ojeda:
Thanks Denys! I am sending the PR for the next upgrade so that it is prepared already.

ojeda:
Done: https://github.com/kernelci/kernelci-core/pull/2068.

PR #2041, opened 2023-08-01
kernelci/build.py: Add apply_patch_mbox method

yurinnick:
Fixes https://github.com/kernelci/kernelci-api/issues/307

Implement `apply_patch_mbox` method that downloads patch from mbox url into a temporary file and apply in to kernel tree

Test run logs:
```
kernelci-pipeline-notifier            | 07/31/2023 11:33:19 PM UTC [INFO] 2023-07-31 23:33:19.800845  HEAD          64c844bfe8ad32503e18af61  Running    -         checkout
kernelci-pipeline-tarball             | 07/31/2023 11:33:19 PM UTC [INFO] Updating repo for mainline
kernelci-pipeline-tarball             | From https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux
kernelci-pipeline-tarball             |  * branch                      HEAD       -> FETCH_HEAD
kernelci-pipeline-tarball             | From git://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux
kernelci-pipeline-tarball             |  * branch                      HEAD       -> FETCH_HEAD
kernelci-pipeline-tarball             | HEAD is now at 5d0c230f1de8 Linux 6.5-rc4
kernelci-pipeline-tarball             | 07/31/2023 11:33:27 PM UTC [INFO] Repo updated
kernelci-pipeline-tarball             | 07/31/2023 11:33:27 PM UTC [INFO] Applying patch to the repo, patch mbox url: https://patchwork.kernel.org/project/linux-hardening/patch/20230726231139.never.601-kees@kernel
.org/mbox/
kernelci-pipeline-tarball             | 07/31/2023 11:33:28 PM UTC [INFO] Making tarball linux-mainline-master-v6.5-rc4-1-g66f6b7deab6b.tar.gz
kernelci-pipeline-tarball             | 07/31/2023 11:33:28 PM UTC [INFO] set -e
kernelci-pipeline-tarball             | cd /home/kernelci/data/src/linux
kernelci-pipeline-tarball             | git archive --format=tar --prefix=linux-mainline-master-v6.5-rc4-1-g66f6b7deab6b/ HEAD | gzip > ../../output/linux-mainline-master-v6.5-rc4-1-g66f6b7deab6b.tar.gz
kernelci-pipeline-tarball             |
kernelci-pipeline-tarball             |
kernelci-pipeline-tarball             | 07/31/2023 11:33:55 PM UTC [INFO] Tarball created
```

### DIFF
--- a/kernelci/build.py
+++ b/kernelci/build.py
@@ -24,6 +24,7 @@ import platform
 import re
 import shutil
 import tarfile
+import tempfile
 import time
 import urllib.parse
 
@@ -319,6 +320,37 @@ retrying in {i} seconds')
             continue
 
     return False
+
+
+def apply_patch_mbox(
+    kdir,
+    mbox_url,
+    git_username="kernelci-tsc",
+    git_email="kernelci-tsc@groups.io"
+):
+    """Download patch mbox from URL and apply with 3-way merge
+
+    *kdir* is the path to a kernel source directory
+    *mbox_url* is the URL to patch mbox content
+    *git_username* is the username used to apply the patch
+    *git_email* is the email used to apply the patch
+    """
+    with tempfile.NamedTemporaryFile(prefix="kernel-patch-") as tmp_f:
+        if not _download_file(mbox_url, tmp_f.name):
+            raise FileNotFoundError(f"Error downloading patch mbox {mbox_url}")
+
+        shell_cmd("""
+set -e
+cd {kdir}
+git config user.name "{git_username}"
+git config user.email "{git_email}"
+git am --3way {mbox_file}
+""".format(
+            kdir=kdir,
+            mbox_file=tmp_f.name,
+            git_username=git_username,
+            git_email=git_email
+        ))
 
 
 def pull_tarball(kdir, url, dest_filename, retries, delete):


gctucker:
The issue right now is that `kernelci.build` is not used by the new API & Pipeline.  I think this is going a bit too fast, see my question on the issue https://github.com/kernelci/kernelci-api/issues/307 as the first part is still missing - namely, how to connect Patchwork with the API.

kernelci/build.py
```diff
@@ -321,6 +322,20 @@ def _download_file(url, dest_filename, chunk_size=1024):
     return False
 
 
+def apply_patch_mbox(path, mbox_url):
+    with tempfile.NamedTemporaryFile(prefix="kernel-patch-") as tmp_f:
```
gctucker, line 326: Please add a docstring.

kernelci/build.py
```diff
@@ -321,6 +322,20 @@ def _download_file(url, dest_filename, chunk_size=1024):
...
+        if not _download_file(mbox_url, tmp_f.name):
+            raise FileNotFoundError(f"Error downloading patch mbox {mbox_url}")
+
+        shell_cmd("""
+set -e
+cd {path}
+git config user.name "kernelci-tsc"
+git config user.email "kernelci-tsc@groups.io"
```
gctucker, line 334: I think we should remove this and let the user define it instead.

kernelci/build.py
```diff
@@ -26,6 +26,7 @@
 import tarfile
 import time
 import urllib.parse
+import tempfile
```
gctucker, line 29: Please keep imports in alphabetical order.

gctucker (commented):
I think we should be able to merge this one, just added some comments about a few minor things to tweak.  It's not used in the code base yet but it makes sense to have it as a utility method for the rest, and the way kernel builds are done will need to be reworked anyway for the new pipeline so this will all get consolidated again in the near future.

kernelci/build.py
yurinnick: Made it function args with default values

kernelci/build.py
```diff
@@ -321,6 +322,37 @@ def _download_file(url, dest_filename, chunk_size=1024):
...
+    """
+    with tempfile.NamedTemporaryFile(prefix="kernel-patch-") as tmp_f:
+        if not _download_file(mbox_url, tmp_f.name):
+            raise FileNotFoundError(f"Error downloading patch mbox {mbox_url}")
+
+        shell_cmd("""
+set -e
+cd {kdir}
+git config user.name "{git_username}"
+git config user.email "{git_email}"
```
gctucker, line 346: No I mean, I think we should actually remove these lines.  Say, if you're doing this over your working kernel source tree and you don't specify command line arguments with your current user name and email then it will replace it.  And then next time you make a commit in the kernel source tree it'll be with the `kernelci-tsc@groups.io` email.

So it should be left to the user to manage this outside of `kci_build`.

gctucker (changes requested):

kernelci/build.py
yurinnick: You mean it should be part of KernelCI config files or something? Sorry, I am not quite following what do you mean. Git won't allow to apply patches without `name` and `email` being set though.

gctucker: It should be part of the user's Git config, yes.  Or we could maybe have a separate command to set the user name and email explicitly if some CI systems need to do that, as a handy wrapper.

> Git won't allow to apply patches without name and email being set though.

Yes and that's fine.  If there's no user name and email configured then git will fail and the issue will be on the user to solve it.  Just like running any other invalid command (say, if `kdir` is not a Git repo etc.).

pawiecz:
@yurinnick Please correct me if I'm wrong: this PR is a dependency to kernelci/kernelci-pipeline#295 which has been closed.

Can this one be closed as well or is it still intended to be used somewhere? If it is being repurposed, could you point me where it's used (I wasn't able to locate any)?

yurinnick:
@pawiecz there was some back-n-forth on a best way to apply patches, I think it's safe to close for now and re-open if you'll need to get back to it.

PR #2042, opened 2023-08-01
Add python project gitignore

yurinnick:
Use default Github `.gitignore` template for python projects

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,153 @@
-*.pyc
-*~
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+*$py.class
+
+# C extensions
+*.so
+
+# Distribution / packaging
+.Python
+build/
+develop-eggs/
+dist/
+downloads/
+eggs/
+.eggs/
+lib/
+lib64/
+parts/
+sdist/
+var/
+wheels/
+share/python-wheels/
+*.egg-info/
+.installed.cfg
+*.egg
+MANIFEST
+
+# PyInstaller
+#  Usually these files are written by a python script from a template
+#  before PyInstaller builds the exe, so as to inject date/other infos into it.
+*.manifest
+*.spec
+
+# Installer logs
+pip-log.txt
+pip-delete-this-directory.txt
+
+# Unit test / coverage reports
+htmlcov/
+.tox/
+.nox/
+.coverage
+.coverage.*
+.cache
+nosetests.xml
+coverage.xml
+*.cover
+*.py,cover
+.hypothesis/
+.pytest_cache/
+cover/
+
+# Translations
+*.mo
+*.pot
+
+# Flask stuff:
+instance/
+.webassets-cache
+
+# Scrapy stuff:
+.scrapy
+
+# Sphinx documentation
+docs/_build/
+
+# PyBuilder
+.pybuilder/
+target/
+
+# Jupyter Notebook
+.ipynb_checkpoints
+
+# IPython
+profile_default/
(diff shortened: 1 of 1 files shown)


gctucker (commented):
Why would you do this?

yurinnick:
After a few day working with `kernelci-core` there are temporary python files created, that are not ignored:
```
$ git check-ignore **/*
build/bdist.linux-x86_64
build/lib
build/scripts-3.11
kernelci.egg-info/dependency_links.txt
kernelci.egg-info/PKG-INFO
kernelci.egg-info/requires.txt
kernelci.egg-info/SOURCES.txt
kernelci.egg-info/top_level.txt
kernelci/__pycache__
```

gctucker:
Ah yes, I guess that falls into the category of making this a proper Python package.

nuclearcat (approved):
Not sure if *.so is good here, but i think we should avoid using .so in project anyway, verified "check-ignore" at my side, LGTM.

gctucker:
> Not sure if *.so is good here, but i think we should avoid using .so in project anyway, verified "check-ignore" at my side, LGTM.

It's here in case we have any native Python modules implemented in C I guess.  And yes I don't see a use-case where we would keep executable binaries under source control in an LGPL project anyway.

.gitignore
```diff
@@ -1,2 +1,160 @@
...
+lib64/
+parts/
+sdist/
+var/
+wheels/
+share/python-wheels/
+*.egg-info/
+.installed.cfg
+*.egg
+MANIFEST
```
gctucker, line 27: Is this always meant to be an automatically-generated file?

.gitignore
```diff
@@ -1,2 +1,160 @@
...
+.hypothesis/
+.pytest_cache/
+cover/
+
+# Translations
+*.mo
+*.pot
+
+# Django stuff:
+*.log
```
gctucker, line 59: This one might bite us at some point in the future, for example if we wan to keep sample log files in the documentation.  Probably we can just remove the Django section.

.gitignore
```diff
@@ -1,2 +1,160 @@
...
+celerybeat-schedule
+celerybeat.pid
+
+# SageMath parsed files
+*.sage.py
+
+# Environments
+.env
+.venv
+env/
```
gctucker, line 125: In other projects there are `env` directories to keep sample environment files, or encrypted actual ones used in deployments.  While ignoring `.env` seems safe, I'm not sure `env/` is a good idea.

gctucker (commented):
A few things to consider.  The main risk here is to not have files added to PRs by mistake because of zealous gitignore settings.  Ideally, I think we should only be adding only entries for the files we positively don't want to have in git rather than blindly applying some default list.

.gitignore
yurinnick: Based on [python documentation](https://docs.python.org/3/distutils/sourcedist.html) - yes:

```
The manifest template is just a list of instructions for how to generate your manifest file, MANIFEST, which is the exact list of files to include in your source distribution. The sdist command processes this template and generates a manifest based on its instructions and what it finds in the filesystem.
```

yurinnick:
Remove Django section and `env` directory from `.gitignore`

.gitignore
gctucker: Ah yes thanks, I think I confused it with the manifest template.

gctucker (approved):

yurinnick:
@gctucker I don't have permissions to merge, so if it looks good, please merge it.

PR #2044, opened 2023-08-04
CHROMEOS tast_parser.py: Raise critical error if tast run fails

nuclearcat:
Often tast execution fail and parser quit silently, not even raising error, which leads to hidden failures that are caught too late. This patch fixes that.

### DIFF
--- a/config/docker/cros-sdk.jinja2
+++ b/config/docker/cros-sdk.jinja2
@@ -21,7 +21,11 @@ RUN apt-get update \
         ssh \
         wget
 
-RUN useradd -u 996 -ms /bin/sh cros
+# We have uid conflict with jenkins useradd, WORKAROUND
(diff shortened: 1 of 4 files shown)


nuclearcat:
Tested manually: https://staging.lava.collabora.dev/scheduler/job/514926
and on dashboard: https://storage.chromeos.kernelci.org/kernelci/chromeos-stable/chromeos-stable-20230808.0/arm64/cros---chromeos-6.1-arm64-chromiumos-qualcomm.flavour.config+arm64-chromebook+CONFIG_MODULE_COMPRESS_GZIP=n/clang-14/lab-collabora/cros-tast-mm-decode-v4l2-stateful-vp9-group1-sc7180-trogdor-lazor-limozeen_chromeos.html

PR #2045, opened 2023-08-07
docker.py: Return non-zero exit code on build failure

nuclearcat:
If Docker image build failed, we should return non-zero, so deploy scripts are aware about failure.
Also print some debug information about failure.

### DIFF
--- a/kernelci/cli/docker.py
+++ b/kernelci/cli/docker.py
@@ -120,13 +120,10 @@ class cmd_build(DockerBuildGenerateCommand):  # pylint: disable=invalid-name
     def _build_image(cls, dockerfile, tag, buildargs, nocache):
         client = docker.from_env()
         dockerfile_obj = io.BytesIO(dockerfile.encode())
-        try:
-            return client.images.build(
-                fileobj=dockerfile_obj, tag=tag, buildargs=buildargs,
-                nocache=nocache,
-            )
-        except docker.errors.BuildError as exc:
-            return exc.msg, exc.build_log
+        return client.images.build(
+            fileobj=dockerfile_obj, tag=tag, buildargs=buildargs,
+            nocache=nocache,
+        )
 
     @classmethod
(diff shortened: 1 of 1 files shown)


kernelci/cli/docker.py
```diff
@@ -170,9 +167,15 @@ def __call__(self, configs, args):
...
-        _, build_log = self._build_image(
-            dockerfile, name, buildargs, args.no_cache
-        )
+        try:
+            _, build_log = self._build_image(
+                dockerfile, name, buildargs, args.no_cache
+            )
+        except docker.errors.BuildError as exc:
+            if args.verbose:
+                self._dump_log(exc.build_log)
```
gctucker, line 176: The issue here is that the log is now only printed when the exception is caught.  Instead, the `build_log` variable should be populated here and dumped in one place e.g. a `finally` block.

gctucker (commented):
Just one small thing.

kernelci/cli/docker.py
gctucker: The `finally` block is run regardless of whether there was an exception or not, so no need to duplicate the code (well there should never be a need for duplicating code anyway to keep it [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)).

nuclearcat: Done, fixed. 
`build_log` exist only if exception didnt happen so i used else:

gctucker: Erm, that's not an ideal implementation but I think I'll merge and send a change with what I mean to avoid playing nuclearcat and mouse ;)

gctucker (approved):
OK that fixes the issue

kernelci/cli/docker.py
nuclearcat: Actually pylint hinted me it is not necessary, because we have return in case of exception, so we will not reach printing build_log anyway

PR #2048, opened 2023-08-08
Allow users to update their own password with `kci` tool

pawiecz:
This PR depends on changes from kernelci/kernelci-api#320.

Changes have been tested against local instance of API.

Two TODOs left in the command implementation might be blockers for merging this change.

### DIFF
--- a/kernelci/api/latest.py
+++ b/kernelci/api/latest.py
@@ -51,6 +51,18 @@ class LatestAPI(API):
     def password_hash(self, password: str) -> dict:
         return self._post('/hash', {'password': password}).json()
 
+    def change_password(self, username: str, current: str, new: str) -> dict:
+        return self._post(
+            '/password',
+            {
+                'current_password': {'password': current},
+                'new_password': {'password': new},
+            },
+            {
+                'username': username,
+            },
+        ).json()
+
     def create_token(self, username: str, password: str,
                      scopes: Optional[Sequence[str]] = None) -> str:
         data = {
--- a/kernelci/cli/user.py
+++ b/kernelci/cli/user.py
@@ -50,6 +50,20 @@ class cmd_password_hash(APICommand):  # pylint: disable=invalid-name
         return True
 
 
+class cmd_change_password(APICommand):  # pylint: disable=invalid-name
+    """Change a password for a given user"""
+    args = APICommand.args + [Args.username]
+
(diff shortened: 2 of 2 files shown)


pawiecz:
This PR adds features required for kernelci/kernelci-api#298 that were missing in kernelci/kernelci-api#320.

gctucker:
> Two TODOs left in the command implementation might be blockers for merging this change.

Yes, `pylint` won't like that.  But that's fine, we can work on it to get them resolved.

pawiecz:
TODOs resolved. Patch retested manually (I have to look into test automation/gating in this repo), aligned with messages from `passwd` I have on my machine.

kernelci/api/latest.py
```diff
@@ -51,6 +51,19 @@ def whoami(self) -> dict:
...
+        """Change a password for a given user"""
+        return self._post(
+            '/password',
+            {
+                'current_password': {'password': current},
+                'new_password': {'password': new},
+            },
+            {
+                'username': username,
+            },
```
gctucker, line 64: Now I'm wondering if having `/password/<username>` would be a better design for the API endpoint.  But that's just an implementation detail, wanted to share it here as a thought.

gctucker (commented):
I think it's missing the abstract method in the base API class, right?  I only see the concrete method in the `latest` version.

kernelci/api/latest.py
gctucker: (or `/<username>/password`)

JenySadadia: > (or `/<username>/password`)

+1 on this.

gctucker: I think the original idea was that the username wasn't required if the query was using an API token for authentication.  But since it requires the old password we don't rely on a token so the user endpoint makes more sense.

gctucker:
Alright we can merge this now, and add the abstract method in the base class as a follow-up.

gctucker (approved):

kernelci/api/latest.py
pawiecz: > > (or `/<username>/password`)
> 
> +1 on this.

Do you mean a new API path or should it be a part of `/user`, i.e. `/user/<username>/password`?

I can adjust it to also accept API token instead of only the old password.

PR #2049, opened 2023-08-09
CHROMEOS test-configs-chromeos.yaml: Disable -fixed tests

nuclearcat:
Temporary disable fixed tests, as they force devices to go into self_repair mode and wipe stateful partition.


nuclearcat:
Tested, visible on dashboard

PR #2050, opened 2023-08-09
Add missing abstract method for changing user's password

pawiecz:
This addresses concern raised in [#2048](https://github.com/kernelci/kernelci-core/pull/2048#pullrequestreview-1568153072).

Linter will not raise a missing docstring error if it's set in the abstract method.

### DIFF
--- a/kernelci/api/__init__.py
+++ b/kernelci/api/__init__.py
@@ -66,6 +66,10 @@ class API(abc.ABC):
         """Get an encryption hash for a given password"""
 
     @abc.abstractmethod
+    def change_password(self, username: str, current: str, new: str) -> dict:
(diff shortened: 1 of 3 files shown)


pawiecz:
Linter no longer complains about docstrings, but now there's `too-many-public-methods` (21>20).

Should this number be actually limited or this error be ignored (with a linter hint)?

gctucker:
Yes, as discussed today we can just disable this check with a "hint" comment in the code.  According to the pylint documentation, this is because classes with too many methods are likely to be doing too many things.  We might rework the code later following this guidance but it's definitely not a blocking issue right now.

gctucker:
Replaced with #2075

Issue #1999, opened 2023-07-11
Add documentation for TOML settings 


JenySadadia:
Since we have enabled TOML support for user settings, add related documentation on `https://kernelci.org/docs/core/settings/`.

- [ ] Update `doc/settings.md`

PR #2052, opened 2023-08-10
Add documentation for TOML settings 

JenySadadia:
Fixes https://github.com/kernelci/kernelci-core/issues/1999

### DIFF
--- a/doc/toml_settings.md
+++ b/doc/toml_settings.md
@@ -1,28 +1,21 @@
 ---
-title: "[Deprecated] User settings"
-date: 2021-08-05
+title: "User settings with TOML"
+date: 2023-08-10
 draft: false
 description: "User-defined local settings"
 weight: 2
 ---
 
-> **Warning**
-> This document describes user settings file format `.conf` which will soon be deprecated. Please use the latest settings file format i.e. `.toml`. More information on it can be found [here](../toml_settings).
-
-
 The user settings file is intended to be created by end-users with options
 specific to their local setup.  It can also be used by automated CI systems, in
-particular to hide secret API tokens.  The format is similar to
-[INI](https://en.wikipedia.org/wiki/INI_file) although it is parsed by the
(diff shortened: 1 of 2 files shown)


JenySadadia:
Updated staging https://static.staging.kernelci.org/docs/core/ for the review.

JenySadadia:
@gctucker Should I update `kci_*.md` documentation as well or wait until `.conf` gets deprecated?

gctucker:
> @gctucker Should I update `kci_*.md` documentation as well or wait until `.conf` gets deprecated?

I think it's not really necessary as the old `kci_*` tools should be deprecated at the same time as the old `.conf` settings format.  Also we might have different settings options for the old and new tools, so in practice we would basically have a new way of configuring things in TOML rather than just a translation of the old settings.

gctucker (approved):
Oh I don't think we should be deprecating the page about user settings.  Just the old format is being deprecated, the new TOML format is still user settings.

I think that's just a matter of organising things in the documentation structure.  We can probably merge it like this anyway as I would like to create a "legacy" section about the things that are going to be deprecated, and the old settings format would be going there anyway.  Just the TOML page would become the new "user settings" file.

gctucker:
Let's just wait for this to land on the staging static website to double check formatting etc. and I think it'll be good to get merged.

JenySadadia:
> Let's just wait for this to land on the staging static website to double check formatting etc. and I think it'll be good to get merged.

Pushed the changes on staging https://static.staging.kernelci.org/docs/core/.

gctucker:
> > Let's just wait for this to land on the staging static website to double check formatting etc. and I think it'll be good to get merged.
> 
> Pushed the changes on staging https://static.staging.kernelci.org/docs/core/.

Yes, just double-checked that.  In principle it should get automatically deployed with the rest of the staging update every 8h but I'm not sure if it's always working as intended.  In any case, that's all sorted now.

PR #2053, opened 2023-08-14
test-configs.yaml: Limit trees we test on FVP

broonie:
The FVP is extremely slow and we've got some quite CPU intensive tests
scheduled for it, including things like the KVM selftests which are very
CPU intensive and end up taking on the order of six hours to run.  Since
we only have a few donated FVPs we're routinely dropping jobs on the
floor, restrict to just running core upstream trees plus the arm64
architecture tree to give us more chance of making it through the jobs
we schedule.  I've included stable but it might need to be removed too
given that we tend to get a lot of stable updates together.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -1097,7 +1097,7 @@ device_types:
     boot_method: fvp
     filters:
       - passlist: {defconfig: ['defconfig']}
-
+      - regex: {'tree': '^(next|mainline|stable|arm64)$'}
 
   hi6220-hikey:
     mach: hisilicon


config/core/test-configs.yaml
```diff
@@ -1097,7 +1097,7 @@ device_types:
     boot_method: fvp
     filters:
       - passlist: {defconfig: ['defconfig']}
-
+      - regex: {'tree': '(next|mainline|stable|arm64)'}
```
gctucker, line 1100: Just to avoid matching other trees which contain these strings (e.g. `rt-stable`):
```suggestion
      - regex: {'tree': '^(next|mainline|stable|arm64)$'}
```

gctucker (commented):
LGTM - just a small thing to avoid unexpected corner cases.

broonie:
Are there any blockers to getting this merged?

gctucker:
Is there any particular reason for not adding the trailing `$`?

gctucker:
It matters because things like `stable-rc` would match this regex implicitly:
```
>>> re.match(r'^stable', 'stable-rc')
<re.Match object; span=(0, 6), match='stable'>
>>> re.match(r'^stable$', 'stable-rc')
```

gctucker (approved):

PR #2054, opened 2023-08-15
test-configs.yaml: Add start of line anchors to regexes

broonie:
While many examples of regex matches use a start of line anchor to
ensure there aren't trees matched unexpectedly there are quite a few
examples that appear to unintentionally omit them.  Update them to have
the anchors.


nuclearcat (approved):
Looks like staging works fine, let's hope on production will work too :)

PR #2056, opened 2023-08-17
CHROMEOS build_board.sh: Add step to remove orphan_files feature

nuclearcat:
As described in b/295364868 if this flag is enabled in mke2fs, and we test in KernelCI newest kernel (post-5.15), it will convert fs using this flag and then when we load pre-5.15,
on old kernel, it will not be able to mount filesystem, and as result initiate self_repair with wipe of stateful partition.


nuclearcat:
Tested manually, built octopus and coral boards.

PR #2057, opened 2023-08-17
CHROMEOS test-configs-chromeos.yaml: Run cros-tast-mm-decode on trogdor

laura-nao:
Enable the cros-tast-mm-decode test plan on trogdor Chromebooks.


nuclearcat (approved):
Results are available on dashboard: https://chromeos.kernelci.org/test/plan/id/64e4328861a72d89fd1508f8/

PR #2058, opened 2023-08-17
CHROMEOS Update ChromeOS base version to R116

a-wai:
This commit adds the manifest, rootfs configs and fixes scripts for the upgrade to R116.

Test configs will be updated once we have uploaded the new builds.


nuclearcat (approved):
Thanks, LGTM

PR #2062, opened 2023-08-22
clang-17.jinja2: Fix clang-17 repository

nuclearcat:
clang-17 repo name slightly changed, update it


nuclearcat:
Tested manually

PR #2063, opened 2023-08-22
CHROMEOS build-configs-chromeos.yaml: Switch to clang-17

nuclearcat:
Current ChromeOS release using clang-17 for kernel builds, it is better to be up to date on compiler and update it from obsolete clang-14.

### DIFF
--- a/config/core/build-configs-chromeos.yaml
+++ b/config/core/build-configs-chromeos.yaml
@@ -8,8 +8,8 @@ trees:
 
 
 chromeos_5.15_variants: &chromeos_5_15_variants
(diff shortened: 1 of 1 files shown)


nuclearcat:
Adding skip as docker image kernelci/cros-clang-17 need to be created in docker repository

nuclearcat:
Builds are completed using clang-17
https://chromeos.kernelci.org/build/kernelci/branch/linux-6.1.y-arm64-chromeos/kernel/linux-6.1.y-arm64-chromeos-20230825.0/

PR #2064, opened 2023-08-23
[WIP] Enable bpf-next tree build and test

yurinnick:
I'd like to enable `bpf-next` tree build and testing on KernelCI. 

BPF subsystem [requires](https://www.kernel.org/doc/readme/tools-testing-selftests-bpf-README.rst) cutting edge `llvm` (comes with clang) and `pahole` (added) to maximize passing tests. I added most of the parts required for build, not sure if I tie them together correctly though.

### Progress
- [x] Configure `bpf-next` tree checkout
- [x] Add llvm + pahole build image
- [x] Add `bpf-next` build config
- [x] Test `bpf-next` build configuration with `kci_build`
- [ ] Configure `bpf-next` kselftest
- [ ] Test `bpf-next` kselftest configuration with `kci_test`

### Testing

1. Build clang-17 + pahole image
```
$ ./kci docker build clang-17 pahole kselftest kernelci --prefix=yurinnick/
$ docker images
REPOSITORY                       TAG                         IMAGE ID       CREATED              SIZE
yurinnick/clang-17               pahole-kselftest-kernelci   50697711d09e   36 seconds ago       1.4GB
```

2. Add `kernelci.toml` config
```
$ cat kernelci.toml
...
[kci_build]
mirror = "bpf-next.git"
kdir = "linux-bpf"
output = "linux-bpf/build-x86"
build_env = "gcc-10"
build_config = "bpf-next"
arch = "x86_64"
j = 13
install = true
...
```

3. Following [kci_build docs](https://kernelci.org/docs/core/kci_build/) build `bpf-next` kernel
```
$ ./kci_build update_mirrot
$ ./kci_build update_repo
$ ./kci_build generate_fragments
$ ./kci_build init_bmeta
$ docker run -it -v $PWD:/home/kernelci yurinnick/clang-17:pahole-kselftest-kernelci /bin/bash
./kci_build make_config --defconfig=x86_64_defconfig+bpf
./kci_build fetch_firmware
./kci_build make_kernel
./kci_build make_modules
```

### DIFF
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -22,6 +22,9 @@ trees:
   arnd:
     url: "https://git.kernel.org/pub/scm/linux/kernel/git/arnd/playground.git"
 
+  bpf-next:
+    url: "https://git.kernel.org/pub/scm/linux/kernel/git/bpf/bpf-next.git"
+
   broonie-misc:
     url: "https://git.kernel.org/pub/scm/linux/kernel/git/broonie/misc.git"
 
@@ -144,7 +147,6 @@ trees:
 
   weiny2:
     url: "https://github.com/weiny2/linux-kernel"
-
 
 fragments:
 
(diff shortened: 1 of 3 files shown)


config/core/build-configs.yaml
```diff
@@ -983,6 +1019,11 @@ build_configs:
     branch: 'to-build'
     variants: *minimal_variants
 
+  bpf-next:
+    tree: bpf-next
+    branch: 'for-next'
+    variants: *bpf_variants
```
gctucker, line 1025: It might be better to directly declare the variants here rather than elsewhere in the file with an anchor as it's only used with the BPF tree.

config/core/test-configs.yaml
```diff
@@ -285,6 +285,12 @@ test_plans:
       kselftest_collections: "arm64"
     filters: *kselftest_no_fragment
 
+  kselftest-bpf:
+    <<: *kselftest
+    params:
+      job_timeout: '10'
+      kselftest_collections: "bpf"
```
gctucker, line 292: I don't think we can actually do this because, like you mentioned, BPF requires an unreasonably cutting-edge toolchain.  It's actually one of the reasons why it got disabled by default:
```
commit 7a6eb7c34a78498742b5f82543b7a68c1c443329 (tag: linux-kselftest-fixes-5.11-rc4)
Author: Mark Brown <broonie@kernel.org>
Date:   Thu Dec 10 18:52:33 2020 +0000

    selftests: Skip BPF seftests by default
```

So a prerequisite before being able to run this test automatically would be to have a dedicated job to build a custom toolchain Docker image.  This sounds like something to do with the new API, at some point in 2024 I guess.

gctucker (commented):
I believe we're going to face practical issues with this due to the LLVM version requirements.  Does the `bpf` fragment already require this or can we use a standard LLVM 17 toolchain?

nuclearcat:
Hi, if we need to enable bpf-tree, we need to do it in kernelci-pipeline repo, as we migrated to new system. Feel free to message here if there is still interest in that.

PR #2065, opened 2023-08-23
CHROMEOS build_board.sh: Reorganize USE flag and add kernelci

nuclearcat:
Reorganize USE flags for better readability, and add chromeos_kernelci_builder flag that will enable some tast tests specific for KernelCI.

Fixes "missing SoftwareDeps: chromeos_kernelci"


nuclearcat:
Adding temporary skip, cause i dont want to affect R116 build process and to have inconsistent/different images for platforms.

PR #2066, opened 2023-08-24
CHROMEOS test-configs-chromeos.yaml: Add more tast tests

nuclearcat:
We discovered more useful tast tests that can be used for KernelCI.


nuclearcat (commented):
Tests are visible on dashboard

PR #2070, opened 2023-08-28
CHROMEOS test-configs-chromeos.yaml: Update build images to R116

nuclearcat:
As we prepared everything and built images, it's time to switch to R116.


nuclearcat:
Tested and available on dashboard

PR #2071, opened 2023-08-28
config/lava/boot: add context in generic-qemu-boot-template.jinja2

lufy90:
I'm trying to submit job to my local lava lab, met: Job error: Invalid job data: ["2.1.1 execute-qemu: Missing 'arch' in job context"]

企业微信截图_16932170361436

I think arch is needed in job template.

### DIFF
--- a/config/lava/boot/generic-qemu-boot-template.jinja2
+++ b/config/lava/boot/generic-qemu-boot-template.jinja2
@@ -31,6 +31,12 @@
 {% endblock %}
 {% block actions %}
 
+{% if context %}
+context:
+{% for key, value in context.items() %}  {{ key }}: {{ value }}
+{% endfor %}
+{% endif %}
+
 actions:
 - deploy:
     timeout:


broonie:
We have a *lot* of qemu jobs running which do indeed specify the architecture, eg:

https://lava.sirena.org.uk/scheduler/job/75636

- probably there's something else that needs fixing here, there's other paramters specified by context that we need to go through?

lufy90:
> We have a _lot_ of qemu jobs running which do indeed specify the architecture, eg:
> 
> https://lava.sirena.org.uk/scheduler/job/75636
> 
> * probably there's something else that needs fixing here, there's other paramters specified by context that we need to go through?

Thanks very much for guiding. Context should be specified in device_types, right?

broonie:
Broadly, yes - there may be some things that can be usefully set there which apply everywhere but I can't think of them right now (IIRC you can actually set arch: all the time, it just only does something for qemu since obviously most platforms can't change their arch).

lufy90:
How long would it take to merge this PR ?

gctucker (changes requested):
The "Change add context in qemu job template." commit doesn't have any message to describe what this change is for.  Also the commit subject should follow the convention, please take a look a the git log for some examples.

lufy90:
> The "Change add context in qemu job template." commit doesn't have any message to describe what this change is for. Also the commit subject should follow the convention, please take a look a the git log for some examples.
Thanks for pointing out. I fixed my commits by git rebase.

config/lava/boot/generic-qemu-boot-template.jinja2
```diff
@@ -31,6 +31,12 @@
 {% endblock %}
 {% block actions %}
 
+{% if context %}
+context:
+{% for key, value in context.items() %}  {{ key }}: {{ value }}
+{% endfor %}
+{% endif %}
+
```
gctucker, line 39: But this is already done in the base template:
https://github.com/kernelci/kernelci-core/blob/main/config/lava/base/kernel-ci-base.jinja2#L36

And this template inherits from it.  Could you please compare the job definitions generated with and without this patch?

As far as I can tell, the context is already included as expected in QEMU jobs:
https://lava.collabora.dev/scheduler/job/11436140/definition#defline75

gctucker (changes requested):

gctucker:
I think the issue is somewhere else, probably something in the YAML config or command line used to generate the job.

config/lava/boot/generic-qemu-boot-template.jinja2
lufy90: Eh, this is true. I tried again, and read `generic-qemu-boot-template.jinja2`. It's my yaml file problem. Sorry for your time.

lufy90:
This is a invalid PR. The issue is caused by lack of context in my yaml file at the beginning.

PR #2072, opened 2023-08-29
Configure pre-commit validation

yurinnick:
About pre-commit: [pre-commit.com](https://pre-commit.com/)

In the current configuration `pre-commit` runs:
- pre-commit hooks:
      - `check-executables-have-shebangs`
      - `check-merge-conflict`
      - `check-yaml`
      - `check-toml`
      - `end-of-file-fixer`
      - `trailing-whitespace`
- `black` formatter

```
$ pre-commit run --all-files
check that executables have shebangs.....................................Passed
check for merge conflicts................................................Passed
check yaml...............................................................Passed
check toml...............................................................Passed
fix end of files.........................................................Passed
trim trailing whitespace.................................................Passed
black....................................................................Passed
```


yurinnick:
We can move some validations from `Makefile` into `pre-commit` and run it using [pre-commit action](https://github.com/pre-commit/action)

gctucker (changes requested):
> * `black` formatter

Please don't do that, we considered it a while ago and decided to stick to `pycodestyle` instead.  Reformatting the whole codebase basically prevents everybody else from making changes, and causes a lot of churn which breaks the git history.  Could you please rewrite this PR without all the reformatting?

gctucker:
About pre-commit, the current approach has been to run `make test` to run some checks locally.  What's the pros / cons of `pre-commit` vs the classic `make test` approach?

yurinnick:
While `make` simply runs commands, `pre-commit` also takes care of installing requested tools and provides interface to configure them. Also it comes with a few useful checks like [these](https://github.com/kernelci/kernelci-core/pull/2072/files#diff-63a9c44a44acf85fea213a857769990937107cf072831e1a26808cfde9d096b9R4-R11). 

Bonus points, there is [pre-commit Github action](https://github.com/marketplace/actions/pre-commit) and there's an option to install `pre-commit` script and a hook for `git commit`: [pre-commit-install-hooks](https://pre-commit.com/#pre-commit-install-hooks)

gctucker:
Adding skip label for now just for clarity, even though this obviously can't be tested on staging since it conflicts with basically everything.

gctucker:
@yurinnick Could you please split this PR into the several things you want to propose, and drop `black` reformatting as that's not practical?

PR #2076, opened 2023-08-30
Add support for creating new API users

gctucker:
Add an API method and `kci user add` command to create new API users.

### DIFF
--- a/kernelci/api/latest.py
+++ b/kernelci/api/latest.py
@@ -22,7 +22,7 @@ class NodeStates(enum.Enum):
     DONE = 'done'
 
 
-class LatestAPI(API):
+class LatestAPI(API):  # pylint: disable=too-many-public-methods
     """Latest API version
 
     The 'latest' version is used to refer to the current development version,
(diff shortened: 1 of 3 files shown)


kernelci/api/latest.py
```diff
@@ -167,6 +167,15 @@ def get_user_profiles(
         return self._get_api_objs(params=params, path='users/profile',
                                   limit=limit, offset=offset)
 
+    def create_user(self, username: str, password: str, profile: dict) -> dict:
+        data = {
+            'password': password,
+        }
+        params = {
+            'email': profile['email'],
+        }
```
JenySadadia, line 176: This should also have a provision to add the user to a group.

gctucker: Yes, however I think we're missing the part about creating groups and listing groups etc.  So maybe this can be ignored for now as we need to have this feature merged today to be able to create user accounts for the early access phase next week.

gctucker: OK I've added an issue to keep track of that: #2084

JenySadadia: We can use `kci user find_groups` to list available groups.
But yes we need to add a command to create a user group before we can add users to new groups.
Thanks for creating the issue.

PR #2078, opened 2023-08-31
test-configs.yaml: Move decoder tests for trogdor to separate plan

laura-nao:
The decoder conformance tests require the chromebook-decoders-filter, specifically defined for these tests only. Move the tests to a dedicated test plan. This re-enables baseline tests for trogdor Chromebooks on the stable kernel tree.


nuclearcat (approved):
Tested on staging: https://staging.kernelci.org/test/plan/id/6557465915fd15b52fa3727f/

PR #2080, opened 2023-08-31
CHROMEOS packagefix-release-R116-15509.B.sh: Add tpm workarounds

nuclearcat:
During the testing we noticed that some boards still need TPM flag to be enabled manually, to work properly.


nuclearcat:
Builds are tested by hand

PR #2085, opened 2023-09-01
CHROMEOS test-configs-chromeos.yaml: Remove missing tests

nuclearcat:
Some tests while shown in list - doesn't exist in reality. Remove them, so they dont break testing


nuclearcat (commented):
mm-misc working again https://chromeos.kernelci.org/test/job/kernelci/branch/chromeos-stable/kernel/chromeos-stable-20230904.0/plan/cros-tast-mm-misc/

PR #2090, opened 2023-09-04
kernelci.cli: kci: fix default value for --indent

gctucker:
When not providing a value for --indent when a command returns some JSON data, do not indent at all rather than indenting with 0 spaces.

Fixes: de31ee73e926 ("kernelci.cli.base: add ._print_json() helper method")


nuclearcat (approved):
LGTM

PR #2094, opened 2023-09-05
CHROMEOS packagefix-release-R116-15509.B.sh: Add puff tpm2 workaround

nuclearcat:
According to test results and SDK files this board also missing tpm2 flag, so we need to activate workaround.


nuclearcat:
Flashed and tested

PR #2095, opened 2023-09-07
kernelci.cli:kci:catch JSONDecodeError exception

lufy90:
Reduce error message when `kci` try to request a invalid url, by catch and ignore JSONDecodeError exception.
I'm using a dockerd kernelci-backend, version is 2022.2.0 .

Here is the **new** output with catch and pass `JSONDecodeError`:

企业微信截图_16940795251081

Compared with the **old** output with `JSONDecodeError` raised, I think this is useless message and may not very friendly to user:

企业微信截图_16940794794198

### DIFF
--- a/kernelci/cli/base.py
+++ b/kernelci/cli/base.py
@@ -18,7 +18,7 @@ import os.path
 import sys
 import toml
 
-from requests.exceptions import HTTPError
+from requests.exceptions import HTTPError, JSONDecodeError
 
 import kernelci.config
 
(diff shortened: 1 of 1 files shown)


nuclearcat:
As i understand it will be replaced by https://github.com/kernelci/kernelci-core/pull/2234 and will cover all cli tools

gctucker:
@lufy90 The `kci` tool is not meant to be used with `kernelci-backend` but with the new `kernelci-api`.

PR #2097, opened 2023-09-12
CHROMEOS test-configs-chromeos.yaml: Update tast tests

nuclearcat:
After viewing diff of R114->R116 tests i found similar to existing tests might be added, or some that doesn't take significant lab time.


nuclearcat:
Tests are available on dashboard and very useful

PR #2098, opened 2023-09-13
CHROMEOS rootfs-configs-chromeos.yaml: Update cros-flash to bookworm

nuclearcat:
First candidate to upgrade to bookworm is cros-flash, as on ChromeOS sometimes used ext4 features that doesn't exist in bullseye ext4 tools.


nuclearcat:
Tested, working, published on storage.

PR #2099, opened 2023-09-13
CHROMEOS test-configs-chromeos.yaml: Update x86 kernel

nuclearcat:
We need better support of onboard ethernet, we can use one of KernelCI built kernels for that.


nuclearcat:
Tested for example at https://storage.chromeos.kernelci.org/kernelci/chromeos-stable/chromeos-stable-20230918.0/x86_64/cros---chromeos-6.1-x86_64-chromeos-intel-pineview.flavour.config+x86-chromebook+CONFIG_MODULE_COMPRESS_GZIP=n/clang-17/lab-collabora/cros-tast-kernel-acer-cb317-1h-c3z6-dedede_chromeos.html

PR #2100, opened 2023-09-13
Tweak format of `kci show` and avoid unnecessary API queries

gctucker:
Add underlines for each job (or group of results) and only show the node names rather than the full path to improve readability.  Also avoid doing queries to get child nodes when the maximum recursion depth has been reached.

image


JenySadadia (approved):
Tested OK locally.

PR #2101, opened 2023-09-15
CHROMEOS test-configs-chromeos.yaml: Update bullseye to bookworm in flash rootfs

nuclearcat:
Flash rootfs is used to install modules and for ChromeOS imaging. Let's try newer distro, as also we need newer ext4 tools.


nuclearcat (commented):
Tested by flashing few chromebooks

PR #2102, opened 2023-09-15
Revert "test-configs.yaml: block minnowboard-turbot in lab-collabora"

gctucker:
This reverts commit 642c5d3fb715e57cad6816c874cd587645263c0c.

Enable minnowboard-turbot again in lab-collabora as other projects have stopped running tests on it so it has more capacity again.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -1583,8 +1583,6 @@ device_types:
     mach: x86
     arch: x86_64
     boot_method: grub
-    filters:
-      - blocklist: {lab: ['lab-collabora']}
 
   mt8173-elm-hana:
     mach: mediatek


gctucker:
@pawiecz ^ Let's see if we get jobs on staging today.

pawiecz:
LGTM

Jobs sent so far were not completed because of too short timeout (5m) for boot stage which takes a bit more (~5:30-6m in TestJobs completed successfully) - I'll look into tuning this.

gctucker:
Right, here's a basic job: https://lava.collabora.dev/scheduler/job/11533870

gctucker:
Yes we might be able to just increase the boot timeout which is currently 5min but the whole job timeout is 10min.  So in cases where it takes 9min to boot and 30s to run the test, we wouldn't be running longer jobs but allow more platforms to boot.  So maybe we can change the boot timeout to 8min or even just set all timeouts to 10min (except the tests which should take less than 1min).

There's a known issue with Minnowboard, the firmware gets stuck in a weird state zero-ing all the RAM very slowly sometimes.  I think this is what we're seeing here.  There's a firmware option that can be disabled to bypass this I believe and then it's pretty fast to boot.

gctucker:
I've resubmitted it with a 10min timeout and that "worked":
https://lava.collabora.dev/scheduler/job/11534452

So I guess we could just do this for now.  At least we can set the 10min timeout in the new pipeline LAVA job template.

This job was with a staging kernel build job, it would be worth trying again with artifacts from the new API.

gctucker:
@pawiecz Actually, we could also just add it to the pipeline YAML config used with the new API and update the timeout in the new base LAVA template.  That way, we wouldn't have to change the system-wide boot timeout for all the jobs currently being run in the legacy system.

pawiecz:
> This job was with a staging kernel build job, it would be worth trying again with artifacts from the new API.

Sure thing, I'll also try submitting modified TestJob definition

gctucker:
OK let's close this and deal with it in the new pipeline instead.  Then we can add a `x86-boards` fragment to enable boards that require a few extra kernel configs such as [aaeon-UPN-EHLX4RE-A10-0864](https://lava.collabora.dev/scheduler/device_type/aaeon-UPN-EHLX4RE-A10-0864).

PR #2105, opened 2023-09-15
kernelci.cli: storage: always print the artifact URL

gctucker:
Always print the URL of the uploaded artifact as otherwise there may not be any easy way to find it again later.

Fixes: c514b5884e44 ("kernelci.cli.storage: add command to upload files etc.")


JenySadadia (changes requested):
`Args.verbose` can be dropped from the list of optional arguments now.

gctucker:
Yes. I guess I was thinking we might use it for other things such as upload progress but can always deal with later. So I'll drop it.

JenySadadia (approved):

PR #2107, opened 2023-09-15
CHROMEOS rootfs: backport griphics hwprobe tool fix

10ne1:
This backports the commit from CL 4863202 into our fork of platform/graphics, in preparation for the R118 uprev, so we don't forget about this patch.

The CL has landed in R119, so we can drop this post R119.


nuclearcat (approved):
LGTM

PR #2108, opened 2023-09-18
Revert "test-configs.yaml: reduce coverage to bare minimal"

gctucker:
This reverts commit 491a3dba7b3b2c433fc1f6518600ac0eb33492ab.

Enable all the test configurations again to measure the added load and related costs in the infrastructure.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -2288,26 +2288,22 @@ test_configs:
   - device_type: acer-cb317-1h-c3z6-dedede
     test_plans:
       - baseline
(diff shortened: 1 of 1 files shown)


gctucker:
The number of results for the linux-next staging jobs roughly doubled since this was enabled again:
https://staging.kernelci.org/job/kernelci/branch/staging-next/

As this is going back to what was used before, I think it's OK to have it deployed in production now.  This week we can compare the amount of I/O with previous weeks to measure the impact of restoring the functional test coverage.

PR #2114, opened 2023-09-21
Rename x86-chromebook to x86-board and make it more generic

pawiecz:
As per discussion in #1893 this PR renames the x86-chromebook fragment and reduces it to use generic x86 options to allow booting boards like [aaeon-UPN-EHLX4RE-A10-0864](https://lava.collabora.dev/scheduler/device_type/aaeon-UPN-EHLX4RE-A10-0864).


pawiecz:
Latest patch set version fixes DCO (missing `Signed-off-by:` in one of the patches).

pawiecz:
This patch set is probably still missing adding `build-config-chromeos.yaml` to some `CONFIG_LIST` (to enable those build config fragments) but I wasn't sure where to put it.

pawiecz:
Latest patch set version fixes links to related PR (typo; redirected to another PR).

gctucker:
OK the full linux-next weekly build on staging is not re-enabled yet, and in any case I think we need to update the staging config to use the new `-boards` fragments instead of the `-chromebook` ones.

gctucker:
Thanks for tackling this.  I think the part about adding back the `x86-chromebook` and `arm64-chromebook` fragments should be left out of this PR and merged only on the `chromeos` branch where the `build-configs-chromeos.yaml` file already exists.

gctucker:
The proposed idea in https://github.com/kernelci/kernelci-core/pull/1893#issuecomment-1519601942 was to just rename `x86-chromebooks` as `x86-boards` in the YAML file and all references to it.  Then some follow-up changes can be made to remove any configs that are only useful for the `chromeos` use-case.  Then the `chromeos` branch can be rebased on top of that to define its own `x86-chromebook` fragment as a plain copy of what was before on the `main` branch, and from this point onwards they can both evolve separately.  To make things simpler, we can have a first PR just for x86 and another one after that just for arm64.  I believe that's the best way forward.

pawiecz:
@gtucker Thank you for clarifying this for me! This PR will rename the `x86-chromebook` fragment and reduce it to use generic x86 options.

> I think the part about adding back the x86-chromebook and arm64-chromebook fragments should be left out of this PR and merged only on the chromeos branch where the build-configs-chromeos.yaml file already exists.

I dropped that patch from this PR.

> I think we need to update the staging config to use the new -boards fragments instead of the -chromebook ones.

I failed to find relevant configuration option. Code search returned only https://github.com/kernelci/kernelci-deploy/blob/2c0b380e6e7a39efc91254b90f10868e2759d059/patches/kernelci-core/staging.kernelci.org/0001-STAGING-add-build-configs-staging.yaml.patch#L57 which might not be the config I should modify but I'm not 100% sure.

pawiecz:
References to "x86-chromebooks" in kcidb and kernelci-project repositories were used only as a test/example data.

nuclearcat:
It looks like it is breaking staging (due staging patches in kernelci-deploy), you can verify it by:
./kci_test validate

```
  File "/root/kci-rootfs/kernelci-core/./kci_test", line 405, in <module>
    configs = kernelci.config.load(opts.get_yaml_configs())
  File "/root/kci-rootfs/kernelci-core/kernelci/config/__init__.py", line 199, in load
    return load_data(data)
  File "/root/kci-rootfs/kernelci-core/kernelci/config/__init__.py", line 174, in load_data
    config.update(kernelci.config.build.from_yaml(data, filters))
  File "/root/kci-rootfs/kernelci-core/kernelci/config/build.py", line 462, in from_yaml
    build_configs = {
  File "/root/kci-rootfs/kernelci-core/kernelci/config/build.py", line 463, in <dictcomp>
    name: BuildConfig.load_from_yaml(
  File "/root/kci-rootfs/kernelci-core/kernelci/config/build.py", line 399, in load_from_yaml
    variants = [
  File "/root/kci-rootfs/kernelci-core/kernelci/config/build.py", line 400, in <listcomp>
    BuildVariant.load_from_yaml(variant, name, fragments, b_envs)
  File "/root/kci-rootfs/kernelci-core/kernelci/config/build.py", line 315, in load_from_yaml
    kw['architectures'] = list(
  File "/root/kci-rootfs/kernelci-core/kernelci/config/build.py", line 316, in <genexpr>
    Architecture.load_from_yaml(data or {}, name, fragments)
  File "/root/kci-rootfs/kernelci-core/kernelci/config/build.py", line 179, in load_from_yaml
    kw['fragments'] = [fragments[name] for name in cf] if cf else None
  File "/root/kci-rootfs/kernelci-core/kernelci/config/build.py", line 179, in <listcomp>
    kw['fragments'] = [fragments[name] for name in cf] if cf else None
KeyError: 'x86-chromebook'
```
I am setting temporary skip flag until this is updated on deploy.

config/core/test-configs-chromeos.yaml
```diff
@@ -13,14 +13,14 @@ device_types:
...
+    filters: &x86-board-filters
+      - passlist: {defconfig: ['x86-board']}
     params: &octopus-params
       cros_flash_initrd: 'https://storage.staging.kernelci.org/images/rootfs/debian/bullseye-cros-flash/20220523.1/amd64/initrd.cpio.gz'
       cros_flash_nfs: 'https://storage.staging.kernelci.org/images/rootfs/debian/bullseye-cros-flash/20220523.1/amd64/full.rootfs.tar.xz'
       cros_flash_image: 'https://storage.chromeos.kernelci.org/images/rootfs/chromeos/chromiumos-octopus/20220520.0/amd64/cros-flash.tar.gz'
-      cros_flash_kernel: 'http://storage.chromeos.kernelci.org/images/kernel/v5.10.112-x86-chromebook/kernel/bzImage'
-      cros_flash_modules: 'http://storage.chromeos.kernelci.org/images/kernel/v5.10.112-x86-chromebook/modules.tar.xz'
+      cros_flash_kernel: 'http://storage.chromeos.kernelci.org/images/kernel/v5.10.112-x86-board/kernel/bzImage'
+      cros_flash_modules: 'http://storage.chromeos.kernelci.org/images/kernel/v5.10.112-x86-board/modules.tar.xz'
```
nuclearcat, line 23: I think we need to just empty this file or even delete it, it is not relevant to main instance, as we dont run tast tests.
@gctucker what you think about this?

config/core/build-configs.yaml
```diff
@@ -506,134 +506,12 @@ fragments:
...
-      amdgpu/stoney_pfp.bin
-      amdgpu/stoney_rlc.bin
-      amdgpu/stoney_sdma.bin
-      amdgpu/stoney_uvd.bin
-      amdgpu/stoney_vce.bin
-      i915/glk_dmc_ver1_04.bin
-      i915/kbl_dmc_ver1_04.bin
-      rtl_nic/rtl8153a-4.fw
-      rtl_nic/rtl8153b-2.fw
-      "'
```
nuclearcat, line 515: This change will break bootability of all chromebooks on mainline tests.
For example GPU tests (non-tast) need firmware files, etc.

config/core/test-configs-chromeos.yaml
nuclearcat: Or maybe simpler, just dont change only these parameters, and i will deal with this file in separate commit

config/core/build-configs.yaml
nuclearcat: Maybe better to do this commit as follow-up, with separate testing on staging, making sure if removed options affecting or not chromebooks test results.

config/core/test-configs-chromeos.yaml
pawiecz: Change reverted in the new patch set, thanks for catching it early!

config/core/build-configs.yaml
pawiecz: ACK; new patch set only extends the fragment to support generic x86 boards as well

nuclearcat:
Kernels being built: https://storage.staging.kernelci.org/kernelci/staging-mainline/staging-mainline-20231018.0/x86_64/x86_64_defconfig%2Bx86-board/gcc-10/
and tested: https://staging.kernelci.org/test/job/kernelci/branch/staging-mainline/kernel/staging-mainline-20231018.0/plan/baseline/

PR #2115, opened 2023-09-22
cli/base.py: Implement KCI_SETTINGS env variable handling

nuclearcat:
By default we can specify config file by --api-config, if not, kci will search config at predefined places.
Now we can also set config file path by environment variable KCI_SETTINGS.

### DIFF
--- a/kernelci/cli/base.py
+++ b/kernelci/cli/base.py
@@ -554,6 +554,15 @@ class Options:
                     path = default_path
                     break
 
+        # If path is set, means it is either retrieved from KCI_SETTINGS
+        # environment variable or set by --settings option. In both cases,
+        # check if the file exists. If not, raise FileNotFoundError, so we are
+        # aware file is missing.
+        # If path is not set, this means "default" - it is permitted that
+        # settings file does not exist, it will appear as empty dict.
+        if path and not os.path.exists(path):
+            raise FileNotFoundError(f"Settings file not found: {path}")
+
         if path and path.endswith('.conf'):
             self._deprecated_settings = True
 
--- a/kernelci/cli/base.py
+++ b/kernelci/cli/base.py
@@ -519,10 +519,14 @@ class Options:
         a section called `db:<db-config-name>` with `db-config-name` the value
         passed in `db_config`, or in the `--db-config` command line argument.
 
-        *path* is the path to the config file, which by default is
-                `kernelci.conf` or
-               `~/.config/kernelci/kernelci.conf` or
-               `/etc/kernelci/kernelci.conf`
+        *path* is the path to the config file, if not provided it will be
+                looked up in the following order:
+                1) from the `KCI_SETTINGS` environment variable
+                2) look for following files in order:
+                     kernelci.toml, ~/.config/kernelci/kernelci.toml,
+                     /etc/kernelci/kernelci.toml, kernelci.conf,
+                     ~/.config/kernelci/kernelci.conf,
+                     /etc/kernelci/kernelci.conf
 
         *command* is a `Command` object for the command being run
 
@@ -533,6 +537,9 @@ class Options:
                   way to have default values for each CLI tool
         """
         self._deprecated_settings = False
+        if path is None:
+            path = os.environ.get('KCI_SETTINGS')
+
         if path is None:
             default_paths = [
                 'kernelci.toml',


gctucker (commented):
See also the implementation I made for the `kci-typer` RFC:
```python
            default_paths = [
                os.getenv('KCI_SETTINGS', ''),
                'kernelci.toml',
                os.path.expanduser('~/.config/kernelci/kernelci.toml'),
                '/etc/kernelci/kernelci.toml',
                'kernelci.conf',
                os.path.expanduser('~/.config/kernelci/kernelci.conf'),
                '/etc/kernelci/kernelci.conf',
            ]
```

gctucker:
> By default we can specify config file by --api-config

You meant `--settings` here I think:

```python
    parser.add_argument(
        "--settings",
        help="Path to the settings file",
    )
```

nuclearcat:
> See also the implementation I made for the `kci-typer` RFC:
> 
> ```python
>             default_paths = [
>                 os.getenv('KCI_SETTINGS', ''),
>                 'kernelci.toml',
>                 os.path.expanduser('~/.config/kernelci/kernelci.toml'),
>                 '/etc/kernelci/kernelci.toml',
>                 'kernelci.conf',
>                 os.path.expanduser('~/.config/kernelci/kernelci.conf'),
>                 '/etc/kernelci/kernelci.conf',
>             ]
> ```
I think such way might lead to very hard to troubleshoot issue, if we set mandatory KCI_SETTINGS variable - we should not try using other "default" files.
Otherwise, if file set in KCI_SETTINGS doesn't exist (type or some other issue) - kci might silently use one of default files, which will be hard to notice and troubleshoot.

gctucker:
Yeah I guess, also it depends how env variables should be treated.  A command-line argument is pretty explicit, environment variables are more implicit as they're set up by a different means than invoking the process.

I think in this PR I would move the part that looks for the KCI_SETTINGS env variable before the list of default paths to make it clearer it's not a "custom default" but actually something the user is explicitly requesting.  I'm wondering if this maybe should actually be done outside this function, to make it a command-line tool feature only.  Probably it's fine to leave it here for now at least, and the docstring should be updated accordingly.

nuclearcat:
docstring updated, and made code more clear according to your suggestions, moved before condition

kernelci/cli/base.py
```diff
@@ -533,6 +537,13 @@ def __init__(self, path, command, cli_args, section=None):
                   way to have default values for each CLI tool
         """
         self._deprecated_settings = False
+        if path is None and os.environ.get('KCI_SETTINGS'):
+            path = os.environ.get('KCI_SETTINGS')
+            if not os.path.exists(path):
+                print(f"Error: KCI_SETTINGS points to a non-existing "
+                      f"file: {path}", file=sys.stderr)
+                sys.exit(1)
```
gctucker, line 545: Arguably this should be throwing an exception, exiting in a class constructor is rather unexpected I think.  The logic is fine though, as the program should print this error and abort.  Maybe for services that aren't CLI tools it would need to handle the error differently etc.

gctucker (commented):
Thanks for the update.  Just one more small thing to consider I think.

kernelci/cli/base.py
nuclearcat: Maybe better i check after everything path variable and make sure file exist? And if doesnt i can throw exception

gctucker: Yes, that might be better as a separate commit.  I think right now there will be an exception anyway when trying to open a file that doesn't exist.

nuclearcat: Done, please check if it is better now

gctucker (commented):
LGTM :)  I'll just give it a try by hand to double check it's working as expected since there's no unit test for this yet.

Also it looks like pylint is finding issues with parts of the code that aren't changing, that's a bit annoying.  I think we might just "silence" the Git check here and address it in a separate PR as it seems entirely unrelated.

gctucker (commented):
Ah no actually the failing GitHub checks are about a valid problem:
```
FileNotFoundError: Settings file not found: None
make: *** [Makefile:50: validate-yaml] Error 1
Error: Process completed with exit code 2.
```
It's not a requirement to have a settings file.  So if `path` is not provided, that's fine and the settings should just be an empty dictionary.

kernelci/cli/base.py
```diff
@@ -533,6 +537,9 @@ def __init__(self, path, command, cli_args, section=None):
                   way to have default values for each CLI tool
         """
         self._deprecated_settings = False
+        if path is None and os.environ.get('KCI_SETTINGS'):
```
gctucker, line 540: Alright I think this should tick all the boxes, and the rest can be left unchanged:
```suggestion
        if path is None:
            env_path = os.environ.get('KCI_SETTINGS')
            if env_path:
                if not os.path.exists(env_path):
                    raise FileNotFoundError(env_path)
                path = env_path
```

Does that make sense?  Sorry for all the ping-pong, maybe it would have been better to just use `--settings` rather than add support for `KCI_SETTINGS`.  Now we have it so I think that's fine.

nuclearcat:
> Ah no actually the failing GitHub checks are about a valid problem:
> 
> ```
> FileNotFoundError: Settings file not found: None
> make: *** [Makefile:50: validate-yaml] Error 1
> Error: Process completed with exit code 2.
> ```
> 
> It's not a requirement to have a settings file. So if `path` is not provided, that's fine and the settings should just be an empty dictionary.

I added explanation in condition. So still if we have set --settings or env variable it is better to raise error if we cannot access file, so we don't have silent failure

kernelci/cli/base.py
nuclearcat: I think environment variable handy in some cases, especially docker

gctucker (approved):
Tested OK by hand with `KCI_SETTINGS` environment variable as well as `--settings` command line argument:

* the argument takes precedence over the env variable as intended
* if `KCI_SETTINGS` is set but the file doesn't exist, and there's no `--settings`, an exception is raised
* if `--settings` is set but the file doesn't exist, an exception is raised regardless of `KCI_SETTINGS`
* if no settings file is specified then it looks for the default paths etc. like before

PR #2116, opened 2023-09-27
runtime/docker: Don't add to env comments

nuclearcat:
docker.run wont tolerate in env variable any comments or empty data, as it will try to interpret it as variable with unacceptable name. So we can just skip comments and empty lines.

### DIFF
--- a/kernelci/runtime/docker.py
+++ b/kernelci/runtime/docker.py
@@ -32,7 +32,15 @@ class Docker(Runtime):
     def _load_env(self):
         if self.config.env_file and os.path.isfile(self.config.env_file):
             with open(self.config.env_file, encoding='utf-8') as env:
-                return [line.strip() for line in env.readlines()]
+                env_list = []
+                for line in env.readlines():
+                    env_line = line.strip()
+                    # Skip empty lines and comments
+                    if not env_line or env_line.startswith('#'):
+                        continue
+                    env_list.append(env_line)
+
+                return env_list
         return None
 
     @classmethod


JenySadadia (commented):
I haven't used `docker` package yet, but doesn't it provide a way to load `env` file?

nuclearcat:
Over python SDK seems not.
https://docker-py.readthedocs.io/en/stable/containers.html
```
environment (dict or list) – Environment variables to set inside the container, as a dictionary or a list of strings in the format ["SOMEVARIABLE=xxx"].
```

JenySadadia:
> Over python SDK seems not. https://docker-py.readthedocs.io/en/stable/containers.html
> 
> ```
> environment (dict or list) – Environment variables to set inside the container, as a dictionary or a list of strings in the format ["SOMEVARIABLE=xxx"].
> ```

Okay. Thanks for checking.

kernelci/runtime/docker.py
```diff
@@ -32,7 +32,9 @@ def __init__(self, *args, **kwargs):
     def _load_env(self):
         if self.config.env_file and os.path.isfile(self.config.env_file):
             with open(self.config.env_file, encoding='utf-8') as env:
-                return [line.strip() for line in env.readlines()]
+                # skip # and empty lines
```
JenySadadia, line 35: Just a minor suggestion.
```suggestion
                # Skip comments and empty lines
```

kernelci/runtime/docker.py
```diff
@@ -32,7 +32,9 @@ def __init__(self, *args, **kwargs):
...
+                return [line.strip() for line in env.readlines()
+                        if not line.startswith('#') and line.strip()]
```
gctucker, line 37: This is calling `line.strip()` twice, it's OK but a bit sub-optimal.  A neater implementation would be two nested list comprehensions to iterate over the stripped lines.  Not a big deal though, it's arguably easier to read the way it's currently implemented.

JenySadadia (approved):

PR #2119, opened 2023-09-28
[RFC] kci-argparse: add kci PoC using argparse

gctucker:
Add a kci-argparse proof-of-concept replacement for kci using the standard argparse module.  It comes with a sample
kernelci-argparse.toml settings file to illustrate how the values can be combined with the command line arguments.

This implementation isn't quite complete as nested commands can't really be implemented with argparse as easily as with Click or derivatives such as Typer.  While it may be possible, this PoC is about comparing solutions and it's clear that using argparse would require a lot of addtional work to provide the same user experience as with Click.

### DIFF
--- a/kci-argparse
+++ b/kci-argparse
@@ -1,0 +1,246 @@
+#!/usr/bin/env python3
+#
+# SPDX-License-Identifier: LGPL-2.1-or-later
+#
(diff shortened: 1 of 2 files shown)


gctucker:
Closing this RFC now as we've reached a consensus.

PR #2126, opened 2023-10-06
test-configs.yaml: Enable acer-chromebox-cxi5-brask for testing

nuclearcat:
We have new device in staging lab, its one only, so we just run baseline and baseline-nfs for verifying bootability and network operation.


nuclearcat:
baseline, baseline-nfs present, for example https://staging.kernelci.org/test/plan/id/6521552f812f57ea23e193bd/

PR #2127, opened 2023-10-06
kernelci.legacy: move kernelci.cli to legacy

gctucker:
Move kernelci.cli to kernelci.legacy.cli as it's about to be replaced with the new implementation for the kci tool based on Click.


nuclearcat:
Tested on staging, related issues fixed (in pipeline https://github.com/kernelci/kernelci-pipeline/pull/338)

PR #2135, opened 2023-10-17
CHROMEOS test.py: Truncate very long cross:// repo name

nuclearcat:
Use shortening service to reduce job length, as it is excessively long.


nuclearcat:
Need to wait until chromeos-stable_6.6 will have updates.

hardboprobot (approved):
LGTM but I didn't test it. If you tested it and it works, go ahead.

nuclearcat:
Thanks!
I am still testing and looking for some more issues, as it looks like some tasts tests are not running due too long job names.

nuclearcat:
Finally tested and passing all build and test process on chromeos instance.

PR #2136, opened 2023-10-18
kubernetes.jinja2: Increase timeout before container destroy

nuclearcat:
As we are in development process, we need to keep container a bit longer after failure or completion, so we can have enough time to catch logs on early failure.

### DIFF
--- a/config/runtime/base/kubernetes.jinja2
+++ b/config/runtime/base/kubernetes.jinja2
@@ -8,7 +8,9 @@ metadata:
 
 spec:
   completions: 1
-  ttlSecondsAfterFinished: 30
+  # During development we keep logs a bit longer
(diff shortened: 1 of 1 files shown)


gctucker:
Small typo: s/increate/increase/

nuclearcat:
Thanks, updated commit and PR description

gctucker (approved):
OK, as discussed this can be resolved by capturing the Kubernetes logs somewhere and then the timeout value won't be an issue.

One solution for this could be to have a service that listens for nodes with jobs running in Kubernetes and get the log with their state goes to `done`, then add it to the node as an artifact.  We would just need to ensure that the Kubernetes timeout is longer than the node timeout with some margin to let the service grab the log (e.g. node timeout + 1min).

PR #2137, opened 2023-10-19
kernelci.config: handle YAML parser exceptions in validation

gctucker:
Handle YAML parser exceptions while validating the config files to provide a better validation result with the actual error message from the exception.


nuclearcat (approved):
LGTM, tested on staging

PR #2138, opened 2023-10-19
Rework `kci config` with new framework

gctucker:
Rework `kci config` with the new command-line framework based on Click.

Note: This doesn't implement `kci config set` yet, it can be added as a follow-up.

See the discussion about this command: https://github.com/kernelci/kernelci-project/discussions/263#discussioncomment-7167111


gctucker:
Adding `staging-skip` so that #2160 can be tested on staging and merged first.

gctucker (commented):
Alright, we're back to "merge and fixup" mode due to lack of available reviewers.

PR #2139, opened 2023-10-20
kernelci.config: rename 'api_configs' to 'api'

gctucker:
Rename the YAML entry 'api_configs' to 'api' as per the GitHub discussion.  The new kci command line tool will also have an --api option to specify the API config to use.

Update standard YAML configuration and unit tests accordingly.

Link: https://github.com/kernelci/kernelci-core/discussions/1907#discussioncomment-5718754

### DIFF
--- a/config/core/api-configs.yaml
+++ b/config/core/api-configs.yaml
@@ -1,4 +1,4 @@
-api_configs:
+api:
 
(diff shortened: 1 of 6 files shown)


nuclearcat:
As required by other patch

PR #2140, opened 2023-10-20
kernelci.cli: rewrite split_attributes() method

gctucker:
Add a new kernelci.cli.split_attributes() method to be used with the
    Click command line tools.  This is based on the previous
    implementation with symbolic comparison operators but with a revised
    implementation.  Update unit tests accordingly.

### DIFF
--- a/kernelci/cli/__init__.py
+++ b/kernelci/cli/__init__.py
@@ -13,6 +13,9 @@ the Click framework, it adds support for loading default values and secrets
 from TOML settings.
 """
 
+import re
+import typing
+
 import click
 
 import kernelci.settings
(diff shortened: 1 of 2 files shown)


gctucker:
@nuclearcat Would you like to review this one since you introduced the feature of boolean operator symbols?

nuclearcat:
Absolutely!

nuclearcat (approved):
LGTM

kernelci/cli/__init__.py
```diff
@@ -129,3 +132,48 @@ def group(self, *args, cls=None, **kwargs):
 def kci(ctx, settings):
     """Entry point for the kci command line tool"""
     ctx.obj = CommandSettings(settings)
+
+
+def split_attributes(attributes: typing.List[str]):
+    """Split attributes into a dictionary.
+
+    Split the attributes string into a dictionary using space as a delimiter
```
gctucker, line 140: Actually, I just realised there was a problem with the docstring.  I'll send a fixup.  The attributes argument is not a string with spaces but a list of strings already split.

PR #2141, opened 2023-10-20
Rework `kci user` with new framework

gctucker:
Implement parts of `kci user` with the new framework based on Click and remove the corresponding old implementation code.

### DIFF
--- a/kci
+++ b/kci
@@ -13,28 +13,12 @@ features.  See the online documentation for more details:
 https://kernelci.org/docs/api/.
 """
 
-import json
 
-import click
-
-from kernelci.cli import Args, kci
+from kernelci.cli import kci
 from kernelci.cli import (  # pylint: disable=unused-import
     docker as kci_docker,
+    user as kci_user,
 )
-import kernelci.api
-import kernelci.config
-
-
-@kci.command(secrets=True)
-@Args.config
-@Args.api
-def whoami(config, api, secrets):
-    """Use whoami to get current user info with API authentication"""
-    configs = kernelci.config.load(config)
-    api_config = configs['api_configs'][api]
-    api = kernelci.api.get_api(api_config, secrets.api.token)
-    data = api.whoami()
-    click.echo(json.dumps(data, indent=2))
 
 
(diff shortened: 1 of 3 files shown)


JenySadadia:
@gctucker Is this ready for review?
It would be good to merge these changes before I update user-related API bindings with `fastapi-users` integration.

gctucker:
> @gctucker Is this ready for review?

It is now I've rebased it ;)

kci
```diff
@@ -13,28 +13,12 @@ features.  See the online documentation for more details:
...
-import json
 
-import click
-
-from kernelci.cli import Args, kci
+from kernelci.cli import kci
 from kernelci.cli import (  # pylint: disable=unused-import
     docker as kci_docker,
+    user as kci_user,
 )
```
JenySadadia, line 21: These are unused imports.

JenySadadia: Seems like it is required to register sub-commands. A comment about it would have been helpful.

JenySadadia (approved):
Tested OK with early-access instance.

kernelci/cli/user.py
```diff
@@ -0,0 +1,109 @@
...
+    configs = kernelci.config.load(config)
+    api_config = configs['api'][api]
+    api = kernelci.api.get_api(api_config)
+    user_token = api.create_token(username, password, scope)
+    click.echo(json.dumps(user_token, indent=indent))
+
+
+@kci_user.group(name='password')
+def user_password():
+    """Manage user passwords"""
```
JenySadadia, line 72: After I added sub-commands to this group, it became `kci user password reset-password`.
I expected it to be `kci password reset-password`.
What are the benefits of adding groups?

gctucker: Well this was as per the proposed syntax on the GitHub discussion:
https://github.com/kernelci/kernelci-project/discussions/263#discussioncomment-7163554

It means all the user-related commands are grouped under `kci user`.  There might be other passwords for other things, like `kci storage password` or anything else.

PR #2143, opened 2023-10-21
config/core: update public Azure SAS token

gctucker:
Update public Azure SAS token as per ZDI-CAN-22317.

Reported-by: Denys Fedoryshchenko <denys.f@collabora.com>


gctucker:
We still need to sort out the YAML configuration files between kernelci-core and kernelci-pipeline.  I'll create another GitHub issue about this.

gctucker:
I've created this issue as a follow-up: https://github.com/kernelci/kernelci-pipeline/issues/344

PR #2145, opened 2023-10-21
kernelci.legacy: move legacy config implementation

gctucker:
Move the modules that relate to the legacy YAML implementation to kernelci.legacy.config.  This is still loaded as part of the main configuration to provide backwards-compatibility until support is eventually dropped.

Update unit tests, imports and setup.py accordingly.


nuclearcat:
As other PR depending on it, also tested on staging

nuclearcat (approved):
LGTM

PR #2147, opened 2023-10-23
config/runtime: base: import kernelci.config.storage

gctucker:
As the kernelci.config module is being rewritten, it no longer implicitly import other submodules.  Explicitly import kernelci.config.storage in python.jinja2 to avoid missing import errors.


nuclearcat (approved):
Fixes issue on staging, tested, thanks!

PR #2149, opened 2023-10-26
test-configs.yaml: Enable more tests on verdin

broonie:
For some reason when reenabling the tests that were disabled due to load
issues those for verdin were skipped, enable everything now.

Also enable the clone3() selftests, they are not currently covered at all
on arm64 and the board is generally very lightly loaded.


nuclearcat (approved):
LGTM, test results:
https://staging.kernelci.org/test/plan/id/6549c3a06a5853e3b9a372ad/
https://staging.kernelci.org/test/plan/id/6549c10b4aec31a960a3767d/

PR #2150, opened 2023-10-27
kernelci.cli: fix split_attributes() docstring

gctucker:
Fix the split_attributes() docstring as the attributes are meant to be a list of key=value strings rather than a single string with all the attributes.  The command line parser should already split each attribute into a separate string.  Also mention that spaces can be used around operators with examples using double quotes.

Fixes: 0a62fb90d404 ("kernelci.cli: rewrite split_attributes() method")

### DIFF
--- a/kernelci/cli/__init__.py
+++ b/kernelci/cli/__init__.py
@@ -137,14 +137,17 @@ def kci(ctx, settings):
 def split_attributes(attributes: typing.List[str]):
     """Split attributes into a dictionary.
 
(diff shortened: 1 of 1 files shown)


gctucker:
This is a stepping stone for proper documentation, it would be good to have this explained directly in some `--help` message too.  This can be all addressed once a few more PRs that add some real commands have been merged e.g. `kci docker`, `kci user` and  `kci config`.

nuclearcat:
LGTM

PR #2152, opened 2023-10-29
config/core: add back db_configs which were deleted with storage rework

gctucker:
Add back the db_configs entries which got lost by mistake while reworking the storage configs.

Fixes: 68b04800c0f0 ("config/core: move storage_configs to storage in storage.yaml")

### DIFF
--- a/config/core/db-configs.yaml
+++ b/config/core/db-configs.yaml
@@ -1,0 +1,17 @@
+db_configs:
+
+  chromeos.kernelci.org:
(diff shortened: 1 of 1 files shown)


gctucker:
Tested manually (previously it would not return anything):
```
$ ./kci_data list_configs 
chromeos.kernelci.org
kernelci.org
localhost
staging.kernelci.org
```

nuclearcat:
tested on staging

PR #2153, opened 2023-10-31
Add basic parameters to LAVA job template

pawiecz:
This PR passes necessary information and adds basic parameters to the LAVA TestJob template.

Verification steps will follow in the comment below.

### DIFF
--- a/config/runtime/base/lava.jinja2
+++ b/config/runtime/base/lava.jinja2
@@ -7,7 +7,6 @@
 #
 # Things left to add:
 # * instance name (staging, AKS...)
-# * context
 # * priority
 # * timeout
 # * modules
@@ -24,10 +23,10 @@ visibility: public
 
 priority: medium
 
-context:
-  arch: x86_64
-  cpu: qemu64
-  guestfs_interface: ide
+{% if platform_config.context %}
+context:{% for key, value in platform_config.context | items %}
+  {{ key }}: {{ value }}{% endfor %}
+{% endif %}
 
 timeouts:
   action:
--- a/config/runtime/base/lava.jinja2
+++ b/config/runtime/base/lava.jinja2
@@ -64,31 +64,8 @@ metadata:
   storage_config_name: {{ storage_config.name }}
 
 actions:
-
-- deploy:
-    images:
-      kernel:
-        image_arg: -kernel {kernel} -append "console=ttyS0,115200 root=/dev/ram0 debug
-          verbose console_msg_format=syslog earlycon"
-        url: {{ node.artifacts.bzImage }}
-      ramdisk:
-        image_arg: -initrd {ramdisk}
-        url: http://storage.kernelci.org/images/rootfs/buildroot/buildroot-baseline/20230421.0/x86/rootfs.cpio.gz
-    os: oe
-    timeout:
-      minutes: 3
-    to: tmpfs
-
-- boot:
-    docker:
-      binary: qemu-system-x86_64
-      image: kernelci/qemu
-    media: tmpfs
-    method: qemu
-    prompts:
-      - '/ #'
-    timeout:
-      minutes: 5
+{% set boot_template = 'boot/' + platform_config.boot_method + '.jinja2' %}
+{% include boot_template %}
 
 - test:
     definitions:
--- a/config/runtime/base/lava.jinja2
+++ b/config/runtime/base/lava.jinja2
@@ -7,7 +7,6 @@
 #
 # Things left to add:
 # * instance name (staging, AKS...)
-# * device type
 # * context
 # * priority
 # * timeout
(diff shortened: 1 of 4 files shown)


nuclearcat:
Can you fix pycodestyle warning please?
Do you think it is ready for staging test?

pawiecz:
I fixed the issues that surfaced so far. Please note that this change will probably also require pulling kernelci/kernelci-pipeline#349 to run properly on staging.

pawiecz:
> Verification steps will follow in the comment below.

I verified these changes by running a local API/Pipeline instance with default config and examining the generated LAVA TestJob definitions for qemu and Minnowboard.

I intended to run those definitions in the LAVA lab but took a shortcut by removing Scheduler Job dependency on `kbuild-gcc-10-x86` and used just `kver` instead. Node `kver` does not provide `node.artifacts.bzImage` so the definitions were incomplete.

nuclearcat:
Staging failing for now, due way how staging works... applying patch. I believe we need to implement it somehow without patching code on staging (thats more task for me and might need some guidance from @gctucker )
```
Applying patch: patches/kernelci-core/staging.kernelci.org/0007-STAGING-use-KernelCI-staging-in-LAVA-jobs.patch
error: patch failed: config/runtime/base/lava.jinja2:17
error: config/runtime/base/lava.jinja2: patch does not apply
hint: Use 'git am --show-current-patch=diff' to see the failed patch
WARNING: Failed to apply patch
Aborting, all patches must apply.
```

nuclearcat:
One of LAVA jobs seems failed: https://lava.collabora.dev/scheduler/job/11917121
QEMU did fine: https://lava.collabora.dev/scheduler/job/11917120

config/runtime/base/lava.jinja2
```diff
@@ -19,16 +17,16 @@
 
 job_name: "[KernelCI AKS] {{ node.id }} {{ node.name }} {{ node.revision.describe }}"
 
-device_type: qemu
+device_type: {{ platform_config.base_name|default(platform_config.name) }}
```
gctucker, line 20: The base name is the only valid one to use here.  The `name` attribute is only something in YAML config to have several variants.

Note: I wonder if it might actually be better to ultimately decouple the templates from the config object attributes, so just having `{{ platform_type }}` would work better, and each Runtime implementation can populate accordingly.  But that's a bit beyond the scope of this PR.

kernelci/runtime/__init__.py
```diff
@@ -125,13 +125,13 @@ def get_params(self, job, api_config=None):
...
             'api_config': api_config or {},
             'storage_config': job.storage_config or {},
+            'platform_config': job.platform_config.to_dict() or {},
             'name': job.name,
             'node': job.node,
             'runtime': self.config.lab_type,
             'runtime_image': job.config.image,
         }
         params.update(job.config.params)
-        params.update(job.platform_config.params)
```
gctucker, line 134: I believe the idea was that this would enable platform configs to override any value already set.

gctucker:
> Staging failing for now, due way how staging works... applying patch.

Yes, reducing the number of floating patches is a good idea as a way to keep improving things there.  In the meantime, it's easy enough to just rebase the series of patches and then an improvement can be done at any time without blocking development.

pawiecz:
> One of LAVA jobs seems failed: https://lava.collabora.dev/scheduler/job/11917121

Fixed boot/deploy section template (images, prompts) in the latest patch set (https://lava.collabora.dev/scheduler/job/11939283 worked fine after applying changes)

nuclearcat:
Thanks, after we fix staging i will test it on it.

nuclearcat:
Congrats, it is working!
https://lava.collabora.dev/scheduler/job/11958601

config/runtime/base/lava.jinja2
gctucker: @nuclearcat This hasn't been addressed, the PR was merged too early.

kernelci/runtime/__init__.py
gctucker: Same thing here.

gctucker (commented):
There were unresolved issues, this PR shouldn't have been merged without at least a reply.  Now we need to create GitHub issues to keep track of this and some follow-up PRs to fix them potentially.

pawiecz:
I'll create the issues for tracking this PR's defects.

config/runtime/base/lava.jinja2
pawiecz: Tracked by #2168, fixed in #2169

kernelci/runtime/__init__.py
pawiecz: Tracked by #2170, fixed in #2171

PR #2155, opened 2023-11-01
`kci user` rework due to `fastapi-users` integration

JenySadadia:
Depends on https://github.com/kernelci/kernelci-api/pull/377

`kci user` sub-commands rework.


gctucker:
Why is there the `staging-skip` label?  Is this because we're missing migrations?

JenySadadia:
> Why is there the `staging-skip` label? Is this because we're missing migrations?

Yes, I'll fix it soon.

gctucker (approved):
Tested OK locally.

PR #2156, opened 2023-11-02
CHROMEOS: rootfs: add and enable R118 manifest

10ne1:
The src/scripts/build_image was removed in favor of one installed in $PATH, so we need to replace the ./build_image invocation with a plain build_image.

minigbm:
 - Dropped mediatek backend in favor of drv_dumb backend, no more divergence for the platform/minigbm component.

chromiumos-overlay:
 - Reworked the minigbm patch into a single two liner patch to activate drv_dumb
 - Droppped the chromeos-kernel-5_10 bump in favor of the 5.10 shipped in R118
 - Fixed broken portage manifest for media-libs/cros-camera-libfs
 - Masked newer chromeos-chrome ebuild version which doesn't have a binpkg, to avoid unnecessary long build times.
 - Bumped chromeos-installer ebuild as a result of forking the platform2 overlay.

platform2:
 - Forked to disable failing installer tpm check, until a fix is landed for b:291036452.

graphics:
 - Backported patch from R119 to avoid a conflict, still need to keep this patch in R118

board-overlays:
 - Forward ported the last remaining commit without conflicts. Will keep it for now but gradually will reduce its footprint as MTK SoCs start using the newer Google kernels with 6.6
 - Bumped octopus to 5.15 because tast fails with 4.14.

kernel v5.10:
 - Removed fork in favor of cros-kernel 5.10

packagefix-release-R118-15604.B.sh:
 - Dropped some of the tpm2 fixes which are not necessary anymore
 - More of these fixes landed post R118 and some are still pending for gurnt, zork and hatch (see CL 4868824).


a-wai:
LGTM (also checked the patches on our downstream repos, nothing to complain about there either).

nuclearcat (approved):
LGTM, also did upload to storage, and tests got verified as well

PR #2157, opened 2023-11-02
Add `kci user` sub-commands

JenySadadia:
Depends on https://github.com/kernelci/kernelci-core/pull/2155

Add below sub-commands to `kci user`:
- `kci user verify EMAIL`
- `kci user password reset EMAIL`
- `kci user password update USERNAME`
- `kci user get  USER_ID`
- `kci user update-by-id USER_ID --username --email --groups`

### DIFF
--- a/kernelci/api/__init__.py
+++ b/kernelci/api/__init__.py
@@ -154,7 +154,7 @@ class API(abc.ABC):  # pylint: disable=too-many-public-methods
         """Create a new user"""
 
     @abc.abstractmethod
-    def update_user(self, username: str, profile: dict) -> dict:
+    def update_user(self, user: dict) -> dict:
         """Update a user"""
 
     # -------------------------------------------------------------------------
@@ -194,6 +194,16 @@ class API(abc.ABC):  # pylint: disable=too-many-public-methods
         url = self._make_url(path)
         jdata = json.dumps(data)
         resp = requests.put(
+            url, jdata, headers=self._headers,
+            params=params, timeout=self._timeout
+        )
+        resp.raise_for_status()
+        return resp
+
+    def _patch(self, path, data=None, params=None):
+        url = self._make_url(path)
+        jdata = json.dumps(data)
+        resp = requests.patch(
             url, jdata, headers=self._headers,
             params=params, timeout=self._timeout
         )
--- a/kernelci/api/__init__.py
+++ b/kernelci/api/__init__.py
@@ -173,6 +173,10 @@ class API(abc.ABC):  # pylint: disable=too-many-public-methods
     def reset_password(self, token: str, password: str):
         """Reset password"""
 
+    @abc.abstractmethod
+    def get_user(self, user_id: str) -> dict:
+        """Get the user matching the given user id"""
+
     # -------------------------------------------------------------------------
     # Private methods
     #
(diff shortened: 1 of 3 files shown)


gctucker:
This isn't aligned with the proposed syntax on the GitHub discussion:
https://github.com/kernelci/kernelci-project/discussions/263

Could you please post a comment there first explaining the changes you want to make, and then we can revise the syntax design?

Having as few commands as possible with short and unambiguous names is the most important thing I think.  Some commands may be interactive, for example to complete a password reset procedure it could be waiting for the user to provide a token sent over email.

JenySadadia:
> Could you please post a comment there first explaining the changes you want to make, and then we can revise the syntax design?

Sure.

JenySadadia:
> Having as few commands as possible with short and unambiguous names is the most important thing I think. Some commands may be interactive, for example to complete a password reset procedure it could be waiting for the user to provide a token sent over email.

I agree. I updated `reset` and `verify` commands to be interactive.

JenySadadia:
Added command to update user password.

gctucker:
FYI I've updated the CLI syntax accordingly on the GitHub discussion.  Please take a look to confirm this is OK.

JenySadadia:
> FYI I've updated the CLI syntax accordingly on the GitHub discussion. Please take a look to confirm this is OK.

Yes, it looks good.
Also, we need to add `kci user update-by-id` command to the discussion. But I am not quite sure about the same of the command though. Can you suggest a better one?

gctucker:
Why do we need a command to update a user by its id rather than username?

JenySadadia:
> Why do we need a command to update a user by its id rather than username?

`fastapi-users` has a route PATCH `/user/{user-id}` for updating user by id. I implemented https://github.com/JenySadadia/kernelci-api/blob/fastapi-users-integration/api/main.py#L250 based on it.
That's why we need ID for the command. If we want to have username then we need to send GET request from command handler first to get user by id and then send username to PATCH request. 
Or else we can change the endpoint to have a username instead of ID.

JenySadadia:
Just to clarify now we have two commands for update:
- Update own user profile with authorization token
  `kci user update --username --email --groups`
- Update other user account by admin
  `kci user update-by-id --username --email --groups`

gctucker:
I think the command should not use the ID on the user interface side as that's not practical, but if we need to provide the user ID to the API then the implementation can do that.  So now I don't see a need for having `update-by-id` on the command line.  The `kci user update` implementation can get the user data using the username and that will contain the id, then submit it again via the endpoint with the ID.

gctucker:
Generally speaking, the command-line tool syntax and even the API endpoints should not be related to how things are implemented.  They need to be designed so they make sense from a user point of view, and the implementation can take care of all the details to make it work.

JenySadadia:
> Generally speaking, the command-line tool syntax and even the API endpoints should not be related to how things are implemented. They need to be designed so they make sense from a user point of view, and the implementation can take care of all the details to make it work.

Makes sense. I agree with this.
I have updated the command and renamed it to `update-by-username`. We can not have the same command names for 2 different commands as per the proposal https://github.com/kernelci/kernelci-project/discussions/263#discussioncomment-7163554. Hence chose a different name.

gctucker:
I don't get this, we just need `kci user update` surely?

Anyway I'll review it and maybe send some follow-up rework if needed.

gctucker:
Replaced with #2183 (rebased with same commits).

PR #2162, opened 2023-11-03
kernelci.storage: azure: don't create new shares

gctucker:
Remove the part of the code that would create a share if it didn't exist.  This worked with over-permissive tokens which had access to all the shares, but the permission scheme is now to provide tokens limited to a single share for each user.  So the share has to exist, and if it doesn't then an exception will be raised as intended.

### DIFF
--- a/kernelci/storage/azure.py
+++ b/kernelci/storage/azure.py
@@ -30,17 +30,6 @@ class StorageAzureFiles(Storage):
             credential=self.credentials
         )
 
-    def _get_share(self):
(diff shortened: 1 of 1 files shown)


nuclearcat (approved):
LGTM. as i understand it is for compatibility with new SAS tokens per user

gctucker:
Yes, now tokens are on a per-share basis so the expectation is that the share already exists.  The previous implementation would create it with a token that had the permission to do so.

PR #2165, opened 2023-11-04
build-configs-android.yaml: Update clang-17 and add more builds

nuclearcat:
By request of maintainer, we are adding more coverage for android trees, and update clang version.


nuclearcat:
Merging as changes are trivial

PR #2166, opened 2023-11-07
setup.py: add missing kernelci.cli package

gctucker:
Add the missing kernelci.cli package to the list of packages to install.

This was missing when adding the new kernelci.cli implementation and went unnoticed when using the code from the Git checkout.

Fixes: ec7f5c61ae53 ("kernelci.cli: rewrite with Click framework")

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -36,6 +36,7 @@ setuptools.setup(
     packages=[
         "kernelci",
         "kernelci.api",
+        "kernelci.cli",
         "kernelci.config",
         "kernelci.db",
         "kernelci.legacy",


gctucker:
Additionally, it would be good to have some GitHub checks that install the package.  I think we should do that when porting it to `pyproject.toml`.

nuclearcat (approved):
Works on staging (was not):
```
docker exec -it b83ec0fb7d38 /bin/sh
$ kci
Usage: kci [OPTIONS] COMMAND [ARGS]...

  Entry point for the kci command line tool

Options:
  -s, --toml-settings TEXT  Path to the TOML user settings
  --help                    Show this message and exit.

Commands:
  api      Run generic queries with API instances
  docker   Build and manage Docker images
  storage  Interact with KernelCI storage services
  user     Manage user accounts
```

PR #2167, opened 2023-11-07
runtime: Add instance ID for LAVA runtime

nuclearcat:
Add environment variable KCI_INSTANCE that will adjust job name prefix for LAVA.
Probably in future similar way we can customize k8s and docker names as well.

### DIFF
--- a/config/runtime/base/lava.jinja2
+++ b/config/runtime/base/lava.jinja2
@@ -15,7 +15,7 @@
 # * rootfs URL
 # * Docker image for QEMU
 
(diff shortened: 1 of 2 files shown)


nuclearcat:
Works on staging: https://lava.collabora.dev/scheduler/job/11959202
```
Description
    [kci-staging] 6549e58cc054242fbccbc208 baseline-x86 staging-mainline-20231107.0
```

gctucker:
Please make it "ready for review" when it's ready.

As long as this works I think it's fine to merge it, however we probably should be either loading the API instance name from YAML or have a command line argument which can have a value in TOML settings.  I guess this is fine for now.

Issue #2168, opened 2023-11-07
Field `device_type` in LAVA TestJob definition unnecessarily falls back to platform name


pawiecz:
The base name is the only valid one to use here.  The `name` attribute is only something in YAML config to have several variants.

_Originally posted by @gctucker in https://github.com/kernelci/kernelci-core/pull/2153#discussion_r1379016807_

from Slack channel:

>[6:57](https://kernelci.slack.com/archives/C015N3UF7C1/p1699336655331919) Yes so the 2nd part needs to be removed
>[6:57](https://kernelci.slack.com/archives/C015N3UF7C1/p1699336667077049) Yes it's a bug because it's the wrong definition of what this value should be
>[6:58](https://kernelci.slack.com/archives/C015N3UF7C1/p1699336687328709) It's not OK to use the name variants from YAML
>[6:58](https://kernelci.slack.com/archives/C015N3UF7C1/p1699336697950819) See the legacy template: device_type: {{ base_device_type }}

pawiecz:
I verified that field `base_name` is always available even if not explicitly defined in platform config (`device_types`) like in case of Minnowboard (https://github.com/kernelci/kernelci-pipeline/blob/6c4947b73cfaa5b13637f68893695cf5550414f8/config/pipeline.yaml#L171 but `'platform_config': {'base_name': 'minnowboard-turbot-E3826', 'name': 'minnowboard-turbot-E3826', (...)`).

This has most probably been populated by: https://github.com/kernelci/kernelci-core/blob/c84053ae5aec15c4fba99edd0b51dedf08cdd63f/kernelci/legacy/config/test.py#L51

PR #2169, opened 2023-11-07
config/runtime: base: drop wrong device_type defaults

pawiecz:
Fixes: #2168


nuclearcat (approved):
Trivial change and tested on staging, thanks!

Issue #2170, opened 2023-11-07
Reenable platform configs to override any value already set


pawiecz:
I believe the idea was that this would enable platform configs to override any value already set.

_Originally posted by @gctucker in https://github.com/kernelci/kernelci-core/pull/2153#discussion_r1379020998_

PR #2171, opened 2023-11-07
config/runtime: base: reenable platform configs to override previously set values

pawiecz:
Fixes: #2170

### DIFF
--- a/kernelci/runtime/__init__.py
+++ b/kernelci/runtime/__init__.py
@@ -134,6 +134,7 @@ class Runtime(abc.ABC):
             'runtime_image': job.config.image,
         }
         params.update(job.config.params)
+        params.update(job.platform_config.params)
         return params
 
     @classmethod


kernelci/runtime/__init__.py
```diff
@@ -134,6 +134,7 @@ def get_params(self, job, api_config=None):
             'runtime_image': job.config.image,
         }
         params.update(job.config.params)
+        params.update(job.platform_config.params)
```
gctucker, line 137: Right, and then we can remove the other `platform_config` then?

pawiecz: Note: I might be missing some context - please correct me if I'm wrong.

L137 adds to the template rendering `params` only the platform (`device_types`) `params` section, e.g.

```
device_types:
  minnowboard-turbot-E3826:
    arch: x86_64
    boot_method: grub
    mach: x86
    params:
      <added>: <param>
```

Required fields of the platform configuration (like `boot_method`, kernelci/legacy/config/test.py:L23) would still be unavailable unless duplicated in the `params` section.

I thought `platform_config.params` were meant to override any previously set value for a job (e.g. in a `job.config`).

I'll verify these assumptions and either post container image used or delete this comment in order not to cause any further confusion.

gctucker: Ah yes sorry, it's a different set of parameters.  I think I got mislead because the previous PR added the other line while removing this one.  So this is fine.

nuclearcat (approved):
Works on staging

PR #2173, opened 2023-11-07
CHROMEOS: test-configs-chromeos: add hp-14-db0003na-grunt

10ne1:
While flashing & testing Chrmoiumos R116 -> R118 upgrade,
it was noticed there is a grunt board which does not run
CrOS tests, so this adds the necessary definitions to be
able to install ChromeOS and run tests on it.

### DIFF
--- a/config/core/test-configs-chromeos.yaml
+++ b/config/core/test-configs-chromeos.yaml
@@ -582,7 +582,7 @@ test_plans:
 
 device_types:
 
-  acer-R721T-grunt_chromeos:
+  acer-R721T-grunt_chromeos: &chromebook-grunt-type
     base_name: acer-R721T-grunt
     mach: x86
     arch: x86_64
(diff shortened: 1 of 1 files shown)


config/core/test-configs-chromeos.yaml
```diff
@@ -790,6 +790,22 @@ device_types:
         modules: 'https://storage.chromeos.kernelci.org/images/rootfs/chromeos/chromiumos-sarien/20231106.0/amd64/modules.tar.xz'
       block_device: detect
 
+  hp-14-db0003na-grunt_chromeos:
+    <<: *chromebook-x86-type
```
a-wai, line 794: Maybe it would be interesting to have a `chromebook-grunt-type` ref? we have 3 of those boards with this commit, all sharing the same exact parameters, except for `base_name`.

config/core/test-configs-chromeos.yaml
```diff
@@ -1077,6 +1069,9 @@ test_configs:
       - cros-tast-video
       - cros-tast-video-fixed
 
+  - device_type: hp-14-db0003na-grunt_chromeos
```
a-wai, line 1072: Maybe move this a few lines below, to keep the list alphabetically sorted?

config/core/test-configs-chromeos.yaml
```diff
@@ -790,21 +790,13 @@ device_types:
         modules: 'https://storage.chromeos.kernelci.org/images/rootfs/chromeos/chromiumos-sarien/20231106.0/amd64/modules.tar.xz'
       block_device: detect
 
+  hp-14-db0003na-grunt_chromeos:
```
a-wai, line 793: Same comment, moving it after `hp-11A-G6-EE-grunt_chromeos` would keep the list alphabetically sorted.

a-wai: Ah, stupid linter, yes, it must consider `14` is a number and `11A` isn't, therefore it puts the former in 1st position... Not a big problem, so yeah, let's follow the linter recommendation, however stupid it looks to us ;)

a-wai:
Perfect, feel free to merge then :)

nuclearcat:
We need to get results on dashboard first, to make sure no mistakes are in config

nuclearcat:
Tests is on dashboard: https://chromeos.kernelci.org/test/plan/id/655359d5aef0c4491a954a25/

PR #2175, opened 2023-11-07
CHROMEOS: test-configs-chromeos: fix reference cherry dtb name

10ne1:
When this entry got added back in august, a copy+paste error resulted in the asurada dtb being used for cherry, leading to boot failures while flashing R118.


nuclearcat (approved):
LGTM

PR #2176, opened 2023-11-08
config/runtime: Add depthcharge boot method

nuclearcat:
Add depthcharge method to support Chromebooks booting.


gctucker:
The commit subject prefix is inconsistent with the rest of the history.  Maybe `config/runtime`...

pawiecz:
Verified together with kernelci/kernelci-pipeline#352 (and a minor fixup PR) on a local instance:

1. LAVA job definition was properly rendered
2. LAVA TestJob completed successfully: https://lava.collabora.dev/scheduler/job/12007421

LGTM

PR #2177, opened 2023-11-08
python.jinja2: Handle failure on json decode properly

nuclearcat:
Sometimes we have non-json response during kbuild, for example
```
[_upload_artifacts] Uploading artifacts to kbuild-gcc-10-x86-board-654b70f8c054242fbccbd592
    return complexjson.loads(self.text, **kwargs)
  File "/usr/lib/python3.9/json/__init__.py", line 346, in loads
    return _default_decoder.decode(s)
  File "/usr/lib/python3.9/json/decoder.py", line 337, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib/python3.9/json/decoder.py", line 355, in raw_decode
    raise JSONDecodeError("Expecting value", s, err.value) from None
json.decoder.JSONDecodeError: Expecting value: line 1 column 1 (char 0)

  File "/home/kernelci/data/output/tmplcp0maoa/kbuild-gcc-10-x86-board", line 383, in <module>
    results = main(sys.argv)
  File "/home/kernelci/data/output/tmplcp0maoa/kbuild-gcc-10-x86-board", line 371, in main
    detail = ex.response.json().get('detail')
  File "/usr/local/lib/python3.9/dist-packages/requests/models.py", line 975, in json
    raise RequestsJSONDecodeError(e.msg, e.doc, e.pos)
requests.exceptions.JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```
We need to handle it better for easier troubleshooting.

### DIFF
--- a/config/runtime/base/python.jinja2
+++ b/config/runtime/base/python.jinja2
@@ -111,7 +111,10 @@ def main(args):
             job.submit(results, NODE, API_CONFIG_YAML)
     except requests.exceptions.HTTPError as ex:
         print(ex, file=sys.stderr)
-        detail = ex.response.json().get('detail')
+        try:
+            detail = ex.response.json().get('detail')
+        except requests.exceptions.JSONDecodeError:
+            print(f"Non-JSON response, body: {ex.response.text}", file=sys.stderr)
         if detail:
             print(detail, file=sys.stderr)
         results = None


config/runtime/base/python.jinja2
```diff
@@ -111,7 +111,11 @@ def main(args):
             job.submit(results, NODE, API_CONFIG_YAML)
     except requests.exceptions.HTTPError as ex:
         print(ex, file=sys.stderr)
-        detail = ex.response.json().get('detail')
+        try:
+            detail = ex.response.json().get('detail')
+        except Exception:
+            detail = None
+            print(f"Non-JSON response, body: {ex.response}", file=sys.stderr)
```
gctucker, line 118: Sounds like the real issue here is when the detail is not JSON.  Maybe this happens when the API is returning a plain text response?

I don't think we should be handling this exception, it's just not meant to be happening and is more likely a bug in the API.  If we want a catch-all exception thing we can have it outside this whole part of the code.

nuclearcat: I was happening to me when one of tokens was missing, and it was really hard to troubleshoot even looking to logs. So it is definitely real issue and cover real case.

gctucker: Yes but every random failure will raise a different exception in another part of the code.  So maybe this was nice to help you debug things, I really think it should be dropped.  The real broken error handling seems to be on the API side for the reasons I've explained before.

nuclearcat: It might be various errors, sometimes request might not even reach python code, including 502 reponse from ingress/nginx, or something else, so we need to handle non-json response for such cases, so we dont guess what was inside the body of this reply.

gctucker: Right but that's not the same as failing to parse the JSON we're expecting from the API.  Anyway, I don't really mind...

gctucker: See #2182 (a new version of what we had in the legacy command line tools)

And yes it looks like we can't easily control the responses sent by the API when there's a problem.  We can add custom exception handlers and implement them to always return `JSONResponse` but I don't think that's really a good idea.

Maybe we could reuse the decorator in the Python job template, but not as a `kernelci.cli` decorator as that's really for the command line tools using Click and the jobs aren't in this category.  So it's probably better to just implement a similar error handling in the template for now and see if things can be consolidated later.

nuclearcat:
As i'm reworking now kbuild, i might work on this PR on new code, taking in consideration https://github.com/kernelci/kernelci-core/pull/2182

PR #2178, opened 2023-11-08
kernelci.cli: add pagination options and helper

gctucker:
Add the --page-length / -l and --page-number / -n options for paginated API data.  Provide a kernelci.cli.get_pagination() function to convert these arguments into the offset / limit values used with the API as pagination is easier with relative numbers on the command line.  Add unit tests accordingly to test this function.

### DIFF
--- a/kernelci/cli/__init__.py
+++ b/kernelci/cli/__init__.py
@@ -35,6 +35,14 @@ class Args:  # pylint: disable=too-few-public-methods
         '--indent', type=int,
         help="Intentation level for structured data output"
     )
+    page_length = click.option(
+        '-l', '--page-length', type=int,
+        help="Page length in paginated data"
+    )
+    page_number = click.option(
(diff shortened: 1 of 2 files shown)


JenySadadia:
I think this PR should also include changes to use `get_pagination` method for getting API objects in `_get_api_objs`.

gctucker:
If it's not preventing this PR from being merged then I wouldn't spend more time just to shuffle commits around.  I don't think it's related, the `get_paginated()` rework is separate from the command line arguments.

JenySadadia:
> If it's not preventing this PR from being merged then I wouldn't spend more time just to shuffle commits around. I don't think it's related, the `get_paginated()` rework is separate from the command line arguments.

Okay. So I believe it will be included in a separate PR.
We can merge this.

JenySadadia (approved):
LGTM.

PR #2180, opened 2023-11-08
kci.test: Fix generation of URLs for DTBs on 32 bit Arm

broonie:
Following a reorganisation of the DTBs into subdirectories for 32 bit Arm
we have had issues generating jobs for that architecture. We fixed
generation of jobs with 970c1982f3cb8561a9 ("kernelci.test: match configs
by dtb basename") but we are still generating jobs with URLs to DTBs that
omit the subdirectories the DTBs have been installed into, resulting in
none of the jobs actually running.

Fix this by substituting in the full path of the DTB from the contents list
in the metadata for the build. If a full path is already specified (eg, for
architectures like arm64 which always used subdirectories) then this will
do nothing.

### DIFF
--- a/kernelci/test.py
+++ b/kernelci/test.py
@@ -111,9 +111,17 @@ def get_params(meta, target, plan_config, storage, device_id):
     variant = target.variant
     dtb = dtb_full = target.dtb
     if dtb:
+        # Find which subidrectory (if any) of the kernel output the DTB is in
+        dtb_list = meta.get_single_artifact('dtbs', attr='contents')
+        for d in dtb_list:
+            (p, f) = os.path.split(d)
+            if f == target.dtb:
+                dtb = dtb_full = d
+
+        # Add the prefix we installed the DTBs into
         dtb_dir = meta.get_single_artifact('dtbs', attr='path')
         if dtb_dir:
-            dtb = dtb_full = os.path.join(dtb_dir, target.dtb)
+            dtb = dtb_full = os.path.join(dtb_dir, dtb)
             dtb = os.path.basename(dtb)  # hack for dtbs in subfolders
     publish_path = kernel['publish_path']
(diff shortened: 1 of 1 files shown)


broonie:
@jluebbe This hopefully will get the 32 bit Arm DTBs working again, it does the right thing for me in local testing.

jluebbe:
The change looks plausible to me, but I've not tested it.

nuclearcat:
Lets check on next staging run, i will be thankful if you can hint me for which devices to look, or can paste here such device/test example, then i will merge it.

broonie:
This affects all arm (32 bit, not arm64) devices that have a DTB.

nuclearcat:
It looks like it got tested:
Nov 09 00:02:28 kernelci-staging staging.kernelci.org[10080]: 2180 broonie          arm-dtb-subdirs                  OK

https://storage.staging.kernelci.org/kernelci/staging-stable/staging-stable-20231109.0/arm/multi_v7_defconfig/gcc-10/lab-clabbe/baseline-sun7i-a20-cubieboard2.html
https://storage.staging.kernelci.org/kernelci/staging-stable/staging-stable-20231109.0/arm/multi_v7_defconfig/gcc-10/lab-clabbe/baseline-sun8i-h3-orangepi-pc.html

jluebbe:
Thanks, it seems to work for our boards: https://storage.staging.kernelci.org/kernelci/staging-mainline/staging-mainline-20231108.1/arm/multi_v7_defconfig/gcc-10/lab-pengutronix/baseline-imx6dl-riotboard.html

broonie:
It's also working for the 32 bit Arm boards in my lab:for example  https://staging.kernelci.org/test/plan/id/654ca33b98720bd61ea3728b/ and https://staging.kernelci.org/test/plan/id/654cab09e611ace303a3728b/

PR #2182, opened 2023-11-10
kernelci.cli: add catch_http_error decorator

gctucker:
Port the catch_http_error decorator from the legacy implementation to the new kernelci.cli package based on Click.  This can be used with command functions to show a user error whenever an HTTP exception is caught.

### DIFF
--- a/kernelci/cli/__init__.py
+++ b/kernelci/cli/__init__.py
@@ -13,10 +13,13 @@ the Click framework, it adds support for loading default values and secrets
 from TOML settings.
 """
 
+import email.policy
+import functools
 import re
 import typing
 
 import click
+import requests
 
 import kernelci.settings
 
@@ -55,6 +58,28 @@ class Args:  # pylint: disable=too-few-public-methods
         '-v', '--verbose/--no-verbose', default=None,
         help="Print more details output"
     )
+
+
+def catch_http_error(func):
+    """Decorator to catch HTTPError exceptions and raise a ClickException"""
+    @functools.wraps(func)
+    def call(*args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+        except requests.exceptions.HTTPError as ex:
+            raw_content_type = ex.response.headers.get('content-type')
+            header = email.policy.EmailPolicy.header_factory(
+                'content-type', raw_content_type
+            )
+            detail = (
+                ex.response.json().get('detail')
+                if header.content_type == 'application/json'
+                else None
+            )
+            raise click.ClickException(
+                '\n'.join((str(ex), detail)) if detail else ex
+            ) from ex
+    return call
 
 
 class CommandSettings:


kernelci/cli/__init__.py
```diff
@@ -49,6 +52,28 @@ class Args:  # pylint: disable=too-few-public-methods
...
 
+def catch_http_error(func):
+    """Decorator to catch HTTPError exceptions and raise a ClickException"""
+    @functools.wraps(func)
+    def call(*args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+        except requests.exceptions.HTTPError as ex:
+            raw_content_type = ex.response.headers.get('content-type')
+            header = email.policy.EmailPolicy.header_factory(
```
JenySadadia, line 63: Sorry, I am not aware of `EmailPolicy`. Can you please explain what is it and why are we using it?

gctucker: It's part of the standard Python library and it's the only way to parse content types without adding third-party dependencies or doing it ourselves.

JenySadadia: When I tried to use raw content type instead, it also worked.
```
try:
            return func(*args, **kwargs)
except requests.exceptions.HTTPError as ex:
            raw_content_type = ex.response.headers.get('content-type')
            detail = (
                ex.response.json().get('detail')
                if raw_content_type == 'application/json'
                else None
            )
            raise click.ClickException(
                '\n'.join((str(ex), detail)) if detail else ex
            ) from ex
```

JenySadadia: Also from the documentation https://docs.python.org/3/library/email.policy.html it looks like this library's primary focus is to form and handle email messages which is not our use-case atm.

gctucker: Yes it works if it's not JSON, just if it's JSON then there should be a `detail` part.  If not then it's just using the string from the HTTP error.

And yes the email package is for handling emails, but like I wrote it's the only part of the Python standard library that has a way to parse content type strings.  I don't think there's any side effects here, it's a stable part of the library and we're not doing anything specific to emails.

JenySadadia: > When I tried to use raw content type instead, it also worked.
> 
> ```
> try:
>             return func(*args, **kwargs)
> except requests.exceptions.HTTPError as ex:
>             raw_content_type = ex.response.headers.get('content-type')
>             detail = (
>                 ex.response.json().get('detail')
>                 if raw_content_type == 'application/json'
>                 else None
>             )
>             raise click.ClickException(
>                 '\n'.join((str(ex), detail)) if detail else ex
>             ) from ex
> ```

I tested the above code snippet with `JSONResponse` and `PlainTextResponse`. It worked with both cases.
Values of `raw_content_type` and `header` are the same just the data type is different.

In the case of JSON:
```
raw_content_type: application/json
header: application/json
```

In case of plain text response:
```
raw_content_type: text/plain; charset=utf-8
header: text/plain; charset="utf-8"
```
So when I removed `header=email.policy.EmailPolicy.header_factory` and used `raw_content_type` directly to get `detail`, it worked OK.

gctucker: Well yes, but then you need to parse `text/plain; charset=utf-8` and `application/json`.  So the whole point of using the standard library method is to avoid having to do this ourselves.

gctucker: There's quite a bit of code there we can rely upon:
https://github.com/python/cpython/blob/main/Lib/email/headerregistry.py#L562

Doing this ourselves wouldn't seem wise I think.  For example, it takes care of the part of the content-type header that specifies encoding and also accepts both upper-case / lower-case, basically it implements it as specified by internet standards.

JenySadadia: Okay. Makes sense.

JenySadadia (approved):
Tested OK locally.

PR #2183, opened 2023-11-10
Add `kci user` commands (was #2157)

gctucker:
Replaces #2157

PR #2184, opened 2023-11-10
`kci user` command fixup

gctucker:
Fix and improve the `kci user` command implementation.

### DIFF
--- a/kernelci/cli/user.py
+++ b/kernelci/cli/user.py
@@ -200,7 +200,7 @@ def update_by_username(username, email,  # pylint: disable=too-many-arguments
         raise click.ClickException(
             "Sorry, user does not exist with given username")
     user_id = users[0]["id"]
-    data = api.update_user_by_id(user_id, user)
+    data = api.update_user(user, user_id)
     click.echo(json.dumps(data, indent=indent))
 
 
@@ -214,7 +214,7 @@ def activate(user_id, config, api, secrets, indent):
     configs = kernelci.config.load(config)
     api_config = configs['api'][api]
     api = kernelci.api.get_api(api_config, secrets.api.token)
-    data = api.update_user_by_id(user_id, {"is_active": 1})
+    data = api.update_user({"is_active": 1}, user_id)
     click.echo(json.dumps(data, indent=indent))
 
 
@@ -228,7 +228,7 @@ def deactivate(user_id, config, api, secrets, indent):
     configs = kernelci.config.load(config)
     api_config = configs['api'][api]
     api = kernelci.api.get_api(api_config, secrets.api.token)
-    data = api.update_user_by_id(user_id, {"is_active": 0})
+    data = api.update_user({"is_active": 0}, user_id)
     click.echo(json.dumps(data, indent=indent))
 
 
--- a/kernelci/cli/user.py
+++ b/kernelci/cli/user.py
@@ -14,7 +14,7 @@ import click
 
 import kernelci.api
 import kernelci.config
-from . import Args, kci, split_attributes
+from . import Args, kci, split_attributes, catch_http_error
 
 
 @kci.group(name='user')
@@ -26,6 +26,7 @@ def kci_user():
 @Args.config
 @Args.api
 @Args.indent
+@catch_http_error
 def whoami(config, api, indent, secrets):
     """Get the current user's details with API authentication"""
     configs = kernelci.config.load(config)
@@ -40,6 +41,7 @@ def whoami(config, api, indent, secrets):
 @Args.config
 @Args.api
 @Args.indent
+@catch_http_error
 def find(attributes, config, api, indent):
     """Find user profiles with arbitrary attributes"""
     configs = kernelci.config.load(config)
@@ -55,57 +57,50 @@ def find(attributes, config, api, indent):
 @click.argument('username')
 @Args.config
 @Args.api
-@Args.indent
-def token(username, config, api, indent):
+@catch_http_error
+def token(username, config, api):
     """Create a new API token using a user name and password"""
-    password = getpass.getpass()
     configs = kernelci.config.load(config)
     api_config = configs['api'][api]
     api = kernelci.api.get_api(api_config)
-    user_token = api.create_token(username, password)
-    click.echo(json.dumps(user_token, indent=indent))
-
-
-@kci_user.group(name='password')
-def user_password():
-    """Manage user passwords"""
+    password = getpass.getpass()
+    api_token = api.create_token(username, password)
+    click.echo(api_token['access_token'])
 
 
 @kci_user.command(secrets=True)
-@click.option('--username')
-@click.option('--email')
-@click.option('--group', multiple=True, help="User group(s)")
+@click.argument('attributes', nargs=-1)
+@click.option('--username', help="Username of the user to update (admin only)")
 @Args.config
 @Args.api
-@Args.indent
-def update(username, email, config,  # pylint: disable=too-many-arguments
-           api, secrets, group, indent):
-    """Update own user account"""
-    user = {}
-    if username:
-        user['username'] = username
-    if email:
-        user['email'] = email
-    if group:
-        user['groups'] = group
-    if not user:
-        raise click.ClickException("Sorry, nothing to update")
+@catch_http_error
+def update(attributes, username, config, api, secrets):
+    """Update user account data with the provided attributes"""
+    fields = split_attributes(attributes)
+    if not fields:
+        click.echo("No user details to update.")
+        return
+
     configs = kernelci.config.load(config)
     api_config = configs['api'][api]
     api = kernelci.api.get_api(api_config, secrets.api.token)
-    data = api.update_user(user)
-    click.echo(json.dumps(data, indent=indent))
+    if username:
+        users = api.get_users({"username": username})
+        if not users:
+            raise click.ClickException(f"User not found: {username}")
+        user_id = users[0]['id']
+    else:
+        user_id = None
+    api.update_user(fields, user_id)
 
 
 @kci_user.command(secrets=True)
 @click.argument('username')
 @click.argument('email')
-@click.option('--group', multiple=True, help="User group(s)")
 @Args.config
 @Args.api
-@Args.indent
-def add(username, email, config,  # pylint: disable=too-many-arguments
-        api, secrets, group, indent):
+@catch_http_error
+def add(username, email, config, api, secrets):
     """Add a new user account"""
     password = getpass.getpass()
     retyped = getpass.getpass("Confirm password: ")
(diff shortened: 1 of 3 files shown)


gctucker:
I'll fix the Pylint errors on Monday...

kernelci/cli/user.py
```diff
@@ -55,57 +57,50 @@ def find(attributes, config, api, indent):
...
-    if email:
-        user['email'] = email
-    if group:
-        user['groups'] = group
-    if not user:
-        raise click.ClickException("Sorry, nothing to update")
+@catch_http_error
+def update(attributes, username, config, api, secrets):
+    """Update user account data with the provided attributes"""
+    fields = split_attributes(attributes)
```
JenySadadia, line 79:
```
$ ./kci user update
  File "/home/jeny/kernelCI/kernelci-core_env/src/gctucker/kernelci-core/./kci", line 28, in <module>
    kci()  # pylint: disable=no-value-for-parameter
  File "/home/jeny/kernelCI/kernelci-core_env/lib/python3.10/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/home/jeny/kernelCI/kernelci-core_env/lib/python3.10/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/home/jeny/kernelCI/kernelci-core_env/lib/python3.10/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/jeny/kernelCI/kernelci-core_env/lib/python3.10/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/jeny/kernelCI/kernelci-core_env/src/gctucker/kernelci-core/kernelci/cli/__init__.py", line 132, in invoke
    super().invoke(ctx)
  File "/home/jeny/kernelCI/kernelci-core_env/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/jeny/kernelCI/kernelci-core_env/lib/python3.10/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/home/jeny/kernelCI/kernelci-core_env/src/gctucker/kernelci-core/kernelci/cli/__init__.py", line 60, in call
    return func(*args, **kwargs)
  File "/home/jeny/kernelCI/kernelci-core_env/src/gctucker/kernelci-core/kernelci/cli/user.py", line 79, in update
    fields = split_attributes(attributes)
  File "/home/jeny/kernelCI/kernelci-core_env/src/gctucker/kernelci-core/kernelci/cli/__init__.py", line 197, in split_attributes
    for attribute in attributes:
TypeError: 'NoneType' object is not iterable
```

A check seems to be missing here.
```
if attributes:
    fields = split_attributes(attributes)
```

gctucker: Yes there's another PR to fix this: #2185

Ah it's merged already, so I'll rebase this PR.

JenySadadia:
User update with `groups` attribute is not working:
```
$ ./kci user update --username test groups=kernelci email=test@gmail.com username=test
  File "/home/jeny/kernelCI/kernelci-core_env/src/gctucker/kernelci-core/kernelci/cli/__init__.py", line 60, in call
    return func(*args, **kwargs)
  File "/home/jeny/kernelCI/kernelci-core_env/src/gctucker/kernelci-core/kernelci/cli/user.py", line 94, in update
    api.update_user(fields, user_id)
  File "/home/jeny/kernelCI/kernelci-core_env/src/gctucker/kernelci-core/kernelci/api/latest.py", line 164, in update_user
    return self._patch(f'user/{user_id or "me"}', fields).json()
  File "/home/jeny/kernelCI/kernelci-core_env/src/gctucker/kernelci-core/kernelci/api/__init__.py", line 241, in _patch
    resp.raise_for_status()
  File "/home/jeny/kernelCI/kernelci-core_env/lib/python3.10/site-packages/requests/models.py", line 1021, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 422 Client Error: Unprocessable Entity for url: http://172.17.0.1:8001/latest/user/652cfd850a9d4984577826c8

  File "/home/jeny/kernelCI/kernelci-core_env/src/gctucker/kernelci-core/./kci", line 28, in <module>
    kci()  # pylint: disable=no-value-for-parameter
  File "/home/jeny/kernelCI/kernelci-core_env/lib/python3.10/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/home/jeny/kernelCI/kernelci-core_env/lib/python3.10/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/home/jeny/kernelCI/kernelci-core_env/lib/python3.10/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/jeny/kernelCI/kernelci-core_env/lib/python3.10/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/jeny/kernelCI/kernelci-core_env/src/gctucker/kernelci-core/kernelci/cli/__init__.py", line 132, in invoke
    super().invoke(ctx)
  File "/home/jeny/kernelCI/kernelci-core_env/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/jeny/kernelCI/kernelci-core_env/lib/python3.10/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/home/jeny/kernelCI/kernelci-core_env/src/gctucker/kernelci-core/kernelci/cli/__init__.py", line 72, in call
    '\n'.join((str(ex), detail)) if detail else ex
TypeError: sequence item 1: expected str instance, list found

```

gctucker:
Ah yes sorry, I didn't test support for groups.  I thought a separate subcommand would be used for this, well it's a grey area.  Probably the group operations should be made part of the user operations.  That would also clarify things around the term `group` to be clear it's about user groups.  For example `kci user group add` is clearer than `kci group add`, and same thing with the API bindings and maybe even the endpoints should reflect this.

JenySadadia:
In https://github.com/kernelci/kernelci-core/pull/2157 I used separate optional argument `--groups` with `multiple=True` option and `Click` automatically converted all the values to a tuple and I was able to pass it to the `requests.patch` directly. It was quite simple and worked fine.

gctucker:
Yes, the problem is that adding arguments for every field of the User model is very fragile and not really practical if new fields get added to the model.

The `split_attributes()` method needs to be fixed anyway in order to handle the same attribute name multiple times, so that could be a more sustainable solution.

JenySadadia:
That's true.

gctucker:
Converted to draft as it depends on #2183

gctucker:
OK this one is ready for review now @JenySadadia

gctucker:
The part about groups isn't complete yet but that will be resolved with follow-up PRs.

kernelci/cli/user.py
```diff
@@ -55,57 +57,50 @@ def find(attributes, config, api, indent):
...
+        user_id = users[0]['id']
+    else:
+        user_id = None
+    api.update_user(fields, user_id)
 
 
 @kci_user.command(secrets=True)
 @click.argument('username')
 @click.argument('email')
-@click.option('--group', multiple=True, help="User group(s)")
```
JenySadadia, line 103: Why `group` has been eliminated?

kernelci/cli/user.py
```diff
@@ -115,127 +110,94 @@ def add(username, email, config,  # pylint: disable=too-many-arguments
         'username': username,
         'email': email,
         'password': password,
-        'groups': group
     }
     configs = kernelci.config.load(config)
     api_config = configs['api'][api]
     api = kernelci.api.get_api(api_config, secrets.api.token)
-    data = api.create_user(user)
-    click.echo(json.dumps(data, indent=indent))
```
JenySadadia, line 124: I think we should keep this to let the user know that a new user has been added.

kernelci/cli/user.py
```diff
@@ -115,127 +110,94 @@ def add(username, email, config,  # pylint: disable=too-many-arguments
...
+    api.create_user(user)
 
 
 @kci_user.command
-@click.argument('email')
```
JenySadadia, line 128: We can send `email` directly for the verification. Why send username instead?

kernelci/cli/user.py
gctucker: Because it' just one of the many attributes you can update in a user profile.  With the rework of `split_attributes()` you can now pass `group=a group=b` and have groups `a` and `b` set for a user (pending further changes to separate the internal model with the database group id and the public schema with just group names on the API side, but that's beyond the scope of this PR).

kernelci/cli/user.py
gctucker: The command provides the user name so it's not necessary to print anything as a result.  We might want to add a verbose flag, but I think it just complicates the code for nothing.  If you want to check what was done after `kci user add bob` you can do `kci user find username=bob`.

kernelci/cli/user.py
gctucker: Because it's easier from a user point of view to provide the username than the email, if the main way of referring to a user is their username.

I think an improvement here might be to have something that can be either a username or an email e.g. a `user` argument and then try to find a user with either the username or the email matching the provided string.  A username doesn't allow characters such as `@` (if I remember correctly) so we can't have a username colliding with another user's email - or we may want to have a unique index on these 2 fields anyway.

gctucker: The key thing here is that the CLI needs to be optimal from a user point of view, and it shouldn't be derived from the API endpoints or bindings or how things might be implemented.  Ideally these things would line up neatly, but whenever there's a gap then usability on the command line should be the priority.  So if a user is more likely to know their username than their email address, or typically would have a shorter username than email address, then relying on the username everywhere is better even if it means an extra query to the API (and this isn't meant to be a high-traffic endpoint like the web dashboard would be doing).

kernelci/cli/user.py
JenySadadia: Yes, but there should be a provision for adding users to groups at the time of creating users.

kernelci/cli/user.py
JenySadadia: Okay.

kernelci/cli/user.py
JenySadadia: IMO username and email both are key fields of any user account. Hence providing email directly would be as easy as providing a username.  Anyway if providing username seems optimal from user's pov, then it's fine.

JenySadadia: > I think an improvement here might be to have something that can be either a username or an email e.g. a user argument and then try to find a user with either the username or the email matching the provided string. A username doesn't allow characters such as @ (if I remember correctly) so we can't have a username colliding with another user's email - or we may want to have a unique index on these 2 fields anyway.

The provision of letting the user provide username or email would complicate the code. I think it should allow just one specific argument.

gctucker: Well it's not that hard:
```python
    users = (
        api.user.find({"username": username}) or
        api.user.find({'email': email})
    )
```

kernelci/cli/user.py
gctucker: Why would this be required?  Is there a case where creating a user first, and then adding it to a group is not going to work?

JenySadadia: Of course, it is going to work. My point is if we can add groups in the same request while creating a user, why should we add an extra request for it?

gctucker: Because then, where do we stop if we want to be able to set up all the fields of a user when creating it?

We may soon have notification preferences, social network profile links, biography, profile picture... these are all at the same level as "groups", they are arbitrary properties.  I guess `man useradd` will show you what adding a Unix user account involves, but that's pretty much set in stone now and simpler than a web service user account.  Specifying the groups was required when `admin` had a special functional meaning, now groups don't really matter.  The only required fields now really are `email` and `username`.

JenySadadia: Basically, when we create a user account on any website, we usually can add all the available fields at the time of account creation. Currently, we just have one additional field i.e. groups. So I guess it is easy to include it in the `add` command as well.
Anyway, this is just another approach to solving the issue. We can go your way too. Merging this now to wrap this up.

JenySadadia (approved):

kernelci/cli/user.py
gctucker: Yes, to be fair this is a bit of a grey area and somewhat subject to personal taste etc.  My approach here is to try and keep the number of commands and arguments to the bare minimum as I expect additional ones to be required when we hit issues in corner cases.  If you look at the legacy command line tools, they have lots of options and are barely usable by hand - only in scripts or Jenkins jobs.  If the new command line tools tend to become more complex too over the months then it helps to start with the simplest possible implementation that still provides all the required features.  No doubt this will need to be reworked a few times anyway as things evolve.

PR #2185, opened 2023-11-13
kernelci.cli: handle no attributes in split_attributes()

gctucker:
As a convenience, since when no attributes are passed on the command line Click gives a None rather than an empty list, deal with the case where attributes are None in split_attributes() as if it was an empty list.

Update unit tests accordingly to take this use-case into account.

Fixes: 0a62fb90d404 ("kernelci.cli: rewrite split_attributes() method")


nuclearcat (approved):
Looks good to me, tests ok, staging ok, and run few commands, obviously ok too.

PR #2186, opened 2023-11-13
CHROMEOS build_board: use board overlays instead of harcoding values

a-wai:
The artifacts extraction currently acts based on hardcoded board names, which can be misleading and lead to incorrect results when working on a new board. Moreover, it copies the kernel image as `bzImage` no matter what the architecture is, although only x86 devices should provide this kind of kernel image.

This change reworks artifact extraction by inspecting the CrOS sources and leverages the board overlay config to find out:
* what the board architecture is so we can extract the correct kernel image (`bzImage` or `Image`)
* what vendor is the chipset from and which DTBs should be extracted on ARM64


nuclearcat (approved):
Tested by team during board builds.

PR #2189, opened 2023-11-14
Rework `kernelci.api.API` class internals to have only abstract bindings

gctucker:
Move the helper methods out of the API() abstract class to make the actual bindings clearer.  This will also make it possible to have nested classes for different parts of the API e.g. `api.node.find()` and `api.node.find()` rather than all the methods in a single ever-growing class.

### DIFF
--- a/kernelci/api/__init__.py
+++ b/kernelci/api/__init__.py
@@ -108,6 +108,30 @@ class Base:
         resp.raise_for_status()
         return resp
 
+    def _get_paginated(self, input_params, path, offset=None, limit=None):
+        params = input_params.copy()
+
+        if any((offset, limit)):
+            params.update({
+                'offset': offset or None,
+                'limit': limit or None,
+            })
+            return self._get(path, params=params)
+
+        objs = []
+        offset = 0
+        limit = 100
+        params['limit'] = limit
+        while True:
+            params['offset'] = offset
+            resp = self._get(path, params=params)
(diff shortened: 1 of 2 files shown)


JenySadadia:
Could you please rebase this? @gctucker

JenySadadia:
Is this ready for review?

gctucker:
Yes, I just made it a draft because #2184 had to be merged first. I'll rebase it, thanks.

kernelci/api/__init__.py
```diff
@@ -31,9 +31,120 @@ def __init__(self, config: kernelci.config.api.API, token: str):
...
+        params['limit'] = limit
+        while True:
+            params['offset'] = offset
+            resp = self._get(path, params=params)
+            items = resp.json()['items']
+            objs.extend(items)
+            if len(items) < limit:
+                break
+            offset += limit
+        return objs
```
JenySadadia, line 133: Shouldn't this be in the `API` class?

gctucker: No, because then nested classes can't access it.  That's why I created a separate Base class.  Then the API class as well as the nested ones for separate parts of the API can all inherit from it.

gctucker:
I'll create the PR with the set of changes that comes after this, it'll make more sense as this one is more a stepping stone in this direction.

JenySadadia (approved):
LGTM.

PR #2190, opened 2023-11-14
Fix linux-firmware dependencies

nuclearcat:
Fix dependencies and also produce error if in future copy-firmware script quit with error.


nuclearcat:
No build failures observed anymore on staging, merging.

PR #2193, opened 2023-11-17
rootfs-images.yaml: Update fluster rootfs image

nuclearcat:
This image built by new rootfs script, verify it in staging, and if no problem observed - merge.

PR #2194, opened 2023-11-17
CHROMEOS rootfs-images.yaml: Update fluster rootfs image

nuclearcat:
This image built by new rootfs script, verify it in staging, and if no problem observed - merge.


nuclearcat:
Tested in https://github.com/kernelci/kernelci-core/pull/2193

PR #2195, opened 2023-11-17
kernelci.cli: allow multiple values for .split_attributes()

gctucker:
Allow the same attribute to be provided several times in .split_attributes().  This is useful when looking for database entries that match one of a set of values e.g. name=foo?name=bar to find entries named 'foo' or 'bar'.

Update unit tests accordingly with an additional one to test the generated request URL with multiple values for the same parameter.

### DIFF
--- a/kernelci/cli/__init__.py
+++ b/kernelci/cli/__init__.py
@@ -208,20 +208,14 @@ def split_attributes(attributes: typing.List[str]):
         if not match:
             raise click.ClickException(f"Invalid attribute: {attribute}")
         name, operator, value = match.groups()
-        ex_op, ex_value = parsed.get(name, (None, None))
-        if ex_value:
-            raise click.ClickException(
-                f"Conflicting values for {name}: \
-                {name}{value}, {ex_op}{ex_value}"
-            )
         opstr = operators.get(operator)
         if opstr is None:
             raise click.ClickException(f"Invalid operator: {operator}")
-        parsed[name] = (opstr, value)
+        parsed.setdefault(''.join((name, opstr)), []).append(value)
 
     return {
-        ''.join((key, opstr)): value
-        for key, (opstr, value) in parsed.items()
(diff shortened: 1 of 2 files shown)


gctucker:
No it's not implemented on the API side yet.

JenySadadia:
Tested locally:
```
$ ./kci user find username=bob username=admin1
[]
```
This doesn't return the user with the name `bob` as I don't have any user named `admin1`. I think it looks only for the last value it receives. I guess this needs a fix on the API side.

JenySadadia:
I thought we were trying to fix the issue I mentioned in https://github.com/kernelci/kernelci-core/pull/2184#issuecomment-1814427201. When multiple values for `--group` are received, it should combine all the values into a list/tuple and send it to API. Just like what `Click` does with options like `@click.option('--group', multiple=True)`.

gctucker:
Yes, I'm not sure if the API can handle things properly.  But at least the logic here on the client side is working to produce the right URL.

JenySadadia:
This works when I send multiple `groups` attributes in a request:
```
$ ./kci user update --username=test1 username=test groups=kernelci groups=admin
fields {'username': 'test', 'groups': ['kernelci', 'admin']}
```

When I send just one `groups`, it is not converting it to a list.  API returns an error as it expects a list for `groups` attribute:
```
$ ./kci user update --username=test groups=kernelci
fields {'groups': 'kernelci'}
requests.exceptions.HTTPError: 422 Client Error: Unprocessable Entity for url: http://172.17.0.1:8001/latest/user/652cfd850a9d4984577826c8
```
I guess `split_attributes` is working as expected and `groups` is a special case and needs to be handled separately.

JenySadadia (approved):
Handling `groups` can be a follow-up PR. Merging this now.

gctucker:
Ah, yes the whole group handling thing will need to be revisited. That can be done as part of the work to add the `kci user group` commands I guess.

PR #2196, opened 2023-11-17
CHROMEOS: test-configs-chromeos: add some new tests in R118

10ne1:
These are new tests added in R118 which are known to work. We can extend coverage more, especially on the storage side but we need some bugs like b:311421846 fixed first.

### DIFF
--- a/config/core/test-configs-chromeos.yaml
+++ b/config/core/test-configs-chromeos.yaml
@@ -380,7 +380,6 @@ test_plans:
     params: &cros-tast-perf-params
       <<: *cros-tast-base-params
       tests: &cros-tast-perf-tests >
-        ui.IdlePerf
         ui.DesksAnimationPerf
         ui.DragTabInTabletPerf.touch
         ui.OverviewWithExpandedDesksBarPerf
--- a/config/core/test-configs-chromeos.yaml
+++ b/config/core/test-configs-chromeos.yaml
@@ -381,23 +381,29 @@ test_plans:
       <<: *cros-tast-base-params
       tests: &cros-tast-perf-tests >
         ui.IdlePerf
-        ui.WindowCyclePerf
-        ui.WindowResizePerf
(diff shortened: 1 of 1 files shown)


a-wai:
I don't see anything wrong here, I assume you've tested those on multiple devices/architectures?

nuclearcat (changes requested):
cros-tast-platform: https://chromeos.kernelci.org/test/plan/id/6559791fe923647cc5954a2c/
cros-tast-perf: Not finishing within 30 minutes, which is very excessive, please remove too "long running tests"
Example: https://storage.chromeos.kernelci.org/kernelci/linux-6.1.y-arm64-chromeos/linux-6.1.y-arm64-chromeos-20231119.0/x86_64/cros---chromeos-6.1-x86_64-chromeos-intel-pineview.flavour.config+x86-chromebook+CONFIG_MODULE_COMPRESS_GZIP=n/clang-17/lab-collabora/cros-tast-perf-asus-C523NA-A20057-coral_chromeos.html
More here: https://chromeos.kernelci.org/test/job/kernelci/branch/linux-6.1.y-arm64-chromeos/kernel/linux-6.1.y-arm64-chromeos-20231119.0/plan/cros-tast-perf/

a-wai:
> I guess we can split the tests on the 30 seconds boundary, separate all tests that take >30s into a new config and keep the quicker tests.

Thanks for having a look! That makes sense, even going with a 1min boundary would be fine by me (it would translate to ~5min of running tests), but I don't have a strong opinion on the limit.

nuclearcat:
i suggest to split tests and create one more tests set, like cros-tast-perf2

nuclearcat:
https://chromeos.kernelci.org/test/job/kernelci/branch/chromeos-stable/kernel/chromeos-stable-20231123.0/plan/cros-tast-perf-long-duration/
Tests appeared, thanks!

PR #2198, opened 2023-11-17
CHROMEOS test-configs-chromeos: run same tests on x86 and arm64 chromebooks

a-wai:
Some Tast tests aren't run on x86 chromebooks, making it difficult to come up with a unified reporting format.

Moreover, having a `-fixed` version of each Tast test plan is useful to track regressions between the ChromeOS kernel and mainline, so add the missing `cros-tast-ui-fixed`.

### DIFF
--- a/config/core/test-configs-chromeos.yaml
+++ b/config/core/test-configs-chromeos.yaml
@@ -501,11 +501,11 @@ test_plans:
       <<: *cros-tast-sound-params
       fixed_kernel: true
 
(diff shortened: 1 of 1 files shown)


nuclearcat (approved):
cros-tast-ui-fixed being run: https://chromeos.kernelci.org/test/job/kernelci/branch/chromeos-stable/kernel/chromeos-stable-20231118.1/plan/cros-tast-ui-fixed/

PR #2200, opened 2023-11-17
build-configs-android.yaml: Enable allmodconfig

nuclearcat:
As per Todd Kjos email we are gradually enabling allmodconfig, then if system can handle load properly, we will enable allyesconfig.


gctucker:
> if system can handle load properly, we will enable allyesconfig.

If not then Todd can provide us with another cluster.

PR #2203, opened 2023-11-21
Rework `kci pubsub` commands with `Click` framework

JenySadadia:
Rework CLI tools to interact with the API PubSub interface with `Click` framework.
Added `kci event` commands.


kernelci/cli/event.py
```diff
@@ -0,0 +1,94 @@
...
+    api_config = configs['api'][api]
+    api = kernelci.api.get_api(api_config, secrets.api.token)
+    api.unsubscribe(sub_id)
+
+
+@kci_event.command(secrets=True)
+@click.option('--event-type',
+              help="CloudEvent type, api.kernelci.org by default")
+@click.option('--source',
+              help="CloudEvent source, https://api.kernelci.org/ by default")
```
gctucker, line 54: These should probably be coming from the API YAML config object.

kernelci/cli/event.py
```diff
@@ -0,0 +1,94 @@
...
+@click.option('--json-data', help="Parse input data as JSON", is_flag=True)
```
gctucker, line 55: I think `--is-json` is better.

kernelci/cli/event.py
```diff
@@ -0,0 +1,94 @@
...
+    attributes['data'] = data
+    api.send_event(channel, attributes)
+
+
+@kci_event.command(secrets=True)
+@click.argument('sub_id')
+@Args.config
+@Args.api
+@Args.indent
+def receive(config, api, indent, sub_id, secrets):
```
gctucker, line 82: How about a timeout?  Maybe the API doesn't handle that yet actually.

gctucker (commented):
OK this doesn't change any of the bindings so it can be done in parallel of the API class rework.  Then when adding the queue commands etc. with new bindings, it would be good to do it directly on top of the reworked API class with a nested class for events.

kernelci/cli/event.py
JenySadadia: That's true.

kernelci/cli/event.py
```diff
@@ -0,0 +1,94 @@
...
+
+import kernelci.api
+import kernelci.api.helper
+import kernelci.config
+from . import Args, kci
+
+
+@kci.group(name='event')
+def kci_event():
+    """Interact with Pub/Sub interface"""
```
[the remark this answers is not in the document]
JenySadadia: Fixed the docstrings.

kernelci/cli/event.py
JenySadadia: Do you mean adding arguments to `kernelci.cli.Args` class? But the class holds common command line arguments and these are specific to this command only.

kernelci/cli/event.py
JenySadadia: `PubSub.listen` handler has set static value for timeout i.e. 1 sec. I also used the same value for blocking pop with Redis List.

kernelci/cli/event.py
gctucker: No I mean removing them as arguments.  There's already the `--api` argument and I think these things should be fixed for each API YAML config entry.

kernelci/cli/event.py
gctucker: Right, so there's no support for it yet.  That's fine, we can add it later.

kernelci/cli/event.py
JenySadadia: Makes sense. We should let API set event type and source instead. I'll drop the arguments.

JenySadadia: We can also add fixed entries to TOML but it makes more sense if API sets the source and type of the events instead.

kernelci/cli/event.py
```diff
@@ -0,0 +1,83 @@
...
+@click.option('--is-json', help="Parse input data as JSON", is_flag=True)
+@Args.config
+@Args.api
+@click.argument('channel')
+def send(config, api, is_json, channel, secrets):
+    """Read some data on stdin and send it as an event on a channel"""
+    configs = kernelci.config.load(config)
+    api_config = configs['api'][api]
+    api = kernelci.api.get_api(api_config, secrets.api.token)
+    data = sys.stdin.read()
```
gctucker, line 60: By the way, it's also possible to have a File argument with Click.  When the file value is `-` it will read from stdin.  This works, but maybe we should consider using File arguments in all the commands.  Here's the implementation I made a couple of weeks ago:
```python
@kci_node.command(secrets=True)
@click.argument('input_file', type=click.File('r'))
@Args.config
@Args.api
@Args.indent
@catch_http_error
def submit(input_file, config, api, indent, secrets):
    data = json.load(input_file)
    api = _get_api(config, api, secrets)
    func = api.node.update if 'id' in data else api.node.add
    node = func(data)
    click.echo(json.dumps(node, indent=indent or None))
```

JenySadadia: This can be included in a follow-up PR.

nuclearcat (approved):
Tested on staging: 
subscribe, receive, unsubscribe, send

PR #2207, opened 2023-11-24
test-configs.yaml: Enable networking selftests

broonie:
We don't currently run the networking selftests since they don't play
nicely with NFS root filesystems as a result of reconfiguring networking.
This isn't an issue with emulated platforms where we use a disk image as
our root filesystem so let's pick one and run the tests there. They're
pretty heavyweight so give it a large timeout.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -414,6 +414,12 @@ test_plans:
       job_timeout: '10'
       kselftest_collections: "net"
 
+  kselftest-net_qemu:
+    <<: *kselftest_qemu
+    params:
+      job_timeout: '90'
+      kselftest_collections: "net"
+
   kselftest-openat2:
     <<: *kselftest
     params:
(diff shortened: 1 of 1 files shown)


broonie:
Tests running in staging https://staging.kernelci.org/test/job/kernelci/branch/staging-next/kernel/staging-next-20231126.0/plan/kselftest-net_qemu/ - there's issues with some of the tests but a lot are working well, the failing tests will need investigating and either fixing in the kernel source or with some extra dependencies in the rootfs.

broonie:
We don't deploy modules into the initramfs which causes a bunch of the issues here, LAVA struggles with CPIO but we could deploy an actual rootfs image instead. Seems like a separate issue though.

nuclearcat (approved):
Sorry for delay and thanks for efforts, it is working for quite a while: https://staging.kernelci.org/test/job/kernelci/branch/staging-mainline/kernel/staging-mainline-20231213.0/plan/kselftest-net_qemu/

PR #2221, opened 2023-12-03
build-configs.yaml: Add rust-for-linux-samples to rust-next

aliceinwire:
rust-for-linux-samples was used for building rust test modules, recently unused.
enable it for rust-next as it have substitute the previous rust branch.

### DIFF
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -1460,7 +1460,7 @@ build_configs:
     variants:
       rustc-1.73:
         build_environment: rustc-1.73
-        fragments: [rust, rust-samples, kselftest]
+        fragments: [rust, rust-for-linux-samples, rust-samples, kselftest]
         architectures:
           x86_64:
             base_defconfig: 'x86_64_defconfig'
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -1463,7 +1463,7 @@ build_configs:
         fragments: [rust, rust-for-linux-samples, rust-samples, kselftest]
         architectures:
           x86_64:
-            base_defconfig: 'x86_64_defconfig'
+            base_defconfig: 'x86_64_defconfig+rust'
 
   samsung:
     tree: samsung


aliceinwire:
looks like the build got the correct defconfig+rust for each one
I should remove rust from the building_configs in this case 
```
+ kci_build tree_branch --build-config=rust-for-linux_rust-next
[Pipeline] echo
0 x86_64 x86_64_defconfig+rust+rust-for-linux-samples rustc-1.73
[Pipeline] echo
1 x86_64 x86_64_defconfig+rust rustc-1.73
[Pipeline] echo
2 x86_64 x86_64_defconfig+rust+rust rustc-1.73
[Pipeline] echo
3 x86_64 x86_64_defconfig+rust+rust-samples rustc-1.73
[Pipeline] echo
4 x86_64 x86_64_defconfig+rust+kselftest rustc-1.73
```

nuclearcat (approved):
LGTM, https://staging.kernelci.org/build/rust-for-linux/branch/rust-next/kernel/rust-6.7/
I can see rust-for-linux-samples build.

gctucker:
@nuclearcat @aliceinwire Out of interest, did you discuss this with @ojeda?

ojeda:
Thanks @gctucker for the ping.

I am confused: the `rust-for-linux-samples` fragment contains configs from the `rust` branch -- it should probably be deleted at this point. I mean, it is possible we could have those exact configs used in the future when we upstream more code, but they could change (or not). It may be a better idea to add them back when we start to use them. For instance, `CONFIG_SAMPLE_RUST_PRINT` can be added to `rust-samples`.

In addition, what does the `+rust` in the other commit do? i.e. how does it affect the build? I don't see it in any other `defconfig:` key. I see it in `extra_configs:`, though.

aliceinwire:
> Thanks @gctucker for the ping.
> 
> I am confused: the `rust-for-linux-samples` fragment contains configs from the `rust` branch -- it should probably be deleted at this point. I mean, it is possible we could have those exact configs used in the future when we upstream more code, but they could change (or not). It may be a better idea to add them back when we start to use them. For instance, `CONFIG_SAMPLE_RUST_PRINT` can be added to `rust-samples`.

yes, you are correct rust-for-linux-samples need to be merged with rust-samples

> 
> In addition, what does the `+rust` in the other commit do? i.e. how does it affect the build? I don't see it in any other `defconfig:` key. I see it in `extra_configs:`, though.

the +rust is needed for enabling rust-for-linux configuration, without that all the kernel build are created without 'CONFIG_RUST=y' https://github.com/aliceinwire/kernelci-core/blob/b8a13ed33d8ab0416b530f20f976b6e151db56f3/config/core/build-configs.yaml#L445

I'm discussing about this on my presentation
https://ossjapan2023.sched.com/event/5d8bea9df5b513d9a11e4e28791e8e17

ojeda:
So `rust` in the `fragments` list is not enough? In other words, what is the difference between adding it in the `fragments` list vs. the one in `defconfig+rust`?

If I run something like `./kci_build list_kernel_configs --build-config rust-for-linux_rust-next --variant rustc-1.73`, I get:

```
x86_64 x86_64_defconfig+rust rustc-1.73
x86_64 x86_64_defconfig+rust+rust rustc-1.73
```

Is the double `+rust` intended?

Would it make sense to document these in `doc/config/build-configs.md`, or should I look somewhere else?

aliceinwire:
First,
we have three rust fragments:

**rust**
https://github.com/aliceinwire/kernelci-core/blob/b8a13ed33d8ab0416b530f20f976b6e151db56f3/config/core/build-configs.yaml#L445
**rust-for-linux-samples**
https://github.com/aliceinwire/kernelci-core/blob/b8a13ed33d8ab0416b530f20f976b6e151db56f3/config/core/build-configs.yaml#L453
**rust-samples**
https://github.com/aliceinwire/kernelci-core/blob/b8a13ed33d8ab0416b530f20f976b6e151db56f3/config/core/build-configs.yaml#L476C3-L476C15

The first fragment **rust** is the only one that contain `CONFIG_RUST=y`

This code with a normal `x86_64_defconfig` https://github.com/aliceinwire/kernelci-core/blob/b8a13ed33d8ab0416b530f20f976b6e151db56f3/config/core/build-configs.yaml#L1463C10-L1463C10
will produce five defconfig combinations here under:
x86_64_defconfig  _(this configuration have not enabled CONFIG_RUST=y so not producing a rust enabled kernel)_
x86_64_defconfig+rust _(only this configuration have enabled the rust fragment and so CONFIG_RUST=y producing a rust enabled kernel)_
x86_64_defconfig+rust-for-linux-samples _(this configuration have not enabled CONFIG_RUST=y)_
x86_64_defconfig+rust-samples _(this configuration have not enabled CONFIG_RUST=y)_
x86_64_defconfig+kselftest

> Is the double +rust intended?

yes, as adding +rust will enable `CONFIG_RUST=y` for each of the five defconfig files creating the following defconfig combinations:
x86_64_defconfig+rust
x86_64_defconfig+rust+rust (this will need to be removed by filtering x86_64_defconfig will do this on one of the next PR)
x86_64_defconfig+rust+rust-for-linux-samples 
x86_64_defconfig+rust+rust-samples 
x86_64_defconfig+rust+kselftest (is really needed? we have any rust-for-linux test in kselftest?)

> Would it make sense to document these in doc/config/build-configs.md, or should I look somewhere else?

I think what we are doing with rust-for-linux is a bit different from the standard defconfig files of linux kernel, I'm not sure if there is need to be documented :woman_shrugging:

ojeda:
If I understand correctly, you are saying both ways enable `CONFIG_RUST=y` properly, but fragments are not enabled all together, and thus some of the combinations did not do anything useful. That makes sense, and then indeed the `+rust` is not needed.

In fact, I think the combinations that would make sense for us would be something like:

  - `x86_64_defconfig`: to check the build environment is OK as a baseline.
  - `x86_64_defconfig+rust`: to check the kernel is still OK with only `CONFIG_RUST` enabled.
  - `x86_64_defconfig+rust+rust-samples+...`: this would be the "production" one where we will add more things, i.e. it will not just be the samples in the future.
  - `x86_64_defconfig+rust+rust-samples+...+rust-debug`: ditto (debug).
  - `x86_64_defconfig+rust+rust-samples+...+rust-kunit+kselftest`: the "testing" one.
  - `x86_64_defconfig+rust+rust-samples+...+rust-kunit+kselftest+rust-debug`: ditto (debug).
  - And perhaps a minimal config and/or a big one and/or random one to catch conditional compilation issues.

where `rust-kunit` is one that enables `CONFIG_RUST_KERNEL_DOCTESTS` (and possibly others in the future), and `rust-debug` enables `CONFIG_RUST_DEBUG_ASSERTIONS` (and maybe others too).

The `rust-samples+...` part could instead be a `rust-base` fragment or similar, where we enable most things we want to regularly build (including `rust-samples`).

The first two combinations in the list above are less important than the others, so if needed, they can be dropped (e.g. it would be better to put those resources into testing other branches/arches in the future, if we have to choose). We could also merge the testing/debug ones if needed too.

Ideally, at least one of those (perhaps the testing+debug one) would be built with `CLIPPY=1`, too. Not sure how build options like that are handled, since it is not a config option.

Finally, we should also duplicate soon these for `arm64`, since that is coming soon too.

Does the above make some sense? Or is this done usually in other ways? Is the number of runs reasonable (@gctucker)? (e.g. 2 arches * 3 branches * 5 configs).

> (is really needed? we have any rust-for-linux test in kselftest?)

I think it makes sense to check that the Rust builds still pass the selftests, even if there are no Rust tests yet, so it would be great to have those run. How does that work? Do they affect the kernel negatively in some way?

In addition, Laura (@laura-nao) recently submitted the first selftest for Rust: https://lore.kernel.org/rust-for-linux/20231215132132.169628-1-laura.nao@collabora.com/

aliceinwire:
> If I understand correctly, you are saying both ways enable `CONFIG_RUST=y` properly, but fragments are not enabled all together, and thus some of the combinations did not do anything useful. That makes sense, and then indeed the `+rust` is not needed.

yes, that's correct!

> 
> In fact, I think the combinations that would make sense for us would be something like:
> 
>     * `x86_64_defconfig`: to check the build environment is OK as a baseline.
> 
>     * `x86_64_defconfig+rust`: to check the kernel is still OK with only `CONFIG_RUST` enabled.
> 
>     * `x86_64_defconfig+rust+rust-samples+...`: this would be the "production" one where we will add more things, i.e. it will not just be the samples in the future.
> 
>     * `x86_64_defconfig+rust+rust-samples+...+rust-debug`: ditto (debug).
> 
>     * `x86_64_defconfig+rust+rust-samples+...+rust-kunit+kselftest`: the "testing" one.
> 
>     * `x86_64_defconfig+rust+rust-samples+...+rust-kunit+kselftest+rust-debug`: ditto (debug).
> 
>     * And perhaps a minimal config and/or a big one and/or random one to catch conditional compilation issues.
> 
> 
> where `rust-kunit` is one that enables `CONFIG_RUST_KERNEL_DOCTESTS` (and possibly others in the future), and `rust-debug` enables `CONFIG_RUST_DEBUG_ASSERTIONS` (and maybe others too).
> 
> The `rust-samples+...` part could instead be a `rust-base` fragment or similar, where we enable most things we want to regularly build (including `rust-samples`).
> 
> The first two combinations in the list above are less important than the others, so if needed, they can be dropped (e.g. it would be better to put those resources into testing other branches/arches in the future, if we have to choose). We could also merge the testing/debug ones if needed too.
> 
> Ideally, at least one of those (perhaps the testing+debug one) would be built with `CLIPPY=1`, too. Not sure how build options like that are handled, since it is not a config option.
> 
> Finally, we should also duplicate soon these for `arm64`, since that is coming soon too.
> 
> Does the above make some sense? Or is this done usually in other ways? Is the number of runs reasonable (@gctucker)? (e.g. 2 arches * 3 branches * 5 configs).


I agree and will send a pull request about this, thanks!

> 
> > (is really needed? we have any rust-for-linux test in kselftest?)
> 
> I think it makes sense to check that the Rust builds still pass the selftests, even if there are no Rust tests yet, so it would be great to have those run. How does that work? Do they affect the kernel negatively in some way?
> 
> In addition, Laura (@laura-nao) recently submitted the first selftest for Rust: https://lore.kernel.org/rust-for-linux/20231215132132.169628-1-laura.nao@collabora.com/

nice! thanks for let me know!

PR #2228, opened 2023-12-06
config: runtime: lab-baylibre: switch to regex for tree

khilman:
Switch the tree filter to a regex with the primary goal of running tests for "stable", but not stable* (e.g. stable-rc-queue) due the number of jobs currently being too much for lab-baylibre.

### DIFF
--- a/config/core/runtime-configs.yaml
+++ b/config/core/runtime-configs.yaml
@@ -11,17 +11,8 @@ runtimes:
     queue_timeout:
       days: 1
     filters:
+      - regex: {'tree': '^(kernelci|mainline|next|stable|arm64|riscv|soc|amlogic|matthiasbgg)$'}
       - passlist:
-          tree:
-            - staging
-            - mainline
-            - next
-            - stable
-            - arm64
-            - riscv
-            - soc
-            - amlogic
-            - matthiasbgg
           plan:
             - baseline
             - kselftest


config/core/runtime-configs.yaml
```diff
@@ -11,17 +11,8 @@ runtimes:
     queue_timeout:
       days: 1
     filters:
+      - regex: {'tree': '^(staging|mainline|next|stable|arm64|riscv|soc|amlogic|matthiasbgg)$'}
```
gctucker, line 14: There's actually no `staging` tree, staging jobs use the `kernelci` tree (I know it's been in the YAML for ages, but spotting this now).

gctucker (commented):
LGTM - not tested.  It should be possible to manually check the filter with `kci_test` and some arbitrary JSON build meta-data.  Just one thing about the `kernelci` tree - did you actually ever get staging jobs running in this lab?

config/core/runtime-configs.yaml
gctucker: > I pushed an updated version with `s/staging/kernelci/`

I don't see it in the diff, was it missing from your commit maybe?

gctucker (approved):
OK I haven't tested this myself but it looks safe and it was definitely already broken before (and it passes YAML validation).

@khilman Could you please just confirm you're getting staging jobs in the BayLibre lab now?

nuclearcat:
https://staging.kernelci.org/test/plan/id/6576ca851a71f71a9aa372b1/
lab-baylibre can be seen in staging

PR #2229, opened 2023-12-07
kernelci.cli: fix context and CommandSettings.secrets

gctucker:
The CommandSettings object is used as the Click command context so use a local "settings" variable in Kci.invoke() to make this clear.  Also drop the .secrets property which was broken and actually not used.

Fixes: ec7f5c61ae53 ("kernelci.cli: rewrite with Click framework")


JenySadadia (approved):
LGTM. Tested OK locally.

PR #2230, opened 2023-12-07
Add `kernelci.cli.get_api()` and update commands

gctucker:
Add `kernelci.cli.get_api()` and `.get_api_helper()` functions and use then in all the `kci` subcommands accordingly to remove repeated code.


JenySadadia (approved):
Tested OK.

PR #2231, opened 2023-12-08
kernelci.cli: use get_pagination() in kci node

gctucker:
Update `kci node find` implementation to use get_pagination() with the standard command line page options.

### DIFF
--- a/kernelci/cli/node.py
+++ b/kernelci/cli/node.py
@@ -14,6 +14,7 @@ import click
 from . import (
     Args,
     get_api,
+    get_pagination,
     kci,
     split_attributes,
 )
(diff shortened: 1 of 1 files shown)


kernelci/cli/user.py
```diff
@@ -44,10 +46,9 @@ def whoami(config, api, indent, secrets):
 @catch_http_error
 def find(attributes, config, api, indent):
     """Find user profiles with arbitrary attributes"""
-    configs = kernelci.config.load(config)
-    api_config = configs['api'][api]
-    api = kernelci.api.get_api(api_config)
-    users = api.get_users(split_attributes(attributes))
+    api = get_api(config, api)
+    attributes = split_attributes(attributes)
+    users = api.get_users(attributes)
```
JenySadadia, line 51: We should also use `get_pagination` here for `get_users`.

gctucker: Yes, I'll let you deal with this as a follow-up I think.  This PR is just about completing the `kci node` command tool.

JenySadadia (approved):
Tested OK.

PR #2233, opened 2023-12-08
kernelci.cli: use click.File for reading input

gctucker:
Use click.File rather than sys.stdin for reading input data.  If the
    file name is '-' then it will use stdin which is standard practice.


JenySadadia (approved):
Tested OK.

PR #2234, opened 2023-12-08
kernelci.cli: use catch_http_error in all API calls

### LINKED ISSUE #2223: Handle API exceptions in `kci`
Example:

https://github.com/kernelci/kernelci-core/blob/bc154c5bca573d395020e1cc3f3a7aee170778cd/kernelci/cli/node.py#L54

If the API raises an exception for an unexpected reason (corrupted or incompatible data format in the DB, for instance) we'll be losing that information and the user would have no idea of what happened. This makes bug fixing much harder.

The API should, hopefully in most cases, return an error message as a request response, so we could show that to the user if the code did something like this instead:

```
try:
    nodes = api.get_nodes(split_attributes(attributes), offset, limit)
except requests.exceptions.HTTPError as e:
    msg = json.loads(e.response.text)
    click.echo(msg['detail'])
    sys.exit(1)
```

This can probably be addressed in a generic way in all `kci` -> API interactions.

gctucker:
Add the catch_http_error decorator to all the commands that make some
    API calls.  This improves the user experience noticeably as it avoids
    having a noisy exception trace and also adds any the details provided
    by the API from the HTTP error data.
    
    For example:
    
        $ ./kci node get --api=staging abcd
        Error: 400 Client Error: Bad Request for url: https://staging.kernelci.org:9000/latest/node/abcd
        'abcd' is not a valid ObjectId, it must be a 12-byte input or a 24-character hex string
    
    The last line there is not part of the standard exception handling,
    and it's needed in order for the user to understand what the actual
    problem was (rather than there was just an HTTP error).

Fixes #2223

### DIFF
--- a/kernelci/cli/api.py
+++ b/kernelci/cli/api.py
@@ -7,6 +7,7 @@
 
 from . import (
     Args,
+    catch_http_error,
     echo_json,
     get_api,
     kci,
@@ -22,6 +23,7 @@ def kci_api():
 @Args.config
 @Args.api
 @Args.indent
+@catch_http_error
 def hello(config, api, indent):
     """Query the API root endpoint"""
     api = get_api(config, api)
--- a/kernelci/cli/event.py
+++ b/kernelci/cli/event.py
@@ -12,6 +12,7 @@ import click
 
 from . import (
     Args,
+    catch_http_error,
     echo_json,
     get_api,
     get_api_helper,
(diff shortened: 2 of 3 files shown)


gctucker:
@hardboprobot Hello, maybe you would like to review this one?  It should be fixing the issue you reported a couple of weeks ago.

JenySadadia (approved):
Tested OK.

gctucker:
> @hardboprobot Hello, maybe you would like to review this one? It should be fixing the issue you reported a couple of weeks ago.

Ah well, next time maybe ;)

PR #2235, opened 2023-12-08
CHROMEOS build_board: cleanup/extract ti50 firmware as well

a-wai:
Recent Chromebooks such as `corsola` or `skyrim` use `ti50` instead of `cr50`. This change ensures this is also taken into account so we don't leave any of those firmware on our images but still extract those as board artifacts.


nuclearcat:
Is it build tested, that it is really removing firmware?

a-wai:
> Is it build tested, that it is really removing firmware?

Yes, and I also confirmed this change moves `ti50.bin.{prepvt,prod}` to the artifacts dir

```
$ tree mnt/opt/google/*50
mnt/opt/google/cr50
└── ro_db
    ├── verify_ro_NBQS.db
    ├── verify_ro_XWJE.db
    └── verify_ro_ZZAF.db
mnt/opt/google/ti50

2 directories, 3 files
```

nuclearcat (approved):
LGTM

PR #2236, opened 2023-12-08
kernelci.cli.user: send token to GET users requests

JenySadadia:
Depends https://github.com/kernelci/kernelci-api/pull/430

Send API token for getting existing users' information as the endpoint has made accessible to only authorized users.

### DIFF
--- a/kernelci/cli/user.py
+++ b/kernelci/cli/user.py
@@ -36,17 +36,17 @@ def whoami(config, api, indent, secrets):
     click.echo(json.dumps(data, indent=indent))
 
 
-@kci_user.command
+@kci_user.command(secrets=True)
 @click.argument('attributes', nargs=-1)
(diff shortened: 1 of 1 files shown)


nuclearcat (approved):
Verified by user find command, works ok on staging

gctucker:
@JenySadadia Why has authentication been added there?  Is this a `fastapi-users` requirement?

nuclearcat:
We had discussion and decided it is not good idea to expose usernames and emails without authentication from privacy standpoint.

gctucker:
OK, this seems a bit arbitrary.  I would have added a field to let users mask their email address if they didn't want to share it publicly.  I guess that can always be done later.

Only real issue as far as I'm concerned is I need to rebase my PRs again so I can finally get my old changes merged ;)

PR #2239, opened 2023-12-09
kernelci.cli.user: send API token to GET user matching ID

JenySadadia:
Depends on https://github.com/kernelci/kernelci-api/pull/431

Send API token for getting user matching given ID as the endpoint has made accessible to only authorized users.

### DIFF
--- a/kernelci/cli/user.py
+++ b/kernelci/cli/user.py
@@ -132,15 +132,15 @@ def verify(username, config, api):
     click.echo("Email verification successful!")
 
 
-@kci_user.command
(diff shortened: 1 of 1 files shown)


nuclearcat:
Tested on staging, but seems patch need rebase.
@JenySadadia can you rebase the patch?

JenySadadia:
> Tested on staging, but seems patch need rebase. @JenySadadia can you rebase the patch?

Done.

nuclearcat:
Re-tested, works fine

PR #2242, opened 2023-12-11
tests: add __init__.py to make it a package

gctucker:
Add tests/__init__.py to make it a Python package so that Pylint can import it, as per the Makefile target which just lists "tests".


nuclearcat:
Thanks, that helps with github checks :)

PR #2245, opened 2023-12-13
rootfs: Add imagesize parameter

nuclearcat:
Some images might exceed 1GB default size, for example fluster tests that include a lot of test vectors.
We need configurable parameters to solve that.


nuclearcat:
Tested and built manually by[ kci-rootfs ](https://github.com/nuclearcat/kci-rootfs)

hardboprobot (approved):

PR #2246, opened 2023-12-13
azure.py: Verify if directory exist if destination file have subdir

nuclearcat:
If destination file needs subdirectory we need to make sure it exist, as Azure wont create them automatically.


nuclearcat:
Tested with new kbuild where filename might have path

PR #2248, opened 2023-12-14
Add API.Node abstract class

gctucker:
Move the Node API bindings to a nested abstract class API.Node and update all the related implementation accordingly.

### DIFF
--- a/kernelci/api/__init__.py
+++ b/kernelci/api/__init__.py
@@ -10,7 +10,7 @@ import enum
 import importlib
 import json
 import urllib
-from typing import Optional, Sequence
+from typing import Dict, Optional, Sequence
 
 from cloudevents.http import CloudEvent
 import requests
@@ -147,9 +147,11 @@ class Base:
 class API(abc.ABC, Base):  # pylint: disable=too-many-public-methods
     """KernelCI API Python bindings abstraction"""
 
+    # pylint: disable=abstract-class-instantiated
     def __init__(self, config: kernelci.config.api.API, token: str):
         data = Data(config, token)
         Base.__init__(self, data)
+        self._node = self.Node(self.data)  # type: ignore[abstract]
 
     @property
     def config(self) -> kernelci.config.api.API:
@@ -165,11 +167,6 @@ class API(abc.ABC, Base):  # pylint: disable=too-many-public-methods
     def version(self) -> str:
         """API version"""
 
-    @property
-    @abc.abstractmethod
-    def node_states(self):
-        """An enum with all the valid node state names"""
-
     @abc.abstractmethod
     def hello(self) -> dict:
         """Get the hello message"""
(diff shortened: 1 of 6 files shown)


gctucker:
Should we increase the minimum Python version?  3.7 is end of life already now:
https://devguide.python.org/versions/

I'll fix the code to make it work anyway but I think 3.8 or 3.9 should be considered as a new minimum.

gctucker:
Also 3.11 should be tested as the latest supported version.  It's the default one in Debian Bookworm (stable) now too.

gctucker:
Alright I've made #2252 to bump the Python versions.  Still this PR now passes the tests with 3.7 so it doesn't depend on it.

JenySadadia (approved):
Tested OK locally.

gctucker:
OK thanks.

In the meantime, @nuclearcat reported a problem on staging, I believe that's because the pipeline services need to be updated to match the changes in the API bindings.  I'll prepare a PR for that and for the user-related ones too if applicable.

Please let me know if any other side-effects are found following these changes.

nuclearcat:
```
    def create_job_node(self, job_config, input_node):
        job_node = {
            'parent': input_node['id'],
            'name': job_config.name,
            'group': job_config.name,
            'artifacts': input_node['artifacts'],
            'revision': input_node['revision'],
        }
        return self._api.create_node(job_node)
```
This code crash on staging:
```
today at 3:46:33 PM12/20/2023 01:46:33 PM UTC [ERROR] Traceback (most recent call last):
today at 3:46:33 PM  File "/home/kernelci/pipeline/base.py", line 69, in run
today at 3:46:33 PM    status = self._run(context)
today at 3:46:33 PM  File "/home/kernelci/./pipeline/scheduler.py", line 113, in _run
today at 3:46:33 PM    self._run_job(job, runtime, platform, input_node)
today at 3:46:33 PM  File "/home/kernelci/./pipeline/scheduler.py", line 75, in _run_job
today at 3:46:33 PM    node = self._api_helper.create_job_node(job_config, input_node)
today at 3:46:33 PM  File "/usr/local/lib/python3.9/dist-packages/kernelci-1.1-py3.9.egg/kernelci/api/helper.py", line 114, in create_job_node
today at 3:46:33 PM    return self._api.create_node(job_node)
today at 3:46:33 PMAttributeError: 'LatestAPI' object has no attribute 'create_node'
```
@gctucker

gctucker:
I think that's already fixed in #2257.  It seems like you're trying to run staging too often, I don't even have the time to send changes in between ;)

nuclearcat:
Yes, i promised to look into one of the PRs and review it, and also I need to check if my own PRs are working or not (some issues are not reproducible on my local setup), and now i am stuck cause broken PR got merged and i cannot even set staging-skip.

gctucker:
Well that's how development happens, when something's broken it gets fixed.  So, is it fixed now with #2257?

nuclearcat:
Staging machine load too high now, i cannot run another staging cycle yet

gctucker:
Alright I'll check again tomorrow.

PR #2249, opened 2023-12-14
kernelci.api: refactor user-facing API models

hardboprobot:
Related to https://github.com/kernelci/kernelci-api/pull/433 (that PR depends on this one)

Bring the API models used by the user-facing endpoints from the kernelci-api repo to this one so that they can be used by the kernelci-api code and the helpers in this repo.

Bring the base model definitions here as well into a separate file (models_base.py) so that they can be used by the models here and by the internal kernelci-api models.

### DIFF
--- a/kernelci/api/models_base.py
+++ b/kernelci/api/models_base.py
@@ -8,6 +8,8 @@
 # need methods
 # pylint: disable=too-few-public-methods
 
+# pylint: disable=no-name-in-module
+
 """Common KernelCI API model definitions"""
 
 from typing import Optional
--- a/kernelci/api/models_base.py
+++ b/kernelci/api/models_base.py
@@ -1,0 +1,67 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+#
+# Copyright (C) 2021-2023 Collabora Limited
+# Author: Guillaume Tucker <guillaume.tucker@collabora.com>
+# Author: Jeny Sadadia <jeny.sadadia@collabora.com>
+
+# Disable below flag as some models are just for storing the data and do not
+# need methods
+# pylint: disable=too-few-public-methods
+
+"""Common KernelCI API model definitions"""
+
+from typing import Optional
+from bson import ObjectId
+from pydantic import (
+    BaseModel,
+    Field,
+)
+
+
+class PyObjectId(ObjectId):
+    """Wrapper around ObjectId to be able to use it in Pydantic models"""
+
+    @classmethod
+    def __get_validators__(cls):
+        yield cls.validate
+
+    @classmethod
+    def __modify_schema__(cls, field_schema):
+        field_schema.update(type='string')
+
+    @classmethod
+    def validate(cls, value):
+        """Validate the value of the ObjectId"""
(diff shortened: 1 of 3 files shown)


kernelci/api/models_base.py
JenySadadia, line 1: How about renaming it to `base_models.py`?

hardboprobot: It's a matter of preference, I guess. When creating multiple related files I prefer to use this naming scheme of "concept" + "specifier" as a suffix rather than the common English grammar: "adjective" + "concept", where the  important part (the key concept) is at the end of the file name. When sorting by name I prefer to have all the model-related files grouped together.

requirements.txt
```diff
@@ -1,10 +1,12 @@
 azure-storage-file-share==12.13.0
+bson
```
JenySadadia, line 2: I think `bson` is one of the dependencies of `pydantic`. Hence, gets installed automatically. Need to confirm though.

JenySadadia:
> @JenySadadia gentle ping for this. Approved?

Did we get staging results along with https://github.com/kernelci/kernelci-api/pull/433 yet? If yes, what did you do about `kernelci` package?

requirements.txt
JenySadadia: @hardboprobot  Could you please check this?

JenySadadia:
> > > @JenySadadia gentle ping for this. Approved?
> > 
> > 
> > Did we get staging results along with [kernelci/kernelci-api#433](https://github.com/kernelci/kernelci-api/pull/433) yet? If yes, what did you do about `kernelci` package?
> 
> I dropped the `staging-skip` tag from this, so if staging is still working I guess we're good to go for this one. [kernelci/kernelci-api#433](https://github.com/kernelci/kernelci-api/pull/433) hasn't been tested on staging because it depends on the kernelci package deployment. I think we discussed this before I left, that the order should be:
> 
> 1. Test this on staging and merge if it's ok
> 2. Deploy a new kernelci pip package
> 3. Update the pending kernelci-api PR to use the new kernelci package (or, at least, pick the release branch for the time being)
> 4. Test kernelci-api PR on staging and merge if it works
> 
> We're stuck in 1. I think this can be merged. About 2, I have no idea. @nuclearcat have you updated the kernelci pip package before?

I'd rather suggest waiting for the staging results along with dependent API PR. This will enable us to find and fix any issues with this one before merging.

requirements.txt
hardboprobot: Fixed

hardboprobot: Oh, now I remember why I had to put that one:

```
************* Module kernelci.api.models
kernelci/api/models.py:18:0: E0401: Unable to import 'bson' (import-error)
************* Module kernelci.api.models_base
kernelci/api/models_base.py:16:0: E0401: Unable to import 'bson' (import-error)
```

I'm putting it back again, then.

JenySadadia: Yes, `bson` comes with `motor`. Can you please pin the version here?

JenySadadia (approved):
LGTM.

PR #2251, opened 2023-12-15
test-configs.yaml: Enable bookworm rootfs on kselftest-net_qemu

nuclearcat:
We use yaml alias which hide that kselftest by default uses bullseye. So we need to change this parameter to bookworm.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -416,6 +416,7 @@ test_plans:
 
   kselftest-net_qemu:
     <<: *kselftest_qemu
+    rootfs: debian_bookworm-kselftest_ramdisk
     params:
       job_timeout: '90'
       kselftest_collections: "net"


config/core/test-configs.yaml
```diff
@@ -416,6 +416,7 @@ test_plans:
 
   kselftest-net_qemu:
     <<: *kselftest_qemu
+    rootfs: debian_bookworm-kselftest_ramdisk
```
hardboprobot, line 419: Why only for `kselftest-net_qemu` and not for `kselftest_a_qemu`?

nuclearcat: We need to verify other tests possible breakages, sometimes some new packages need to be added and etc.
It is possible to try in separate PR, and check the results.
As i recall with previous debos version migration some tests was broken.

nuclearcat:
As discussed - merging, but we still have issue to fix loading modules for this kind of rootfs

PR #2252, opened 2023-12-15
Python 3.8 / 3.11

gctucker:
Bump the Python version range from 3.7 / 3.10 to 3.8 / 3.11.  As per the Python versions timeline, 3.7 is now end-of-life and 3.8 is the minimal version supported until the end of 2024:

https://devguide.python.org/versions/

Then 3.11 is now the standard version used by the new API & Pipeline code, and it's also the default version in Debian Bookworm (stable).

Update GitHub checks and the base Python Docker template accordingly.


nuclearcat:
Staging is ok, merging

PR #2253, opened 2023-12-15
Add API.User abstract class

gctucker:
Similarly to API.Node, refactor the user API endpoint bindings with a new API.User abstract class.  Update the LatestAPI implementation and `kernelci.cli.user` accordingly.  There doesn't appear to be any unit tests for this part of the API.

As a side-effect, the Pylint check for too many public methods can be enabled again as the base API abstract class has now fewer of them since each nested class is taking care of a specific part of the API.

The user groups are still in the API class and should be moved to API.User or a new API.Group class.  This can be done as a follow-up, it wasn't very clearly defined initially so I've left it out for now.

Similarly, there should be API.Event or API.PubSub etc. for the other parts of the API.

### DIFF
--- a/kernelci/api/__init__.py
+++ b/kernelci/api/__init__.py
@@ -144,7 +144,7 @@ class Base:
         return resp
 
 
-class API(abc.ABC, Base):  # pylint: disable=too-many-public-methods
+class API(abc.ABC, Base):
     """KernelCI API Python bindings abstraction"""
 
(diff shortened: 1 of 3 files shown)


gctucker:
Should work on staging with https://github.com/kernelci/kernelci-pipeline/pull/373 so removing skip label for now.

nuclearcat (approved):
Tested on staging

Issue #2254, opened 2023-12-15
Add branch queue/6.6 and linux-6.6.y from stable-rc tree


crazoes:
Which Git branch do you want to add?
⇨ Git repo URL:
https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable-rc.git/

⇨ Git branch name:
queue/6.6 and linux-6.6.y

How much build coverage do you need for your branch?
As other stable-rc branches.

nuclearcat:
Merged to production

PR #2255, opened 2023-12-15
Adding stable and rc-queue for kernel 6.6

nuclearcat:
As per https://github.com/kernelci/kernelci-core/issues/2254 adding 6.6 to kernelci


nuclearcat:
Tested on staging

PR #2256, opened 2023-12-18
kernelci/kbuild.py: Initial kbuild offloading library implementation

nuclearcat:
This is initial implementation of kernel build steps, probably on very early stages and doesn't use properly other kernelci primitives and functions.
Main purpose is to reduce complexity and move out kernel build from jinja2 templates in pipeline. This can be also foundation to implement 3-step kernel build. (prepare, build, upload)

### DIFF
--- a/kernelci/kbuild.py
+++ b/kernelci/kbuild.py
@@ -1,0 +1,566 @@
+# SPDX-License-Identifier: LGPL-2.1-or-later
+#
+# Copyright (C) 2023 Collabora Limited
+# Author: Denys Fedoryshchenko <denys.f@collabora.com>
+"""
+KernelCI kernel build abstraction package
+Initial version, still work in progress
+Roadmap:
+- correct build status - fail/pass/incomplete
+- Code deuglification
+- Reuse API class/functions and other code from KernelCI
+- Add support for multiple CPU architectures (arm64, arm, mips, etc)
+- Add support for multiple compilers (clang)
+- Implement 3-stage build process
+  (fetch-prepare, build, package) with 3 containers
+  (kernelci, compiler-specific, kernelci)
+"""
+
+import os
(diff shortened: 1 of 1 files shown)


kernelci/kbuild.py
```diff
@@ -0,0 +1,573 @@
...
+    '''
+    Build class that represents kernel build
+    args, 1 or 3 args.
+    3 args, create new build object
+        node - node object from API
+        jobname - job name
+        kbuild_params - dict with kbuild parameters
+    1 arg, load build object from json
+    '''
+    def __init__(self, *args, **kwargs):
```
hardboprobot, line 93: Is there a reason to pass the arguments to `__init__` as a variable list and keyword arguments rather than with optional arguments? ie. `def __init__(self, node=None, jobname=None, params=None, jsonobj=None)`, or even better, to split it into a primitive initializer and then use class methods (such as `from_json` above) to create Kbuild objects with other specific initialization params?

hardboprobot (approved):
I understand this basically moves the code-generation code from a jinja2 template to a python class. While I think the intention is good and that the code will be easier to refactor into libraries in this way, I also think this still keeps most of the same inherent problems we were having with templates. I'd like to put some time aside to think of an alternative solution.

I also think this could be way less verbose. Since this is more or less a direct translation of the jinja template I understand it takes some care into reproducing the resulting script exactly, including comments. I'm not sure that'll be necessary in the end.

But for now it's ok for me to merge this if it'll make your life easier. So if you're tested it and it works, then it LGTM

kernelci/kbuild.py
nuclearcat: Done, updated to fixed arguments

hardboprobot (approved):

PR #2257, opened 2023-12-20
Use new API.Node bindings in a few more places

gctucker:
Fix a few more places to use the new API.Node bindings class.


JenySadadia (approved):
Thanks for the fixes.
Tested OK.

PR #2264, opened 2023-12-26
kernelci.cli: enable pagination arguments for `kci user find`

JenySadadia:
Enable user to provide pagination arguments like `page_length` and `page_number` for `kci user find` command.


nuclearcat (approved):
Tested locally, thanks

PR #2265, opened 2023-12-27
kernelci.cli.user: fix user verification

JenySadadia:
Since GET `/users` request is now restricted to authorized users only, it is not possible to retrieve user information in user verification request to get email address. Instead of getting username, directly ask for email address in the command argument to fix the issue.


nuclearcat (approved):
Looks good to me, on my opinion it is perfectly fine, as at some resources for recovery it might ask any of your credentials (username, email, etc).

PR #2266, opened 2023-12-27
kernelci.cli.user: fix password reset

JenySadadia:
Since GET `/users` request is now restricted to authorized users only, it is not possible to retrieve user information in password reset request to get email address. Instead of getting username, directly ask for email address in the
command argument to fix the issue.


nuclearcat (approved):
LGTM, thanks

PR #2269, opened 2024-01-02
test-configs.yaml: Add LibreTech Solitude

broonie:
This is another Meson board from Libretech with some new audio hardware
to test that I have in my lab, enable baseline tests plus tests relevant for
the platform support and also KVM since the board has VHE which we don't
currently have good coverage for.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -1579,6 +1579,12 @@ device_types:
     boot_method: uboot
     flags: ['big_endian']
 
+  meson-sm1-s905d3-libretech-cc:
+    mach: amlogic
+    class: arm64-dtb
+    boot_method: uboot
+    flags: ['big_endian']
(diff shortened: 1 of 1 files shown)


broonie:
Baseline job for mainline: https://staging.kernelci.org/test/plan/id/659603b9c5f1da8ad162ee71/ (a bunch of other baseline jobs have run successfully).

Everything is most likely fine, the issues with the firmware (which will hopefully be resolved next week, we're due another vendor release once the u-boot release happens on Monday) mean that the jobs get scheduled a bit intermittently and my own jobs are getting priority as ever.

nuclearcat:
Works:
https://staging.kernelci.org/test/plan/id/65960c17489b9382dd62ef15/
Many other tests are missing, i suspect it is something to troubleshoot on kernelci side

PR #2270, opened 2024-01-03
rootfs-images.yaml: Update fluster to new version

nuclearcat:
Update rootfs links to newer version, merging also with pending PR from staging.


nuclearcat (commented):
Tested in fluster related PRs

PR #2271, opened 2024-01-03
build-configs.yaml: Update qcom-lt tree URL

nuclearcat:
It seems qcom-lt tree changed location, and emitting error during jenkins monitor job. Updating to new URL.


nuclearcat:
Works fine

PR #2272, opened 2024-01-04
build-configs.yaml: Add scp.img for mt8183/mt8186 to arm64-chromebook

nfraprado:
Embed the scp.img firmware for mt8183 and mt8186 into the arm64-chromebook kernel, like was already done for other MediaTek chromebooks. This ensures the SCP can fully boot on those platforms independent of the rootfs used. In particular it makes the cros-ec-rpmsg check in the dt-kselftest pass.


nuclearcat (approved):
Built successfully: https://storage.staging.kernelci.org/kernelci/staging-next/staging-next-20240105.0/arm64/defconfig+arm64-chromebook/gcc-10/config/kernel.config

PR #2273, opened 2024-01-06
test-configs.yaml: Add LibreTech Alta

broonie:
The Alta is another new board from LibreTech again based on an AMLogic
SoC, it uses the same PCB as Solitude but has a different lower end SoC
mounted.  Enable basic tests relevant to the hardware present on the
board.


nuclearcat (approved):
LGTM: https://staging.kernelci.org/test/plan/id/659a07f324c093d49962ee67/

PR #2276, opened 2024-01-07
test-configs.yaml: Enable V4L decoder tests on Le Potato

broonie:
There is a video CODEC block with upstream support on the Le Potato,
enable the V4L2 decoder tests for the CODECs it supports.  It also
supports MJPEG but that doesn't seem to be a suite we cover.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -3074,6 +3074,9 @@ test_configs:
       # ltp-mm - Runs board out of memory
       - ltp-pty
       - ltp-timers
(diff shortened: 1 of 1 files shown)


nuclearcat (approved):
Tests are working, example: https://storage.staging.kernelci.org/kernelci/staging-next/staging-next-20240313.3/arm64/defconfig/gcc-10/lab-broonie/v4l2-decoder-conformance-vp9-meson-gxl-s905x-libretech-cc.html

PR #2277, opened 2024-01-07
test-configs.yaml: Enable V4L decoder coverage on Solitude

broonie:
The SoC on the Solitude has an upstream supported video decoder IP,
enable coverage for the CODECs it supports.


nuclearcat (approved):
Some tests are failing, but probably its expected?
https://storage.staging.kernelci.org/mainline/master/v6.7-9928-g052d534373b7/arm64/defconfig+videodec/gcc-10/lab-broonie/v4l2-decoder-conformance-h264-meson-sm1-s905d3-libretech-cc.html

PR #2278, opened 2024-01-07
test-configs.yaml: Enable video decoder coverage on Alta

broonie:
The SoC on the LibreTech Alta has a video decoder IP which is supported
upstream, enable coverage of the CODECs it supports.


nuclearcat (approved):
https://staging.kernelci.org/test/plan/id/65a51c50fd78ae63e2844c36/
All tests are failing, but probably this is something to fix with tests environment?

PR #2280, opened 2024-01-07
test-configs.yaml: Block kselftest-dt in older kernels

broonie:
The DT selftests were only introduced in v6.7, don't try to run them on
stable kernels.  We should really block v6.1 as well but we're not doing
regexp matching here so the syntax is awkward, we can always improve
later.


nuclearcat (approved):
LGTM, still running on -next

nuclearcat:
tested

PR #2282, opened 2024-01-08
test-configs.yaml: Use battery tag in kselftest-dt jobs for Chromebooks

laura-nao:
Chromebook/Chromebox devices might operate with or without a battery.

The `kselftest-dt` test checks whether drivers are probed correctly, including the battery gauge driver. 
Use the battery LAVA tag to target only Chromebooks where the battery is actually connected, to avoid reporting false regressions.

Signed-off-by: Laura Nao <laura.nao@collabora.com>

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -317,6 +317,8 @@ test_plans:
     params:
       job_timeout: '10'
       kselftest_collections: "dt"
+      tags:
+        - battery
     filters:
       - combination: *arch_defconfig_filter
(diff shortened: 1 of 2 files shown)


hardboprobot (approved):

nuclearcat:
Please rebase

laura-nao:
@nuclearcat rebased - thanks for the heads up!

nuclearcat:
@laura-nao , I think you need to move tags: under params: section

laura-nao:
> @laura-nao , I think you need to move tags: under params: section

Definitely - I accidentally moved the tags section when I rebased. Should be fixed now, thanks for the pointer!

nuclearcat:
Works, for example: https://lava.collabora.dev/scheduler/job/12619525/definition

PR #2285, opened 2024-01-11
test-configs.yaml: Enable LKDTM selftests on Juno

broonie:
We are not covering the LKDTM selftests on any arm64 platforms since
while it's enabled for some of them we're struggling to run the
kselftest kernel on any of them.  Since Juno does currently seem to be
booting the kselftest kernels fine and has some bandwidth let's schedule
the LKDTM tests there.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -2932,6 +2932,7 @@ test_configs:
       - kselftest-cpufreq
       - kselftest-dt
       - kselftest-ftrace
+      - kselftest-lkdtm
       - ltp-crypto
       # ltp-mm - runs system out of memory
       - smc


nuclearcat:
Looks like one test attempt ended as lava glitch: https://storage.staging.kernelci.org/kernelci/staging-next/staging-next-20240112.0/arm/multi_v7_defconfig+kselftest/gcc-10/lab-broonie/kselftest-lkdtm-stm32mp157a-dhcor-avenger96.html
Will wait for next run

broonie:
That's a legit test failure on a completely different board that's already merged.

nuclearcat:
Thanks, merging

PR #2290, opened 2024-01-12
kbuild.py: Map artifact fields to proper key names

nuclearcat:
In MongoDB, when you need to query nested fields where the field name includes dots (.), such as artifacts.job.txt, you face a common issue. MongoDB interprets the dots as separators for nested documents. We should avoid using dots in field names.

Suppose we have:
```
{
    "artifacts": {
        "job.txt": "data"
    }
}
```
In this case, artifacts.job.txt is intended to be a single field name. However, MongoDB interprets artifacts.job.txt as artifacts -> job -> txt, assuming job is a nested document within artifacts, and txt is a field within job.

When you try to query this field directly, MongoDB will look for a nested structure rather than a field with dots in its name. This leads to confusion and errors in querying and manipulating data. To handle such field names, you typically need to use special syntax or workarounds.

### DIFF
--- a/kernelci/kbuild.py
+++ b/kernelci/kbuild.py
@@ -59,9 +59,20 @@ KERNEL_IMAGE_NAMES = {
     'x86': {'bzImage'},
 }
 
+'''
+Following convention is used for artifacts key
+and filename mapping:
+filename_regex - artifact name_regex
+'''
 ARTIFACT_NAMES = {
-    'modules.tar.gz': 'modules',
-    'modules.tar.xz': 'modules',
+    r'modules\.tar\.gz': 'modules',
+    r'modules\.tar\.xz': 'modules',
+    r'build\.log': 'build_log',
+    r'build_kimage\.log': 'build_kimage_stdout',
+    r'build_kimage_stderr\.log': 'build_kernel_errors',
+    r'build_modules\.log': 'build_modules_stdout',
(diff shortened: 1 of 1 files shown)


kernelci/kbuild.py
```diff
@@ -59,9 +59,20 @@
...
-    'modules.tar.gz': 'modules',
-    'modules.tar.xz': 'modules',
+    r'modules\.tar\.gz': 'modules',
+    r'modules\.tar\.xz': 'modules',
+    r'build\.log': 'build_log',
+    r'build_kimage\.log': 'build_kimage_stdout',
+    r'build_kimage_stderr\.log': 'build_kernel_errors',
+    r'build_modules\.log': 'build_modules_stdout',
+    r'build_modules_stderr\.log': 'build_modules_errors',
+    r'fragment/(\d)\.config': 'fragment/\\1.config',
```
hardboprobot, line 75:
```suggestion
    r'fragment/(\d)\.config': 'fragment/\\1_config',
```

hardboprobot (approved):

kernelci/kbuild.py
nuclearcat: Thanks, thats i missed big one, will update now also better way

nuclearcat:
Tested on staging

PR #2291, opened 2024-01-12
lava.jinja2: Update lava template to new data model

nuclearcat:
New data model have a bit different location for describe, update it accordingly.


nuclearcat:
Tested on staging

PR #2292, opened 2024-01-14
test-configs.yaml: Enable hardware relevant tests on Renegade Elite

broonie:
I've added two Renegade Elites to my lab (and whichever lab used to have
them seems to have removed them) so let's enable all the hardware
relevant tests that had been commented out.  Also enable kselftest-dt
since it's a DT based platform.


nuclearcat (approved):
panfrost for example working: https://staging.kernelci.org/test/plan/id/65a509b52a9bef1a33844a9f/

PR #2296, opened 2024-01-15
test-configs.yaml: Enable DT selftests for Raspberry Pi

broonie:
The Raspberry Pi platforms are DT based, enable the DT selftests for
those boards where we have NFS baseline tests enabled.


broonie:
Tested OK on staging: https://staging.kernelci.org/test/plan/id/65a5a3401f61f69a71844a92/

PR #2298, opened 2024-01-15
CHROMEOS Small fixes for build_board.sh

a-wai:
This PR packs multiple small improvement and fixes to the `build_board.sh` script we use to build CrOS images:
* fixes a build failure when a single board/baseboard has multiple `chipset-*` parents (affects `baseboard-brya`)
* fixes script error when the compressed CrOS image already exists (e.g. when restarting the build after it previously failed at a later stage)
* ensure the repo remote and branch are properly determined even in cases where the repo owner is different from the user executing the script (e.g. a normal user and root, respectively)


nuclearcat (approved):
LGTM, and i'm quite sure it is "battle tested", thanks!

PR #2299, opened 2024-01-16
test-configs.yaml: Add VC4 tests

broonie:
The Raspberry Pi has a GPU called V3D and a display driver called VC4
both of which have specific tests in the IGT GPU tests, add coverage of
them on the Pi 4.  Since there are additional tests we need to duplicate
the test list in the VC4 entry, simply adding the new tests overwrites
the generic KMS tests.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -235,6 +235,34 @@ test_plans:
       <<: *igt-kms-test-list
       drm_driver: "tegra"
 
+  igt-kms-vc4:
+    <<: *igt-kms
+    params:
+      drm_driver: "vc4"
+      tests: >
+        core_auth
+        core_getclient
+        core_getstats
+        core_getversion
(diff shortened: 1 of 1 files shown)


broonie:
Good VC4 results: https://staging.kernelci.org/test/plan/id/65b038918877c9abb9d4e4ca/

V3D seems to have some problem finding the testcases which needs a rootfs rebuild to fix, TBH I don't understand why VC4 works - the issue appears to be that it's searching for the test programs using ${PATH} but the subdirectory with V3D isn't added, I have no idea why the same issue doesn't affect VC4. Still, I'll push a version that should work.

broonie:
I've dropped the V3D changes for now given the issues with rootfs builds - I'll move them to a separate PR so they don't hold up the VC4 part which is working fine.

nuclearcat:
as shown tested on staging

PR #2300, opened 2024-01-16
kbuild.py: Remove hardcoded staging config

nuclearcat:
Cleanup, remove staging config left by mistake.

### DIFF
--- a/kernelci/kbuild.py
+++ b/kernelci/kbuild.py
@@ -104,6 +104,7 @@ class KBuild():
             self._firmware_dir = None
             self._af_dir = None
             self._node = node
+            self._api_yaml = None
             return
         # load class from serialized json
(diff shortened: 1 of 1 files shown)


kernelci/kbuild.py
```diff
@@ -520,10 +520,7 @@ def submit(self, retcode):
         if not api_token:
             raise ValueError("KCI_API_TOKEN is not set")
         config = '/etc/kernelci/core/api-configs.yaml'
```
a-wai, line 522: This line can likely be drop as well

nuclearcat: Thanks for spotting this! Fixed

a-wai (approved):

nuclearcat:
tested on staging

PR #2301, opened 2024-01-16
api.models: Add patchset field to node revision

pawiecz:
This patch was initially submitted as kernelci/kernelci-api#380 but required moving after merging kernelci/kernelci-api#433.

### DIFF
--- a/kernelci/api/models.py
+++ b/kernelci/api/models.py
@@ -87,6 +87,9 @@ class Revision(BaseModel):
     version: Optional[KernelVersion] = Field(
         description="Kernel version"
     )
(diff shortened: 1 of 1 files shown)


pawiecz:
DCO check fail is caused by keeping SoB from original PR - can this change be merged despite this unsuccessful check?

pawiecz:
As discussed with @nuclearcat I amended the patch with additional SoB to fix DCO check and keep track of contributions.

nuclearcat:
Thanks, passing staging tests, i guess it is noop for now, but needed for future PR, merging.

PR #2303, opened 2024-01-19
Implement Job types

hardboprobot:
Add a new "kind" property to Jobs, so that we can define what "kind" of node each define Job should create.


nuclearcat (approved):
Tested on staging

PR #2304, opened 2024-01-19
Refine Node subtype models to match current pipeline usage

hardboprobot:
Make some changes to the Node models to make them fit the current use cases


nuclearcat (approved):
Tested on staging

PR #2308, opened 2024-01-19
test-configs.yaml: Enable more test coverage for Raspberry Pi 4

broonie:
The labs which have Raspberry Pi 4s (including mine) seem to have quite
a bit of free capacity with multiple devices available and frequently
idle so let's use the capacity to fill in some of our coverage, add some
LTP suites which we don't currently cover on arm64 together with the
core LTP smoke tests.  Since we currently only run the preempt RT tests
on systems with in order cores also add them here to ensure coverage
with out of order cores.

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -2538,6 +2538,11 @@ test_configs:
       - kselftest-alsa
       - kselftest-arm64
       - kselftest-dt
+      - ltp-dio
+      - ltp-fsx
+      - ltp-smoketest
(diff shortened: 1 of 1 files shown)


broonie:
LTP 
dio: https://staging.kernelci.org/test/plan/id/65b03859c5a5a88c4ed4e50c/
smoketest: https://staging.kernelci.org/test/plan/id/65b038638877c9abb9d4e4b8/

nuclearcat:
Thanks, merging

PR #2309, opened 2024-01-22
api.models: workaround: remove data restrictions from Node subtypes

hardboprobot:
This is a workaround to remove error conditions during testing. Remove all "data" restrictions from Node subtypes, that is, allow extra fields and make all defined fields optional.


nuclearcat (approved):
Tested on staging, works

PR #2313, opened 2024-01-22
build-configs.yaml: Build sc7180 display clock controller driver as m…

laura-nao:
…odule

Configure the display clock controller for trogdor Chromebooks to be built as a module, by modifying the arm64-chromebook fragment. This fixes a kernel hang observed on lazor and kingoftown Chromebooks when disabling the display clocks while they might still be in use.

See also: https://lore.kernel.org/linux-arm-msm/20231218091806.7155-1-laura.nao@collabora.com/

### DIFF
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -169,7 +169,7 @@ fragments:
       # This is required for sc7180-trogdor-lazor-limozeen and it hasn't been
       # merged upstream yet.
       - 'CONFIG_INTERCONNECT_QCOM_SC7180=y'
-      - 'CONFIG_SC_DISPCC_7180=y'
+      - 'CONFIG_SC_DISPCC_7180=m'
       - 'CONFIG_SC_GPUCC_7180=y'
       - 'CONFIG_SC_LPASS_CORECC_7180=y'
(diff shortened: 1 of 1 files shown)


laura-nao:
Yes, I've tested the change by rerunning `baseline-nfs` with the `next-20240103` kernel revision (which previously failed - see: https://linux.kernelci.org/test/case/id/659509dbc5ec8b704377b580/) manually in LAVA: got 20/20 successful jobs without any hang (e.g. https://lava.collabora.dev/scheduler/job/12574811).

hardboprobot (approved):

a-wai:
Added `staging-skip` to check whether this could be the cause of issues with codec tests

a-wai (changes requested):
This seems to break codec-related Tast tests on trogdor: I couldn't find any successful job ("Complete" with at least some "pass" results) with this PR applied, while there are several passing jobs now, see https://lava.collabora.dev/results/13217203/0_tast for example.

This [patchset](https://lore.kernel.org/lkml/20240327202740.3075378-3-swboyd@chromium.org/T/) mentioned in a private chat would likely be a better fix, didn't have time to test it though.

nuclearcat:

PR #2314, opened 2024-01-23
kbuild.py: Add arm64 support and dtb support

nuclearcat:
We need certain options for cross-compiling, environment variables mostly, and also to support building and packing dtb files.

### DIFF
--- a/kernelci/kbuild.py
+++ b/kernelci/kbuild.py
@@ -513,12 +513,6 @@ class KBuild():
         metadata['build']['srcdir'] = self._srcdir
         metadata['build']['config_full'] = self._config_full
         metadata['artifacts'] = self._artifacts
-        # if we have 'dtbs/' in artifacts, add dtbs to metadata
-        for artifact in self._artifacts:
-            if artifact.startswith("dtbs/"):
-                if 'dtbs' not in metadata:
-                    metadata['dtbs'] = []
-                metadata['dtbs'].append(artifact)
 
         with open(os.path.join(self._af_dir, "metadata.json"), 'w') as f:
             json.dump(metadata, f, indent=4)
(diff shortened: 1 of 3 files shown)


kernelci/kbuild.py
```diff
@@ -528,18 +594,33 @@ def upload_artifacts(self):
                 (artifact_path, artifact), root_path
             )
             artifact_key = self.map_artifact_name(artifact)
-
             node_af[artifact_key] = stored_url
+            # map also bzImage, zImage, Image to kernel
+            # for template simplicity
+            if artifact in KERNEL_IMAGE_NAMES[self._arch]:
+                node_af['kernel'] = stored_url
```
a-wai, line 601:
```suggestion
            # map also bzImage, zImage, Image to kernel
            # for template simplicity
            if artifact in KERNEL_IMAGE_NAMES[self._arch]:
                node_af['kernel'] = stored_url
            else:
                node_af[artifact_key] = stored_url
```

This would avoid having both `Image` and `kernel` in the artifacts list.

nuclearcat: I left both intentionally, as i'm wondering if in some cases we need to check image type present. But if you think it is not necessary, i can drop that and do as you suggesting.

a-wai: I think the image type should be deduced from some other parameter (e.g. if the platform config has a `dtb` key), and with #2319 merged, I think we should move forward with change. Unless `bzImage`/`Image` is used elsewhere?

kernelci/kbuild.py
```diff
@@ -556,6 +654,24 @@ def submit(self, retcode):
...
+        # Add full artifacts path to metadata.json
+        # We do it after full artifacts upload, twice
+        # as we can get urls of artifacts AFTER upload
+        self._update_metadata(job_result)
+        # Upload metadata.json to storage
+        metadata_uri = self.upload_metadata()
+        af_uri['metadata'] = metadata_uri
+
+        # do we have dtbs/ in af_uri?
+        dtb_present = False
```
a-wai, line 666: Ah, one thing I forgot to mention: this `dtb_present` is actually unneeded, as we can just check if `platform_config` has a `dtb` key. It won't hurt to keep it, but this parameter isn't used.

PR #2315, opened 2024-01-25
kernelci.api.models: add `job_id` field to Node data

JenySadadia:
Add field to store job ID received from runtime for `kbuild` and `test` nodes.


nuclearcat:
Tested with https://github.com/kernelci/kernelci-pipeline/pull/396#issuecomment-1914559312

PR #2319, opened 2024-01-26
Handle DTB for devices requiring it

a-wai:
This PR allows to handle DTBs for arm64 devices by executing the following steps:
* download the `metadata.json` file from the `kbuild` job
* extract from this file the URL to the DTB as defined in the device type parameters
* add this to the node's artifacts list with the name `dtb`

The boot templates are modified accordingly and made device-agnostic in the process.


a-wai:
Note: includes #2314 for now, still needs the full artifact URL to be present in `metadata.json`

nuclearcat (approved):
tested on staging

PR #2321, opened 2024-01-28
Rust 1.75

ojeda:
Rust 1.75.0 will likely be the next version supported by the kernel. Therefore, add it as a new build environment.

@gctucker @aliceinwire @nuclearcat This time around, if possible, I would like to keep both the latest and previous latest build environments, so that we can upgrade only `linux-next` and so on, before `mainline`, and so on. Please let me know if this somehow breaks some assumptions -- it is the first time we are doing this. Thanks!

Upgrade PRs:
  - https://github.com/kernelci/kernelci-core/pull/2321.
  - https://github.com/kernelci/kernelci-deploy/pull/121.
  - https://github.com/kernelci/kernelci-deploy/pull/122.

### DIFF
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -723,6 +723,12 @@ build_environments:
     arch_params:
       x86_64: *x86_64_params
 
+  rustc-1.75:
+    cc: clang
+    cc_version: 17
+    arch_params:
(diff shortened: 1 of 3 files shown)


nuclearcat:
Tested on staging

ojeda:
Thanks Denys!

PR #2322, opened 2024-01-29
kernelci.api.helper: unpack and propagate errors in submit_results

hardboprobot:
Until now, if self.api._put() failed we got an exception but no reason about it. This commit unpacks the error provided by the server and propagates the exception to the caller.


JenySadadia (approved):
LGTM.
Let's wait for next staging run.

nuclearcat:
All good on staging, thanks!

PR #2327, opened 2024-01-29
runtime/kubernetes: Add trivial context load-balancing

nuclearcat:
If we have multiple clusters defined in config - pick random one. This is enough for now.


nuclearcat:
tested on staging

PR #2328, opened 2024-01-29
Dry run support for python runtime and kbuild

hardboprobot:
Requirement: https://github.com/kernelci/kernelci-core/pull/2326

If the input node specifies a "dry_run" result in its debug parameters, bypass all the script running, artifacts upload, etc. and simply return the fake result.

### DIFF
--- a/kernelci/kbuild.py
+++ b/kernelci/kbuild.py
@@ -121,7 +121,12 @@ class KBuild():
             self._artifacts = []
             self._current_job = None
             self._config_full = ''
-            self._srctarball = node['artifacts']['tarball']
+            if node.get('artifacts') and 'tarball' in node['artifacts']:
+                self._srctarball = node['artifacts']['tarball']
+            elif node.get('debug') and 'tarball' in node['debug']:
+                self._srctarball = node['debug']['tarball']
+            else:
+                raise ValueError("No tarball artifact in input node")
             self._srcdir = None
             self._firmware_dir = None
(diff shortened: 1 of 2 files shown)


kernelci/kbuild.py
```diff
@@ -110,7 +110,12 @@ def __init__(self, node=None, jobname=None, params=None, jsonobj=None):
             self._artifacts = []
             self._current_job = None
             self._config_full = ''
-            self._srctarball = node['artifacts']['tarball']
+            if node.get('artifacts') and 'tarball' in node['artifacts']:
+                self._srctarball = node['artifacts']['tarball']
+            elif node.get('debug') and 'tarball' in node['debug']:
+                self._srctarball = node['debug']['tarball']
```
a-wai, line 116:
```suggestion
            if node.get('debug') and 'tarball' in node['debug']:
                self._srctarball = node['debug']['tarball']
            elif node.get('artifacts') and 'tarball' in node['artifacts']:
                self._srctarball = node['artifacts']['tarball']
```

Maybe the other way around? So we always prioritize the `debug` parameters if present

a-wai (changes requested):

kernelci/kbuild.py
hardboprobot: Good idea, thanks!

nuclearcat (approved):
LGTM, staging is stable, so should be safe to merge

PR #2329, opened 2024-01-30
kubernetes.jinja2: Remove ssh key from k8s template

nuclearcat:
We are not using this feature/storage type in our environment.


nuclearcat:
tested on staging

Issue #2331, opened 2024-01-30
api: Add `kci docker` recipe for kernelci/api image


pawiecz:
Source [Dockerfile](https://github.com/kernelci/kernelci-api/blob/main/docker/api/Dockerfile.production)

nuclearcat:
I think this is completed?

PR #2332, opened 2024-01-30
Add recipe for kernelci/api-production image

pawiecz:
Verification (on the assumption environment for `kci` tool is set and there is `kernelci-api` repo in the CWD parent):
```
diff <(./kci docker generate api) ../kernelci-api/docker/api/Dockerfile.production
```
Difference should only include:
- stripped license header (moved to a relevant fragment from which Dockerfile is generated)
- cosmetic changes in comments

Then build the image with: `./kci docker build --prefix=kernelci/ api`

Fixes: https://github.com/kernelci/kernelci-core/issues/2331

### DIFF
--- a/config/docker/api.jinja2
+++ b/config/docker/api.jinja2
@@ -1,0 +1,8 @@
+{# SPDX-License-Identifier: LGPL-2.1-or-later
+ #
+ # Copyright (C) 2024 Collabora Limited
+ # Author: Paweł Wieczorek <pawiecz@collabora.com>
+-#}
+
+{%- set fragments = ['fragment/api.jinja2'] + fragments %}
+{%- include 'base/python-fixed.jinja2' %}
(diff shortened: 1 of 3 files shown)


pawiecz:
Please do not merge.

Upon inspection of deployment scripts I realized that split of user permissions fragment will break runtime environment for several other images.

I'll adjust this in the image rebuild script or drop the separation entirely.

nuclearcat:
I will add staging-skip label for now, feel free to remove or tell me to do so when PR is ready

pawiecz:
I added verification step to the PR description.

pawiecz:
I dropped "-production" suffixes or changed it to "-fixed" which also points out locked dependency versions.

Naming has been simplified so that there's would be no `kernelci/staging-api-production` abomination of an image produced by kernelci/kernelci-deploy#126.

PR #2333, opened 2024-01-30
kernelci.api.helper: fix node type in job result submission

hardboprobot:
Set the "kind" of child nodes to the same "kind" of the parent node, which is likely to be a "test".


nuclearcat (approved):
Tested in staging https://staging.kernelci.org:9000/viewer?search=parent%3D65b8ff0811ee0a70c2f58779
Fixes issue

PR #2336, opened 2024-01-31
kernelci.cli: add support for regexes in queries

hardboprobot:
Introduce the `=~` operator (a la perl) to match a query parameter using a mongodb-compatible regular expression.

Requirement: https://github.com/kernelci/kernelci-api/pull/460

Examples:

`./kci node find "data.platform=~qemu"`

`./kci node find "name=~baseline"`

`./kci node fine "name=~x86"`

**NOTE**: Due to the query parameters translation done by the API, not every mongodb-compatible regex expression is supported. For instance: `./kci node find "state=~done|running"` won't work as expected.


nuclearcat (approved):
Tested on staging, works!

PR #2338, opened 2024-02-01
rootfs-images.yaml: Update most of images to recent build

nuclearcat:
We will start gradually updating rootfs images.
First verification - newer build of bullseye and in some cases bookworm.


nuclearcat:
Looking to results, looks like all ok, but on legacy it is not easy to spot issues...

PR #2340, opened 2024-02-01
rootfs-configs.yaml: Add liburing for mm selftests

musamaanjum:
The mm selftests require liburing. The binaries error out if liburing
isn't found at runtime. This will fix the following class of errors:

error while loading shared libraries: liburing.so.1: cannot open
shared object file: No such file or directory
https://lava.collabora.dev/scheduler/job/12674655#L3643

### DIFF
--- a/config/core/rootfs-configs.yaml
+++ b/config/core/rootfs-configs.yaml
@@ -84,6 +84,7 @@ rootfs_configs:
       - libpcre2-8-0
       - libperl5.36
       - libpsl5
+      - liburing2
       - libxtables12
       - netbase
       - openssl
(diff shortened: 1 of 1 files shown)


config/core/rootfs-configs.yaml
```diff
@@ -84,6 +84,7 @@ rootfs_configs:
       - libpcre2-8-0
       - libperl5.36
       - libpsl5
+      - liburing-dev
```
obbardc, line 87: The other library packages do not install `-dev` package (e.g. development headers).

Is it worth following exiting convention and installing the `liburing2` package directly?

obbardc (changes requested):

obbardc (approved):

nuclearcat (approved):
kselftest images rebuilt with this PR

PR #2347, opened 2024-02-05
kbuild.py: Refresh node before submit

nuclearcat:
As node might have new data, retrieve it from API just before submitting updated one.

### DIFF
--- a/kernelci/kbuild.py
+++ b/kernelci/kbuild.py
@@ -652,7 +652,6 @@ class KBuild():
         else:
             af_uri = self.upload_artifacts()
         print("[_submit] Submitting results to API")
-        af_uri = self.upload_artifacts()
         api_token = os.environ.get('KCI_API_TOKEN')
         if not api_token:
             raise ValueError("KCI_API_TOKEN is not set")
@@ -660,6 +659,10 @@ class KBuild():
             yaml.safe_load(self._api_yaml), name='api'
         )
         api = kernelci.api.get_api(api_config, api_token)
+        # Update node from API, as we might have new fields
+        # such as k8s_context
+        node_id = self._node['id']
+        self._node = api.node.get(node_id)
         # TODO/FIXME: real detailed job result
         # pass fail skip incomplete
         if retcode != 0:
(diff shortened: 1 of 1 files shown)


hardboprobot (approved):
LGTM but I'm noticing that line 705:

```
results['node']['data']['kernel_revision'] = node['data']['kernel_revision']  # noqa
```

shouldn't be necessary and is redundant, isn't it? If it's safe to remove it maybe it can be done in this same PR

nuclearcat:
yes, probably leftover from time when i was having issues with submit_data method

nuclearcat:
Worked: https://staging.kernelci.org:9000/viewer?node_id=65c0e5246c0c48387d45916c

I will remove mentioned line, but it will require one more run on staging:
results['node']['data']['kernel_revision'] = node['data']['kernel_revision']  # noqa

kernelci/kbuild.py
```diff
@@ -699,16 +699,17 @@ def submit(self, retcode, dry_run=False):
             },
             'child_nodes': [],
         }
-        # TODO: Fix this hack
+        # Update node from API, as we might have new fields
+        # such as k8s_context
+        nodeid = self._node['id']
+        self._node = api.node.get(nodeid)
```
a-wai, line 705: Can this be done at the start of the function? In #2342 the node data with a new field in upload_artifacts(), IIUC this change would be lost if the node is refreshed afterwards.

kernelci/kbuild.py
```diff
@@ -699,16 +699,17 @@ def submit(self, retcode, dry_run=False):
             },
             'child_nodes': [],
         }
-        # TODO: Fix this hack
+        # Update node from API, as we might have new fields
+        # such as k8s_context
+        nodeid = self._node['id']
```
JenySadadia, line 704: For better readability,
```suggestion
        node_id = self._node['id']
```

kernelci/kbuild.py
```diff
@@ -699,16 +699,17 @@ def submit(self, retcode, dry_run=False):
...
+        self._node = api.node.get(nodeid)
         node = self._node.copy()
         results['node']['data'] = node['data']
-        results['node']['data']['kernel_revision'] = node['data']['kernel_revision']  # noqa
         results['node']['data']['arch'] = self._arch
         results['node']['data']['compiler'] = self._compiler
         results['node']['data']['defconfig'] = self._defconfig
         results['node']['data']['fragments'] = self._fragments
         results['node']['data']['config_full'] = self._config_full
-        results['node']['data']['dtb_present'] = dtb_present
```
JenySadadia, line 711: Why is this removed?

a-wai: This is not actually needed (`platform_config.dtb` is set for devices expecting a dtb, checking this field should be enough)

nuclearcat: I will add it to commit description, it was added shortly to assist @a-wai with ARM devices, but he found better way of detecting dtb.

nuclearcat:
@hardboprobot please confirm if my change to "        af_uri = self.upload_artifacts()" is correct, i think it is redundant.

Issue #2350, opened 2024-02-07
pipeline: Add `kci docker` recipe for kernelci/pipeline image


pawiecz:
Source [Dockerfile](https://github.com/kernelci/kernelci-pipeline/blob/main/docker/pipeline/Dockerfile.production)

PR #2354, opened 2024-02-07
Add recipe for kernelci/pipeline image

pawiecz:
This patch duplicates the whole `kernelci/k8s:kernelci` image recipe so that it can be built without depending on it as a base image.

The same result could be achieved by:

```
kci build --prefix=kernelci/ k8s kernelci pipeline
```

Resulting image would be `kernelci/k8s:kernelci-pipeline`. To get `kernelci/pipeline` image out of it, retagging image would be required (which could result in more confusion in staging/production deployment scripts) or alternatively providing a custom name at build (which would require rewriting name handling in `kci docker` tool).

This duplication will probably require revisiting in the future but for now the key deliverable is to provide versioned production images using `kci docker` tool.

Fixes: #2350

### DIFF
--- a/config/docker/fragment/pipeline.jinja2
+++ b/config/docker/fragment/pipeline.jinja2
@@ -1,0 +1,6 @@
+ARG pipeline_url=https://github.com/kernelci/kernelci-pipeline.git
+ARG pipeline_rev=main
+RUN git clone --depth=1 $pipeline_url pipeline
+WORKDIR pipeline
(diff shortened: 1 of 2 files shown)


nuclearcat:
Tested on staging:
```
./kci docker build --prefix=kernelci/ k8s kernelci pipeline
Building kernelci/k8s:kernelci-pipeline
```
Completed fine

PR #2356, opened 2024-02-07
Rework templates and add Tast tests

a-wai:
Chromebooks use a ChromeOS-specific test suite called "Tast", requiring a different workflow than baseline tests:
* those tests run under ChromeOS, requiring an altered boot process
* they involve more that just running a set of shell commands


a-wai:
Depends on https://github.com/kernelci/kernelci-pipeline/pull/418

a-wai:
Cherry-picks https://github.com/kernelci/kernelci-core/pull/2376 as both PRs target different branches, so we don't end up with a fix missing in one of the main branches.

a-wai:
Marking as ready as it was successfully tested on `puff`: https://lava.collabora.dev/scheduler/job/12768684

nuclearcat (approved):
Excellent, thanks!

PR #2359, opened 2024-02-08
lava.jinja2: Hardcode priority as 10, lower than medium

nuclearcat:
As noted by Mark, priority should be low, to not interfere with other lab jobs. We need to open issue also to make this value configurable at least per lab.

### DIFF
--- a/config/runtime/base/lava.jinja2
+++ b/config/runtime/base/lava.jinja2
@@ -21,7 +21,7 @@ device_type: {{ platform_config.base_name }}
 
 visibility: public
 
-priority: medium
+priority: 10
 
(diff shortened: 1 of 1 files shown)


broonie:
What I did for the existing setup was allow labs to specify a priority range then scale the priority for the jobs into that range, that way the API can still schedule things itself and the lab gets to control KernelCI priority vs the rest of it's workload.

nuclearcat:
Tested on staging: https://lava.collabora.dev/scheduler/job/12724039
@broonie - yes, this is next step, as mentioned in linked issue i plan to specify allowed priorities for each lab, but for now will keep it on safe 10 :)

PR #2361, opened 2024-02-08
models.py: Add node_timeout in errors

nuclearcat:
We need to indicate when node reached timeout,
it is likely infrastructure issue, not real node failure.


nuclearcat:
Tested on staging

PR #2363, opened 2024-02-09
runtime/lava: Add method to get job status

nuclearcat:
We need to retrieve job status, to know if it was executed properly or failed.

### DIFF
--- a/kernelci/runtime/lava.py
+++ b/kernelci/runtime/lava.py
@@ -66,10 +66,10 @@ class Callback:
 
     # LAVA job result names
     LAVA_JOB_RESULT_NAMES = {
-        COMPLETE: "PASS",
-        INCOMPLETE: "FAIL",
-        CANCELED: "UNKNOWN",
-        CANCELING: "UNKNOWN",
+        COMPLETE: "pass",
+        INCOMPLETE: "incomplete",
+        CANCELED: "incomplete",
+        CANCELING: "incomplete",
     }
 
     def __init__(self, data):
@@ -82,6 +82,11 @@ class Callback:
         if self._meta is None:
             self._meta = yaml.safe_load(self._data['definition'])['metadata']
         return self._meta.get(key)
+
+    def get_job_status(self):
+        """Get the job status"""
+        # map over LAVA_JOB_RESULT_NAMES
+        return self.LAVA_JOB_RESULT_NAMES.get(self._data['status'])
 
     def is_infra_error(self):
         """Determine wether the job has hit an infrastructure error"""


kernelci/runtime/lava.py
```diff
@@ -66,10 +66,10 @@ class Callback:
...
     # LAVA job result names
     LAVA_JOB_RESULT_NAMES = {
-        COMPLETE: "PASS",
-        INCOMPLETE: "FAIL",
-        CANCELED: "UNKNOWN",
-        CANCELING: "UNKNOWN",
+        COMPLETE: "pass",
+        INCOMPLETE: "incomplete",
+        CANCELED: "incomplete",
+        CANCELING: "incomplete",
```
JenySadadia, line 72: What status does lava use for failed jobs?

JenySadadia: I mean, how would we denote that the job ran successfully but somehow failed?

nuclearcat: Unfortunately LAVA documentation seems not very consitent on this topic. For example here is described LAVA TestJob state machine:
https://lava.readthedocs.io/en/latest/technical-references/state-machine/#testjobs
But there is nothing about results, it has paragraph "Results" in that doc, but it is empty. 
So what i did, i submitted job with invalid rootfs URL, recorded callback details/content, and noted which result code job will return on invalid rootfs url (that will cause job to fail shortly after submission), and what result code returned on "good" job.

JenySadadia: Yes, it's weird that the documentation is empty.
So what result code did you receive?
Maybe @pawiecz is aware of the different lava job results?

nuclearcat:
```
{
    "id": 12724007,
    "status": 3,
    "status_string": "incomplete",
    "state": 5,
    "state_string": "Finished",
    "health": 2,
    "health_string": "Incomplete",
    "submit_time": "2024-02-09 08:01:24.713834+00:00",
    "start_time": "2024-02-09 08:01:40.110214+00:00",
    "end_time": "2024-02-09 08:02:00.006301+00:00",
    "submitter_username": "nuclearcat",
    "failure_comment": null,
    "priority": 10,
    "description": "[kci-staging] 65c5daef4c1455ded8894700 baseline-x86 staging-stable-20240209.0",
    "actual_device_id": "qemu-hetzner-9",
    "definition": --skip-definitio--
    "metadata": [],
    "token": "kernelci-api-token-staging",
    "log": "-- skip-log--
...
```

nuclearcat:
Tested on staging, works well and helping now with broken LAVA templates and etc

PR #2364, opened 2024-02-09
test-configs.yaml: Enable MM selftests on Avenger96

broonie:
The Avenger96 has a reasonable amount of memory so should be an
interesting candidate for the MM selftests.


nuclearcat:

PR #2367, opened 2024-02-12
models.py: Move error_code to data parent

nuclearcat:
By mistake i defined error_code directly in node root, while this field might not exist for all kinds of nodes. Moving error_code in model definition to kbuild and test nodes.


JenySadadia (approved):
LGTM.

nuclearcat:
Tested on staging

PR #2370, opened 2024-02-12
kbuild: Remove stray break

nuclearcat:
This break will cause only first kernel to be added to artifacts.


nuclearcat:
Tested on staging, now it is producing proper armel image.

PR #2371, opened 2024-02-12
api/helper.py: Inherit data.kernel_type for tests

nuclearcat:
To make rendering of test jobs simpler, without fetching or storing parent node data, inherit parameter kernel_type to test node from parent.


a-wai (approved):

nuclearcat:
armel devices now works fine

PR #2373, opened 2024-02-12
Get platforms/device types from pipeline configuration for the new system

a-wai:
The current platform config is taken exclusively from the legacy `device_types` configuration. We have to parse similar parameters for the new system, avoiding collisions with legacy config.

This PR creates a new class for parsing the new system's config and makes use of it instead of the legacy config.

### DIFF
--- a/kernelci/runtime/__init__.py
+++ b/kernelci/runtime/__init__.py
@@ -138,7 +138,7 @@ class Runtime(abc.ABC):
         params = {
             'api_config': api_config or {},
             'storage_config': job.storage_config or {},
-            'platform_config': job.platform_config.to_dict() or {},
+            'platform_config': job.platform_config or {},
             'instanceid': instanceid,
             'name': job.name,
             'node': job.node,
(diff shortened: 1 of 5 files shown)


kernelci/runtime/__init__.py
```diff
@@ -138,15 +138,16 @@ def get_params(self, job, api_config=None):
         params = {
             'api_config': api_config or {},
             'storage_config': job.storage_config or {},
-            'platform_config': job.platform_config.to_dict() or {},
+            'platform_config': job.platform_config or {},
```
pawiecz, line 141: What would be the reason not to convert it to dictionary? This already covered `None` case fixed in L149 below

a-wai: The reasons are consistency (other config objects aren't converted that way, and it isn't needed) and code simplicity: the `to_dict()` method isn't part of the `YAMLConfigObject` base class but is defined (for an unknown reason) in [_YAMLObject](https://github.com/kernelci/kernelci-core/blob/main/kernelci/config/base.py#L74).

nuclearcat:
Tested on staging

PR #2374, opened 2024-02-13
models: allow `kernel_type` field in kbuild/test node data

a-wai:
According to my local cli-based tests, this could be the missing bit for actually propagating this field to test nodes.

### DIFF
--- a/kernelci/api/models.py
+++ b/kernelci/api/models.py
@@ -339,6 +339,9 @@ class KbuildData(BaseModel):
     job_context: Optional[str] = Field(
         description="Kubernetes cluster name the job submitted to"
     )
+    kernel_type: Optional[str] = Field(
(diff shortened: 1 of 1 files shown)


hardboprobot (commented):
I can understand this field in Kbuild nodes, but why is it needed for test nodes? What's the use case for it?

a-wai:
LAVA jobs for devices using u-boot require this parameter, otherwise they fail like https://lava.collabora.dev/scheduler/job/12748862

nuclearcat:
Tested on staging

a-wai:
@hardboprobot I definitely agree with you, which is why I added the TODO in the code. That's actually something I could look at tomorrow (probably initiating a 1st PR just fetching the data from the node as a dict, and maybe go a bit further using helper functions like yours).

PR #2375, opened 2024-02-13
docker/fragment/api.jinja2: install kernelci-api using pyproject.toml

JenySadadia:
Depends on https://github.com/kernelci/kernelci-api/pull/473

Use `pyproject.toml` file to build and install `kernelci-api`. As the `kernelci` python package is not being updated on PyPI yet, add adjustments to install the package from `git` URL of `kernelci-core` repo. It can be discarded later once we include `kernelci` package as a requirement in the TOML.


nuclearcat:
tested on staging, merging

PR #2376, opened 2024-02-13
CHROMEOS rootfs: debos: cros-flash: fix install-modules for new system

a-wai:
The new KernelCI system provides URLs with appended GET parameters for authentication, causing this script to fail due to some of those being URL-encoded in the base URL, but decoded in the file name.

This change fixes this behavior.


a-wai:
@nuclearcat Once this is merged, the `bookworm-cros-flash` should be re-generated and the URL updated. Please let me know when you have some time to build/upload a new rootfs and I'll handle the rest.

nuclearcat (approved):
Thank you for neat, well commented code!

PR #2377, opened 2024-02-14
fragment/kernelci.jinja2: drop requirements installation using `pip`

JenySadadia:
As the support of building and installing `kernelci-core` package using `pyproject.toml` has been in place, delete a line to explicitly install requirements as the TOML would take care of it internally.

Fixes: cc937787 ("config/docker/fragment: update `kernelci.jinja2`")


nuclearcat:
Tested on staging, merging

PR #2378, opened 2024-02-14
More utility functions for the kernelci-core api helper

hardboprobot:
- `is_test_suite()`: function to tell if a Test node represents a test case or a test suite, as we don't have this information in the models yet
- `get_node_obj()`: returns a node dict as a proper object, optionally with resolved links to nodes also fetched as objects.


kernelci/api/helper.py
```diff
@@ -217,6 +217,47 @@ def submit_results(self, results, root):
...
+
+        generic_node_obj = models.Node.parse_obj(node_dict)
+        node_obj = models.parse_node_obj(generic_node_obj)
+        if get_linked:
+            for linked_node_attr in node_obj._OBJECT_ID_FIELDS:
+                node_id = get_attr(node_obj, linked_node_attr)
+                if node_id:
+                    resp = self.api.node.get(node_id)
+                    linked_node_obj = models.Node.parse_obj(resp)
+                    linked_obj = models.parse_node_obj(linked_node_obj)
```
a-wai, line 257: Would it make sense to improve `models.parse_node_obj()` so it:
* detects if its argument is a dict
* parse it as a generic Node first
* then moves on trying to make it a subclass of Node

hardboprobot: Yes, that'll make this code cleaner. I'll try it.

hardboprobot: Done

a-wai (approved):
LGTM, not sure the linter error can be easily addressed

kernelci/api/helper.py
```diff
@@ -246,6 +259,46 @@ def submit_results(self, results, root):
         except requests.exceptions.HTTPError as error:
             raise RuntimeError(json.loads(error.response.text)) from error
 
+    def get_node_obj(self, node_dict, get_linked=False):
```
JenySadadia, line 262: If the objective here is to get an object from a dictionary, we can simply use `Node(**dict)` for it.
But I assume this would require importing model definitions and that wouldn't be possible due to python version conflicts.

hardboprobot: That would create a `Node` object, with no validation of the `data` fields depending on the node kind. What I want in this case is a `Node` submodel object, that's the whole point of using `models.parse_node_obj()` instead. Additionally, it'll perform validation on the whole set of fields, including those inside `data`, depending on the node kind, which `Node(**dict)` won't do.

JenySadadia: That's true.
It's good to add a comment to replace this function with `model.parse_node_obj` once we resolve python version issue.

hardboprobot: Clarification: `get_node_obj()` uses `model.parse_node_obj()` to build an object from a dict, but it does more things on top of that, as explained in the function docstring. This function is not supposed to substitute `model.parse_node_obj()`.

nuclearcat:
Do we still need this PR?

PR #2379, opened 2024-02-14
helper.py: Merge root node with refreshed node

nuclearcat:
It might create complicated logic if submitter have his own updated data as well, for example new artifacts.
Instead, refresh node in submitter (kunit and lava_callback at moment)
Related-to: https://github.com/kernelci/kernelci-pipeline/pull/434

### DIFF
--- a/kernelci/api/helper.py
+++ b/kernelci/api/helper.py
@@ -13,6 +13,27 @@ import requests
 from . import API
 
 
+def merge(primary: dict, secondary: dict):
+    """Deep merges dicts a and b, returning a new dict containing
+    dictionary a updated with the contents of dictionary b.
+    TODO: This might need to be moved elsewhere, with proper fix to
+    https://github.com/kernelci/kernelci-core/issues/2386
+    """
+    result = {}
+    for key in primary:
+        result[key] = primary[key]
+        if key in secondary:
+            if isinstance(primary[key], dict) and \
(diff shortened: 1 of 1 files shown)


JenySadadia (commented):
Although the PR is still not ready for review, I'd suggest if we drop this line, we'll again end up losing metadata fields of `data`. Instead, we should consider merging `data` here just like https://github.com/kernelci/kernelci-core/pull/2379/files#diff-8c7508fe0635fa037ca61dcac0f618ba15878c5a5a6b81d538110f436f759ca9R143

nuclearcat:
@hardboprobot i used your merge code snippet, just renamed variables, because of pylint:
```
kernelci/api/helper.py:16:10: C0103: Argument name "a" doesn't conform to snake_case naming style (invalid-name)
kernelci/api/helper.py:16:19: C0103: Argument name "b" doesn't conform to snake_case naming style (invalid-name)
```

hardboprobot (approved):
If you have tested this and it fixes the bug you were after and it doesn't break anything, then ILGTM, but I think it's rather urgent to rework the way nodes are submitted because it's starting to get messy with so many different corner cases. IMO there should be **one** path to submit nodes, and any specific use cases should be managed as "decorators" of that, independently from it, in a way that's obvious and clear.

PR #2380, opened 2024-02-14
kbuild: add build config to artifacts

a-wai:
This makes it easier to check whether the final config is correct.


a-wai:
Hmm, actually `config_full` isn't properly handled here, I'll have to fix that first...

PR #2381, opened 2024-02-15
Bump minimum python version to 3.10

hardboprobot:


JenySadadia (approved):
LGTM.

pawiecz:
LGTM too

I had a slight concern about having 3.9 in the execution environment for [prod](https://github.com/kernelci/kernelci-deploy/blob/main/kernelci.org)/[staging](https://github.com/kernelci/kernelci-deploy/blob/main/staging.kernelci.org)/[ChromeOS](https://github.com/kernelci/kernelci-deploy/blob/main/chromeos.kernelci.org) update scripts but it looks like it should be fine for now:

- (prod) 16 calls to `kci` tools: 15 to `kci docker`, one to `kci_test`
- (stage) 16 calls to `kci docker`
- (chromeos) 7 calls to `kci docker`

pawiecz (approved):

pawiecz:
Turns out I did not verify it thoroughly enough: those `kci docker` commands make attempts to install KernelCI Python package using Py 3.9.2, e.g.

1. [./kci docker $args k8s kernelci](https://github.com/kernelci/kernelci-deploy/blob/dd483af654d7f21287c344366e461af7983d1ba5/staging.kernelci.org#L245)
2. uses [k8s recipe](https://github.com/kernelci/kernelci-core/blob/main/config/docker/k8s.jinja2) which is based off [debian.jinja2](https://github.com/kernelci/kernelci-core/blob/main/config/docker/base/debian.jinja2)
3. which still uses `bullseye` as a [base image](https://github.com/kernelci/kernelci-core/blob/b68f435edf231240e724d9c633ed787f4bb88e3d/config/docker/base/debian.jinja2#L9)

It's far from perfect solution but could we keep 3.10 requirement in the GitHub workflows and relax `pyproject.toml` requirements to also allow 3.9?

JenySadadia:
> Turns out I did not verify it thoroughly enough: those `kci docker` commands make attempts to install KernelCI Python package using Py 3.9.2, e.g.
> 
> 1. [./kci docker $args k8s kernelci](https://github.com/kernelci/kernelci-deploy/blob/dd483af654d7f21287c344366e461af7983d1ba5/staging.kernelci.org#L245)
> 2. uses [k8s recipe](https://github.com/kernelci/kernelci-core/blob/main/config/docker/k8s.jinja2) which is based off [debian.jinja2](https://github.com/kernelci/kernelci-core/blob/main/config/docker/base/debian.jinja2)
> 3. which still uses `bullseye` as a [base image](https://github.com/kernelci/kernelci-core/blob/b68f435edf231240e724d9c633ed787f4bb88e3d/config/docker/base/debian.jinja2#L9)
> 
> It's far from perfect solution but could we keep 3.10 requirement in the GitHub workflows and relax `pyproject.toml` requirements to also allow 3.9?

@pawiecz Thank you for your findings.
This means we can't do much about this but use `3.9` until the `bullseye` gets upgraded with a newer python version.
Is that true?
I am wondering what if we install a different python version in `debian.jinja2` template? Or that would break the `debian:bullseye` image?

nuclearcat:
I will add skip-staging label as at moment it is breaking staging

pyproject.toml
```diff
@@ -48,8 +48,6 @@ classifiers = [
   'Operating System :: POSIX :: Linux',
   'Programming Language :: C',
   'Programming Language :: Python',
-  'Programming Language :: Python :: 3.8',
-  'Programming Language :: Python :: 3.9',
```
pawiecz, line 52: I'm unable to make it a GH code suggestion on deleted lines but I'd leave the "3.9" line as it was

pawiecz:
@JenySadadia For a long term solution I believe a way to go would be updating `debian.jinja2` base image to current stable (`bookworm`). This, however, would involve at least:

1. List all images built with `kci docker` based on `debian.jinja2`
2. List key acceptance criteria for those ^ to make sure that images rebuilt on a new base kept all required functionalities
3. Run tests ^ on rebuilt images

But we can do that after M1 - for now let's go with relaxed package requirements and strict GH workflow.

pawiecz (changes requested):

pyproject.toml
hardboprobot: Done, thanks for the suggestions

pyproject.toml
```diff
@@ -12,7 +12,7 @@ name = "kernelci"
 version = "1.1"
 description = "KernelCI core functions"
 readme = "README.md"
-requires-python = ">=3.8"
+requires-python = ">=3.10"
```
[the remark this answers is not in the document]
hardboprobot: Done, thanks!

pawiecz (approved):

pawiecz:
LGTM, I'm taking `staging-skip` label off; I'd merge it but I lack permissions to do so

JenySadadia:
I'll merge it. Thanks.

PR #2382, opened 2024-02-15
kernelci.api.models: use `StrictInt` for version fields

JenySadadia:
For `int` fields, pydantic internally converts fields to integer if possible e.g. request field value `"2"` is converted to `2` in the FastAPI request handler for model parsing. That's why `KernelVersion.version` is converted to `int` when received by API.
Instead of relying on implicit data type conversion for `int` field, use `StrictInt` to enforce defined data type in the request itself.

### DIFF
--- a/kernelci/api/models.py
+++ b/kernelci/api/models.py
@@ -22,6 +22,7 @@ from pydantic import (
     BaseModel,
     Field,
     FileUrl,
+    StrictInt,
 )
 from .models_base import (
     PyObjectId,
@@ -63,13 +64,13 @@ class ErrorCodes(str, enum.Enum):
 
 class KernelVersion(BaseModel):
     """Linux kernel version model"""
-    version: int = Field(
+    version: StrictInt = Field(
         description="Major version number e.g. 4 in 'v4.19'"
     )
-    patchlevel: int = Field(
+    patchlevel: StrictInt = Field(
         description="Minor version number or 'patch level' e.g. 19 in 'v4.19'"
     )
-    sublevel: Optional[int] = Field(
+    sublevel: Optional[StrictInt] = Field(
         description="Stable version or 'sub-level' e.g. 123 in 'v4.19.123'"
     )
     extra: Optional[str] = Field(
--- a/kernelci/api/models.py
+++ b/kernelci/api/models.py
@@ -80,6 +80,16 @@ class KernelVersion(BaseModel):
         description="Version name e.g. People's Front for v4.19"
     )
 
+    _STRICT_INT_FIELDS = ['version', 'patchlevel', 'sublevel']
+
+    @classmethod
+    def translate_version_fields(cls, params):
+        """Translate `StrictInt` field values into `int`"""
+        for key, value in params.items():
+            if key in cls._STRICT_INT_FIELDS:
+                params[key] = int(value)
+        return params
+
 
 class Revision(BaseModel):
     """Linux kernel Git revision model"""


hardboprobot (approved):

nuclearcat:
Causing exception during kci tools use on staging:
```
Build docker images: kernelci
  File "/home/nuclearcat/Documents/kci-easy/kernelci/kernelci-core/./kci", line 17, in <module>
    from kernelci.cli import kci  # noqa: E402
  File "/home/nuclearcat/Documents/kci-easy/kernelci/kernelci-core/kernelci/cli/__init__.py", line 26, in <module>
    import kernelci.api.helper
  File "/home/nuclearcat/Documents/kci-easy/kernelci/kernelci-core/kernelci/api/helper.py", line 14, in <module>
    from .models import KernelVersion
  File "/home/nuclearcat/Documents/kci-easy/kernelci/kernelci-core/kernelci/api/models.py", line 97, in <module>
    class Revision(BaseModel):
  File "/home/nuclearcat/Documents/kci-easy/kernelci/kernelci-core/kernelci/api/models.py", line 102, in Revision
    url: AnyUrl | FileUrl = Field(
TypeError: unsupported operand type(s) for |: 'type' and 'type'
```

kernelci/api/models.py
```diff
@@ -79,6 +80,19 @@ class KernelVersion(BaseModel):
         description="Version name e.g. People's Front for v4.19"
     )
 
+    _STRICT_INT_FIELDS = ['version', 'patchlevel', 'sublevel']
+
+    @classmethod
+    def translate_version_fields(cls, params):
```
hardboprobot, line 86: Ooc why a generator for this instead of a regular function? What would be the use case for this as a generator?

pawiecz:
As @hardboprobot noted, this indeed looks like Py <= 3.10 issue.

@nuclearcat Please provide more details to track down this issue (e.g. command called, exec env - [kci-easy](https://github.com/nuclearcat/kci-easy)?).

nuclearcat:
Yes, it can be reproduced also in latest kci-easy during build on debian bullseye.

JenySadadia:
> @nuclearcat that's supposedly related to something in the staging environment using python < 3.10. @pawiecz and @JenySadadia know about this in more detail. Related: #2381

`|` operator was introduced in python 3.10 and not available in 3.9
Please see https://github.com/tiangolo/typer/issues/371

pawiecz:
> `|` operator was introduced in python 3.10 and not available in 3.9

Right - I'm afraid [importing models in helpers used by cli](https://github.com/kernelci/kernelci-core/pull/2382/commits/ae8491313df9d31ea5ee75ba8bc067dc292a14f1#diff-8c7508fe0635fa037ca61dcac0f618ba15878c5a5a6b81d538110f436f759ca9R14) will be a no-go until we drop 3.9 entirely

JenySadadia:
> Right - I'm afraid [importing models in helpers used by cli](https://github.com/kernelci/kernelci-core/pull/2382/commits/ae8491313df9d31ea5ee75ba8bc067dc292a14f1#diff-8c7508fe0635fa037ca61dcac0f618ba15878c5a5a6b81d538110f436f759ca9R14) will be a no-go until we drop 3.9 entirely

True. Now, we are left with 2 options: one is type cast model fields in the tarball service but that is leaking model info. Another option is we can use `KernelVersion.translate_version_fields` in `api.main.post_node` handler itself to translate fields before parsing. We are using these kind of handlers for `get_nodes` atm.

kernelci/api/models.py
JenySadadia: It should be a regular function. Updated it.

JenySadadia:
> > Right - I'm afraid [importing models in helpers used by cli](https://github.com/kernelci/kernelci-core/pull/2382/commits/ae8491313df9d31ea5ee75ba8bc067dc292a14f1#diff-8c7508fe0635fa037ca61dcac0f618ba15878c5a5a6b81d538110f436f759ca9R14) will be a no-go until we drop 3.9 entirely
> 
> True. Now, we are left with 2 options: one is type cast model fields in the tarball service but that is leaking model info. Another option is we can use `KernelVersion.translate_version_fields` in `api.main.post_node` handler itself to translate fields before parsing. We are using these kind of handlers for `get_nodes` atm.

See https://github.com/kernelci/kernelci-api/pull/483

hardboprobot (approved):

PR #2384, opened 2024-02-15
kernelci.api.helper: unpack and propagate errors in submit_regression

hardboprobot:
Until now, if self.api._post() failed we got an exception but no reason about it. This commit unpacks the error provided by the server and propagates the exception to the caller.


JenySadadia (approved):

nuclearcat:
LGTM, but i am not sure it will help, as server gives error 500, which is likely text, not json, will do test run now.

nuclearcat:
Thanks, lgtm

PR #2385, opened 2024-02-15
test-configs.yaml: Run basic tests for mt8186-corsola-steelix-sku131072

nfraprado:
Starting with next-20240215, the mt8186-corsola-steelix-sku131072 machine is able to boot. Enable the baseline and kselftest-dt tests on it so we can track the base support: booting and device probing.

Baseline LAVA run: https://lava.collabora.dev/scheduler/job/12769714
DT kselftest LAVA run: https://lava.collabora.dev/scheduler/job/12769689

(Not all tests currently pass but that's fine, will be addressed upstream over time, the important thing is that it boots and the tests run).

### DIFF
--- a/config/core/test-configs.yaml
+++ b/config/core/test-configs.yaml
@@ -1689,6 +1689,12 @@ device_types:
     boot_method: depthcharge
     filters: *arm64-chromebook-filters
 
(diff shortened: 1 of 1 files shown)


nfraprado:
The test is already running in the new KernelCI.

PR #2387, opened 2024-02-16
Add data fields to Test and Regression models

hardboprobot:
In order to perform effective api queries to search for regressions and test results we need to have these new fields in the models. In the case of tests, these will be populated from the parent node (either a kbuild or a test case node). In the case of regressions, they'll be taken from the associated test runs.

**WARNING:** Not tested locally.

### DIFF
--- a/kernelci/api/helper.py
+++ b/kernelci/api/helper.py
@@ -116,10 +116,13 @@ class APIHelper:
                 'kernel_revision': input_node['data']['kernel_revision'],
             },
         }
-        # for test kind, if node.data.kernel_type - copy to test node
-        # it is required for LAVA templates
-        if 'kernel_type' in input_node['data'] and job_config.kind == 'test':
-            job_node['data']['kernel_type'] = input_node['data']['kernel_type']
+        # Test-specific fields inherited from parent node (kbuild or
+        # test) if available
+        if job_config.kind == 'test':
+            job_node['data']['kernel_type'] = input_node['data'].get('kernel_type')  # noqa
+            job_node['data']['arch'] = input_node['data'].get('arch')
+            job_node['data']['defconfig'] = input_node['data'].get('defconfig')
+            job_node['data']['compiler'] = input_node['data'].get('compiler')
         # This information is highly useful, as we might
         # extract from it the following, for example:
         # in case of lab: lab-name, device-name
--- a/kernelci/api/models.py
+++ b/kernelci/api/models.py
@@ -422,6 +422,21 @@ class RegressionData(BaseModel):
     pass_node: Optional[PyObjectId] = Field(
         description="Previous passing Node"
     )
+    failed_kernel_revision: Optional[Revision] = Field(
+        description="Kernel repo revision data of the failed test run"
+    )
+    arch: Optional[str] = Field(
+        description="CPU architecture family"
+    )
+    defconfig: Optional[str] = Field(
+        description="Kernel defconfig identifier"
+    )
+    compiler: Optional[str] = Field(
+        description="Compiler used for the build"
+    )
+    platform: Optional[str] = Field(
+        description="Test platform"
+    )
 
 
 class Regression(Node):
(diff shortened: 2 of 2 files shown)


kernelci/api/helper.py
```diff
@@ -116,10 +116,14 @@ def create_job_node(self, job_config, input_node,
...
-        if 'kernel_type' in input_node['data'] and job_config.kind == 'test':
-            job_node['data']['kernel_type'] = input_node['data']['kernel_type']
+        # Test-specific fields inherited from parent node (kbuild or
+        # test) if available
+        if job_config.kind == 'test':
+            job_node['data']['kernel_type'] = input_node['data'].get('kernel_type')  # noqa
+            job_node['data']['arch'] = input_node['data'].get('arch')
+            job_node['data']['defconfig'] = input_node['data'].get('defconfig')
+            job_node['data']['compiler'] = input_node['data'].get('compiler')
+            job_node['data']['arch'] = input_node['data'].get('arch')
```
a-wai, line 126: Duplicate line

hardboprobot: Wow good catch. Fixed

kernelci/api/helper.py
```diff
@@ -116,10 +116,13 @@ def create_job_node(self, job_config, input_node,
...
             },
         }
-        # for test kind, if node.data.kernel_type - copy to test node
-        # it is required for LAVA templates
-        if 'kernel_type' in input_node['data'] and job_config.kind == 'test':
-            job_node['data']['kernel_type'] = input_node['data']['kernel_type']
+        # Test-specific fields inherited from parent node (kbuild or
+        # test) if available
+        if job_config.kind == 'test':
+            job_node['data']['kernel_type'] = input_node['data'].get('kernel_type')  # noqa
```
a-wai, line 122: We have a lot of `#noqa` all over the place, often because of line length checks... Shall we consider adding a config for pycodestyle with an increased max line length? (I mean, it's 2024, we have screens and editors that can fit more than 79 chars per line)

a-wai (approved):

kernelci/api/helper.py
hardboprobot: Most of the pycodestyle warnings I get are moot, dogmatic and useless. So I'm all for disabling some of the checks, but I understand it might be a controversial move, so I'd leave that for a future PR where we can hopefully make the config change and cleanup `noqa` markers

a-wai: Absolutely, I wasn't expecting such changes to land in this PR, I created #2396 so we can further discuss this.

a-wai:
You must use the name of a job defined in `pipeline.yaml`, like `baseline-x86` for example.

a-wai:
Ah, another case of "works for me, not for others", yay... I assume you did point it to pipeline.yaml using `-c /path/to/kernelci-pipeline/config/pipeline.yaml` ? FYI the exact steps I just followed:

* start the api and pipeline services using `docker-compose` in both repos
* `cd /path/to/kernelci-core`
* copy json from https://staging.kernelci.org:9000/viewer?node_id=65bcd41d1893bb5c371d55ec and paste it into `new.json`, removing the `id` and `parent` lines in the process
* `./kci node submit -c ../pipeline/config/pipeline.yaml new.json` -> creates the node with id `65d46f2b5d02a10c3c189bf5`
* `./kci job new -c ../pipeline/config/pipeline.yaml --node-id 65d46f2b5d02a10c3c189bf5 baseline-x86` -> creates the new test node

Note: I have added the following line to the `kernelci-api` volumes in the api docker-compose file so it uses my local copy of kernelci-core:
```
      - '/path/to/kernelci-core/kernelci:/home/kernelci/.local/lib/python3.10/site-packages/kernelci'
```

a-wai:
ooc, what part of the instructions were you missing? so I make sure to emphasize it when writing the doc ;)

nuclearcat:
Tested on staging, thanks!

PR #2389, opened 2024-02-16
rootfs-images.yaml: Update bookworm-kselftest to 20240221

nuclearcat:


nuclearcat:
Verified as required in https://github.com/kernelci/kernelci-core/pull/2392 - /usr/bin/tappy present for example, tests results on staging is ok

PR #2390, opened 2024-02-16
kbuild: look for legacy config in current folder first

a-wai:
During development, executing `kci` and kernel build scripts from the root of the `kernelci-core` source tree is common practice. In such cases, instead of mandating the developer to create/update those under `/etc/kernelci`, we should look for the config file under `config/` first.


nuclearcat:
Thanks, LGTM, also battle tested on ChromiumOS PRs

PR #2391, opened 2024-02-16
kbuild: improve cros defconfig handling

a-wai:
Currently, `cros://` fragments used as a defconfig are special-cased so the defconfig node data is updated with the actual branch. However, this is done after `config_full` is updated, resulting in the latter still containing the `{kver}` placeholder.

This PR reworks this logic so we update all variables properly, and adds the ability to also use the `{arch}` placeholder in fragment names.

### DIFF
--- a/kernelci/kbuild.py
+++ b/kernelci/kbuild.py
@@ -313,10 +313,11 @@ class KBuild():
     def _getcrosfragment(self, fragment):
         """ Get ChromeOS specific configuration fragments """
         # ChromeOS fragment name can be f-strings with placeholders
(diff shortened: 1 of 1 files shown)


a-wai:
Added a fallback to using the latest LTS chromeos branch when building a newer upstream kernel, otherwise bulid will fail due to e.g. the `chromeos-6.8` branch not existing in https://chromium.googlesource.com/chromiumos/third_party/kernel/+refs

nuclearcat:
Thanks! Tested on staging

Issue #2346, opened 2024-02-05
Add branch queue/6.7 and linux-6.7.y from stable-rc tree #2254


crazoes:
Which Git branch do you want to add?
⇨ Git repo URL:
https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux-stable-rc.git/

⇨ Git branch name:
queue/6.7 and linux-6.7.y

How much build coverage do you need for your branch?
As other stable-rc branches.

nuclearcat:
Can you please suggest which stable trees we need to track? 
Because i think we have lot of trees listed, which are long time EOL and maybe we are missing some.

crazoes:
@nuclearcat we need to track 4.19, 5.4, 5.10, 5.15, 6.1, 6.6 and 6.7 releases.
There are two branches `queue` and `linux-x.x.y` like mentioned in the description. Both these branches needs to be tracked for stable-rc releases.

PR #2395, opened 2024-02-19
Add stable 6.7 trees

nuclearcat:
As per https://github.com/kernelci/kernelci-core/issues/2346 Adding two more trees to monitor


nuclearcat:
Merging, as it is trivial change

PR #2399, opened 2024-02-20
kernelci/api/models_base: dataclass to store DB indexes

JenySadadia:
Implement a dataclass `Index` to store model field names and constraints to create indexes. Also, write a method to get indexes to be implemented by child class.


hardboprobot (approved):

JenySadadia:
I'll merge it to enable API to use these changes.

PR #2400, opened 2024-02-20
rootfs: Don't ship .deb files we installed

broonie:
When we install Debian packages with apt these will be placed in
/var/cache/apt/archives, there is no need for them at runtime so we can
just remove them and reduce the size of the images.

### DIFF
--- a/config/rootfs/debos/rootfs.yaml
+++ b/config/rootfs/debos/rootfs.yaml
@@ -122,6 +122,11 @@ actions:
     command: ln -s /usr/bin/systemd /init
 
   - action: run
+    description: Clean installed package files
+    chroot: true
+    command: apt-get clean
+
+  - action: run
     description: Create full archive
     chroot: false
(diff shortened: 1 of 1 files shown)


config/rootfs/debos/scripts/strip.sh
```diff
@@ -44,6 +44,9 @@ apt-get autoremove --yes || true
 # Dropping logs
 rm -rf /var/log/*
 
+# Remove .deb files we installed
+rm -f /var/cache/apt/archives/*.deb
```
a-wai, line 48:
```suggestion
apt-get clean
```

There is actually an apt invocation for this

config/rootfs/debos/scripts/strip.sh
```diff
@@ -44,6 +44,9 @@ apt-get autoremove --yes || true
 # Dropping logs
 rm -rf /var/log/*
 
+# Remove .deb files we installed
+apt-get clean /var/cache/apt/archives/*.deb
```
a-wai, line 48:
```suggestion
apt-get clean
```

This command doesn't take any additional argument

nuclearcat:
Verified on recent builds, no deb stored anymore in that directory

PR #2402, opened 2024-02-21
debos: Don't remove extra_packages

broonie:
The strip.sh script is passed a list of packages to remove.  For very
unclear reasons the list that is passed by debos is the extra_packages
list but this is a list of packages we want in the shipped image, there
is a separate extra_packages_remove which is intended for removal which
we already handle with a separate debos step.  Just remove the argument
and it's handling from strip.sh, ensuring that we don't remove
extra_packages.

This hasn't been apparent because strip.sh only takes effect for ext4
images, we don't use those images for many platforms.  That is a
separate issue which should also be addressed.

### DIFF
--- a/config/rootfs/debos/rootfs.yaml
+++ b/config/rootfs/debos/rootfs.yaml
@@ -168,7 +168,7 @@ actions:
 
   - action: run
     chroot: true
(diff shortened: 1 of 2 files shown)


nuclearcat:
Tested and built rootfs with this changes

PR #2403, opened 2024-02-21
Disable pycodestyle warning for long lines and cleanup noqa markers

hardboprobot:
```
We're littering the code with #noqa markers just to make pycodstyle
happy, particularly wrt the length of some lines. While we try to keep
code lines reasonably short (<80 chars), some other lines are harder to
read or more cumbersome to maintain if they're forced to be
splitted (for instance, long strings).

Disable this check (E501). Please, keep being mindful of the length of
code lines, try to stay below 80 chars except in reasonable
situations. Use good judgement.
```

### DIFF
--- a/kernelci/api/helper.py
+++ b/kernelci/api/helper.py
@@ -119,7 +119,7 @@ class APIHelper:
         # Test-specific fields inherited from parent node (kbuild or
         # test) if available
         if job_config.kind == 'test':
-            job_node['data']['kernel_type'] = input_node['data'].get('kernel_type')  # noqa
+            job_node['data']['kernel_type'] = input_node['data'].get('kernel_type')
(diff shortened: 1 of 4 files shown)


a-wai:
I'm not sure disabling it entirely is the right thing to do, I'd rather just increase the max line length (100 chars seem reasonable) and keep it as a safeguard.

This would leave a few occurrences of `#noqa` (the cros URL for example) but those seem legit to me, and rare enough to not really bother us.

a-wai (approved):
Thanks, LGTM

nuclearcat (approved):
Excellent, thank you, this is very useful and will make workflow easier.

nuclearcat:
If no objections, i will merge it within few hours.

PR #2404, opened 2024-02-21
kernelci.api.models: create_regression method for Regression model

hardboprobot:
Add a Regression object builder that will take care of formatting and returning Regression object for subsequent use, for instance, for submitting it to the API.

This is meant to be used by client code when detecting and submitting regressions given two test or build nodes that mark the breaking point.


nuclearcat:
Passing staging, merging

PR #2405, opened 2024-02-21
config: runtime: tests: tast: use "for" loop to enumerate tests

a-wai:
The legacy implementation relied on Tast tests being listed as a single string in the job config, then uses the `wordwrap()` jinja2 function to output one test per line in the job definition.

This is needlessly complex, storing the list of tests as a string array and using a classic `for` loop makes things more straightforward and understandable to external contributors.


nuclearcat:
Tested on staging

PR #2406, opened 2024-02-21
kernelci.api: rework and enhance the api _post() method

hardboprobot:
Simplify the base case, document the method and make it capable of handling both string and dict payloads (automatically detected).

The default and most frequent use case will still be to pass a dict to it, which will be sent to the server as a json object or as an encoded string, depending on the <json_data> parameter.

The additional capability of accepting a string fits the use cases where we might have an already processed json definition to POST instead of a python dict.


nuclearcat:
Works on staging, thanks, merging

PR #2407, opened 2024-02-22
docker: Update all images to bookworm

nuclearcat:
It's time to upgrade compiler images at least,
as gcc-10 gives really bad false positives.


config/docker/data/kmod_kci_bookworm.patch
```diff
@@ -0,0 +1,35 @@
+diff -Naur kmod-30+20221128/debian/changelog kmod-30+20221128-kernelci/debian/changelog
```
a-wai, line 1: Has this patch been proposed to Debian? It won't help for bookworm, but that would help once we switch to the next stable.

nuclearcat: This is specific to our needs in ChromiumOS, i dont know why, but they use zlib... (while they invented zstd for example). Not sure it can be upstreamed.

JenySadadia (commented):
I believe this will also resolve python version conflicts. Is that right? @nuclearcat

nuclearcat:
I am merging asap(and a bit in hurry) to unblock other issues, please feel free to open new PR to improve this particular PR.

PR #2409, opened 2024-02-22
kernelci.api: remove manual json data handling in requests

hardboprobot:
Fixes 6e8d6c5b8dec943431c1f9d305cba1fd11d0f34d, see also a6df5594ca6508869fd3438310da4b902aa0c785.

The requests library can automatically encode data as json and set the proper Content-Type header. Remove the default header hardcoding in the requests and use the `json` option of the requests calls.

In the `_post` method, this is selectable and it accepts json data (by default) or a www-form-urlencoded string payload.


nuclearcat:
LGTM

PR #2410, opened 2024-02-23
Build `kernelci` package with developer requirements

JenySadadia:
Before `pyproject.toml` support, when `setup.py` was being used for package building, developer requirements were also installed in the package. Keep it the same while using `pyproject.toml` by adding `[dev]` option to the installation command.

### DIFF
--- a/config/docker/fragment/kernelci.jinja2
+++ b/config/docker/fragment/kernelci.jinja2
@@ -12,7 +12,7 @@ RUN git clone --depth=1 $core_url /tmp/kernelci-core
 WORKDIR /tmp/kernelci-core
 RUN git fetch origin $core_rev
 RUN git checkout FETCH_HEAD
-RUN python3 -m pip install .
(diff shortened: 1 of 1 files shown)


config/docker/fragment/kernelci.jinja2
```diff
@@ -12,7 +12,7 @@ RUN git clone --depth=1 $core_url /tmp/kernelci-core
 WORKDIR /tmp/kernelci-core
 RUN git fetch origin $core_rev
 RUN git checkout FETCH_HEAD
-RUN python3 -m pip install .
+RUN python3 -m pip install .[dev]
```
a-wai, line 15:
```suggestion
RUN python3 -m pip install '.[dev]'
```

I'd rather be on the safe side and ensure the brackets are not interpreted by the shell.

JenySadadia: Thanks. Fixed the command.

a-wai (approved):

PR #2411, opened 2024-02-23
Add hostname mapping for docker gateway

nuclearcat:
As in docker-compose.yaml we need to add mapping
in docker runtime, so we can access local instance of API.


nuclearcat:
Tested by kci-easy

PR #2413, opened 2024-02-23
Implement initial filtering capabilities

### LINKED ISSUE #2121: Event filtering in pipeline
Related to: https://github.com/kernelci/kernelci-core/discussions/1907
At current moment we have very basic event filtering which we need to extend.

I reviewed existing solutions and found following:

* JMESPath (very close to what we need but IMHO massive overkill)
* pandas
* dpath
* pyql

Most of them designed to do complex operations with data, not filter matching. There is few more libraries that might be used as filters, but this is not their primary purpose, and they might change "second-class" features at any time.

What we also can do, build our own solution around lexx/yacc or pyparsing.

I propose to add following filters (PoC examples):

- "(node="checkout")AND((branch~=/mainline/)OR(branch~=/stable/))AND(version>6)"
- "(node=checkout)AND(branch="staging-mainline")

Where we can unroll expression to:
```
['data'] must be equal to 'checkout'
AND
(('branch' must match regex '/mainline/') OR ('branch' must match regex '/stable/'))
AND
('version' must be greater than '6')
```
So we should support following operators:
* AND (or &)
* OR (or |)
* NOT (or !)
* =
* ~=
* >
* <
* >=
* <=

What is next? We need to decide if it cover all our use cases, and if it is good enough - we decide how to implement it.
Also i dont rule out using some existing library, if someone know one that suits our needs.

a-wai:
As all devices might not be equally supported by all kernel trees and versions, this PR implements filtering based on rules defined per job, platform and/or runtime. This is achieved by setting rules affecting any field present in the node structure, with the ability to provide both allowed and forbidden values.

If one of those items include at least one such rule, checks will be conducted during node creation, which will be aborted if one of the rules isn't satisfied.

For example, a job can be configured to run only on arm64 devices, with a kernel version >= 6.6, built with any defconfig except `allnoconfig`, and using the `kselftest` fragment but not the `arm64-chromebook` one, by adding the following to the job definition:

```yaml
  rules:
    min_version:
      version: 6
      patchlevel: 6
    arch:
      - 'arm64'
    defconfig:
      - '!allnoconfig'
    fragments:
      - 'kselftest'
      - '!arm64-chromebook'
```

Requires https://github.com/kernelci/kernelci-pipeline/pull/443 to avoid breaking the pipeline.

Note: the term "rules" has been preferred to "filters" in order to avoid confusion: the latter is used in the legacy config using different syntax and semantics, so providing a different implementation with the same name in the new system would likely lead to unexpected weird issues.

Fixes #2121

### DIFF
--- a/kernelci/api/helper.py
+++ b/kernelci/api/helper.py
@@ -123,6 +123,137 @@ class APIHelper:
                    for obj in [node, event]):
                 return node
 
+    def _find_container(self, field, node):
+        """
+        Return the first found dict containing a field of a given name
+        """
+        for item in node:
+            if isinstance(node[item], dict):
+                base_object = self._find_container(field, node[item])
+                if base_object:
+                    return base_object
+            elif field == item:
+                return node
+        return None
+
+    # pylint: disable=too-many-arguments
+    def _is_allowed(self, rules, key, node, parent, name):
+        """
+        Check whether the value of a specific node attribute matches
+        a filtering rule. As the specified attribute might not be present
+        in the node being created, we fall back to checking the value in its
+        parent node in such cases.
+
+        Returns True if the rule allows the current value, False otherwise.
+        """
+
+        # Find the node (or parent node) attribute corresponding to the
+        # rule we're applying
+        base = self._find_container(key, node)
+        if not base:
+            base = self._find_container(key, parent)
+        if not base:
+            return True
+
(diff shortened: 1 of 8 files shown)


a-wai:
Ready to review after a few iterations, I updated the PR description to match the current state and design.

nuclearcat:
Doesn't break staging, didnt really functionally tested it, but should be ok

PR #2416, opened 2024-02-26
Add dtbs check test

nuclearcat:
Add dtbs check as build test and implement properly sub-nodes results.

### DIFF
--- a/kernelci/kbuild.py
+++ b/kernelci/kbuild.py
@@ -9,11 +9,19 @@ Roadmap:
 - correct build status - fail/pass/incomplete
 - Code deuglification
 - Reuse API class/functions and other code from KernelCI
-- Add support for multiple CPU architectures (arm64, arm, mips, etc)
 - Add support for multiple compilers (clang)
 - Implement 3-stage build process
   (fetch-prepare, build, package) with 3 containers
   (kernelci, compiler-specific, kernelci)
+
+Available kbuild parameters:
+- arch: architecture
+- compiler: compiler
+- defconfig: defconfig
+- fragments: list of config fragments
+- cross_compile: cross compile prefix
+- cross_compile_compat: cross compile compat prefix
+- dtbs_check: run "make dtbs_check" ONLY, it is actually a separate test
 """
 
 import os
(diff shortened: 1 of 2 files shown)


kernelci/kbuild.py
```diff
@@ -533,6 +553,15 @@ def _build_kselftest(self):
         self.addcmd("make -j$(nproc) kselftest-gen_tar" + " 1> " +
                     self._af_dir + "/build_kselftest.log" +
                     " 2> " + self._af_dir + "/build_kselftest_stderr.log")
+
+    def _build_dtbs_check(self):
+        """ Check if dtbs are present """
+        self.startjob("build_dtbs_check")
+        self.addcmd("cd " + self._srcdir)
+        self.addcmd("make -j$(nproc) dtbs_check" + " 1> " +
```
a-wai, line 561: This needs a `.config` to be present. IMHO this should be the base defconfig for the target architecture, as defined in the `build_variants` section of [pipeline.yaml](https://github.com/kernelci/kernelci-pipeline/blob/main/config/pipeline.yaml#L444), e.g. `multi_v7_defconfig` for armv7.

a-wai (changes requested):

kernelci/kbuild.py
nuclearcat: You are right, i adjusted code a bit

kernelci/kbuild.py
```diff
@@ -776,17 +822,17 @@ def submit(self, retcode, dry_run=False):
...
         results['node']['data']['compiler'] = self._compiler
         results['node']['data']['defconfig'] = self._defconfig
         results['node']['data']['fragments'] = self._fragments
         results['node']['data']['config_full'] = self._config_full
-        node.update(results['node'])
-        print(json.dumps(node, indent=2))
-        api.node.update(node)
-        print(f"[_submit] Results submitted to API, node: {node['id']}")
+        api_helper = kernelci.api.helper.APIHelper(api)
+        api_helper.submit_results(results, node)
```
hardboprobot, line 836: This results in a different kbuild node contents than what we had before this change. `submit_results` doesn't update all the node fields with the contents `results`, it only updates some of the fields, so we're losing `config_full` for instance.

There were some discussions about the can of worms that's `submit_results()` and how it needs to be fixed and generalized and this is an example of the problem underneath.

a-wai (approved):

kernelci/kbuild.py
hardboprobot: Results are now looking good, no data loss after the results are submitted

hardboprobot (approved):

nuclearcat:
Thanks all, merging

PR #2419, opened 2024-02-28
kernelci.api.helper: make node hierarchies inherit parameters from parent

hardboprobot:
When submitting test results via `submit_results()`, a tree of nodes is created hanging from a root node. These nodes aren't created in the same way as nodes built with `create_job_node`, where certain node types will inherit specific parameters from the parent node. Fix submit_results() so that the child nodes will inherit additional data parameters if they're defined in the parent node.

### DIFF
--- a/kernelci/api/helper.py
+++ b/kernelci/api/helper.py
@@ -231,7 +231,13 @@ class APIHelper:
         parent = self.api.node.get(root['parent'])
         base = {
             'data': {
-                'kernel_revision': root['data']['kernel_revision'],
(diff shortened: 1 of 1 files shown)


a-wai (commented):
Looks good, maybe it would be worth making a function of it and using this function in both places it's needed? So we have a single place to modify next time a similar change is needed.

JenySadadia (approved):
LGTM.

a-wai (approved):
> I'd rather do the refactoring later as a follow-up PR

Alright then, feel free to merge

PR #2421, opened 2024-02-29
runtime/docker: Add mechanism for local images

nuclearcat:
Try to pull image from hub, but if it fails, try to use local image. This is useful for local testing purposes. Also introduce NO_DOCKER_PULL environment variable.


nuclearcat:
Tested on staging, merging ASAP as i need it to continue testing local development environment

PR #2425, opened 2024-03-01
kernelci.api.helper: disable json parsing in error propagation

hardboprobot:
Apparently, not all HTTPErrors raised by requests to the server are formatted as json. Remove the json parsing to propagate the raw response text to the caller.

### DIFF
--- a/kernelci/api/helper.py
+++ b/kernelci/api/helper.py
@@ -155,7 +155,7 @@ class APIHelper:
         try:
             return self._api.node.add(job_node)
         except requests.exceptions.HTTPError as error:
-            raise RuntimeError(json.loads(error.response.text)) from error
+            raise RuntimeError(json.loads(error.response.content)) from error
 
     def submit_regression(self, regression):
         """Post a regression object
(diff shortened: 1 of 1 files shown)


kernelci/api/helper.py
```diff
@@ -155,7 +155,7 @@ def create_job_node(self, job_config, input_node,
         try:
             return self._api.node.add(job_node)
         except requests.exceptions.HTTPError as error:
-            raise RuntimeError(json.loads(error.response.text)) from error
+            raise RuntimeError(error.response.text) from error
```
JenySadadia, line 158: API uses `content` to store error messages. See https://github.com/kernelci/kernelci-api/blob/main/api/main.py#L95
```suggestion
            raise RuntimeError(json.loads(error.response.content)) from error
```
See that's how the pipeline parses error message https://github.com/kernelci/kernelci-pipeline/blob/main/src/scheduler.py#L98

hardboprobot: Ok let me apply that to all cases in the file then

hardboprobot: Done, let's see if it makes any difference in staging

nuclearcat:
Thanks, seems doesnt break staging, we will see how it parses errors in future :)

PR #2426, opened 2024-03-01
kernelci.api.helper: don't inherit parent node artifacts

hardboprobot:
When creating a new job node, we don't want it to automatically inherit the parent node artifacts. This leads to nodes carrying irrelevant information in most cases, such as test nodes including the build artifacts of the parent kbuild node.

@nuclearcat let's see how this works in staging first and then we can decide whether it's worth it or not.

### DIFF
--- a/kernelci/api/helper.py
+++ b/kernelci/api/helper.py
@@ -263,7 +263,7 @@ class APIHelper:
             'name': job_config.name,
             'path': input_node['path'] + [job_config.name],
             'group': job_config.name,
-            'artifacts': input_node['artifacts'],
+            'artifacts': {},
             'data': {
                 'kernel_revision': input_node['data']['kernel_revision'],
             },


kernelci/api/helper.py
```diff
@@ -132,11 +132,14 @@ def create_job_node(self, job_config, input_node,
             'name': job_config.name,
             'path': input_node['path'] + [job_config.name],
             'group': job_config.name,
-            'artifacts': input_node['artifacts'],
+            'artifacts': {},
             'data': {
                 'kernel_revision': input_node['data']['kernel_revision'],
             },
         }
+        # Specific artifacts we want to keep from the parent node
```
a-wai, line 140: Maybe, but why? Is it always needed? Or just for specific jobs? A more detailed comment would be really helpful when we'll try to reduce the tech debt.

hardboprobot: Hard to document, really, considering that the procedure of creating a job node should be more clearly defined. So, initially I thought we didn't want or need to carry over artifacts from the parent node, but then it looks like [other parts of the code](https://github.com/kernelci/kernelci-core/blob/4633cd275fb922020a7ea4e07d1a4cd8e50ef22d/config/runtime/base/python.jinja2#L136) expect some contents there.

Again, the problem is that none of this is consistent, there are a lot of individual cases and we're keeping many hacks in place to get something done quick at the expense of leaving many potential pitfalls that we'll need to find fix later.

I'll explain some of this in the comment

a-wai: Thanks, the more detailed comment is already a good start. This `python.jinja2` template feels wrong in many ways tbh, I'll try to improve it a bit.

a-wai: Tried to address this with https://github.com/kernelci/kernelci-core/pull/2428 so you can hopefully avoid single-casing this artifact ;)

hardboprobot: Thanks, I left some comments there. I totally agree with the template solution being a bad idea for this. The same thing can probably be done in a simpler, more straightforward way, and I feel like we should address the reimplementation of the job generation code some day when we find the time to do it.

a-wai: > I feel like we should address the reimplementation of the job generation code some day when we find the time to do it.

Agreed, sounds like a good task for post-M1 tech debt reduction ;)

a-wai (approved):

PR #2428, opened 2024-03-01
Fetch `tarball` artifact from parent if not present in node

a-wai:
This hopes to ease the work of ensuring nodes only reference the artifacts they create.

Requires https://github.com/kernelci/kernelci-pipeline/pull/453 on the pipeline side.

### DIFF
--- a/config/runtime/base/python.jinja2
+++ b/config/runtime/base/python.jinja2
@@ -59,7 +59,9 @@ WORKSPACE = '/tmp/kci'
 {% block python_base_job -%}
 class BaseJob:
 
-    def __init__(self, workspace):
+    def __init__(self, api_config, node, workspace):
+        self._api = self._get_api(api_config)
+        self._node = node
         self._workspace = workspace
 
     def _get_api(self, api_config_yaml):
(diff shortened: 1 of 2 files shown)


JenySadadia:
Adding `staging-skip` due to below errors on staging:
```
today at 11:48:09Traceback (most recent call last):
today at 11:48:09  File "/home/kernelci/data/output/tmp63nc5a97/kver", line 187, in main
today at 11:48:09    results = job.run()
today at 11:48:09              ^^^^^^^^^
today at 11:48:09  File "/home/kernelci/data/output/tmp63nc5a97/kver", line 102, in run
today at 11:48:09    if 'debug' in self._node and 'dry_run' in self._node['debug']:
today at 11:48:09                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
today at 11:48:09TypeError: argument of type 'NoneType' is not iterable
today at 11:48:09
```

config/runtime/base/python.jinja2
```diff
@@ -112,7 +112,12 @@ class BaseJob:
             return self.dry_run(self._node['debug']['result'])
 
         print("Getting kernel source tree...")
-        tarball_url = node['artifacts']['tarball']
+        tarball_url = None
+        if 'artifacts' in node and 'tarball' in node['artifacts']:
+            tarball_url = node['artifacts']['tarball']
+        else:
+            parent = self._api.node.get(node['parent'])
+            tarball_url = parent['artifacts']['tarball']
```
hardboprobot, line 120: Is there any guarantee that the parent will have a 'tarball' artifact, though?

hardboprobot (commented):
Thanks for taking a stab at this, I agree all of this needs to be simplified and refactored. Actually, I hope we can get rid of this template scheme someday. I think there are a few pitfalls in the changes that need to be addressed if you find the time to do so, I left a hint in another comment.
Also, I think it'll be hard to reshape this into something that works for every case, seeing how part of it is, since the beginning, tightly coupled to certain use cases and to templates in other repos. But I appreciate the effort and, if it works and it makes the code simpler, I'm all for it once we can see it working in staging.

config/runtime/base/python.jinja2
a-wai: Not really, it only works so far as this template is used only by children of `checkout` nodes. I was being too lazy to implement proper research of this artifact (partly due to the fact that I'd like to nuke this template entirely in the future), but this is solved now: we go back following the whole node hierarchy until we find one with this artifact.

a-wai:
> Adding `staging-skip` due to below errors on staging:

Damn, ok, should be fixed now (@hardboprobot thanks for the explanation)

nuclearcat:
Tested on staging

Issue #2429, opened 2024-03-04
k8s DNS resolution issues


nuclearcat:
As examples:
```
Handling uncaught exception: HTTPSConnectionPool(host='staging.kernelci.org', port=9000): Max retries exceeded with url: /latest/node/65e58e209d5adb6aceeae510 (Caused by NameResolutionError("<urllib3.connection.HTTPSConnection object at 0x7f99dabf8a00>: Failed to resolve 'staging.kernelci.org' ([Errno -3] Temporary failure in name resolution)"))
```
```
kubectl logs -f kci-65e58e209d5adb6aceeae510-kbuild-gcc-10-i386-30y7b4fa-t7r64
+ touch /tmp/kci/artifacts/build.log
+ tailpid=10
+ exec
+ tail -f /tmp/kci/artifacts/build.log
+ cd /tmp/kci
+ wget -c -t 10 'https://kciapistagingstorage1.file.core.windows.net/staging/linux-kernelci-staging-stable-staging-stable-20240304.5.tar.gz?sv=2022-11-02&ss=f&srt=sco&sp=r&se=2024-10-17T19:19:12Z&st=2023-10-17T11:19:12Z&spr=https&sig=sLmFlvZHXRrZsSGubsDUIvTiv%2BtzgDq6vALfkrtWnv8%3D' -O linux.tgz
--2024-03-04 09:04:38--  https://kciapistagingstorage1.file.core.windows.net/staging/linux-kernelci-staging-stable-staging-stable-20240304.5.tar.gz?sv=2022-11-02&ss=f&srt=sco&sp=r&se=2024-10-17T19:19:12Z&st=2023-10-17T11:19:12Z&spr=https&sig=sLmFlvZHXRrZsSGubsDUIvTiv%2BtzgDq6vALfkrtWnv8%3D
Resolving kciapistagingstorage1.file.core.windows.net (kciapistagingstorage1.file.core.windows.net)... failed: Temporary failure in name resolution.
wget: unable to resolve host address ‘kciapistagingstorage1.file.core.windows.net’
[_set_api_yaml] CONTENT: 
```
It seems kubernetes nodes on startup might have various DNS issues and this keep introducing false positive build failures that we need to handle better.

PR #2430, opened 2024-03-04
python.jinja2: Add DNS test and retry

nuclearcat:
As reported in issue, we need to make sure DNS working properly in container, before starting real job.

Fixes: https://github.com/kernelci/kernelci-core/issues/2429


nuclearcat:
Tested on staging

PR #2432, opened 2024-03-04
Define config_full in models and propagate it to test and regression nodes

hardboprobot:
Currently, tests and regressions only contain the defconfig info from the kbuild but not the config fragment info. The easiest way to do it is to pass the complete 'config_full' field downwards.

Remember that in KernelCI, knowing the "defconfig" of a kernel build may be only a partial picture of the whole config, since the complete config applied to the kernel may include any number of additional KernelCI-defined config fragments.


a-wai:
That makes sense, but is propagating `defconfig` still needed then? It can likely be inferred from `config_full` at this point (e.g. `config_full.split('+')[0]`)

a-wai (approved):

nuclearcat:
Thanks, merging

PR #2427, opened 2024-03-01
Fix "incomplete" lava job result

JenySadadia:
Fixes https://github.com/kernelci/kernelci-pipeline/issues/451

PR #2434, opened 2024-03-04
Set failed lava job metdata

JenySadadia:
Created on top of https://github.com/kernelci/kernelci-core/pull/2427

Store error code and error message for failed lava jobs.


kernelci/api/models.py
```diff
@@ -62,6 +62,17 @@ class ErrorCodes(str, enum.Enum):
     SUBMIT_ERROR = 'submit_error'
     # Node reached timeout and timeout service forced it to be done
     NODE_TIMEOUT = 'node_timeout'
+    # Lava job error codes copied from source: lava/lava_common/exceptions.py
```
hardboprobot, line 65: I was going to say that having these in a separate class isn't necessary but you just changed it as I was reviewing it ;)

hardboprobot (approved):
If it works fine on staging, the changes look good to me

kernelci/api/models.py
JenySadadia: Python3 doesn't allow it anyway for `Enum` classes.

hardboprobot: What do you mean? It's [supposed to be supported](https://docs.python.org/3/howto/enum.html) (see the first example in that page). Actually, the `ErrorCodes` class in this same file inherits from it.

JenySadadia: I was trying to do something like this and it's not permitted.
See https://docs.python.org/3/howto/enum.html#restricted-enum-subclassing
```
class LavaErrors(str, enum.Enum):
    ...
class ErrorCodes(LavaErrors):
    ...
```

hardboprobot: That makes sense, conceptually, and is basically the reason why I was going to comment that adding another class for this didn't look quite right.

JenySadadia:
Tested OK on staging:
```
$ curl https://staging.kernelci.org:9000/latest/node/65e6ebd1aecb33125f9c679b
{"id":"65e6ebd1aecb33125f9c679b","kind":"test","name":"tast-mm-misc-x86-stoneyridge","path":["checkout","kbuild-gcc-10-x86-chromeos-stoneyridge","tast-mm-misc-x86-stoneyridge"],"group":"tast-mm-misc-x86-stoneyridge","parent":"65e6e8baaecb33125f9c666d","state":"done","result":"fail","artifacts":{"lava_log":"https://kciapistagingstorage1.file.core.windows.net/staging/tast-mm-misc-x86-stoneyridge-65e6ebd1aecb33125f9c679b/log.txt?sv=2022-11-02&ss=f&srt=sco&sp=r&se=2024-10-17T19:19:12Z&st=2023-10-17T11:19:12Z&spr=https&sig=sLmFlvZHXRrZsSGubsDUIvTiv%2BtzgDq6vALfkrtWnv8%3D"},"data":{"error_code":"Infrastructure","error_msg":"matched a bootloader error message: '(Bulk read error(.*)){3}' (9)","test_source":null,"test_revision":null,"platform":"acer-R721T-grunt","runtime":"lava-collabora","job_id":"12945396","job_context":null,"kernel_revision":{"tree":"kernelci","url":"https://github.com/kernelci/linux.git","branch":"staging-stable","commit":"091f516073d7a569ceee080156a7755928cba55c","describe":"staging-stable-20240305.6","version":{"version":6,"patchlevel":1,"sublevel":80,"extra":"-1-g091f516073d7a","name":null},"patchset":null},"arch":"x86_64","defconfig":"cros://chromeos-6.1/x86_64/chromeos-amd-stoneyridge.flavour.config","config_full":"cros://chromeos-6.1/x86_64/chromeos-amd-stoneyridge.flavour.config+x86-board+CONFIG_MODULE_COMPRESS=n","compiler":"gcc-10","kernel_type":"bzimage"},"debug":null,"created":"2024-03-05T09:54:25.278000","updated":"2024-03-05T09:56:09.209000","timeout":"2024-03-05T15:54:25.278000","holdoff":null,"owner":"staging.kernelci.org","user_groups":[]}
```

PR #2436, opened 2024-03-05
runtime: prevent crash on missing metadata

nuclearcat:
In some cases, for example incorrectly set dtb for x86, we might crash cause metadata file wont exist for such platforms.


nuclearcat:
Tested on staging, just prevent crashes in case of minor misconfiguration

PR #2437, opened 2024-03-05
cli: job: import artifacts from parent node before generating job

a-wai:
In most situations, jobs need artifacts generated by their parent. As those are no longer automatically propagated, we need to import them when generating the job.


nuclearcat:
Thanks, merging

PR #2438, opened 2024-03-05
Format config params using objet attributes

a-wai:
Our config files currently have to specify lots of nearly-identical parameters, especially for chromebooks. Those parameters usually depend on the device/job target architecture, name or family.

In order to make things more flexible and reduce config verbosity, this PR implements string formatting for job/platform params, substituting placeholders with the values of the corresponding object attributes.

A special `krev` placeholder is also handled: it doesn't match any object attribute, but it is constructed from the code as the kernel revision in the form `major.minor`, extracted from node data.

Example:
```yaml
  chromebook-device: &chromebook-device
    arch: arm64
    mach: mediatek
    params:
      myconfig: 'config-{arch}-{mach}-{base_name}.config'

  device-1: *chromebook-device

  device-2:
    <<: *chromebook-device
    base_name: dev2
    mach: qcom
```

This config will then be equivalent to:
```yaml
  device-1: *chromebook-device
    arch: arm64
    base_name: device-1
    mach: mediatek
    params:
      myconfig: 'config-arm64-mediatek-device-1.config'

  device-2:
    arch: arm64
    base_name: dev2
    mach: qcom
    params:
      myconfig: 'config-arm64-qcom-dev2.config'
```

### DIFF
--- a/config/runtime/base/lava.jinja2
+++ b/config/runtime/base/lava.jinja2
@@ -17,7 +17,7 @@
 
 job_name: "[{{ instanceid|default('KernelCI AKS')}}] {{ node.id }} {{ node.name }} {{ node.data.kernel_revision.describe }}"
 
-device_type: {{ platform_config.base_name }}
+device_type: {{ platform_config.name }}
 
 visibility: public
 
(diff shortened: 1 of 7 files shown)


nuclearcat:
Looking to pipeline it seems working, but a bit hard to check in details every corner case: https://staging.kernelci.org:9000/viewer?node_id=65eac74577d9d88b36edbb6b

a-wai:
@nuclearcat can you please merge this? Otherwise kernelci-pipeline will soon break due to https://github.com/kernelci/kernelci-pipeline/pull/459 depending on it ;)

PR #2439, opened 2024-03-05
Apply rules to scheduler entries as well

a-wai:
The filtering capabilities implemented in #2413 only apply to job and platform entries. It could, however, be useful to add such capabilities to scheduler entries as well.

This PR ensures the generic mechanisms for doing so are in place, but the actual filtering will be executed within [kernelci-pipeline](https://github.com/kernelci/kernelci-pipeline)


nuclearcat:
Amazingly flexible implementation, thanks! Looks like tested on staging...

PR #2442, opened 2024-03-06
kernelci.api.models: enable multiple operator queries

JenySadadia:
To support multiple operators for a query parameter, convert fields with operator and timestamp fields to a dictionary instead of a tuple. For example, `{field__operator1: value, field__operator2: value}` will be converted to `{field: {operator1: value, operator2: value}}`.


hardboprobot (approved):

nuclearcat:
Amazing news Jeny, thanks for quick fix and certain portion of magic with it :) Merging

PR #2446, opened 2024-03-08
rootfs: bullseye - migrate to bookworm

nuclearcat:
It is a time to migrate all rootfs images to bookworm, as many test images getting broken due missing or outdated packages.


nuclearcat:
Just to keep documented, following issues were found and fixed in the last week:
1) Some of packages installed by pip converted to apt python3-name packages
2) Remaining needed --break-system-packages flag
3) debos script  for installing extra firmware needed update due rdfind package requirements
4) Usual package version bumping (mostly perl 5.32 to perl 5.36, and few others python related libraries and etc)
5) Used newer libhugetlbfs git snapshot (why it is fixed to specific git commit? we need to make list of such software and update them from time to time)

PR #2447, opened 2024-03-08
kbuild: Remove custom name mapping, making more consistent

nuclearcat:
As we are automatically replacing . by _, by default most of log files are renamed properly. Let's make things simpler and let automagical replacement do the job, and same time it will make names more consistent.


nuclearcat:
Tested on staging

PR #2449, opened 2024-03-11
build-configs-stable.yaml: Some fixes and an update for v6.8

broonie:
- build-config-stable.yaml: Add coverage for v6.7 stables
- build-configs-stable.yaml: Correct indentation for v6.7 queue entry
- build-configs-stable.yaml: Add v6.8 stable coverage


nuclearcat:
Tested on staging, https://staging.kernelci.org/job/

PR #2450, opened 2024-03-11
config: runtime: boot: Use latest kselftest rootfs to avoid tap issue

nfraprado:
Update the kselftest rootfs URL to the latest one which includes tappy, and is thus able to parse the kselftest results without throwing an error:

ModuleNotFoundError: No module named 'tap'

### DIFF
--- a/config/runtime/boot/u-boot.jinja2
+++ b/config/runtime/boot/u-boot.jinja2
@@ -11,10 +11,10 @@
 {%- endif %}
 {%- if boot_commands == 'nfs' %}
     nfsrootfs:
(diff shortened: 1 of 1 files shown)


nuclearcat:
I am in process of rebuilding all rootfs images, including tappy and migratin also to bookworm. I will ping when you might need to update your PR

pawiecz (approved):
LGTM

It could be useful to bump those URLs when fresh rootfses are published (like @nuclearcat [mentioned](https://github.com/kernelci/kernelci-core/pull/2450#issuecomment-1989770343))

Thanks for [verifying this in Collabora's LAVA](https://lava.collabora.dev/scheduler/job/13020048)

nuclearcat:
If verified - better i merge now, and we update again when i complete new builds

PR #2451, opened 2024-03-12
kernelci.api.models: fix event payload data type

JenySadadia:
Fixes https://github.com/kernelci/kernelci-api/issues/451

When using `Any` for event payload `data` field allows all the values including `None`. Events with `None` payload will break pipeline services as per current design. Update the field to be mandatory and dictionary type to make it compatible with pipeline service usage.

### DIFF
--- a/kernelci/api/models.py
+++ b/kernelci/api/models.py
@@ -25,6 +25,7 @@ from pydantic import (
     Field,
     FileUrl,
     StrictInt,
+    validator,
 )
 from .models_base import (
     PyObjectId,
@@ -612,6 +613,15 @@ class PublishEvent(BaseModel):
     attributes: Optional[Dict] = Field(
         description="Extra Cloudevents Extension Context Attributes"
     )
+
+    # suppress pylint error below
+    # It's a known issue: https://github.com/pylint-dev/pylint/issues/6900
+    @validator('data')
+    def validate_data(cls, val):  # pylint: disable=no-self-argument
+        """Do not allow 'None' as event payload data"""
+        if not val:
+            raise ValueError('None is not allowed as event payload')
+        return val
 
 
 def parse_node_obj(node: Node):


JenySadadia:
It would be necessary to declare `data` as `dict` type to avoid below errors in pipeline services:

When `data=None`:
```
kernelci-pipeline-monitor |   File "/home/kernelci/./pipeline/monitor.py", line 65, in _run
kernelci-pipeline-monitor |     if 'kernel_revision' in obj['data']
kernelci-pipeline-monitor |                             ~~~^^^^^^^^
kernelci-pipeline-monitor | TypeError: 'NoneType' object is not subscriptable
```

When `data=1`:
```
kernelci-pipeline-monitor |   File "/home/kernelci/./pipeline/monitor.py", line 65, in _run
kernelci-pipeline-monitor |     if 'kernel_revision' in obj['data']
kernelci-pipeline-monitor |                             ~~~^^^^^^^^
kernelci-pipeline-monitor | TypeError: 'int' object is not subscriptable
```

When `data={"data": 1}`
```
kernelci-pipeline-monitor |   File "/home/kernelci/./pipeline/monitor.py", line 65, in _run
kernelci-pipeline-monitor |     if 'kernel_revision' in obj['data']
kernelci-pipeline-monitor |                             ~~~^^^^^^^^
kernelci-pipeline-monitor | TypeError: string indices must be integers, not 'str'
```

JenySadadia:
> if event.data == 'BEEP': 

Oh yes, it somehow slipped through the cracks.

> If what you want to avoid "None" payloads, why not use the solution proposed in https://github.com/kernelci/kernelci-api/issues/451 ? -> "The issue can be fixed by not allowing None for PublishEvent.data field."

Seems like that is the only way forward.

JenySadadia:
Tested OK locally when `data=None` is sent to API:
```
$ ./kci event send input.json node
Error: 422 Client Error: Unprocessable Entity for url: http://172.17.0.1:8001/latest/publish/node
[{'loc': ['body', 'data'], 'msg': 'None is not allowed as event payload', 'type': 'value_error'}]
```

hardboprobot (approved):
Good. But remember that the pipeline monitor still expects this to be a dict.

JenySadadia:
> Good. But remember that the pipeline monitor still expects this to be a dict.

Yes, I'll create a separate issue to tackle this.

JenySadadia:
Tested OK on staging.

JenySadadia:
> > Good. But remember that the pipeline monitor still expects this to be a dict.
> 
> Yes, I'll create a separate issue to tackle this.

See https://github.com/kernelci/kernelci-pipeline/issues/481

PR #2452, opened 2024-03-12
kbuild: Output stdout and stderr to a single file for dtbs_check

nfraprado:
In order to be able to identify which Device Tree source files have generated which errors when running `make dtbs_check` we need both stdout and stderr to be in a single file.


nfraprado:
Better approach done in #2441.

PR #2453, opened 2024-03-13
kbuild: Fix back dtbs_check "magic word"

nuclearcat:
After fixing stderr/stdout file we got lost DTBS_CHECK_OK/FAIL result trigger in build.log output, return it back.


nuclearcat:
Tested at https://staging.kernelci.org:9000/viewer?node_id=65f2783e2231f89f50242d94

PR #2459, opened 2024-03-15
baseline: Disable bootrr tests

nuclearcat:
Unfortunately bootrr tests introduce more noise than useful results. Disable them for now.


nuclearcat:
tested on staging

PR #2467, opened 2024-03-19
T40464 config/docker/api.jinja2: use python 3.11

JenySadadia:
Depends on https://github.com/kernelci/kernelci-api/pull/504

Fixes https://github.com/kernelci/kernelci-api/issues/479

Since Python 3.11 support has been enabled for API, use the same version for building `api` docker image.
Remove conditional block from `base/python.jinja2` that was used to decide Python version as a part of a temporary solution for API until it supports Python 3.11.

### DIFF
--- a/config/docker/api.jinja2
+++ b/config/docker/api.jinja2
@@ -5,5 +5,4 @@
 -#}
 
 {%- set fragments = ['fragment/api.jinja2'] + fragments %}
-{%- set is_api = true %}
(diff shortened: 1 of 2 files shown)


pawiecz (approved):
LGTM (to be merged when https://github.com/kernelci/kernelci-api/pull/504 lands)

JenySadadia:
Tested OK on staging.

PR #2471, opened 2024-03-22
Allow listing multiple dtbs for a single platform

a-wai:
Device-tree files can be renamed upstream, like it happened for the `trogdor-kingoftown` Chromebook during the 6.4 developement cycle. In such cases, we have to provide several possible file names and select the appropriate one based on which files are present in the test kernel artifacts.

This PR makes it possible to store either a single string or a list of values in the platform config `dtb` field, and stores the selected value as a top-level `device_dtb` parameter to be used by templates. It also includes the corresponding rework of the existing templates.

To be merged in sync with https://github.com/kernelci/kernelci-pipeline/pull/495


nuclearcat (approved):
Tested on staging

PR #2472, opened 2024-03-22
kernelci: api: helper: add new `max_version` rule

a-wai:
Similar to the existing `min_version` rule, implement the needed bits for a new `max_version` rule. This can be useful if fundamental aspects of a given platform support is modified, or if support for a given platform is deprecated upstream.

### DIFF
--- a/kernelci/api/helper.py
+++ b/kernelci/api/helper.py
@@ -208,20 +208,27 @@ class APIHelper:
               - '!blocked-value2'
               ...
 
-        An additional rule can be defined to check for a minimum kernel version,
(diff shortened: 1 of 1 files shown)


a-wai:
Note: this was originally created so we could have 2 configs for `trogdor-kingoftown`: one to be used with kernels <= 6.3, the other one for newer kernels (>= 6.4).

I ended up going a [different route](https://github.com/kernelci/kernelci-pipeline/pull/495) for this specific case, but figured I could submit that change anyway as it was implemented and working.

PR #2474, opened 2024-03-25
models: rename regression.data.failed_kernel_revision to kernel_revision

hardboprobot:
In hindsight, using a different name than the rest of the Node submodels just for the sake of clarity was a bad idea.

This change will require running the db migration script in https://github.com/kernelci/kernelci-api/pull/508
https://github.com/kernelci/kernelci-pipeline/pull/501 should be applied after this one


nuclearcat:
I think it is not relevant anymore?

PR #2475, opened 2024-03-26
kbuild: compress log artifacts

nuclearcat:
To save space and egress fees it is better to compress log artifacts


nuclearcat:
Tested on staging

PR #2476, opened 2024-03-26
kernelci.api.models: node_sequence field in regressions data

hardboprobot:
Encode the concept of "active" and "inactive" regressions. A regression is "active" if the test is still failing after the failure was introduced. The regression has "fail" as a result. A regression is "inactive" if the test has passed after the regression was created. The regression result in this case is "pass".

A new "node_sequence" field to stores the list of failed test runs after the first failure for active regressions. For inactive regressions it stores up to the first passing test run.

### DIFF
--- a/kernelci/api/models.py
+++ b/kernelci/api/models.py
@@ -385,6 +385,9 @@ class KbuildData(BaseModel):
     kernel_type: Optional[str] = Field(
         description="Kernel image type (zimage, bzimage...)"
     )
+    regression: Optional[PyObjectId] = Field(
+        description="Regression node related to this build instance"
+    )
 
 
 class Kbuild(Node):
@@ -398,6 +401,10 @@ class Kbuild(Node):
     data: KbuildData = Field(
         description="Kbuild details"
     )
+
+    _OBJECT_ID_FIELDS = Node._OBJECT_ID_FIELDS + [
+        'data.regression',
+    ]
 
 
 class TestData(BaseModel):
@@ -426,6 +433,9 @@ class TestData(BaseModel):
     )
     job_context: Optional[str] = Field(
         description="Kubernetes cluster name the job submitted to"
+    )
+    regression: Optional[PyObjectId] = Field(
+        description="Regression node related to this test run"
     )
 
     # Fields inherited from the parent kbuild or test case node
@@ -463,6 +473,10 @@ class Test(Node):
         description="Test details"
     )
 
+    _OBJECT_ID_FIELDS = Node._OBJECT_ID_FIELDS + [
+        'data.regression',
+    ]
+
 
 class RegressionData(BaseModel):
     """Model for the data field of a Regression node"""
@@ -473,6 +487,7 @@ class RegressionData(BaseModel):
         description="Previous passing Node"
     )
     node_sequence: Optional[List[PyObjectId]] = Field(
+        default=[],
         description=("Instances of this same job ran after the initial "
                      "failure. The last run in the sequence may be a "
                      "passed run, which means the regression is no longer"
@@ -516,6 +531,7 @@ class Regression(Node):
         const=True
     )
     result: Optional[ResultValues] = Field(
+        default=ResultValues.FAIL.value,
         description=("PASS if the regression is 'inactive', that is, if the "
                      "test has ever passed after the regression was created. "
                      "FAIL if the regression is still 'active', ie. the test "
--- a/kernelci/api/models.py
+++ b/kernelci/api/models.py
@@ -472,6 +472,14 @@ class RegressionData(BaseModel):
     pass_node: Optional[PyObjectId] = Field(
         description="Previous passing Node"
     )
+    node_sequence: Optional[List[PyObjectId]] = Field(
+        description=("Instances of this same job ran after the initial "
+                     "failure. The last run in the sequence may be a "
+                     "passed run, which means the regression is no longer"
+                     "active. If the sequence is empty or if all the runs "
+                     "in the sequence failed, that means the job is still "
+                     "failing and the regression is active")
+    )
     error_code: Optional[ErrorCodes] = Field(
         description="Error code of the failed job"
     )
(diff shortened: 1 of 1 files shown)


JenySadadia (changes requested):
Could you please reduce the length of the commit title [`43234f6ac8205e90ce3bab1fe1559d1f21cb27a6`](https://github.com/kernelci/kernelci-core/pull/2476/commits/43234f6ac8205e90ce3bab1fe1559d1f21cb27a6)?
It exceeds the allowed length.

kernelci/api/models.py
```diff
@@ -507,6 +529,12 @@ class Regression(Node):
         description='Type of the object',
         const=True
     )
+    result: Optional[ResultValues] = Field(
+        description=("PASS if the regression is 'inactive', that is, if the "
+                     "test has ever passed after the regression was created. "
+                     "FAIL if the regression is still 'active', ie. the test "
+                     "is still failing")
+    )
```
JenySadadia, line 537: Related to https://github.com/kernelci/kernelci-pipeline/pull/503/files#r1544459852
```suggestion
    result: ResultValues = Field(
        default=ResultValues.FAIL.value,
        description=("PASS if the regression is 'inactive', that is, if the "
                     "test has ever passed after the regression was created. "
                     "FAIL if the regression is still 'active', ie. the test "
                     "is still failing")
    )
```

kernelci/api/models.py
```diff
@@ -472,6 +486,14 @@ class RegressionData(BaseModel):
...
         description="Previous passing Node"
     )
+    node_sequence: Optional[List[PyObjectId]] = Field(
+        description=("Instances of this same job ran after the initial "
+                     "failure. The last run in the sequence may be a "
+                     "passed run, which means the regression is no longer"
+                     "active. If the sequence is empty or if all the runs "
+                     "in the sequence failed, that means the job is still "
+                     "failing and the regression is active")
+    )
```
JenySadadia, line 496: Related to https://github.com/kernelci/kernelci-pipeline/pull/503/files#r1544460596
```suggestion
    node_sequence: List[PyObjectId] = Field(
        default=[],
        description=("Instances of this same job ran after the initial "
                     "failure. The last run in the sequence may be a "
                     "passed run, which means the regression is no longer"
                     "active. If the sequence is empty or if all the runs "
                     "in the sequence failed, that means the job is still "
                     "failing and the regression is active")
    )
```

kernelci/api/models.py
hardboprobot: I agree on having "FAIL" as a default value. About making this field mandatory, the reason I set it `Optional` has more to do with test & deployment logistics: there are already defined regressions in staging that don't contain this field, so having it mandatory would make this fail unless the DB is migrated accordingly. That means that in order to test and deploy this change on staging I'd have to involve all three repos, add a DB migration script, have the DB migrated and then test this PR and https://github.com/kernelci/kernelci-pipeline/pull/484. If I set the new fields as `Optional` for now this can be automatically tested on staging without any hassle.

kernelci/api/models.py
hardboprobot: Same reason as above, but I agree on the default value.

kernelci/api/models.py
hardboprobot: In summary, I know how hard it is sometimes to have certain changes integrated that require DB migrations, so this is a way to simplify the deployment.

kernelci/api/models.py
```diff
@@ -385,6 +385,9 @@ class KbuildData(BaseModel):
     kernel_type: Optional[str] = Field(
         description="Kernel image type (zimage, bzimage...)"
     )
+    regression: Optional[PyObjectId] = Field(
+        description="Regression node related to this build instance"
+    )
```
JenySadadia, line 390: We are already storing node IDs that caused regressions to `Regression` node. And that is enough to link failing nodes with regression. Then why are we storing regression ID here as well?

hardboprobot: > that is enough to link failing nodes with regression

Not in all cases, let me elaborate:

The purpose of this field here is to make it possible to check in a single query if a test run is "related to a regression". "Related to a regression" means that the test run _caused_ a regression _or_ that the test run is still failing after a certain regression was detected and created (so the node id is part of the regression `node_sequence`).

In the first case we can tell if a test case is related to a regression by doing a single query: "find regressions whose `fail_node` has this `node ID`". For the second case we can't because we can't do a query like "find regressions whose `node_sequence` contain this `node ID`". So this is an easy way to keep the bi-directional link between test nodes and regression nodes that works for all cases.

About why to do this at all: I believe it'll be helpful when we browse through test results to tell if a failed test caused a regression or if it's a known error that raised a regression some time ago.

JenySadadia: Okay. So basically this bi-directional link is for querying regressions.

> For the second case we can't because we can't do a query like "find regressions whose node_sequence contain this node ID"

In that case, this worked:
```
$ curl https://staging.kernelci.org:9000/latest/nodes?data.node_sequence=6603a4d35ca93487c2da3c76
```
The equality `=` operator works for querying an array field as well. See https://www.mongodb.com/docs/v3.4/tutorial/query-arrays/#query-an-array-for-an-element

hardboprobot: > In that case, this worked:

Aaah, I didn't know that, thanks for the tip! Ok, so at least we can query for the existence of an element in a list, that'll probably be helpful in some use cases :+1:

JenySadadia: Yes, would we still need this `regression` field?

hardboprobot: As with other fields in the models, it depends. It doesn't store additional information but it will allow certain features that wouldn't be viable without it.

Example: when searching for test results, with this field, once the query is done the results will automatically contain the links to the regressions. Without the field we can get the same information, but to do so we'd have to go through each result and run an additional query, which in practice would make the feature inviable.

So in this case I'd prefer to keep it.

JenySadadia (approved):

PR #2478, opened 2024-03-27
config: docker: data: tast_parser: split functionality

a-wai:
This script was initially meant to run only once, running Tast tests and parsing the corresponding results in a single operation. However, there isn't any obligation to enforce such a behaviour. It can even be useful to split those functionality and choose to either run the tests or parse the results using command-line options.

This change introduces the `--run` and `--results` flags to this effect: if either of those is present, only the corresponding feature will be executed; otherwise, we revert to the previous behaviour in order to maintain compatibility with the legacy system.

When called with `--results`, the script will return the number of failed tests, allowing us to mark the whole test suite as failed in this case.

### DIFF
--- a/config/docker/data/tast_parser.py
+++ b/config/docker/data/tast_parser.py
@@ -123,12 +123,14 @@ def parse_measurements(results_chart):
     return measurements
 
 
-def main(tests):
-    run_tests(tests)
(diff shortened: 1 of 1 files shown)


nuclearcat (approved):
LGTM

Issue #2464, opened 2024-03-18
Consider adding OS/system information to test results where relevant


a-wai:
@hardboprobot mentioned in an OOB discussion:

> Currently, the data we keep about a test result includes the kernel version used for it, but for other types of tests where the userspace is a key ingredient (ChromeOS - Yocto - Android), we might want to also encode the ChromeOS userspace version used as part of the test results. Does it makes sense?

Retrieving and storing such information would probably make sense (for ChromeOS it could be the content of `/etc/os-release` -- which is already `cat`'ed in LAVA jobs -- for example).

This needs to be carefully designed as it involves multiple components (e.g. making this information a test result in LAVA) and should be generic enough to be used across various userspaces (Android likely has different way of storing OS-related information).

Issue #2455, opened 2024-03-14
Issues with the representation of test hierarchies, result semantics and runtime errors


hardboprobot:
## Introduction

One of the most urgent problems to address before we move on with the next features, is the standardization and stabilization of the current API models, definitions and behaviors. Since there are a few tests running now we could get the impression that things are working and that we can move forward and continue developing additional features, but upon deeper inspection of the results, and from the perspective of a real user analyzing and building tools on top of them, it's clear that some things need to be refined.

After some analysis I managed to sum the problems in two parts: modeling of test results and test organization as hierarchies, and modeling of runtime and test fixture errors.

## Modeling of test results and test organization as hierarchies

In the current models, there's only one type of `Node` (`Test`) to represent a test result. This Test type is used throughout a test
hierarchy, that is, the same model is used for "test suite" nodes, "test case" nodes and everything in between.

However, we're finding that we need to model certain things differently depending on the type of node. Specifically, the "result" of a Test node might mean different things depending on whether the node is a "test case" or a "test suite" (and we haven't decided anything about intermediate nodes).

### Test cases

This is the most basic case. These will be the leaf nodes of a test node hierarchy, the individual test parts of a suite that can pass or fail (or allow for other result values: skip, etc).

From a user, developer perspective, and for regression detection purposes, these are the results we're interested in. Everything above this level is only a way to group tests together. In other words, for a developer, it's meaningful that a particular test case failed, but it may not be very meaningful that a test suite failed (what does it mean that a test suite failed?).

So the "result" field in these nodes is pretty clear right now. It translates directly to the result of the test case.

There's an exception to this: certain test suites may define test cases that work as sanity checks or fixtures. If these test cases fail it could mean that the test suite failed to run properly or that some requirement wasn't met. For our purposes, we can consider these as test cases to keep the same test suite semantics and translate them to KernelCI directly without transformations.

### Test suites
Test suites are collections of test cases that share the same general purpose, testing a single functionality, subsystem or feature.

In KernelCI they're currently being modeled as separate test nodes.

#### Test suite nodes and test case nodes are modeled using the same Node type
However, we don't use them for the same purpose, and the meaning of the "result" field could be interpreted differently:
- in a test case: the result of the test case
- in a test suite: whether the test suite could run completely or not. For instance, if there was a runtime problem that stopped the suite and left some test cases pending to run.

[...]

PR #2479, opened 2024-03-27
Improve results reporting for Tast tests

a-wai:
This PR adds the OS version as a "measurement" for the `os-release` test case, so it can be stored as part of the test results and provide valuable information when investigating a regression.

It also ensures the whole test suite is marked as failed if at least one of the Tast tests fail, making it easier to determine the actual status of a test run.

Depends on the docker images to be re-generated/published after #2478 is merged.

### DIFF
--- a/config/runtime/tests/tast.jinja2
+++ b/config/runtime/tests/tast.jinja2
@@ -30,13 +30,16 @@
               && cp cros /usr/libexec/tast/bundles/remote/
             - while ! ping -c 1 -w 1 $(lava-target-ip); do sleep 1; done
             - >-
-              lava-test-case os-release --shell
               ./ssh_retry.sh
               -o StrictHostKeyChecking=no
               -o UserKnownHostsFile=/dev/null
               -i /home/cros/.ssh/id_rsa
               root@$(lava-target-ip)
-              cat /etc/os-release
+              cat /etc/os-release > /tmp/osrel.tmp
+            - cat /tmp/osrel.tmp
+            - >-
+              lava-test-case os-release --result pass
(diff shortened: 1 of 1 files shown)


hardboprobot (commented):
Great, I understand this links to https://github.com/kernelci/kernelci-core/issues/2464 right?

How would the "measurement" look like in the KernelCI job node? Does this require an additional change in KernelCI code to interpret the Lava result and store it as part of the node data?

hardboprobot (commented):
Good idea, related to https://github.com/kernelci/kernelci-core/issues/2455

Is there any way we could keep the information about whether the test suite ran entirely or not, as an additional node field or in any other way? Prior to this change it was encoded as the "result". This change replaces that and uses "result" to encode a different thing, but now we'd be losing the initial information we were storing in "result", which could also be useful.

a-wai:
> How would the "measurement" look like in the KernelCI job node? Does this require an additional change in KernelCI code to interpret the Lava result and store it as part of the node data?

So it turns out that yes, this would require a change as measurements are basically discarded for now.

As an example:
* [LAVA job](https://lava.collabora.dev/scheduler/job/13168655): the `filemanager.UIPerf.directory_list` test has 2 child nodes, each one of those reporting a measurement
* The [corresponding sub-node](https://staging.kernelci.org:9000/viewer?node_id=6603fce9f3d47fcde0b3dea8) has `null` as a result (it is a test *set* to LAVA, not a test *suite*, so it doesn't have its dedicated result)
* The [leaf node](https://staging.kernelci.org:9000/viewer?node_id=6603fce9f3d47fcde0b3dea9) doesn't include the measurement information

a-wai:
Closing as the important bit is now part of #2498

PR #2480, opened 2024-03-27
kernelci: kbuild: handle case where cros configs aren't in a subdir

a-wai:
It seems there has been a change in https://chromium.googlesource.com/ causing the config files to no longer be under a `chromeos` subdir. Until we can ensure how the generated tarball is laid out, we can try both with and without the subdir and move on with whichever option succeeds.


nuclearcat (approved):
LGTM, tested and solving important bug

PR #2484, opened 2024-03-29
models: Add jobfilter field

nuclearcat:
If this field have defined list, schedulers in pipeline can schedule jobs ONLY specified in list.
This might be useful for bisection, so custom checkout will make node that will run only specific build(s) and specific test(s).

### DIFF
--- a/kernelci/api/helper.py
+++ b/kernelci/api/helper.py
@@ -272,6 +272,7 @@ class APIHelper:
     def create_job_node(self, job_config, input_node,
                         runtime=None, platform=None):
         """Create a new job node based on input and configuration"""
+        jobfilter = input_node.get('jobfilter')
         job_node = {
             'kind': job_config.kind,
             'parent': input_node['id'],
@@ -279,10 +280,16 @@ class APIHelper:
             'path': input_node['path'] + [job_config.name],
             'group': job_config.name,
             'artifacts': {},
+            'jobfilter': jobfilter,
             'data': {
                 'kernel_revision': input_node['data']['kernel_revision'],
             },
         }
+        # if jobfilter not null, verify if job_config.name exist in jobfilter
+        if jobfilter and job_config.name not in jobfilter:
+            print(f"Filtered: Job {job_config.name} not found in jobfilter")
+            return None
+
         if not self.should_create_node(job_config.rules, job_node, input_node):
             print(f"Not creating node due to job rules for {job_config.name}")
             return None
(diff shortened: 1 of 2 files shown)


kernelci/api/helper.py
```diff
@@ -279,10 +279,17 @@ def create_job_node(self, job_config, input_node,
             'path': input_node['path'] + [job_config.name],
             'group': job_config.name,
             'artifacts': {},
+            'jobfilter': input_node['jobfilter'],
```
a-wai, line 282: You should use a variable, like `jobfilter = input_node.get('jobfilter')`: this way we ensure `jobfilter` (the variable) always exists, but is `None` if `input_node` doesn't have a `jobfilter` key. Otherwise this line (and the `if` below) would raise a `KeyError` exception., causing a crash.

kernelci/api/helper.py
```diff
@@ -279,10 +279,17 @@ def create_job_node(self, job_config, input_node,
...
             'data': {
                 'kernel_revision': input_node['data']['kernel_revision'],
             },
         }
+        # if jobfilter not null, verify if job_config.name exist in jobfilter
+        if input_node['jobfilter']:
+            if job_config.name not in input_node['jobfilter']:
```
a-wai, line 289:
```suggestion
        if jobfilter:
            if job_config.name not in jobfilter:
```

According to the above comment

a-wai (changes requested):

kernelci/api/helper.py
nuclearcat: Updated

kernelci/api/helper.py
nuclearcat: Updated

kernelci/api/helper.py
```diff
@@ -272,17 +272,25 @@ def should_create_node(self, rules, node, parent=None):
...
             'group': job_config.name,
             'artifacts': {},
+            'jobfilter': jobfilter,
             'data': {
                 'kernel_revision': input_node['data']['kernel_revision'],
             },
         }
+        # if jobfilter not null, verify if job_config.name exist in jobfilter
+        if jobfilter:
+            if job_config.name not in jobfilter:
```
a-wai, line 290:
```suggestion
        if jobfilter and job_config.name not in jobfilter:
```

Minor nit, optional

a-wai (approved):
LGTM, just a minor nitpick you can ignore if you want

kernelci/api/helper.py
nuclearcat: No, i appreciate for such nits, i should not be lazy/low attention on such issues and produce more accurate code.
So, thanks for spotting, updated

PR #2493, opened 2024-04-05
CHROMEOS test-configs-chromeos: drop failing/absent tests

a-wai:
We experience a lot of failures due to `platform.Histograms` no longer existing. Moreover, recent investigations confirmed that `platform.Memd` and `platform.DLCService*` simply cannot run on mainline kernels, as they rely on downstream ChromeOS features.


nuclearcat:
Tested on staging, thanks!

PR #2494, opened 2024-04-08
api/models: Add submitter

nuclearcat:
Add submitter field that represents hash of token


nuclearcat:
Tested on staging

JenySadadia:
Please reflect this change in the API tests.

PR #2496, opened 2024-04-10
CHROMEOS config: docker: data: tast_parser: split functionality

a-wai:
This script was initially meant to run only once, running Tast tests and parsing the corresponding results in a single operation. However, there isn't any obligation to enforce such a behaviour. It can even be useful to split those functionality and choose to either run the tests or parse the results using command-line options.

This change introduces the `--run` and `--results` flags to this effect: if either of those is present, only the corresponding feature will be executed; otherwise, we revert to the previous behaviour in order to maintain compatibility with the legacy system.

This is basically #2478 but targetting the `chromeos` branch, which is the one used for actually building the docker images.

### DIFF
--- a/config/docker/data/tast_parser.py
+++ b/config/docker/data/tast_parser.py
@@ -24,6 +24,7 @@ import os
 import json
 import pwd
 
+FAILED_RUN_FILE = "failed_run"
 RESULTS_DIR = "/tmp/results"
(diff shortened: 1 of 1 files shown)


a-wai:
Dropped the part where we error out if any test failed: this is too opinionated and should wait for a consensus in #2455

Issue #2500, opened 2024-04-11
Builroot-generated rootfs should add a symlink for x86_64


a-wai:
For multiple reasons, 64-bit x86 platforms defined in the new system use the `x86_64` string as their `arch` param. However, the legacy system use `x86` everywhere for the same arch.

As a consequence, buildroot-generated images only contain the `x86` subfolder, except for `20230623.0` which has a `x86_64`  symlink pointing to `x86`. It would be nice to have this done automatically and avoid a whole class of [failures](https://lava.collabora.dev/scheduler/job/13343362).

Note: this doesn't affect Debian-based images as we use the debian architecture name here (`amd64`) and rely less on the `arch` param for autocompleting URLs.

PR #2501, opened 2024-04-11
config: runtime: boot: use the same buildroot image for all

a-wai:
This is the image used  for the `depthcharge` boot method, and also the only one containing a `x86_64` subfolder (or rather, symlink). Changing this avoid job failures on the new system due to how we handle the `arch` parameter there.


a-wai:
Example of failure: https://lava.collabora.dev/scheduler/job/13343362

Please note this is only a workaround, I've also opened #2500 so we don't forget to implement a proper fix.

a-wai:
Closing in favor of #2504

PR #2504, opened 2024-04-16
Add system-specific "arch" fields to account for differences

### LINKED ISSUE #2500: Builroot-generated rootfs should add a symlink for x86_64

a-wai:
This PR adds new, "virtual" (i.e. not used directly in config files), attributes for platform configuration in order to make it easier to account for differences in architecture strings used by the different systems (e.g. `x86` for Buildroot, `amd64` for Debian, `x86_64` for ChromeOS). More specifically, the following attributes are now available:
* `crosarch` for the ChromeOS-compatible architecture string
* `debarch` for the Debian version
* `karch` for the one used by the Linux kernel and Buildroot

The `boot` templates are modified to use those new properties instead of plain `arch`.

Closes #2500

### DIFF
--- a/config/runtime/boot/depthcharge.jinja2
+++ b/config/runtime/boot/depthcharge.jinja2
@@ -34,7 +34,7 @@
 {%- else %}
     ramdisk:
       compression: gz
-      url: 'http://storage.kernelci.org/images/rootfs/buildroot/buildroot-baseline/20230623.0/{{ platform_config.arch }}/rootfs.cpio.gz'
+      url: 'http://storage.kernelci.org/images/rootfs/buildroot/buildroot-baseline/20230623.0/{{ platform_config.karch }}/rootfs.cpio.gz'
 {%- endif %}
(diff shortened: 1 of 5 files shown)


nuclearcat:
Very good idea! I remember in old kernelci it was hardcoded hacks, your way is way more configurable.

nuclearcat:
Tested on staging

PR #2506, opened 2024-04-18
docker: Add lava-callback fragment

nuclearcat:
To run lava-callback in k8s we need proper fragment containing flask.


nuclearcat:
Tested on staging

PR #2508, opened 2024-04-19
docker/pipeline: Add gcloud tools

nuclearcat:
Right now kubectl emit error in scheduler-k8s:
```
04/19/2024 12:27:47 AM UTC [ERROR] [Errno 2] No such file or directory: 'gke-gcloud-auth-plugin'
```
As gke auth plugin is missing.


nuclearcat:
Tested on k8s

PR #2511, opened 2024-04-22
build-configs.yaml: Disable media tree

nuclearcat:
media tree is broken for several month and seems nobody is going to repair it, so disabling for now as it is breaking jenkins jobs:
```
+ kci_build update_repo --build-config=media --kdir=/data/workspace/bot.kernelci.org/build-trigger/workspace/kernel-build-trigger__2/configs/media --mirror=/data/workspace/bot.kernelci.org/build-trigger/workspace/kernel-build-trigger__2/linux.git
Cloning into '/data/workspace/bot.kernelci.org/build-trigger/workspace/kernel-build-trigger__2/configs/media'...

warning: remote HEAD refers to nonexistent ref, unable to checkout
fatal: couldn't find remote ref HEAD
```

### DIFF
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -1266,25 +1266,25 @@ build_configs:
           x86_64:
             base_defconfig: 'x86_64_defconfig'
 
-  media:
-    tree: media
-    branch: 'master'
-    variants:
-      gcc-10:
-        build_environment: gcc-10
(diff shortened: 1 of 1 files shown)


nuclearcat:
Can't be tested on staging, as it conflicts with staging patches, merging as is to not miss production update. Will update staging patches as well after merging.

PR #2512, opened 2024-04-22
kernelci: update system-specific arch params in runtime.get_params()

a-wai:
We currently allow using the `crosarch`, `debarch` & `karch` virtual
params in configuration files, but those should also be accessible from
job templates. This used to be possible with the initial implementation
as those were part of the `Platform` config object, but this was changed
as it caused negative side effects.

As the current implementation is more robust, stick to it but make it
available to other modules so we can make those params usable from
`Runtime` objects.

### DIFF
--- a/config/runtime/boot/u-boot.jinja2
+++ b/config/runtime/boot/u-boot.jinja2
@@ -11,15 +11,15 @@
 {%- endif %}
 {%- if boot_commands == 'nfs' %}
     nfsrootfs:
-      url: 'http://storage.kernelci.org/images/rootfs/debian/bookworm-kselftest/20240221.0/{{ platform_config.debarch }}/full.rootfs.tar.xz'
+      url: 'http://storage.kernelci.org/images/rootfs/debian/bookworm-kselftest/20240221.0/{{ debarch }}/full.rootfs.tar.xz'
       compression: xz
     ramdisk:
-      url: 'http://storage.kernelci.org/images/rootfs/debian/bookworm-kselftest/20240221.0/{{ platform_config.debarch }}/initrd.cpio.gz'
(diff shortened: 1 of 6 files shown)


config/runtime/boot/u-boot.jinja2
```diff
@@ -11,15 +11,15 @@
...
+      url: 'http://storage.kernelci.org/images/rootfs/debian/bookworm-kselftest/20240221.0/{{ debarch }}/full.rootfs.tar.xz'
       compression: xz
     ramdisk:
-      url: 'http://storage.kernelci.org/images/rootfs/debian/bookworm-kselftest/20240221.0/{{ platform_config.debarch }}/initrd.cpio.gz'
+      url: 'http://storage.kernelci.org/images/rootfs/debian/bookworm-kselftest/20240221.0/{{ debarch }}/initrd.cpio.gz'
       compression: gz
     os: debian
 {%- else %}
     ramdisk:
-      url: 'http://storage.kernelci.org/images/rootfs/buildroot/buildroot-baseline/20230703.0/{{ platform_config.debarch }}/rootfs.cpio.gz'
```
pawiecz, line 22: Shouldn't buildroot images use `karch` instead? Let me double check

a-wai: Dammit you're right, pushing a fix right now!

pawiecz (approved):
LGTM

a-wai:
Added `brarch` as well as Builroot doesn't always follow the kernel practice (e.g. 32-bit arm is either `armel` or `armeb`, while the kernel only uses `arm`).

nuclearcat:
Thanks! I guess it fixes buildroot issues

PR #2515, opened 2024-04-25
config: runtime: tests: tast: fail early for unreachable devices

a-wai:
Currently, we try to ping the DUT indefinitely, leading to the job to timeout after 30 mins. This is way too long and impacts other users, so we should limit the number of retries and raise an error if the device is unreachable.

This change ensures the job fails after 2 mins (60 iterations, 1s for `ping` + 1s for `sleep` per iteration).


nuclearcat (approved):
LGTM, thanks for pointing to results on staging!

PR #2517, opened 2024-04-26
kernelci.api.helper: fix `submit_results`

JenySadadia:
Enable sending `data` dictionary fields from root node while submitting hierarchy of results.


nuclearcat (approved):
LGTM, results on staging seems fine

PR #2535, opened 2024-04-30
config: runtime: tests: Add v4l2 fluster tests lava template

musamaanjum:
Copy file template file v4l2-decoder-conformance.jinja2 and edit it.

Note: This is my first PR. Please review it and I'll improve it.

### DIFF
--- a/config/runtime/tests/v4l2-decoder-conformance.jinja2
+++ b/config/runtime/tests/v4l2-decoder-conformance.jinja2
@@ -1,0 +1,35 @@
+{%- if testsuite -%}
+    {% set testsuite_arg = "-ts " + testsuite %}
+{%- endif -%}
+{%- if decoders -%}
+    {%- set decoder_list = [ ] -%}
+    {%- for decoder in decoders -%}
+        {{ decoder_list.append(decoder) or "" }}
+    {%- endfor -%}
+    {%- set decoders_arg = "-d " + decoder_list|join(" ") -%}
+{%- endif -%}
+{%- if videodec_timeout -%}
+    {% set videodec_timeout_arg = "-t " + videodec_timeout|string %}
+{%- endif -%}
+{%- if videodec_parallel_jobs -%}
+    {% set videodec_parallel_jobs_arg = "-j " + videodec_parallel_jobs|string %}
+{%- endif -%}
+- test:
+    timeout:
+      minutes: {{ job_timeout }}
+    definitions:
+    - repository:
+        metadata:
+          format: Lava-Test Test Definition 1.0
+          name: {{ plan }}
+          description: "V4L2 decoder conformance test plan"
+          os:
+          - debian
+          scope:
+          - functional
+        run:
+          steps:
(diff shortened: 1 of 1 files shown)


musamaanjum:
I've added the staging-skip until someone reviews to not disturb the staging.

laura-nao:
Looks ok to me as it's basically just a copy of the old template (except for `plan` being replaced with `node.name`). I'll let someone from the core team review and approve this too.

config/runtime/tests/v4l2-decoder-conformance.jinja2
```diff
@@ -0,0 +1,35 @@
+{%- if testsuite -%}
+    {% set testsuite_arg = "-ts " + testsuite %}
+{%- endif -%}
+{%- if decoders -%}
+    {%- set decoder_list = [ ] -%}
+    {%- for decoder in decoders -%}
+        {{ decoder_list.append(decoder) or "" }}
+    {%- endfor -%}
+    {%- set decoders_arg = "-d " + decoder_list|join(" ") -%}
```
a-wai, line 9:
```suggestion
    {%- set decoders_arg = "-d " + decoders|join(" ") -%}
```

`decoders` is already a list.

config/runtime/tests/v4l2-decoder-conformance.jinja2
```diff
@@ -0,0 +1,35 @@
...
+          os:
+          - debian
+          scope:
+          - functional
+        run:
+          steps:
+          - python3 /usr/bin/fluster_parser.py {{ testsuite_arg }} {{ decoders_arg }} {{ videodec_timeout_arg }} {{ videodec_parallel_jobs_arg }}
+      from: inline
+      name: {{ plan }}
+      path: inline/{{ plan }}.yaml
```
a-wai, line 35: Where is `plan` defined? I suspect this was available in the legacy system but not something we have in the new system.

a-wai (changes requested):

config/runtime/tests/v4l2-decoder-conformance.jinja2
musamaanjum: @laura-nao @nuclearcat any ideas about this?

laura-nao: Right, I missed this. `plan` doesn't seem to be used by the new system indeed, I think we can use `node.name`  instead here as well.

musamaanjum: Okay. Put name: node.name and removed path.

laura-nao: @musamaanjum the path is a requirement for inline test definitions in LAVA (https://docs.lavasoftware.org/lava/writing-tests.html#using-inline-test-definitions), I think we should keep it and use `inline/{{ node.name }}.yaml`.

musamaanjum: Updated the path.

nuclearcat:
Thank you @musamaanjum , this is very demanded test, tested on staging, merging.

PR #2537, opened 2024-04-30
kernelci.config.job: add attribute 'kcidb_test_suite'

JenySadadia:
Every KernelCI test job will have mapping to KCIDB test suite name to unify the test names with other CI systems.
Add `kcidb_test_suite` attribute to job configurations that will store the mapped KCIDB test suite name.


JenySadadia:
./staging

nuclearcat:
Command /staging works only for website, for kernelci-core and other projects i will add it a bit later.

PR #2538, opened 2024-04-30
build-configs.yaml: Add fragment with Collabora lab specific configs

nfraprado:
Some devices in Collabora's lab are connected to the wired network using an ASIX AX88772B ethernet adapter. The kernel config for it needs to be enabled in order for tests that require working network to succeed. Furthermore it needs to be builtin for the same reason as commit ed888aa7f200 ("build-configs.yaml: Prevent deferred probe timeout errors").

Since this is specific to the hardware used at the Collabora lab, add a new config fragment for it.

@laura-nao Talking to you I got the impression the ASIX adapter is something standalone and specific to Collabora's setup. Can you confirm that? The alternative would be that the ASIX adapter is present in a newer revision of the servo board, in which case I'd consider it a generic setup, and include this config in the `arm64-chromebook` fragment instead.

Example runs:
* mt8186-corsola-steelix-sku131072 running baseline-nfs with the ChromeOS config, which doesn't include this config. It times out while waiting for a network interface to come up: https://lava.collabora.dev/scheduler/job/13560437
* With this config enabled it works: https://lava.collabora.dev/scheduler/job/13589287

### DIFF
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -351,6 +351,7 @@ fragments:
       - 'CONFIG_DRM_MEDIATEK_DP=m'
       - 'CONFIG_SND_SOC_SOF_MT8195=m'
       - 'CONFIG_USB_RTL8152=y' # Allows kernel ip-config. Prevents deferred probe timeout errors.
+      - 'CONFIG_USB_USBNET=y' # Needed for ASIX AX88772B ethernet adapter used at Collabora's lab. Builtin to allow kernel ip-config, and prevent deferred probe timeout errors.
       - 'CONFIG_EXTRA_FIRMWARE="
       mediatek/mt8173/vpu_d.bin
       mediatek/mt8173/vpu_p.bin
@@ -610,6 +611,7 @@ fragments:
       - 'CONFIG_USB_ETH=y'
       - 'CONFIG_USB_GADGET=y'
       - 'CONFIG_USB_RTL8152=y'
+      - 'CONFIG_USB_USBNET=y' # Needed for ASIX AX88772B ethernet adapter used at Collabora's lab. Builtin to allow kernel ip-config, and prevent deferred probe timeout errors.
       - 'CONFIG_USB_VIDEO_CLASS=m'
       - 'CONFIG_WILCO_EC_DEBUGFS=m'
       - 'CONFIG_WILCO_EC_EVENTS=m'


laura-nao:
> @laura-nao Talking to you I got the impression the ASIX adapter is something standalone and specific to Collabora's setup. Can you confirm that? The alternative would be that the ASIX adapter is present in a newer revision of the servo board, in which case I'd consider it a generic setup, and include this config in the `arm64-chromebook` fragment instead.

Yes, I confirm the Asix adapters are specific for our lab configuration and are external to the ServoV4. I believe this fragment could host the R8152 options too, given that the ServoV4 is also part of the lab setup and that we're using external R8152 usb-to-ethernet adapters on some devices too. (not something to tackle necessarily in this PR though)

nfraprado:
@laura-nao True, the ServoV4 board is part of the lab setup, but it seems to me like that would be the go-to board to connect Chromebooks to in a lab, hence why I thought made sense to keep that one in a generic fragment, common to chromebooks (`arm64-chromebook`).

laura-nao:
> @laura-nao True, the ServoV4 board is part of the lab setup, but it seems to me like that would be the go-to board to connect Chromebooks to in a lab, hence why I thought made sense to keep that one in a generic fragment, common to chromebooks (`arm64-chromebook`).

Technically, SuzyQables could also be used instead of the ServoV4 (which is actually the case for the majority of devices in the Collabora lab). I guess it depends on the purpose of the arm64-chromebook fragment - if it's meant to contain only the configurations necessary for booting devices and enabling peripheral testing, then I would exclude all network boot-related configs as those are specific to the lab setup. However, if the fragment is intended to include all configs needed to boot and test the devices in a particular lab environment, then I would just put everything there.

nfraprado:
@laura-nao Yeah, that makes sense. I think it's better to keep all the lab setup separate, it'll make the fragments easier to maintain if there are other labs testing chromebooks with their own adapters. So I can move the r8152 config to the collabora fragment, but I'll do that in a separate PR after the needed changes in kernelci-pipeline to use this new fragment are applied to avoid introducing regressions in the results.

@a-wai @nuclearcat If you don't have any concerns with the introduction of this new fragment, this is ready to merge as is.

a-wai:
I'm not sure it's really worth adding a new fragment for that, especially as this could also potentially be needed for non-Chromebook devices.

My take on this would be to either:
* add this option to *both* `x86-board` and `arm64-chromebook fragments
* move *all* the lab-setup related options (NFS, R8152, this one and maybe others?) to a single separate fragment and use it on top of device-specific ones; having per-lab fragments definitely feels a bit overkill to me

nfraprado:
@a-wai FWIW the point that this could be needed by non-chromebook devices I think gives an extra reason for having it in a non-specific chromebook fragment (unlike `arm64-chromebook`). That said, I don't know of non-chromebooks that are missing configs for their adapters, so I just went ahead with the easiest option and added the config to both `x86-board` and `arm64-chromebook` fragments. The highest priority for me right now is to get rid of all the noise from the failed jobs.

a-wai (approved):
> @a-wai FWIW the point that this could be needed by non-chromebook devices I think gives an extra reason for having it in a non-specific chromebook fragment

Agreed, hence my suggestion to have a separate `lab-setup` (or `test-setup` or whatever) fragment including all options we currently have in both `x86-board` and `arm64-chromebook` which are:
* arch/device-agnostic
* needed for testing in LAVA

This way, we can keep only the options related to **hardware support** in those fragments, and not bother with options related to **testing equipment support**.

That can be done as a low priority follow-up item

PR #2539, opened 2024-05-01
api/helper.py: Don't crash if kernel_revision is missing, but report error

nuclearcat:
Probably due some bug this key was missing in some cases.

### DIFF
--- a/kernelci/api/helper.py
+++ b/kernelci/api/helper.py
@@ -273,18 +273,23 @@ class APIHelper:
                         runtime=None, platform=None):
         """Create a new job node based on input and configuration"""
         jobfilter = input_node.get('jobfilter')
-        job_node = {
-            'kind': job_config.kind,
-            'parent': input_node['id'],
-            'name': job_config.name,
-            'path': input_node['path'] + [job_config.name],
-            'group': job_config.name,
-            'artifacts': {},
-            'jobfilter': jobfilter,
-            'data': {
-                'kernel_revision': input_node['data']['kernel_revision'],
-            },
-        }
+        try:
+            job_node = {
+                'kind': job_config.kind,
+                'parent': input_node['id'],
+                'name': job_config.name,
+                'path': input_node['path'] + [job_config.name],
+                'group': job_config.name,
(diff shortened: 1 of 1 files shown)


kernelci/api/helper.py
```diff
@@ -273,6 +273,11 @@ def create_job_node(self, job_config, input_node,
                         runtime=None, platform=None):
         """Create a new job node based on input and configuration"""
         jobfilter = input_node.get('jobfilter')
+        kernel_revision = input_node['data'].get('kernel_revision')
+        if not kernel_revision:
+            print("No kernel revision found in input node")
+            return None
```
a-wai, line 279: Why not enclose the `job_node = ...` assignment in a `try: except:` block? This would feel appropriate.

nuclearcat: Thats always my thoughts, which way more appropriate, generally "safe" .get or try/except. Is there any common practices or something, or it is matter of choice?

a-wai: My feeling is that `.get()` should be used if we accept `None` as a valid value and continue processing even if the key is missing from the dict.

If a missing key is a serious enough issue for us to stop normal processing and return early and/or do something else, then using `var['key']` in a try/except block should be preferred IMHO.

nuclearcat: Fair enough, will update

a-wai: Actually my suggestion was more to do things this way:
```python
        try:
            job_node = {
                'kind': job_config.kind,
                'parent': input_node['id'],
                'name': job_config.name,
                'group': job_config.name,
                'artifacts': {},
                'jobfilter': jobfilter,
                'data': {
                },
            }
        except KeyError as key:
            print(f"Input node missing '{key}' attribute, aborting node creation")
            return None
```

Missing `kernel_revision` is a bug, but we could envision a similar bug affecting the `id`, `path` or even `data` attributes, so we could catch all of those with the above solution.

nuclearcat: agree, updated

a-wai (approved):

PR #2545, opened 2024-05-07
config/runtime: Add support for parameter "platforms"

nuclearcat:
We might have limited set of platforms available in certain lab. Lab owners should have option to define such list. 
As per suggestion: https://github.com/kernelci/kernelci-project/issues/350#issuecomment-2096625245

### DIFF
--- a/kernelci/config/runtime.py
+++ b/kernelci/config/runtime.py
@@ -85,7 +85,7 @@ class RuntimeLAVA(Runtime):
     # pylint: disable=too-many-arguments
     def __init__(self, url, priority=None, priority_min=None,
                  priority_max=None, queue_timeout=None, notify=None,
-                 **kwargs):
+                 platforms=None, **kwargs):
         super().__init__(**kwargs)
 
         def _set_priority_value(value, default):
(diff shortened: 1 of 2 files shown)


a-wai:
What would be the point of this, as in, what problem would it actually solve ? Scheduler entries already link to a runtime instance **and** list the platforms on which jobs should run.

I'm not sure adding `platforms` to runtime definitions would add anything useful here.

nuclearcat:
We might have for example raspberry pi device defined as platform, available in baylibre and collabora, as many other common SBC, and we have job that will execute for example generic arm64 build on number of platforms and labs:
```
  - job: baseline-arm64
    event:
      channel: node
      name: kbuild-gcc-10-arm64
      result: pass
    runtime:
      type: lava
      name: lava-collabora
    platforms:
      - bcm2711-rpi-4-b
      - meson-g12b-a311d-khadas-vim3
      - rk3399-gru-kevin
      - rk3399-rock-pi-4b
      - rk3588-rock-5b
      - sun50i-h6-pine-h64
```
(we nee to add support of list of runtime names btw)
But for example baylibre don't want to allow KernelCI to run jobs on their Raspberry PI, so with this patch they can define list of allowed platforms only.

nuclearcat:
Also this is easy and safe place  (less error-prone) where lab owners can specify mandatory "permit" list, so nobody by mistake put jobs on devices that kernelci not supposed to run.

nuclearcat:
As per feedback closing for now

PR #2546, opened 2024-05-07
config/rootfs: dmesg.sh: Change exit code for dmesg test script

laura-nao:
Update the exit code for the dmesg test script to return 1 in the event of one or more test cases failing. This ensures that when the test is executed in LAVA, the overall pass/fail status of the dmesg test suite reflects the outcome of the individual test cases (i.e. status should be fail if any test cases fail).

### DIFF
--- a/config/rootfs/debos/overlays/baseline/opt/kernelci/dmesg.sh
+++ b/config/rootfs/debos/overlays/baseline/opt/kernelci/dmesg.sh
@@ -2,6 +2,8 @@
 
 set -e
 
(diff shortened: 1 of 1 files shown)


laura-nao:
Manual LAVA job for reference: https://lava.collabora.dev/scheduler/job/13675753

Here the result for the `dmesg_0` test suite is `fail`, as the `emerg` test case failed.

nuclearcat:
As change is trivial and tested on LAVA - merging

PR #2547, opened 2024-05-07
CHROMEOS upgrade to R124 images

a-wai:
Draft as it requires final images to be built & uploaded to storage.chromeos.kernelci.org


a-wai:
Added the equivalent of #2553 for legacy

Undrafting as we're now good to go.

nuclearcat:
Thanks merging

PR #2549, opened 2024-05-08
build-configs.yaml: Add lab-setup fragment

### LINKED ISSUE #2540: Create separate kconfig fragments for testing equipment/features support
>> @a-wai FWIW the point that this could be needed by non-chromebook devices I think gives an extra reason for having it in a non-specific chromebook fragment
>
>Agreed, hence my suggestion to have a separate `lab-setup` (or `test-setup` or whatever) fragment including all options we currently have in both `x86-board` and `arm64-chromebook` which are:
>* arch/device-agnostic
>* needed for testing in LAVA
>
>This way, we can keep only the options related to **hardware support** in those fragments, and not bother with options related to **testing equipment support**.
>
>That can be done as a low priority follow-up item

_Originally posted by @a-wai in https://github.com/kernelci/kernelci-core/pull/2538#pullrequestreview-2035942121_

Such options include, for example:
* CONFIG_USB_RTL8152
* CONFIG_USB_USBNET
* CONFIG_NFS_FS
* CONFIG_ROOT_NFS
* ...

nfraprado:
I'm still seeing long boot delays on mt8186-steelix: https://lava.collabora.dev/scheduler/job/13696724 . Upon closer inspection I see `CONFIG_USB_USBNET` is `=y`, but `CONFIG_USB_NET_AX8817X` is `=m`, even though it's supposed to be builtin by default. So I'm enabling it explicitly here, which should hopefully fix the issue. Since I was adding one more config for the ethernet adapters it seemed a good moment to do the move to a separate lab-setup fragment. Fixes #2540.

---

Add a new fragment to contain configurations required for the lab setup which, at least for now, consists of builtin network and NFS support.

These configs come from the arm64-chromebook and x86-board fragments, but aren't removed from those fragments for now in order to not impact legacy KernelCI.

An extra config, CONFIG_USB_NET_AX8817X, is also added which is needed for an ethernet adapter.

### DIFF
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -449,6 +449,25 @@ fragments:
       - 'CONFIG_NET_IPGRE=m'
       - 'CONFIG_NET_IPGRE_DEMUX=m'
 
+  lab-setup:
+    path: "kernel/configs/lab-setup.config"
+    configs:
+      - 'CONFIG_NFS_FS=y'
+      - 'CONFIG_NFS_V2=y'
+      - 'CONFIG_NFS_V3=y'
+      - 'CONFIG_NFS_V3_ACL=y'
+      - 'CONFIG_NFS_V4=y'
+      # Disable kernel IP config to avoid needing builtin network support,
+      # prevent long boot delays and deferred probe timeout errors. Rely on
+      # ramdisk to do the IP config.
+      - 'CONFIG_IP_PNP=n'
+      # Network adapters
+      - 'CONFIG_USB_ETH=m'
+      - 'CONFIG_USB_GADGET=m'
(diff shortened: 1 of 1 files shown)


nfraprado:
I've updated the fragment to disable IP config in the kernel (`CONFIG_IP_PNP`). That means that a ramdisk will need to be supplied for jobs that boot to an NFS rootfs, since the ramdisk needs to do the IP config, but as far as I can see, all NFS jobs already supply a ramdisk, so it shouldn't have any impact.

The benefit of this is that we won't need to add builtin network support for all combinations of network adapters and platforms, just having them enabled as modules is enough.

nuclearcat:
Merging as seems ok on staging

PR #2550, opened 2024-05-09
kernelci.api.helper: get `is_hierarchy` from event data

JenySadadia:
Depends on https://github.com/kernelci/kernelci-api/pull/529

PubSub event data has a field called `is_hierarchy` to distinguish events of submission of hierarchy of results from a single node submission.
Return this field value from `APIHelper.receive_event_node` method along with node.


nuclearcat:
Is it tested and i can merge both?

JenySadadia:
Yes, tested OK on staging.
Please merge its dependency PR first https://github.com/kernelci/kernelci-api/pull/529.

PR #2551, opened 2024-05-09
kbuild: Add option to not install modules

nuclearcat:
Some configuration doesn't build modules,
we should not attempt to build them.


nuclearcat:
Tested on staging

PR #2552, opened 2024-05-14
config/core: rename staging entry

JenySadadia:
Update YAML configurations for staging instance.
Rename staging instance entries from `staging.kernelci.org` to `staging` to simplify TOML section names.
The the same convetion has been used here for similar reasons: https://github.com/kernelci/kernelci-pipeline/commit/1991f2cfd18fe0fde12c3a13c8738f381f483aa5


pawiecz (approved):
LGTM

PR #2553, opened 2024-05-14
config: runtime: tests: tast: drop useless "sync"

a-wai:
Calling `sync` after running Tast tests was a leftover from an old version, but it's been made useless as we now properly power off the device, with the kernel automatically performing this operation during shutdown.

This call failing caused wrong results to be reported, so let's drop it entirely.


nuclearcat:
It was just attempt to fight with filesystem corruption :)

PR #2554, opened 2024-05-16
CHROMEOS config: docker: data: tast_parser: report results then exit

a-wai:
In some cases (e.g. the DUT crashes while running one testcase), the Tast run might not finish but still yield important results, which we're losing if we error out before parsing them.

This change ensure we report existing results and only then error out in such cases.

Manually tested in LAVA: https://lava.collabora.dev/scheduler/job/13841947
As expected, the job fails ("Incomplete") but reports the results of tests which actually ran.


nuclearcat (approved):
LGTM

PR #2556, opened 2024-05-21
build-configs.yaml: Enable more configs for x86 Intel Chromebooks

laura-nao:
Enable additional configurations for x86 Intel Chromebooks in the x86-board fragment. These missing configurations were identified based on the preliminary results of the ACPI kselftest on Intel Chromebooks.


a-wai (approved):

nuclearcat (approved):
LGTM, thanks a lot!

PR #2558, opened 2024-05-22
config: runtime: allow to retry boot actions

a-wai:
There are numerous cases where LAVA either fails to recognize the login (and/or shell) prompt, leading to a job failure even if everything actually goes well.

Allowing `boot` actions to be retried is an effective way to work around those issues and reduce our waste our resources time.

### DIFF
--- a/config/runtime/boot/depthcharge.jinja2
+++ b/config/runtime/boot/depthcharge.jinja2
@@ -46,6 +46,7 @@
     namespace: {{ boot_namespace }}
 {%- endif %}
     method: depthcharge
+    failure_retry: 3
     prompts:
     - '/ #'
     timeout:
--- a/config/runtime/boot/depthcharge.jinja2
+++ b/config/runtime/boot/depthcharge.jinja2
@@ -50,4 +50,9 @@
     prompts:
     - '/ #'
     timeout:
-      minutes: 5
+      minutes: 10
+    timeouts:
+      bootloader-commands:
+        minutes: 3
+      auto-login-action:
+        minutes: 2
--- a/config/runtime/boot/grub.jinja2
+++ b/config/runtime/boot/grub.jinja2
@@ -22,3 +22,10 @@
     failure_retry: 3
     prompts:
     - '/ #'
+    timeout:
+      minutes: 10
+    timeouts:
+      bootloader-commands:
+        minutes: 3
+      auto-login-action:
+        minutes: 2
--- a/config/runtime/boot/grub.jinja2
+++ b/config/runtime/boot/grub.jinja2
@@ -19,5 +19,6 @@
 - boot:
     method: grub
     commands: ramdisk
+    failure_retry: 3
     prompts:
     - '/ #'
--- a/config/runtime/boot/qemu.jinja2
+++ b/config/runtime/boot/qemu.jinja2
@@ -27,4 +27,9 @@
     prompts:
       - '/ #'
     timeout:
-      minutes: 5
+      minutes: 10
+    timeouts:
+      bootloader-commands:
+        minutes: 3
+      auto-login-action:
+        minutes: 2
--- a/config/runtime/boot/qemu.jinja2
+++ b/config/runtime/boot/qemu.jinja2
@@ -23,6 +23,7 @@
       image: kernelci/qemu
     media: tmpfs
     method: qemu
+    failure_retry: 3
     prompts:
       - '/ #'
     timeout:
--- a/config/runtime/boot/u-boot.jinja2
+++ b/config/runtime/boot/u-boot.jinja2
@@ -30,6 +30,7 @@
 - boot:
     method: u-boot
     commands: {{ boot_commands | default('ramdisk', true) }}
+    failure_retry: 3
     prompts:
     - '/ #'
     timeout:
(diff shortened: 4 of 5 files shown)


pawiecz (approved):
LGTM

laura-nao:
@a-wai I think we need to fix the action timeouts accordingly (and potentially reduce the current boot action timeout, 5 minutes is a long time for depthcharge) I bumped into some instances of  `No time left for remaining 2 retries. 1 retries out of 3 failed for depthcharge-retry` errors on staging (see e.g. https://lava.collabora.dev/scheduler/job/13963436).

a-wai:
> @a-wai I think we need to fix the action timeouts accordingly (and potentially reduce the current boot action timeout, 5 minutes is a long time for depthcharge): I bumped into some instances of `No time left for remaining 2 retries. 1 retries out of 3 failed for depthcharge-retry` errors on staging (see e.g. https://lava.collabora.dev/scheduler/job/13963436).

Based on my experiments, there's no easy way to avoid those, unfortunately:
* if we reduce the timeout for the `boot` action, then it will time out quicker, but won't retry anyway as the timeout has been reached already
* changing the global `action` timeout won't be effective as the `boot` action overrides it with its own timeout

In other words: retries work fine when an error occurs, but are ignored (with this annoying error message) on timeouts.

laura-nao:
> > @a-wai I think we need to fix the action timeouts accordingly (and potentially reduce the current boot action timeout, 5 minutes is a long time for depthcharge): I bumped into some instances of `No time left for remaining 2 retries. 1 retries out of 3 failed for depthcharge-retry` errors on staging (see e.g. https://lava.collabora.dev/scheduler/job/13963436).
> 
> Based on my experiments, there's no easy way to avoid those, unfortunately:
> 
>     * if we reduce the timeout for the `boot` action, then it will time out quicker, but won't retry anyway as the timeout has been reached already
> 
>     * changing the global `action` timeout won't be effective as the `boot` action overrides it with its own timeout
> 
> 
> In other words: retries work fine when an error occurs, but are ignored (with this annoying error message) on timeouts.

I think we might need to adjust the timeouts for individual actions like `depthcharge-retry` and `depthcharge-action` - I'll see if I can find some useful reference or throw some quick tests to our staging LAVA instance (I recall we used a similar approach for other test jobs as well). Timeouts should be configurable to allow for retries, if this doesn't work as expected we should address the issue in LAVA itself. I'm mentioning this not to be pedantic, but because timeouts at the bootloader level should be raised as `InfrastructureError` exceptions rather than `JobError` exceptions.

laura-nao:
@a-wai overriding just the `bootloader-commands` timeout should do (at least until [this MR](https://gitlab.com/lava/lava/-/merge_requests/2428) gets merged upstream):
```
actions:
  bootloader-commands:
    minutes: 1 
```
I did a quick test on staging and it seems to work (I forced a `bootloader-commands` action timeout by overriding `bootloader_prompt` in the job def): 
- without timeout override: https://staging.lava.collabora.dev/scheduler/job/686585
- with timeout override: https://staging.lava.collabora.dev/scheduler/job/686617 - the action is retried 3 times as expected and an `InfrastructureError` is raised if all attempts failed

 One minute should work for depthcharge, not sure if that's enough for the other bootloaders though (if not, we can just increase the action timeout accordingly).

Does this look reasonable to you? can you please re-run one of your previous test jobs with this and see if it still works?

a-wai:
Interesting, thanks for looking into it! Not sure 1 minute is enough, as IIUC `bootloader-commands` include the TFTP upload time, I'd go with 3 minutes to keep some margin.

Other stuff I noted:
* if we define a timeout for the `boot` action, then the `bootloader-commands` timeout is ignored: https://lava.collabora.dev/scheduler/job/13987231
* `bootloader-commands` only lasts until the kernel starts, so we should probably override `auto-login-action` as well, for cases where the kernel boots but eventually hangs before we get to a login prompt

So I guess the way forward would be to:
* drop timeout we set for `boot` actions
* add a 3 minutes timeout for `bootloader-commands`
* add a 2 minutes timeout for `auto-login-action`

Tested this approach on 3 jobs:
* https://lava.collabora.dev/scheduler/job/13987790 (passed after a retry)
* https://lava.collabora.dev/scheduler/job/13987789 (failed, but 3 attempts were made)
* https://lava.collabora.dev/scheduler/job/13987726 (passed, no retry needed)

a-wai:
Oh that's even better, thanks! A pity all those neat tricks are poorly documented (or at least, not so easily found)

nfraprado:
I've come across a job, https://lava.collabora.dev/scheduler/job/14038815, where boot only succeeded after a few tries, and the result stored in the KernelCI database was a failure, https://staging.kernelci.org:9000/viewer?node_id=6654b8e5e4f27102d46d4b27, (which lead to a regression node being generated) though it should have been a pass, since that was the last result.

So it seems that the KernelCI code that parses the LAVA results will need to be tweaked to take the retries into account.

a-wai:
> So it seems that the KernelCI code that parses the LAVA results will need to be tweaked to take the retries into account.

Right, I guess the issue is that LAVA results aren't presented in chronological order IIRC, so we could have a FAIL from an early failed attempt overwrite a PASS from the last attempt.

I'll fix that asap.

nuclearcat:
Is this PR ready for merging or not yet?

PR #2559, opened 2024-05-22
CHROMEOS config: lava: chromeos: cros-boot: retry boot actions

a-wai:
There are numerous cases where LAVA either fails to recognize the login (and/or shell) prompt, leading to a job failure even if everything actually goes well.

Allowing `boot` actions to be retried is an effective way to work around those issues and reduce our waste of resources time.


a-wai:
Based on #2558

pawiecz (approved):
LGTM

PR #2560, opened 2024-05-24
CHROMEOS config: docker: data: tast_parser: handle missing results file

a-wai:
When the test run failed before yielding a single result, the JSON file containing the results is missing. This leads to an exception being raised, rather than properly erroring out with a meaningful message.

Avoid this issue by ensuring the file is present before opening it.


nuclearcat (approved):
LGTM

PR #2561, opened 2024-05-28
runtime/lava: Introduce method to get log to variable

nuclearcat:
We might need to retrieve log as variable in some cases in LAVA callback.


nuclearcat:
Already used in lava callback, so merging

PR #2562, opened 2024-05-28
Add watchdog reset test template and enable watchdog configs

laura-nao:
Add template for watchdog reset test. Enable the required watchdog configs to run the test on x86_64 and arm64 Chromebooks.

### DIFF
--- a/config/runtime/tests/watchdog-reset.jinja2
+++ b/config/runtime/tests/watchdog-reset.jinja2
@@ -1,0 +1,24 @@
+{# This test verifies the watchdog reset functionality.
+ # It opens the watchdog device and waits for the timeout to expire, triggering a device reset.
+ # If the reset occurs as expected, a bootloader string should be matched.
+ # If the command returns successfully without a reset, the prompt string will be matched instead.
+ # This test assumes no other watchdog service is running on the system.
+-#}
+- test:
+    timeout:
+      minutes: 5
+    interactive:
+    - name: wdt-reset
+      prompts: ['/ #']
+      script:
+      - name: wdt-get-timeout
+        command: "wdt_timeout=$(cat /sys/class/watchdog/{{ wdt_dev|default('watchdog0') }}/timeout)"
+        wait_for_prompt: true
+        failures:
+        - message: "can't open '/sys/class/watchdog/watchdog\\d+/timeout': No such file or directory"
+          exception: JobError
+      - name: wdt-trigger-reset
+        command: "sleep $((wdt_timeout + 5)) > /dev/watchdog"
+        wait_for_prompt: false
+        successes:
+        - message: {{ bl_message }}
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -223,6 +223,7 @@ fragments:
       - 'CONFIG_QCOM_RMTFS_MEM=y'
       - 'CONFIG_QCOM_SOCINFO=y'
       - 'CONFIG_QCOM_SYSMON=y'
+      - 'CONFIG_QCOM_WDT=m'
       - 'CONFIG_QRTR_SMD=y'
       - 'CONFIG_QRTR=y'
       - 'CONFIG_RMNET=y'
@@ -332,6 +333,7 @@ fragments:
       - 'CONFIG_SND_SOC_SOF_TOPLEVEL=y'
       - 'CONFIG_SPI_GPIO=y'
       - 'CONFIG_SPI_MTK_NOR=y'
+      - 'CONFIG_MEDIATEK_WATCHDOG=m'
       - 'CONFIG_MTK_THERMAL=y'
       - 'CONFIG_MTK_SOC_THERMAL=m'
       - 'CONFIG_MTK_LVTS_THERMAL=m'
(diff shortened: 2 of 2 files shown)


config/runtime/tests/watchdog-reset.jinja2
```diff
@@ -0,0 +1,12 @@
+- test:
+    timeout:
+      minutes: {{ test_timeout|default('2') }}
+    interactive:
+    - name: wdt-reset
+      prompts: ['{{ rootfs_prompt }}']
+      script:
+      - name: wdt-expect-reset
+        command: "sleep {{ wdt_timeout|default('30') | int + 5 }} > {{ wdt_dev|default('/dev/watchdog')}}"
```
a-wai, line 9: A comment explaining how/why this command works would be nice (I assume the shell command is blocking the watchdog device file, so the OS can't reset it)

OOC, could the WDT timeout be retrieved from somewhere else, like a sysfs entry?

config/runtime/tests/watchdog-reset.jinja2
```diff
@@ -0,0 +1,12 @@
+- test:
+    timeout:
+      minutes: {{ test_timeout|default('2') }}
```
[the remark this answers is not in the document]
laura-nao: I thought of making this timeout configurable because the sleep duration for the test might vary (based on the watchdog timeout) and the action timeout needs to be set accordingly. On second thought, I think this can be a fixed overarching timeout. I'll fix this.

config/runtime/tests/watchdog-reset.jinja2
laura-nao: This command opens the watchdog device and just waits for the timeout to expire, causing the device to reset. If the reset occurs as expected, a bootloader string should be matched ; if the command returns successfully without a reset, the prompt string will be matched instead. I can add a comment in the template itself to clarify this.

> OOC, could the WDT timeout be retrieved from somewhere else, like a sysfs entry?

Yes, according to [the docs](https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-class-watchdog) the timeout is exposed by `/sys/class/watchdog/watchdogX/timeout`. Afaik there's no way to define variables in LAVA interactive tests, but we could change the command to `sleep $(( $(cat /sys/class/watchdog/watchdogX/timeout) + 2 )) > /dev/watchdog` - it's less readable, but the user would just need to provide the name of the watchdog device. I'll run a quick manual test to check if this even works when integrated in the LAVA definition..

config/runtime/tests/watchdog-reset.jinja2
```diff
@@ -0,0 +1,12 @@
...
+    interactive:
+    - name: wdt-reset
+      prompts: ['{{ rootfs_prompt }}']
```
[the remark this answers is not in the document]
a-wai: Right, you referred to the template for the legacy system, hence the confusion (we really need to fork it and make it a fully separate project).

The depthcharge template for the new system is [here](https://github.com/kernelci/kernelci-core/blob/main/config/runtime/boot/depthcharge.jinja2) (and more generally, templates used by the new system are only found under `config/runtime` in this repo)

config/runtime/tests/watchdog-reset.jinja2
a-wai: > This command opens the watchdog device and just waits for the timeout to expire, causing the device to reset.

Does that mean no other process is likely to reset the watchdog?

> Afaik there's no way to define variables in LAVA interactive tests

Too bad it doesn't accept a `commands:` param with a list of commands like we have in other tests...

> we could change the command to [...]

I'll trust your judgement here, FWIW this question was really driven by curiosity, I'd be fine with a comment explaining how this works so future devs don't need to scratch their head.

config/runtime/tests/watchdog-reset.jinja2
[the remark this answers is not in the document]
laura-nao: Thanks! Seems like all the templates in `config/runtime` are using a fixed `/ #` prompt - I'll do the same here.

config/runtime/tests/watchdog-reset.jinja2
laura-nao: > Does that mean no other process is likely to reset the watchdog?

This test assumes there's no other watchdog service running on the system - this applies to our current `bookworm` rootfs, so I didn't feel the need to create an ad-hoc rootfs for this test.

> Too bad it doesn't accept a commands: param with a list of commands like we have in other tests...

Actually, I realized it's possible to include more than one command and define a variable within that context. I force-pushed some changes - would you mind reviewing them and see if they look good to you?

I manually tested the changes and they seem to work as expected - e.g.: https://lava.collabora.dev/scheduler/job/14061709

a-wai: > This test assumes there's no other watchdog service running on the system - this applies to our current `bookworm` rootfs

Ah, that's worth mentioning in the comment as well.

> Actually, I realized it's possible to include more than one command and define a variable within that context.

Awesome, thanks for checking!

a-wai (approved):
LGTM

laura-nao:
Added failure message matching to the template when the timeout is read from sysfs:
```
failures:
 - message: "can't open '/sys/class/watchdog/watchdog\\d+/timeout': No such file or directory"
   exception: JobError
```

This will cause the job to fail with a JobError whenever the test requirements are not met, such as when the watchdog device is missing or the watchdog sysfs interface is not enabled.

Example run where the watchdog sysfs interface was intentionally disabled: https://lava.collabora.dev/scheduler/job/14147118

This ensures that the reset test is not executed when the watchdog timeout is unknown.

a-wai (approved):

PR #2563, opened 2024-05-29
rootfs-configs.yaml: Increase image size for igt

nuclearcat:
During armhf build we noticed disk space issues.
https://github.com/kernelci/kernelci-core/actions/runs/9280775475/job/25535588226
```
2024-05-29T05:35:03.8358631Z 2024/05/29 05:35:01 a630_sqe.fw | cat: write error: No space left on device
```


nuclearcat:
It solve problem with disk space, but still igt image have problems on armhf, i will open separate issue for that

PR #2565, opened 2024-05-30
config: add qcs6490-rb3gen2 device type

mwasilew:
qcs6490-rb3gen2 corresponds to gen 2 of Qualcomm robotics kit. This device type is pre-configured to use "fastboot boot" as boot method.

### DIFF
--- a/config/runtime/boot/fastboot.jinja2
+++ b/config/runtime/boot/fastboot.jinja2
@@ -1,0 +1,39 @@
+- deploy:
+    images:
+      image:
+        url: '{{ node.artifacts.kernel }}'
+      dtb:
+        url: '{{ node.artifacts.dtb }}'
+      ramdisk:
+        url: 'http://storage.kernelci.org/images/rootfs/buildroot/buildroot-baseline/20230703.0/{{ brarch }}/rootfs.cpio.gz'
+        compression: gz
+        format: cpio.newc
+        overlays:
+          lava: true
+    postprocess:
+      docker:
+        image: ghcr.io/mwasilew/docker-mkbootimage:master
+        steps:
+        - mkbootimg --header_version 2 --kernel Image --dtb qcs6490-rb3gen2.dtb --cmdline "earlycon clk_ignore_unused pd_ignore_unused audit=0" --ramdisk rootfs.cpio.gz --output boot.img
+    to: downloads
+
+- deploy:
+    images:
+      boot:
+        url: 'downloads://boot.img'
+    timeout:
+      minutes: 2
+    to: download
+
+- boot:
+    prompts:
+    - '/ #'
+    failure_retry: 3
+    timeout:
+      minutes: 10
+    timeouts:
+      bootloader-commands:
+        minutes: 3
+      auto-login-action:
+        minutes: 2
+    method: fastboot


mwasilew:
@JenySadadia @pawiecz I think this is required for the pipeline changes I will post in a moment.

config/runtime/boot/fastboot.jinja2
```diff
@@ -0,0 +1,29 @@
...
+    images:
+      boot:
+        url: 'downloads://boot.img'
+    timeout:
+      minutes: 20
+    to: download
+
+- boot:
+    prompts:
+    - root@(.*):[/~]#
```
a-wai, line 26:
```suggestion
    - '/ #'
```

As mentioned in https://github.com/kernelci/kernelci-pipeline/pull/623#issuecomment-2139635297

config/core/test-configs.yaml
```diff
@@ -1797,6 +1797,11 @@ device_types:
     params:
       boot_timeout: '10'
 
+  qcs6490-rb3gen2:
```
a-wai, line 1800: You can simply drop changes to this file, it's for the legacy system only, which is currently on life support. Better not to add anything new there at this point.

config/runtime/boot/fastboot.jinja2
```diff
@@ -0,0 +1,29 @@
...
+        steps:
+        - mkbootimg --header_version 2 --kernel Image --dtb qcs6490-rb3gen2.dtb --cmdline "earlycon clk_ignore_unused pd_ignore_unused audit=0" --ramdisk rootfs.cpio.gz --output boot.img
+    to: downloads
+
+- deploy:
+    images:
+      boot:
+        url: 'downloads://boot.img'
+    timeout:
+      minutes: 20
```
a-wai, line 21: This feels a bit excessive, a value of `1` or `2` would likely be enough here.

config/runtime/boot/fastboot.jinja2
```diff
@@ -0,0 +1,29 @@
...
+    to: download
+
+- boot:
+    prompts:
+    - root@(.*):[/~]#
+    timeout:
+      minutes: 15
```
a-wai, line 28:
```suggestion
    failure_retry: 3
    timeout:
      minutes: 10
    timeouts:
      bootloader-commands:
        minutes: 3
      auto-login-action:
        minutes: 2
```

This would make it aligned with https://github.com/kernelci/kernelci-core/pull/2558, allowing multiple boot attempts in case something goes wrong (either on the board or on the LAVA side).

a-wai (changes requested):

config/runtime/boot/fastboot.jinja2
mwasilew: Yes, this is just copying the file.

config/runtime/boot/fastboot.jinja2
mwasilew: yes, this fixes it. Thanks!

config/runtime/boot/fastboot.jinja2
mwasilew: fixed.

config/runtime/boot/fastboot.jinja2
mwasilew: fixed

config/core/test-configs.yaml
mwasilew: OK. Dropped.

mwasilew:
I manage to add LAVA overlay to ramdisk. This version of the patch supports running tests.

config/runtime/boot/fastboot.jinja2
```diff
@@ -0,0 +1,39 @@
...
+        url: '{{ node.artifacts.kernel }}'
+      dtb:
+        url: '{{ node.artifacts.dtb }}'
+      ramdisk:
+        url: 'http://storage.kernelci.org/images/rootfs/buildroot/buildroot-baseline/20230703.0/{{ brarch }}/rootfs.cpio.gz'
+        compression: gz
+        format: cpio.newc
+        overlays:
+          lava: true
+     postprocess:
```
pawiecz, line 13: Attempt to use this template (as described in https://github.com/kernelci/kernelci-pipeline/pull/623#issuecomment-2147008153) results in:

```py
yaml.parser.ParserError: while parsing a block mapping
  in "<unicode string>", line 61, column 5
did not find expected key
  in "<unicode string>", line 68, column 6
```

```suggestion
    postprocess:
```

Aligning indentation allowed me to render TestJob (otherwise [YAML dumping](https://github.com/kernelci/kernelci-core/blob/260d3a5283a8c1a9b07b10fdf8eba37f9e8ad1aa/kernelci/runtime/lava.py#L250) failed)

pawiecz (changes requested):

config/runtime/boot/fastboot.jinja2
mwasilew: oh, I missed that. Let me fix it quickly.

mwasilew: Fixed now.

pawiecz (approved):
New template works fine for TestJob submission: https://staging.kernelci.org:9000/node/66602410d0c90ecf5c348089 (TestJob itself failed with an InfraError but that's out of scope for Maestro)

mwasilew:
I re-ran the job and it passed. I think there is an issue with board reset. I'll try to fix it.

PR #2566, opened 2024-05-30
build-configs.yaml: Enable firmware and config for sound on MT8186

nfraprado:
Add SOF firmware needed to get the ADSP probed on the mt8186-corsola-steelix-sku131072 platform. This fixes a test case in the DT kselftest and is a pre-requirement to get working audio. Also enable the new config needed for the sound card on MT8186, since it has been renamed.


nuclearcat:
LGTM on staging, merging

PR #2568, opened 2024-06-03
kbuild.py: Add more clarity to build script logs

nuclearcat:
1)Do not try to build dtbs for architectures that doesnt have it.
2)Add more clarity to non-critical scripts
3)Add notification that build script ended in log and script


nuclearcat:
Tested on staging

PR #2569, opened 2024-06-04
config/docker: add libubsan to kselftest fragment

musamaanjum:
The libubsan is missing and is required for some kselftest suites such as openat2. The error is as following:

[error while loading shared libraries: libubsan.so.1: cannot open shared object file: No such file or directory](https://storage.kernelci.org/next/master/next-20240529/arm/multi_v7_defconfig/gcc-10/lab-broonie/kselftest-openat2-beaglebone-black.html)


musamaanjum:
Request to review @nuclearcat @a-wai

nuclearcat:
Tested on staging

PR #2571, opened 2024-06-04
Introduce `Job` model and re-structure test hierarchy

JenySadadia:
- Introduce `Job` model for job/test suite nodes
- Introduce `setup` test suite for pre-check runs for `tast` test
- Tweaked LAVA callback data parsing to re-structure test hierarchy
- Calculate job node results based on child test nodes

### DIFF
--- a/kernelci/runtime/lava.py
+++ b/kernelci/runtime/lava.py
@@ -155,8 +155,10 @@ class Callback:
         for suite_name, suite_results in self._data['results'].items():
             tests = yaml.safe_load(suite_results)
             if suite_name == 'lava':
-                results['login'] = self._get_login_case(tests)
-                results['kernelmsg'] = self._get_kernelmsg_case(tests)
+                results['setup'] = {
+                    'login': self._get_login_case(tests),
+                    'kernelmsg': self._get_kernelmsg_case(tests)
+                }
             else:
                 suite_name = suite_name.partition("_")[2]
                 results[suite_name] = self._get_suite_results(tests)
@@ -182,8 +184,10 @@ class Callback:
             if isinstance(value, dict):
                 item['child_nodes'] = self._get_results_hierarchy(value)
                 node['result'] = self._get_stage_result(node['name'])
+                node['kind'] = 'job'
             elif isinstance(value, str):
                 node['result'] = value
+                node['kind'] = 'test'
             hierarchy.append(item)
         return hierarchy
 
--- a/kernelci/runtime/lava.py
+++ b/kernelci/runtime/lava.py
@@ -16,7 +16,10 @@ from urllib.parse import urljoin
 import requests
 import yaml
 
-from kernelci.runtime import Runtime
+from kernelci.runtime import (
+    Runtime,
+    evaluate_test_suite_result
+)
 
 
 # This will go away when adding get_html_log()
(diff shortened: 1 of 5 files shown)


kernelci/runtime/lava.py
```diff
@@ -182,8 +186,10 @@ def _get_results_hierarchy(self, results):
             if isinstance(value, dict):
                 item['child_nodes'] = self._get_results_hierarchy(value)
                 node['result'] = self._get_stage_result(node['name'])
+                node['kind'] = 'job'
```
hardboprobot, line 189: Will this be applied to all nodes that contain child nodes (node groups included) or only to the top-level test suite nodes?

Re: https://github.com/kernelci/kernelci-core/issues/2455#issuecomment-2121984882

hardboprobot (commented):
I'd remove the commented lines altogether before merging this.

kernelci/runtime/lava.py
JenySadadia: Technically, all the nodes with child nodes will have `kind=job`.
So they'll be job nodes and their result will be calculated based on child node results.
See this hierarchy https://staging.kernelci.org:9000/viewer?node_id=666102ecd0c90ecf5c35640e
What do you think about it?

hardboprobot: Fine by me, that's already better that what we have now, I think.

hardboprobot (approved):

JenySadadia:
`tast` test hierarchy with `setup` test suite: https://staging.kernelci.org:9000/viewer?node_id=6661733f451ad0645154c1c2

kernelci/runtime/lava.py
```diff
@@ -164,15 +164,28 @@ def get_results(self):
...
+        if 'fail' in result_list:
+            result = 'fail'
+        elif all(result == 'pass' for result in result_list):
+            result = 'pass'
+        elif all(result == 'skip' for result in result_list):
+            result = 'skip'
+        elif 'pass' in result_list:
+            result = 'pass'
+        elif 'skip' in result_list:
+            result = 'skip'
```
pawiecz, line 188: I'd suggest simplifying this condition: `elif all(result == '$result' for result in result_list) or '$result' in result_list` leaving the current order of conditions.

This way if 1st condition is not met there's only `incomplete/skip/pass`, 2nd would exclude `pass`es and 3rd would check if there were any `skip`s. There still would be the edge case were all the results could be none of these kinds and the returned result from function would be `None` - would that be a valid value?

pawiecz (changes requested):

kernelci/runtime/lava.py
JenySadadia: > I'd suggest simplifying this condition: elif all(result == '$result' for result in result_list) or '$result' in result_list leaving the current order of conditions (fail -> pass -> skip even though as a rule I'd suggest leaving pass at the very end).

Thanks for the suggestion. I'll update it accordingly.

> There still would be the edge case were all the results could be none of these kinds and the returned result from function would be None - would that be a valid value?

Yes, `None` will be stored as `result=null` and that's valid as the field is optional.

pawiecz (approved):
LGTM

nuclearcat:
Let's check for today, and if all ok, i merge at evening

PR #2573, opened 2024-06-05
kernelci.cli.job: fix `kci job submit` command

JenySadadia:
Fixes https://github.com/kernelci/kernelci-core/issues/2574

`kci job submit` is not able to find `runtime_token` for the TOML configuration.
Fix the bug by providing `api` config entry as it will look for `api.<api-config>.runtime_token` in the
`kci.secrets` section of the TOML file.


JenySadadia:
Tested OK on staging.

PR #2580, opened 2024-06-20
build-configs.yaml: Add kcidebug fragment

nuclearcat:
It might be useful to enable lightweight(relatively) kernel debug options.


broonie:
This looks like it might be useful, we might need to pare some of these depending on how bad the performance is.

Might also be worth taking a look at https://git.kernel.org/pub/scm/linux/kernel/git/mark/linux.git/log/?h=testing/v6.10-rc3 which is a similar idea.

nuclearcat:
Tested on staging

PR #2582, opened 2024-06-20
config/runtime/base/python.jinja2: update `BaseJob._submit`

JenySadadia:
Update `_submit` method in `BaseJob` and drop its returned value as it's not being used anywhere.


pawiecz (approved):
LGTM

PR #2584, opened 2024-06-24
workflows: Add initial version of docker images building workflows

nuclearcat:
Initial version of docker workflows to test if docker images are buildable


nuclearcat:
Tested manually on my own tree, merging, should not affect other workflows

PR #2586, opened 2024-06-24
kernelci/docker: Add timeout value

nuclearcat:
Add timeout value to docker API on push, otherwise rust image fails on default timeout.


nuclearcat:
Tested on staging, no disruptions

PR #2589, opened 2024-06-25
Use LAVA stage result

JenySadadia:
`kernelci.runtime.lava.py`: add `_get_stage_result`
    
Get suite/stage result from LAVA. If it's `pass`, evaluate manually from child node results to overcome
an issue with LAVA that it set `pass` to test suites even if one of the sub-tests is failing.

### DIFF
--- a/config/runtime/tests/tast.jinja2
+++ b/config/runtime/tests/tast.jinja2
@@ -12,7 +12,7 @@
       location: /home/cros/lava
     definitions:
     - from: inline
-      name: setup
+      name: tast
       path: inline/cros-tast.yaml
       repository:
         metadata:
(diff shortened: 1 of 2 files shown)


JenySadadia:
Seems like working on staging:
- Even if child `setup` is `pass`, `tast` has `fail` as result as that's the LAVA stage result:
https://staging.kernelci.org:9000/viewer?node_id=667abf35739857c4a8b570e7
- When the stage result for `tast` is `pass`, it manually evaluated test suite result from child nodes:
https://staging.kernelci.org:9000/viewer?search=parent%3D667ac076739857c4a8b574f5

pawiecz (approved):
LGTM

PR #2590, opened 2024-06-26
config/storage: Fix legacy option

nuclearcat:
Legacy require storage.kernelci.org, so 200bfa86296c023959f059568dc3c070b7069de7 breaks legacy setup for staging. Adding back option


nuclearcat:
Tested on staging

PR #2596, opened 2024-07-02
pipeline.jinja2: Add kcidb dependency to pipeline

nuclearcat:
As we are moving kcidb bridge to production, for kubernetes we need this dependency.


nuclearcat:
Tested on staging

PR #2599, opened 2024-07-03
config/v4l2-decoder-conformance: Replace dots in LAVA test set/case names

laura-nao:
Dots are used in KCIDB to denote test hierarchy; this can lead to misinterpretation when they're used in LAVA test sets and case names. Modify the fluster parser script to replace dots with dashes to prevent this issue.

### DIFF
--- a/config/rootfs/debos/overlays/fluster/usr/bin/fluster_parser.py
+++ b/config/rootfs/debos/overlays/fluster/usr/bin/fluster_parser.py
@@ -110,13 +110,15 @@ def main(args):
         cmd['case'], 'validate-fluster-results', '--result', 'pass'])
 
     # parse test results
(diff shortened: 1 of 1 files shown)


JenySadadia:
Thanks a lot @laura-nao for the fix.
Let's wait for the results on staging in the next run.

laura-nao:
@JenySadadia I opened https://github.com/kernelci/kernelci-pipeline/pull/675 to test the changes on staging. I'll check the results in a while.

laura-nao:
Seems to be working fine on staging - see e.g.: https://staging.kernelci.org:9000/viewer?node_id=668b02578e761c46b24e6b11

I think this can be merged @JenySadadia , then we can build the definitive version of the rootfs for both staging and production.

JenySadadia (approved):
Thanks for the fix.

PR #2600, opened 2024-07-03
config: runtime: boot: add support for barebox bootloader

pawiecz:
Barebox is used by platforms from Pengutronix LAVA laboratory.


a-wai:
This doesn't take into account the case where `boot_commands` == `nfs`; I'd rather suggest to just copy the u-boot template and `s/u-boot/barebox/` there, as it seems they should be otherwise identical.

pawiecz:
Superseded by https://github.com/kernelci/kernelci-core/pull/2782

PR #2603, opened 2024-07-05
base/lava.jinja2: Add instance_callback retrieved from environment

nuclearcat:
Right now we define callback URL for LAVA labs in config, which doesn't make sense, as this URL is pipeline instance specific, and complicates deployment of staging vs production instances. It is better to define callback URL as environment variable, for now.


nuclearcat:
Tested on staging

JenySadadia:
As per the discussion, alternatively, we can think about using TOML file to define this variable later on.

PR #2608, opened 2024-07-10
kbuild: Allow list in defconfig option

nuclearcat:
In some cases we need to have complex targets, such as: make defconfig allmodconfig hardened.config
or
make imx_v6_v7_defconfig allmodconfig

Now we can set it as, for example:

kbuild-gcc-12-arm-android-allmodconfig:
    <<: *kbuild-gcc-12-arm-android-job
    params:
      <<: *kbuild-gcc-12-arm-params
      defconfig:
        - imx_v6_v7_defconfig
        - allmodconfig


pawiecz (approved):
LGTM

PR #2609, opened 2024-07-10
rootfs:overlays:fluster Add verbose option for fluster parser

denisyuji:
add `--verbose` option for fluster parser to make fluster more verbose to facilitate debug


laura-nao:
@denisyuji you should use `main` branch as the target branch for this PR.

Did you get a chance to try the `--verbose` option manually in a LAVA job? I wonder how much verbosity it adds and whether it significantly slows down the test execution.

nuclearcat:
I am adding staging skip as this PR might corrupt staging.

PR #2613, opened 2024-07-16
Implement tree/branch combinations in rules

### LINKED ISSUE #2548: Allowing filtering on tree/branch pairs
Currently, our `rules` system allows one to ensure jobs will run only for allowed values of each specified node attribute.

This makes it possible to run jobs only for a given set of trees and/or branches, but doesn't allow to enter a list of tree/branch combinations, which can be useful to have more fine-grained conditions.

An example situation could be the following:
* the `mainline` tree is built on branch `master`
* the `media` tree is built on branches `master` and `fixes`
* we want to run a job on both `mainline/master` and `media/fixes`, but not on `media/master`

This could be achieved by allowing `tree` and `branch` rules to be formatted as `<tree>/<branch>` and process those entries in a special way if they contain a `/` character. This solution feels the most user-friendly one, however implementing this might be quite complex.

Alternative solutions could be:
* add a new `build_config` rule type, which will look at the build configs defined in the pipeline configuration to identify the tree/branch combinations, and then proceed to checking the corresponding node attributes
*  add a new rule type (name TBD) which will be a special case (just like current `*_version` rules) in the filtering code working as described above (accept entries in the `<tree>/<branch>` format and filter based on the node's tree and branch)

a-wai:
Currently the rules can only specify allowed/denied trees or branches as
a whole, but we lack a way to authorize (or exclude) only a specific
branch for a given tree.

This change implements tree/branch combination-based rules, by allowing
to specify such combinations using the `tree:branch` format under the
`tree` and/or `branch` rules.

Note: the `:` character has been chosen as a separator because git
forbids its use in branch names.

Fixes #2548

### DIFF
--- a/kernelci/api/helper.py
+++ b/kernelci/api/helper.py
@@ -191,6 +191,95 @@ class APIHelper:
 
         return True
 
+    def _extract_combos(self, rules, key):
+        """
+        Process tree/branch rules and extract "combo" values.
+
+        Tree and branch rules can be formatted as `<tree>:<branch>`, meaning
+        only a given branch is allowed for a specific tree. When prepended with
+        `!`, it indicates a forbidden tree/branch combination.
+
+        In order to simplify further processing, this function processes those
+        rules and returns lists of allowed/denied single values and combos.
+
+        Returns a tuple of 4 arrays:
+          * allowed single values
+          * allowed combinations
+          * denied single values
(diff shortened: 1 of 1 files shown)


nuclearcat:
As always, thank you very much for your exceptionally professional and useful contributions!
Tested on staging

PR #2616, opened 2024-07-17
config: core: add SCHED_CONFIG needed for cyclicdeadline rt-test

musamaanjum:
I've reproduced the failure locally. The cyclicdeadline requires sched_feature or sched/feature directory. Add these directories by including SCHED_DEBUG config in the kernel.

### DIFF
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -538,6 +538,7 @@ fragments:
     path: "kernel/configs/preempt_rt.config"
     configs:
       - 'CONFIG_EXPERT=y'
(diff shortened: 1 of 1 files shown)


musamaanjum:
I've confirmed locally that this configuration fixes this test. Lets wait to get the results from staging as well.

musamaanjum:
Confirmed that CONFIG_SCHED_DEBUG is being included in preempt-rt fragment in overall build config of the kernel: https://staging.kernelci.org:9000/viewer?node_id=66a2128b41e62083cda5f7d6

PR #2617, opened 2024-07-18
v4l2-decoder-conformance: Enable Fluster tests on ChromeOS

denisyuji:
This PR aims to combine Tast test and v4l2-decoder-conformance test structures to execute Fluster test on ChromeOS. It re-use the ChromeOS kernel module loading structure from Tast and Fluster parser script calling from v4l2-decoder-conformance test.


a-wai (changes requested):

config/runtime/tests/v4l2-decoder-conformance-chromeos.jinja2
```diff
@@ -0,0 +1,59 @@
...
+              && cp remote_test_runner /usr/bin/remote_test_runner
+              && mkdir -p /usr/libexec/tast/bundles/remote/
+              && cp cros /usr/libexec/tast/bundles/remote/
+            - for i in $(seq 1 60); do ping -c 1 -w 1 $(lava-target-ip) && break || sleep 1; done
+            - ping -c 1 -w 1 $(lava-target-ip) || lava-test-raise "cros-device-unreachable"
+            - >-
+              ./ssh_retry.sh
+              -o StrictHostKeyChecking=no
+              -o UserKnownHostsFile=/dev/null
+              -i /home/cros/.ssh/id_rsa
```
pawiecz, line 57: From [changes to the pipeline config](https://github.com/kernelci/kernelci-pipeline/pull/705/files#diff-e055f21b05c11c155060b7dd48fa22812a879fada49f942fbf4368d9afbd5761R384) it looks like this TestJob Action should be ran on `bookworm-gst-fluster` rootfs.

When is the corresponding pubkey deployed to that image?

[I wasn't able to verify these changes](https://lava.collabora.dev/scheduler/job/15329364#bottom)

pawiecz (changes requested):

config/runtime/tests/v4l2-decoder-conformance-chromeos.jinja2
```diff
@@ -0,0 +1,59 @@
...
+{%- endfor %}
+{%- endif %}
+            - lava-test-set start setup
+            - >-
+              lava-test-case tast-tarball --shell
+              curl -s '{{ platform_config.params.tast_tarball }}'
+              \| tar xzvf -
+              && cp remote_test_runner /usr/bin/remote_test_runner
+              && mkdir -p /usr/libexec/tast/bundles/remote/
+              && cp cros /usr/libexec/tast/bundles/remote/
```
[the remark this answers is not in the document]
denisyuji: Thank you for the review @a-wai . No , these are not needed. removed it.

config/runtime/tests/v4l2-decoder-conformance-chromeos.jinja2
```diff
@@ -0,0 +1,59 @@
...
+            - for i in $(seq 1 60); do ping -c 1 -w 1 $(lava-target-ip) && break || sleep 1; done
+            - ping -c 1 -w 1 $(lava-target-ip) || lava-test-raise "cros-device-unreachable"
+            - >-
+              ./ssh_retry.sh
+              -o StrictHostKeyChecking=no
+              -o UserKnownHostsFile=/dev/null
+              -i /home/cros/.ssh/id_rsa
+              root@$(lava-target-ip)
+              python3 /usr/bin/fluster_parser.py {{ testsuite_arg }} {{ decoders_arg }} {{ videodec_timeout_arg }} {{ videodec_parallel_jobs_arg }}
```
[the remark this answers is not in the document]
denisyuji: Makes sense! I [splited it](https://github.com/kernelci/kernelci-core/pull/2617/commits/1ef7fe5603a59294bbe0d80e5516aceec49b05ff) and added the [shut down](https://github.com/kernelci/kernelci-core/pull/2617/commits/2b4e75839064ffafb251e55def86ddb1b4aa69ae)

config/runtime/tests/v4l2-decoder-conformance-chromeos.jinja2
denisyuji: Hi @pawiecz . Thank you for your review!
Yes, you are right, good catch. I removed the line which was wrongly using the debian rootfs. See https://github.com/kernelci/kernelci-pipeline/pull/705/files#diff-e055f21b05c11c155060b7dd48fa22812a879fada49f942fbf4368d9afbd5761

config/runtime/tests/v4l2-decoder-conformance-chromeos.jinja2
```diff
@@ -0,0 +1,59 @@
...
+    results:
+      location: /home/cros/lava
+    definitions:
+    - from: inline
+      name: {{ node.name }}
+      path: inline/{{ node.name }}.yaml
+      repository:
+        metadata:
+          format: Lava-Test Test Definition 1.0
+          name: cros-tast
```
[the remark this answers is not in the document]
denisyuji: @a-wai , thank you so much for your review! you are right. I fixed it. Please, see [config/runtime/tests/fluster-chromeos.jinja2](https://github.com/kernelci/kernelci-core/pull/2617/files#diff-54aeac138d72c9eb9e88b5179e2490cac5b4b75c3c8a139020ed9b3bf7ec6938)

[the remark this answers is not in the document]
denisyuji: Ps: I also renamed from `config/runtime/tests/v4l2-decoder-conformance-chromeos.jinja2` to `config/runtime/tests/fluster.jinja2`.

config/runtime/tests/fluster-chromeos.jinja2
```diff
@@ -0,0 +1,61 @@
...
+              root@$(lava-target-ip)
+              python3 /usr/bin/fluster_parser.py {{ testsuite_arg }} {{ decoders_arg }} 
+                            {{ videodec_timeout_arg }} {{ videodec_parallel_jobs_arg }}
+            - >-
+              ./ssh_retry.sh
+              -o StrictHostKeyChecking=no
+              -o UserKnownHostsFile=/dev/null
+              -i /home/cros/.ssh/id_rsa
+              root@$(lava-target-ip)
+              python3 /usr/bin/fluster_parser.py --results
```
a-wai, line 54: This must be executed *after* poweroff (which implies that the test results should be available locally, and fluster_parser as well): if the test fails, then fluster_parser will error out, ending the LAVA job immediately, resulting in a hard shutdown.

a-wai (changes requested):
Still a few changes needed; FWIW I prefer the approach I took with tast_parser:
* if executed with `--run`, it only runs the tests and doesn't parse the results
* if executed with `--results`, it only parses the results, assuming tests have run already
* if executed with neither of those, it falls back to the old behaviour: runs the tests *and* parses the results
 
This approach allows not modifying files from the legacy KernelCI, which is still running (hopefully not for long).

config/runtime/tests/fluster-chromeos.jinja2
denisyuji: Thank you @a-wai . See https://github.com/kernelci/kernelci-core/pull/2617/commits/8312746098685b34342f98b66e5316051e07efe2

denisyuji:
> Still a few changes needed; FWIW I prefer the approach I took with tast_parser:
> 
> * if executed with `--run`, it only runs the tests and doesn't parse the results
> * if executed with `--results`, it only parses the results, assuming tests have run already
> * if executed with neither of those, it falls back to the old behaviour: runs the tests _and_ parses the results
> 
> This approach allows not modifying files from the legacy KernelCI, which is still running (hopefully not for long).

@a-wai , makes sense, this is done!

denisyuji:
> Do we have this tests enabled in pipeline or in PR?

Hi @nuclearcat . Yes. This is tied to https://github.com/kernelci/kernelci-pipeline/pull/705/. I plan to merge both altogether

config/runtime/tests/fluster-chromeos.jinja2
```diff
@@ -45,17 +45,26 @@
...
-              ./ssh_retry.sh
+              scp
               -o StrictHostKeyChecking=no
               -o UserKnownHostsFile=/dev/null
               -i /home/cros/.ssh/id_rsa
-              root@$(lava-target-ip)
-              python3 /usr/bin/fluster_parser.py --results
+              root@$(lava-target-ip):/opt/fluster/results.xml /opt/fluster/
+            - >-
+              scp
```
pawiecz, line 57: Would it be useful to ship this tool in the TestJob environment by default, i.e. add it to the `kernelci/cros-tast` Docker container image?

Definitely out of scope for this PR but @denisyuji please share your thoughts

pawiecz (approved):
LGTM

config/runtime/tests/fluster-chromeos.jinja2
```diff
@@ -0,0 +1,70 @@
+{%- include 'chromeos/base.jinja2' %}
+
+{%- if testsuite -%}
```
pawiecz, line 3: Error message [in the relevant Node](https://staging.kernelci.org:9000/viewer?node_id=66d8184915e96f36b9e14b22) (`block sequence entries are not allowed in this context\n  in \"\", line 157, column 27`) was caused by right-stripping whitespace from the template - rendered TestJob definition looked as follows:

```
      password: "test0000"- test:
```

instead of:

```
      password: "test0000"
- test:
```

```suggestion
{%- if testsuite %}
```

(or just run `sed -i 's/-%}/%}/' config/runtime/tests/fluster-chromeos.jinja2` to get rid of all of them)

pawiecz (changes requested):

denisyuji:
Thank you @pawiecz . This is done.

config/runtime/tests/fluster-chromeos.jinja2
pawiecz: @denisyuji I probably should have made this more explicit: *all* `rstrips()` (dashes before closing tag: `-%}`) need to go away.

I dropped one in my code suggestion and advised to run `sed` on this file so that all of them would be removed.

pawiecz: This has been addressed by https://github.com/kernelci/kernelci-core/pull/2658

PR #2620, opened 2024-07-23
Enable LTP in New KernelCI

musamaanjum:


config/runtime/tests/ltp.jinja2
```diff
@@ -0,0 +1,15 @@
...
+    timeout:
+      minutes: {{ job_timeout }}
+    definitions:
+       - repository: https://github.com/kernelci/test-definitions
+         from: git
+         revision: kernelci.org
+         path: automated/linux/ltp/ltp.yaml
+         name: {{ node.name }}
+         parameters:
+            TST_CMDFILES: "{{ tst_cmdfiles }}"
```
a-wai, line 13:
```suggestion
            TST_CMDFILES: "{{ tst_cmdfiles|default('') }}"
```

Without this we're forced to add the `tst_cmdfiles`, while it can be optional IIUC

config/runtime/tests/ltp.jinja2
```diff
@@ -0,0 +1,15 @@
+
+- test:
+    name: {{ node.name }}
+    timeout:
+      minutes: {{ job_timeout }}
```
a-wai, line 5: Please provide a default timeout value here

a-wai (changes requested):

config/runtime/tests/ltp.jinja2
musamaanjum: Every LTP test specifies job_timeout. It is configurable on the test's end. That is why job_timeout is being used here.

musamaanjum: @a-wai Am I missing something?

musamaanjum: I understood from your other comment that we need to specify the default timeout value here. Updated.
> job_timeout should be an int (and the default value should actually be in the template IMHO).

config/runtime/tests/ltp.jinja2
```diff
@@ -0,0 +1,15 @@
+
```
a-wai, line 1: Please drop this blank line

a-wai (changes requested):
LGTM, just one minor nit left.

a-wai (approved):

PR #2621, opened 2024-07-23
Allow configurable job timeouts

### LINKED ISSUE #2576: Need a way to set the job timeout per job
In the legacy system, we could set the job timeout for each test case.

This is no longer the case as there was no urgent need to implement it so far. However, codec-related Tast tests on ChromeBooks can now take a huge amount of time:
* a bit less than 1 hour on [jacuzzi](https://lava.collabora.dev/scheduler/job/14182221)
* almost 2 hours on [cherry](https://lava.collabora.dev/scheduler/job/14204661)

This is way over our default 30 minutes timeout, which otherwise work fine for most jobs.

As a consequence, we should reinstate the ability to set different timeout values as job params.

Note: the 30 minutes timeout is currently hardcoded in both the [base LAVA template](https://github.com/kernelci/kernelci-core/blob/main/config/runtime/base/lava.jinja2#L38) (global job timeout) and the [tast tests template](https://github.com/kernelci/kernelci-core/blob/main/config/runtime/tests/tast.jinja2#L6) (test timeout); custom job timeouts should therefore affect both those definitions (and likely other test templates as well, for completeness).

a-wai:
A number of tests already use the `job_timeout` parameter to set a custom test timeout. This PR ensures all tests templates are aware of this parameter so we can use it more widely.

Additionally, it ensures the global timeout defined for the whole LAVA job is increased accordingly in case `job_timeout` exceeds the default value of 30 minutes.

Fixes #2576

### DIFF
--- a/config/runtime/tests/baseline.jinja2
+++ b/config/runtime/tests/baseline.jinja2
@@ -23,4 +23,4 @@
 {%- endfilter %}
 
     timeout:
-      minutes: 1
+      minutes: {{ job_timeout|default('1') }}
--- a/config/runtime/tests/sleep.jinja2
+++ b/config/runtime/tests/sleep.jinja2
@@ -1,6 +1,6 @@
 - test:
     timeout:
-      minutes: 5
+      minutes: {{ job_timeout|default('5') }}
     definitions:
     - repository: https://github.com/kernelci/kernelci-core
       branch: kernelci.org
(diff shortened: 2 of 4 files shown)


a-wai:
Example definition of a LAVA job with custom timeout (180): https://lava.collabora.dev/scheduler/job/15115477/definition

@nuclearcat can you please merge?

PR #2622, opened 2024-07-24
kernelci/runtime/lava.py: handle `incomplete` baseline tests

JenySadadia:
If `setup` suite fails for baseline test, mark the job node `fail` as it means device failed to boot.

### DIFF
--- a/kernelci/runtime/lava.py
+++ b/kernelci/runtime/lava.py
@@ -228,6 +228,14 @@ class Callback:
 
         child_nodes = self._get_results_hierarchy(results)
 
+        if job_result == 'incomplete' and 'baseline' in job_node['name']:
+            for child in child_nodes:
+                if child['node']['name'] == 'setup' and child['node']['result'] == 'fail':
+                    print(f"DEBUG: Setup failed for {job_node['id']}. Transitting job node \
+result: incomplete -> fail")
+                    job_result = 'fail'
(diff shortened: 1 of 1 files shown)


JenySadadia:
Tested OK on staging:
```
24/07/2024
15:54:16
DEBUG: Setup failed for 66a0d0e741e62083cda50e90. Transitting job node result: incomplete -> fail
```
Job node: https://staging.kernelci.org:9000/viewer?node_id=66a0d0e741e62083cda50e90

laura-nao (commented):
I wonder if this applies to `baseline` tests only. I guess other jobs have a `setup` test case that includes the kernel boot and login actions too, right? If so, maybe we can apply the same approach to every test and drop the `'baseline' in job_node['name']` constraint.

JenySadadia:
> I wonder if this applies to `baseline` tests only. I guess other jobs have a `setup` test case that includes the kernel boot and login actions too, right? If so, maybe we can apply the same approach to every test and drop the `'baseline' in job_node['name']` constraint.

I think this is specific to `boot` tests only.
For other tests such as `tast`, if the `setup` fails, it means the device couldn't boot and the main suite couldn't run. That's why the job node result should be `incomplete` instead. As it means the test run couldn't completed.

laura-nao (approved):

PR #2623, opened 2024-07-24
build-configs.yaml: Drop removed firmware file for SOF on MT8195

nfraprado:
The firmware mediatek/sof-tplg/sof-mt8195-mt6359-rt1019-rt5682-dts.tplg has been removed upstream [1]. The plan seems to be to use another firmware for MT8195 Tomato, but a kernel change [2] needs to land before that can happen. In the mean time, consider audio functionality for MT8195-Tomato broken, and drop the file from the config so we can build again. Fixes
https://github.com/kernelci/kernelci-pipeline/issues/726.

[1] https://lore.kernel.org/all/20240719022132.30990-1-xiangzhi.tang@mediatek.com/
[2] https://lore.kernel.org/all/20240318183004.3976923-1-cujomalainey@chromium.org/#t

### DIFF
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -368,7 +368,7 @@ fragments:
       mediatek/mt8192/scp.img
       mediatek/mt8195/scp.img
       mediatek/sof-tplg/sof-mt8186.tplg
-      mediatek/sof-tplg/sof-mt8195-mt6359-rt1019-rt5682-dts.tplg
+      mediatek/sof-tplg/sof-mt8195-mt6359-rt1019-rt5682.tplg
       mediatek/sof/sof-mt8186.ri
       mediatek/sof/sof-mt8195.ri
       mediatek/WIFI_MT7961_patch_mcu_1_2_hdr.bin


config/core/build-configs.yaml
```diff
@@ -368,7 +368,6 @@ fragments:
       mediatek/mt8192/scp.img
       mediatek/mt8195/scp.img
       mediatek/sof-tplg/sof-mt8186.tplg
-      mediatek/sof-tplg/sof-mt8195-mt6359-rt1019-rt5682-dts.tplg
```
a-wai, line 371: Do you expect we'll have to rely on `sof-mt8195-mt6359-rt1019-rt5682.tplg` once the driver support it? If that's the case, maybe simply replace the current filename with the non-dts one?

nfraprado: That's what [2] suggests, however I haven't had time to test locally and validate that topology actually works on this machine.

a-wai: Right, so there are 2 options here, I'll let you choose the one you prefer:
* include `sof-mt8195-mt6359-rt1019-rt5682.tplg` right now as a substitute for future use
* merge this as-is, in which case I'll let you create and self-assign a new issue as a reminder to look into it once [2] gets merged

a-wai (approved):

config/core/build-configs.yaml
nfraprado: I've updated to option 1 so I don't need to remember to update the firmware later.

a-wai (approved):

PR #2628, opened 2024-08-08
fix(runtime/lava): Present rendered job template if loading it fails

pawiecz:
```
while scanning a plain scalar
  in "<unicode string>", line 121, column 23
found a tab character that violates indentation
  in "<unicode string>", line 122, column 1
```

Not being able to reproduce this error locally provide more information from the running pipeline.

### DIFF
--- a/kernelci/runtime/lava.py
+++ b/kernelci/runtime/lava.py
@@ -300,7 +300,14 @@ class LAVA(Runtime):
             return None
 
         # yaml round-trip to process e.g. multi-line commands
-        return yaml.dump(yaml.load(rendered, Loader=yaml.CLoader))
+        try:
+            loaded = yaml.load(rendered, Loader=yaml.CLoader)
+        except yaml.scanner.ScannerError as exc:
+            print(f"Error loading rendered job template as YAML: {exc}," +
+                  f"{rendered}")
+            return None
+
+        return yaml.dump(loaded)
 
     def submit(self, job_path):
         with open(job_path, 'r', encoding='utf-8') as job_file:


nuclearcat:
Please check: https://github.com/kernelci/kernelci-core/pull/2626

pawiecz:
> Please check: #2626

Why didn't https://github.com/kernelci/kernelci-core/blob/4d95e4eef39e73f62d5fb586f0d66818818a8471/kernelci/runtime/lava.py#L296 catch the undefined variable during job template rendering? Shouldn't it prevent such a corruption, then return no job for submission and end up in: [`Failed to generate job definition, aborting...`](https://github.com/kernelci/kernelci-pipeline/blob/8fbb5cb79910cf93068e49b3e470c927edd632e4/src/scheduler.py#L149)?

nuclearcat:
Rendering was fine, but it rendered corrupted yaml file and crashed at: https://github.com/kernelci/kernelci-core/blob/4d95e4eef39e73f62d5fb586f0d66818818a8471/kernelci/runtime/lava.py#L303
Corrupted because:
```
      parameters:
        TESTPROG_URL: '{{ node.artifacts.kselftest_tar_gz }}'
        SKIPFILE: /dev/null
        TST_CMDFILES: '{{ collections }}'
        TST_CASENAME: '{{ tests }}'
        SKIP_INSTALL: True
	ENV: '{{ env }}'
```
env was not defined

pawiecz:
Thanks for locating the job template corruption! The `\t` character `yaml.loads()` complained about was in fact in the `ENV` line (`0x09`) while `env` not being defined still allowed template rendering (substituting it with an empty string):

```
$ hexdump -C <(tail -1 kselftest.jinja2)
00000000  09 45 4e 56 3a 20 27 7b  7b 20 65 6e 76 20 7d 7d  |.ENV: '{{ env }}|
00000010  27 0a                                             |'.|
00000012
```

nuclearcat:
Weird, i was unable to find this tab symbol using grep yesterday... need to figure out better ways.

PR #2629, opened 2024-08-09
kernelci.api: enable `promisc` flag for subscription

JenySadadia:
Update `API.Node.subscribe` method to provide `promisc` field while calling API `subscribe/` endpoint.


nuclearcat:
Thanks! Merging

PR #2638, opened 2024-08-15
config/core: rename `early-access` entries

JenySadadia:
Rename configurations for `early-access` to `production`. Early access instance was setup for the initial release phase. It's been used as production now.

### DIFF
--- a/config/core/api-configs.yaml
+++ b/config/core/api-configs.yaml
@@ -3,7 +3,7 @@ api:
   docker-host:
     url: http://172.17.0.1:8001
 
-  early-access:
+  production:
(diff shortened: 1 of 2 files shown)


JenySadadia:
@nuclearcat Should we also update this https://github.com/kernelci/kernelci-core/blob/main/config/runtime/base/kubernetes.jinja2#L58 ?

JenySadadia:
> Replace all the early-access mentions in docs

Thanks for pointing this out.
Created https://github.com/kernelci/kernelci-project/pull/488 and https://github.com/kernelci/kernelci-project/pull/489

PR #2640, opened 2024-08-16
rootfs/bookworm-gst-fluster: Rework error handling for missing test suites

pawiecz:
This patch changes missing test suite error handling to:
- fail (with message) if no test suite was passed to downloader function
- allow readlink to fail and then handle its error with debugging message


nuclearcat (approved):
LGTM

PR #2641, opened 2024-08-20
config: runtime: boot: u-boot: use `nfsroot` if defined

a-wai:
Device/job definitions can specify a NFS rootfs URL as the `nfsroot` param. However, this one is currently ignored when using the `u-boot` boot method. This patch ensures the `nfsroot` param is used when booting with an NFS rootfs, falling back to the default/hardcoded value only if this param is missing.

Needed for https://github.com/kernelci/kernelci-pipeline/pull/700


pawiecz (approved):
LGTM

a-wai:
Thanks for the approval!

Note: this is completely untested, so let's wait for the next staging deployment before merging ;)

a-wai:
Succesfully tested on staging:
* job with default rootfs: https://lava.collabora.dev/scheduler/job/15283435/definition
* overridden rootfs URL: https://lava.collabora.dev/scheduler/job/15283431/definition

PR #2645, opened 2024-08-21
build-configs.yaml: Add MT6359_AUXADC config to arm64-chromebook

nfraprado:
Enable the CONFIG_MEDIATEK_MT6359_AUXADC config option (which provides the mt6359-auxadc driver) in the arm64-chromebook fragment. This makes the test case /soc/pwrap@10024000/pmic/adc in the DT kselftest pass on the mediatek chromebooks, which has been failing in linux-next since the addition of commit b0a4ce81f327 ("arm64: dts: mediatek: Add ADC node on MT6357, MT6358, MT6359 PMICs") in the kernel.


nfraprado:
Before (2 failures): https://lava.collabora.dev/scheduler/job/15253895
After (1 failure): https://lava.collabora.dev/scheduler/job/15259079

nuclearcat:
Thanks!

PR #2646, opened 2024-08-22
feat(kbuild): Add kselftest node

nuclearcat:
Often we have kselftest build failing, but it doesn't register as failure anywhere. Now we will add kbuild sub-node with kselftest suffix to indicate success or failure of kselftest build.

### DIFF
--- a/kernelci/kbuild.py
+++ b/kernelci/kbuild.py
@@ -871,6 +871,13 @@ class KBuild():
 
         # TODO(nuclearcat):
         # Add child_nodes for each sub-step
+        # do we have kselftest_tar_gz in artifact keys? then node is ok
+        kselftest_result = 'fail'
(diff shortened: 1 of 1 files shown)


JenySadadia:
This is setting node `id` and `parent` with the same id.
Such as https://staging.kernelci.org:9000/viewer?node_id=66c72dcd7c8a38c2eb18a604
That's breaking kcidb bridge while getting test data.

nuclearcat:
I fixed that recently

JenySadadia:
> I fixed that recently. I guess we need to add safeguards in API as well, and few other places, to not allow parent to be pointed to itself.

Thanks. Yes, I agree.

nuclearcat:
Tested on staging, works

PR #2647, opened 2024-08-22
feat(job): Add compat field to job

nuclearcat:
Compat field used to report to kcidb parameters/info about device.


JenySadadia:
This is already covered in https://github.com/kernelci/kernelci-pipeline/pull/762

PR #2655, opened 2024-09-04
feat(models): Add treeid field

nuclearcat:
For easier tree tracking we introduce unique tree identified, treeid, which is sha256 hash of (treeurl + branch + time).

### DIFF
--- a/kernelci/api/models.py
+++ b/kernelci/api/models.py
@@ -212,6 +212,9 @@ class Node(DatabaseModel):
     submitter: Optional[str] = Field(
         description="Token md5 hash to identify node origin(submitter token)"
     )
+    treeid: Optional[str] = Field(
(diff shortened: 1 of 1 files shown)


nuclearcat:
Note, production will require updates in mongo over mongo cli:
```
db.node.createIndex({ treeid: 1 }, { sparse: true })
const batchSize = 1000;
let updatedCount = 0;

do {
  const result = db.node.updateMany(
    { treeid: { $exists: false } }, 
    { $set: { treeid: null } },
    { limit: batchSize }
  );
  updatedCount = result.matchedCount;
} while (updatedCount === batchSize);
```

JenySadadia:
This PR title needs correction:
`s/submitter/treeid`

nuclearcat:
Thanks Jeny, fixed

PR #2659, opened 2024-09-06
fix(api/helper): Fix child nodes fields treeid and submitter

nuclearcat:
treeid and submitter should be borrowed from parent node


JenySadadia (commented):
Thanks for the fix.
I am wondering if we also need to tweak `submit_results` method accordingly.

nuclearcat:
We can check current results, and if data missing - yes

PR #2660, opened 2024-09-09
config: runtime: tests: Use absolute path to ssh_retry.sh script

pawiecz:


denisyuji (approved):

nuclearcat (approved):
LGTM, tast tests seems still working

pawiecz:
I won't be able to amend merged patch but let me at least provide a bit of context here:

1. LAVA runs the test execution environment [as follows](https://lava.collabora.dev/scheduler/job/15540614#L4858):

```
docker run --interactive --tty --rm --init --name=lava-docker-test-shell-15540614-6.3 --mount=type=bind,source=/var/lib/lava/dispatcher/tmp/15540614/lava-create-overlay-go7jp1d8/lava-15540614,destination=/lava-15540614 '--env=PS1=docker-test-shell:$ ' kernelci/cros-tast bash --norc -i
```

(here the `ssh_retry.sh` would work properly)

```
docker-test-shell:$ pwd
/home/cros
```

2. Next LAVA applies [its environment](https://lava.collabora.dev/scheduler/job/15540614#L4873)

```
. /lava-15540614/environment
```

which also changes CWD (e.g. to `/lava-15540614/0/tests/0_fluster-chromeos-av1`)

This PR was a quick approach to fixing it, definitely not the best one. This script could be put into one of $PATH dirs or $PATH itself could be amended to lookup this script but that would either pollute the TestJob definition or require rootfs rebuild.

PR #2661, opened 2024-09-09
fix(kbuild): Detect if module dont have modules support

nuclearcat:
Right now we rely on config option, which turns to be too much of burden to maintain. Let's detect lack of modules support dynamically.


nuclearcat:
Tested on staging, logic working as expected. Now will merge and cleanup custom config option i created before,as it is not needed anymore

PR #2664, opened 2024-09-10
fix(README): Remove obsolete information from README

nuclearcat:
This README contains lot of invalid info, relevant to old website, legacy infrastructure and etc.
For now truncate it to minimum, leaving only valid information.


nuclearcat:
Thanks!

PR #2665, opened 2024-09-11
config: rt-tests: Add default timeout and duration values

musamaanjum:
Add default timeout value to template and set it to 10 minutes. Also increase the duration of test to 10 minutes.

Close https://github.com/kernelci/kernelci-project/issues/417

### DIFF
--- a/config/runtime/tests/rt-tests.jinja2
+++ b/config/runtime/tests/rt-tests.jinja2
@@ -1,7 +1,7 @@
 - test:
     name: {{ node.name }}
     timeout:
-      minutes: {{ job_timeout }}
+      minutes: {{ job_timeout|default('600s') }}
 
     definitions:
     - repository: https://github.com/kernelci/test-definitions.git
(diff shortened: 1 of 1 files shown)


config/runtime/tests/rt-tests.jinja2
```diff
@@ -13,7 +13,7 @@
         ARTIFACTORIAL_TOKEN: {{ artifactorial_token|default('') }}
         ARTIFACTORIAL_URL: {{ artifactorial_url|default('') }}
         AFFINTIY: {{ affinity|default('0-1') }}
-        DURATION: {{ duration|default('300s') }}
+        DURATION: {{ duration|default('600s') }}
```
musamaanjum, line 16: @nuclearcat If the timeout is set to 10 minutes and the test duration is also the same, wouldn't it cause the forceful termination of the tests? Should I decrease the duration to a lower value? What value would be best here?

nuclearcat: I think it should be fine, total job value is safeguard that job doesnt occupy chromebook for too long time, so it should be fine. More comprehensive approach will be estimating if there is any other tasks before and after this timeout, then subtracting their average duration from this 600s.

nuclearcat:
LGTM, merging

PR #2668, opened 2024-09-11
config: rt-tests: fix MLOCKALL getting set to True

musamaanjum:
When MLOCKALL is set to 'true', the lava job has MLOCKALL becomes true [1] which in turn translates to "True" while job execution [2]. It causes error as the test expects true instead of True.

  ./pi-stress.sh -D 60s -m True -r false -w hackbench
  ./pi-stress.sh: 37: True: not found

Fix it by using double quotes instead.

[1] https://lava.collabora.dev/scheduler/job/15562666/definition#defline43
[2] https://lava.collabora.dev/scheduler/job/15562666#L220

### DIFF
--- a/config/runtime/tests/rt-tests.jinja2
+++ b/config/runtime/tests/rt-tests.jinja2
@@ -25,7 +25,7 @@
         PRIORITY: {{ priority|default('98') }}
         THREADS: {{ threads|default('2') }}
       {% elif tst_cmd == 'pi-stress' %}
-        MLOCKALL: {{ mlockall|default('true') }}
+        MLOCKALL: {{ mlockall|default('"true"') }}
       {% elif tst_cmd == 'rt-migrate-test' %}
         PRIORITY: {{ priority|default('90') }}
       {% elif tst_cmd == 'signaltest' %}


musamaanjum:
Let's confirm from the execution results if this works.

musamaanjum:
This fix didn't work: https://lava.collabora.dev/scheduler/job/15747114

@pawiecz please can you help solve this issue?

musamaanjum:
> This fix didn't work: https://lava.collabora.dev/scheduler/job/15747114
> 
> @pawiecz please can you help solve this issue? In job (https://lava.collabora.dev/scheduler/job/15747239), I've manually added double quotes around true and job worked fine.

@pawiecz Please have a look.

config/runtime/tests/rt-tests.jinja2
```diff
@@ -25,7 +25,7 @@
         PRIORITY: {{ priority|default('98') }}
         THREADS: {{ threads|default('2') }}
       {% elif tst_cmd == 'pi-stress' %}
-        MLOCKALL: {{ mlockall|default('true') }}
+        MLOCKALL: {{ mlockall|default("true") }}
```
pawiecz, line 28:
```suggestion
        MLOCKALL: {{ mlockall|default('"true"') }}
```

Verified with:
```
$ ./kci job generate --api staging -c ../rampup/kernelci-pipeline/config --runtime=lava-collabora 66ec2b36ebd9a94074b9dfc9 | lavacli jobs submit -
```
[15790058](https://lava.collabora.dev/scheduler/job/15790058)

Why?
1. First set of quotations allows Jinja template engine to treat it as string
2. Second set is for LAVA to also treat is as string and prevent YAML parser from making it a bool

Why does it even matter?
1. Test runner script [expects command name to run in a condition check](https://github.com/Linaro/test-definitions/blob/6c6ed7499e40c29da7dabfa2149350d13fb6496d/automated/linux/pi-stress/pi-stress.sh#L37)
2. This test runner script should be rewritten so that it converts truthy/falsy arguments and doesn't allow external commands to control its flow - it's a vector for command injection from bad actors

Bottom line:
1. Double quotations will allow this suite to be executed
2. Long-term solution would be fixing upstream test runner script

pawiecz (changes requested):

pawiecz:
Also: timeout for this test suite has to be extended to provide some buffer for initialization - it shouldn't be equal to the requested test duration.

config/runtime/tests/rt-tests.jinja2
musamaanjum: @pawiecz Thanks a lot for taking a look and fixing it. I couldn't figure out the parsing.

Created: https://github.com/Linaro/test-definitions/issues/532

musamaanjum:
> Also: timeout for this test suite has to be extended to provide some buffer for initialization - it shouldn't be equal to the requested test duration.

@pawiecz we had probably discussed this in some meetings that it is okay to have the same timeouts. Or I may be mistaken. Currently, the total test duration is set to 10 minutes and the test's execution is also set to 10 minutes. Should we reduce the test duration by 30 or 60 seconds?

musamaanjum:
Result: https://lava.collabora.dev/scheduler/job/15790240

@nuclearcat let's merge this PR.

pawiecz:
> > Also: timeout for this test suite has to be extended to provide some buffer for initialization - it shouldn't be equal to the requested test duration.
> 
> @pawiecz we had probably discussed this in some meetings that it is okay to have the same timeouts. Or I may be mistaken. Currently, the total test duration is set to 10 minutes and the test's execution is also set to 10 minutes.

Let me give an example:

- timeout "countdown" starts when [TestAction](https://lava.collabora.dev/scheduler/job/15790058#action_3-1) begins
- test execution starts after:
  - logging in
  - setting the environment
  - [triggering test runner script](https://lava.collabora.dev/scheduler/job/15790058#L11159)

Due to these slight delays [it might not be able to finish successfully](https://lava.collabora.dev/scheduler/job/15790058#L11164). Providing a slight margin (e.g. 11 minutes instead of 10 for a test expected to go for 10 minutes) could reduce flakiness

> Should we reduce the test duration by 30 or 60 seconds?

I'm not sure how long this test case should run to provide meaningful results :cry:

pawiecz (approved):
LGTM

(Setting right timeouts is a separate item)

PR #2669, opened 2024-09-12
Add Debian bookworm Tast image

10ne1:
This image can be used to run Tast tests on Debian.

It can be built for both arm64 and amd64 for use on chromebooks.

[Here is an example LAVA job](https://lava.collabora.dev/scheduler/job/15648037) successfully running a video codec test by downloading these images from my webserver.

The built binary images can be found on oasis under `/home/aratiuR/workspace/kernelci-core/temp/_install_/bookworm-tast/`


nuclearcat:
Thanks!

PR #2658, opened 2024-09-05
v4l2:fluster: Fix config/runtime/tests/fluster-chromeos.jinja2

pawiecz:
Fix remaining right-stripping white space from the template and line breaking (details in the commit messages).

PR #2670, opened 2024-09-13
fluster-chromeos: Re-mount read-only filesystem and small fixes

denisyuji:
(This PR depends on the merging of https://github.com/kernelci/kernelci-core/pull/2658 first)

Add the following commands to be executed on DUT:

- Execute `mount -o remount,rw /` to re-mount the DUT filesytem in read-write mode. Having the filesystem in read-only mode is causing error when writing the `results.xml` file.
- Execute `cat /etc/os-release > /tmp/osrel.tmp` in loop until the `ssh` host to be available.
- Add the `--run` flag on the `fluster_parser.py` execution to follow the current struct of the program in the repo. The
execution is expected to fail if the `fluster_parser.py` version on the DUT is not updated.

In addition, install the dependency `python3-junitparser` to parse `fluster_parser.py` results.

### DIFF
--- a/config/runtime/tests/fluster-chromeos.jinja2
+++ b/config/runtime/tests/fluster-chromeos.jinja2
@@ -2,16 +2,16 @@
 
 {%- if testsuite %}
     {% set testsuite_arg = "-ts " + testsuite %}
-{%- endif -%}
-{%- if decoders -%}
-    {%- set decoders_arg = "-d " + decoders|join(" ") -%}
-{%- endif -%}
-{%- if videodec_timeout -%}
+{%- endif %}
+{%- if decoders %}
+    {%- set decoders_arg = "-d " + decoders|join(" ") %}
+{%- endif %}
+{%- if videodec_timeout %}
     {% set videodec_timeout_arg = "-t " + videodec_timeout|string %}
-{%- endif -%}
-{%- if videodec_parallel_jobs -%}
+{%- endif %}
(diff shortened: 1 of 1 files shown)


a-wai:
> Execute mount -o remount,rw / to re-mount the DUT filesytem in read-write mode. Having the filesystem in read-only mode is causing error when writing the results.xml file.

This is risky, as we already had many issues with FS corruption on Chromebooks. It'd be best to ensure the results file is in a writable location (such as /tmp)

nuclearcat:
> > Execute mount -o remount,rw / to re-mount the DUT filesytem in read-write mode. Having the filesystem in read-only mode is causing error when writing the results.xml file.
> 
> This is risky, as we already had many issues with FS corruption on Chromebooks. It'd be best to ensure the results file is in a writable location (such as /tmp)

+1
@10ne1 this is what i mentioned in our discussion. Because if filesystem become corrupted, we will have very painful and long rootfs reflashing (and its big burden in maintenance)

a-wai:
> /tmp could work or even one of the RW data partitions.

`/tmp` would be best, IMHO the less we write to internal storage, the better ;)

denisyuji:
That makes sense. I will create a new pull request to use `/tmp` instead of re-mounting the filesystem.

denisyuji:
I will create a new pull request to use /tmp instead of re-mounting the filesystem.

PR #2673, opened 2024-09-13
fluster: Use /tmp as result directory and small fixes

denisyuji:
This commit will:

- Replace the result directory from `/opt/fluster` to `/tmp`. Chromeos uses a read-only filesystem and the program is failing when writing on the previous directory.
- Uses `wget` to fetch the latest version of `fluster_parser.py` from the kernelci-core repo
- Execute `cat /etc/os-release > /tmp/osrel.tmp` in loop to wait the `ssh` host to be available.
- Add the `--run` flag on the `fluster_parser.py` execution to follow the current struct of the program in the repo. Execute the same script with the `--results` flag to collect the results.
- Install the dependency `python3-junitparser` to parse `fluster_parser.py` results.

### DIFF
--- a/config/runtime/tests/fluster-chromeos.jinja2
+++ b/config/runtime/tests/fluster-chromeos.jinja2
@@ -36,28 +36,36 @@
             - lava-test-set start setup
             - for i in $(seq 1 60); do ping -c 1 -w 1 $(lava-target-ip) && break || sleep 1; done
             - ping -c 1 -w 1 $(lava-target-ip) || lava-test-raise "cros-device-unreachable"
+            - >-
+            ./ssh_retry.sh
+            -o StrictHostKeyChecking=no
+            -o UserKnownHostsFile=/dev/null
+            -i /home/cros/.ssh/id_rsa
+            root@$(lava-target-ip)
+            cat /etc/os-release > /tmp/osrel.tmp
+            - cat /tmp/osrel.tmp
             - lava-test-set stop setup
+            - wget https://github.com/kernelci/kernelci-core/raw/main/config/rootfs/debos/overlays/fluster/usr/bin/fluster_parser.py
+            - >-
+              scp
+              -o StrictHostKeyChecking=no
+              -o UserKnownHostsFile=/dev/null
+              -i /home/cros/.ssh/id_rsa
+              ./fluster_parser.py root@$(lava-target-ip):/tmp
             - >-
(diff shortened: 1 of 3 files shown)


config/runtime/tests/fluster-chromeos.jinja2
```diff
@@ -66,4 +74,5 @@
               -i /home/cros/.ssh/id_rsa
               root@$(lava-target-ip)
               poweroff && sleep 30 || true
-            - python3 /usr/bin/fluster_parser.py --results
```
nuclearcat, line 69: You are calling this commands after issuing poweroff

a-wai: @nuclearcat poweroff is issued on the DUT, this command runs on lava/docker

config/runtime/tests/fluster-chromeos.jinja2
```diff
@@ -66,4 +74,5 @@
...
+            - apt install -y python3-junitparser
```
a-wai, line 77: Please have this package installed directly in the docker image. Otherwise this will run for each job, wasting both time and bandwidth.

config/runtime/tests/fluster-chromeos.jinja2
nuclearcat: Ah ok then, i guess it should not disrupt anything.

config/runtime/tests/fluster-chromeos.jinja2
```diff
@@ -36,28 +36,36 @@
...
+            - >-
+            ./ssh_retry.sh
+            -o StrictHostKeyChecking=no
+            -o UserKnownHostsFile=/dev/null
+            -i /home/cros/.ssh/id_rsa
+            root@$(lava-target-ip)
+            cat /etc/os-release > /tmp/osrel.tmp
+            - cat /tmp/osrel.tmp
             - lava-test-set stop setup
+            - wget https://github.com/kernelci/kernelci-core/raw/main/config/rootfs/debos/overlays/fluster/usr/bin/fluster_parser.py
```
a-wai, line 48: Please add this file to the docker image to avoid useless downloads in jobs.

a-wai:
@nuclearcat this wasn't fit for merging, see my comments ;)

@denisyuji please follow-up with fixes addressing my comments...

PR #2676, opened 2024-09-16
feat(docker/cros-tast): Add package to cros-tast image

nuclearcat:
Right now we are installing python3-junitparser by apt, it is better to pre-build it with docker image.


a-wai:
LGTM, please add `fluster_parser` as well (see https://github.com/kernelci/kernelci-core/pull/2673#discussion_r1760774951)

a-wai (approved):

PR #2683, opened 2024-09-19
feat(runtime/lava.py): Add more LAVA submit error logging

nuclearcat:
Sometimes LAVA doesn't like job, but HTTP error code not enough, to understand why. Let's log also response.

### DIFF
--- a/kernelci/runtime/lava.py
+++ b/kernelci/runtime/lava.py
@@ -342,6 +342,8 @@ class LAVA(Runtime):
         resp = self._server.session.post(
             jobs_url, json=job_data, allow_redirects=False
         )
+        if resp.status_code >= 400:
+            print(f"Error submitting job: {resp.status_code}, {resp.text}")
         resp.raise_for_status()
         return resp.json()['job_ids'][0]
 


kernelci/runtime/lava.py
```diff
@@ -342,6 +342,8 @@ def _submit(self, job):
         resp = self._server.session.post(
             jobs_url, json=job_data, allow_redirects=False
         )
+        if resp.status_code >= 400:
```
JenySadadia, line 345: I'd suggest using a specific code i.e. for a successful response to use here to cover the rest of the codes.
I believe we receive HTTP 200 on success?
```suggestion
        if resp.status_code != 200:
```

JenySadadia: > It might be 201, 202, which are valid too. Error codes are starting from 400.

Yes, that's true. I am not aware of different HTTP codes returned by LAVA jobs, but I thought it might be just one code for successful submission.

nuclearcat:
Tested on staging:
```
kernelci-pipeline-scheduler-lava-5637b705932c0d321e152d0d063b1de9451a5bdaa629f1cd2e2fcf6ea76ebc62.log:Error submitting job: 400, {"message":"Devices unavailable: Device type 'qemu-x86' is unavailable."}
kernelci-pipeline-scheduler-lava-5637b705932c0d321e152d0d063b1de9451a5bdaa629f1cd2e2fcf6ea76ebc62.log:Error submitting job: 403, {"detail":"Invalid token."}
kernelci-pipeline-scheduler-lava-5637b705932c0d321e152d0d063b1de9451a5bdaa629f1cd2e2fcf6ea76ebc62.log:Error submitting job: 403, {"detail":"Invalid token."}
kernelci-pipeline-scheduler-lava-7f44fab62fbbb2a9556e86feca0824d1b35631fa9f63bf5fe4126aac1008ddb2.log:Error submitting job: 403, {"detail":"Invalid token."}
```

PR #2687, opened 2024-09-22
tast-on-debian: Create a new test method for Tast on Debian fs

denisyuji:
Linked to https://github.com/kernelci/kernelci-pipeline/pull/803

Add support for `Tast` tests on Debian. Specifically:

- Create a template `tast-debian.jinja2` based on `tast.jinja2` to execute Tast against DUT running Debian root filesystem.
- Add the option `-checktestdeps=false` for `tast_parser.py` to allow the execution on non-chromeos DUTs.
- Hack: Create a shell script utility named `fake_initctl.sh` to simulate the required `initctl` command from ChromeOs on Debian.
- Hack: Create a shell script utility named `fake_crossystem.sh` to simulate the required `crossystem` command from ChromeOs on Debian.

### DIFF
--- a/config/docker/data/tast_parser.py
+++ b/config/docker/data/tast_parser.py
@@ -87,6 +87,7 @@ def run_tests(args):
         'run',
         f'-resultsdir={RESULTS_DIR}',
         '-sysinfo=false',
+        '-checktestdeps=false',
         '-build=false',
         remote_ip
     ]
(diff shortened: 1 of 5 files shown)


nuclearcat:
There is a lot of things retrieved from internet, they should be pre-added to docker image instead (especially google cloud cli over 100MB), for two reasons - reduce traffic to LAVA, and make tests reproducible and not hard depending on content in internet.

PR #2696, opened 2024-10-01
config: rt-tests: reduce the duration of the test

musamaanjum:
The timeout of the job is set to 10 minutes. We should run the test for less than that. Reduce the total duration of the test to 540 seconds. Leave 60 seconds for logging in, setting the environment and triggering the test runner script.

We wanted to have total time of this job to be 10 minutes and run the test for as much duration as possible. So reducing to 540 seconds or 9 minutes is okay.


musamaanjum:
@pawiecz Thanks for explaining [here](https://github.com/kernelci/kernelci-core/pull/2668#issuecomment-2376748179)

musamaanjum:
@nuclearcat @pawiecz please can we merge this PR? Only the timeout is being reduced.

pawiecz (approved):
LGTM

PR #2697, opened 2024-10-02
CHROMEOS: update R124 xml to include gstreamer matroska fix

10ne1:
Further testing in KernelCI revealed we're also missing the matroska plugin, so this updates to a commit which enables it:

https://gitlab.collabora.com/chromiumos-kernelci/chromiumos-overlay/-/commit/e97d251b1c06bbbb370aa2e9718a3709bc7daf1b


laura-nao:
> btw @laura-nao does this missing matroska plugin affect only cherry? or are there any other boards affected, so we know which ones to reflash?

All the platforms we run the fluster tests on are affected, as we use the same VP9 test vectors on all of them (and the majority of test vectors in the VP9 suite are in webm format). So we'll need to reflash trogdors (both lazor and kingoftown), asurada, cherry, jacuzzi and corsola.

laura-nao (approved):

nuclearcat:
Thanks a lot, i will proceed with builds and reflashing shortly. Merging now.

PR #2699, opened 2024-10-08
fix(kbuild): On kbuild failures generate failure for node

nuclearcat:
Generate incomplete state with proper error message for easier diagnostics, for example missing fragment.


nuclearcat:
tested on staging

PR #2700, opened 2024-10-09
feat(helper/models): Add platformfilter

nuclearcat:
Add filter that allows to restrict tests on specific platform. This feature required for proper bisection implementation, as some tests might be scheduled on multiple platforms.

### DIFF
--- a/kernelci/api/helper.py
+++ b/kernelci/api/helper.py
@@ -385,26 +385,26 @@ class APIHelper:
                         runtime=None, platform=None):
         """Create a new job node based on input and configuration"""
         jobfilter = input_node.get('jobfilter')
+        platform_filter = input_node.get('platform_filter')
         treeid = input_node.get('treeid')
         submitter = input_node.get('submitter')
-        try:
-            job_node = {
-                'kind': job_config.kind,
-                'parent': input_node['id'],
-                'name': job_config.name,
-                'path': input_node['path'] + [job_config.name],
-                'group': job_config.name,
-                'artifacts': {},
-                'jobfilter': jobfilter,
-                'treeid': treeid,
-                'submitter': submitter,
-                'data': {
-                    'kernel_revision': input_node['data']['kernel_revision'],
-                },
-            }
-        except KeyError as error:
-            print(f"Missing key {error} in input node")
-            return None
+        job_node = {
+            'kind': job_config.kind,
+            'parent': input_node['id'],
+            'name': job_config.name,
+            'path': input_node['path'] + [job_config.name],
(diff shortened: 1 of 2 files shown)


kernelci/api/models.py
```diff
@@ -203,6 +203,9 @@ class Node(DatabaseModel):
     jobfilter: Optional[List[str]] = Field(
         description="Restrict jobs that can be scheduled by this node"
     )
+    platformfilter: Optional[List[str]] = Field(
```
JenySadadia, line 206: I'd suggest using `_` between words for better readability as per python naming conventions.
`platform_filter`.

JenySadadia, line 206: I'd suggest using `_` between words for better readability as per python naming conventions.
`platform_filter`.

nuclearcat: Done. I will prefer to keep in json node consistent with previous styling, without underscore, and i will rename later also in code treeid and jobfilter as well in python code (keeping in json same, as migration is not working at moment).

kernelci/api/helper.py
```diff
@@ -385,26 +385,26 @@ def create_job_node(self, job_config, input_node,
                         runtime=None, platform=None):
         """Create a new job node based on input and configuration"""
         jobfilter = input_node.get('jobfilter')
+        platform_filter = input_node.get('platformfilter')
```
JenySadadia, line 388:
```suggestion
        platform_filter = input_node.get('platform_filter')
```

nuclearcat: It is already correct, because inside node, when it is decoded from json to dict, it will be available by key platformfilter, but variable names according to PEP8 is platform_filter.

JenySadadia: But I see you have already changed the model field name in `kernelci.api.models` to `platform_filter`.
And the same will be used in the `input_node`.

nuclearcat: I used alias parameter, so while python variable is PEP8, field name in json is same as jobfilter.
https://docs.pydantic.dev/latest/concepts/fields/#field-aliases
I am not entirely sure, but it should work.

JenySadadia: Is this tested OK?
I don't think that field alias name would work as we use `allow_population_by_field_name` in model config option that would use field name instead of alias while getting or sending the object to DB.

nuclearcat: Yes, it seems it will use platform_filter due this parameter. I guess i will have to live with that and discrepancy in field names, we will migrate jobfilter later as well.

PR #2702, opened 2024-10-10
feat(kbuild): Add kselftest disable option

nuclearcat:
Some kernel builds does not require to build kselftest.


nuclearcat:
Depends on https://github.com/kernelci/kernelci-pipeline/pull/835

PR #2705, opened 2024-10-11
fix(docker/k8s): Fix image naming

nuclearcat:
We had previously hack with hardcoding ghcr.io.
Now we make it more flexible, so people can set their own image url, and our staging instance automatically add staging- prefix in appropriate place.


pawiecz (approved):
As discussed offline I'd prefer static configuration but this also works.

LGTM

PR #2708, opened 2024-10-14
feat(runtime/lava): Add priority for manual submits

nuclearcat:
Manually submitted jobs need slightly better priority to complete faster.


nuclearcat:
Verified on staging

PR #2713, opened 2024-10-18
feat(storage/azure.py): Add content_type option for .gz files

nuclearcat:
As requested by kcidb team, it will be easier to handle files if they have proper content-type set.


nuclearcat:
Tested on staging

PR #2721, opened 2024-10-25
fluster-on-chromeos: Simplify ssh between client and dut

denisyuji:
Simplify `ssh` connection between client and DUT by setting the `/root/.ssh/config` to avoid using `-o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i /home/cros/.ssh/id_rsa` flags on every connection.
Keeps only the first `ssh` connection flags which waits until the DUT to be available.


nuclearcat:
Is  there any job example where it is tested?

nuclearcat:
ping

denisyuji:
Hi @nuclearcat , sorry, lost this one. Sure! See: https://lava.collabora.dev/scheduler/job/16551784

PR #2722, opened 2024-10-28
feat(models.py): Add EventHistory collection model

nuclearcat:
Add model with index that will cause documents to expire.


nuclearcat:
tested on staging

PR #2729, opened 2024-11-04
fix(kbuild): Do not build kselftest if not set in jobfilter

nuclearcat:
To speed up bisection do not build kselftest if it is not set in jobfilter.

### DIFF
--- a/kernelci/kbuild.py
+++ b/kernelci/kbuild.py
@@ -156,6 +156,10 @@ class KBuild():
                 self._kfselftest = False
             else:
                 self._kfselftest = True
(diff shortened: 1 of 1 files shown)


nuclearcat:
I need to implement also auto-enabling kbuild selftest if in jobfilter present any test that require kselftest, otherwise this will create more confusion. This feature requires more logic in kernelci-pipeline

nuclearcat:
Converted into an issue for proper design (auto-enabling kselftest based on requested tests, which needs logic in kernelci-pipeline). Tracking in #3127.

Issue #2730, opened 2024-11-04
Builds failing on ChromeOS config retrieval


nuclearcat:
Log snippet:
```
2024-11-04 19:27:04.817173 kci-672903754fd588c32f3690b4-kbuild-gcc-12-arm64-chro-j6im2j6nf default Handling uncaught exception: HTTPSConnectionPool(host='chromium.googlesource.com', port=443): Max retries exceed
ed with url: /chromiumos/third_party/kernel/+archive/refs/heads/chromeos-6.6/chromeos/config.tar.gz (Caused by NameResolutionError("<urllib3.connection.HTTPSConnection object at 0x7fe96a0f3390>: Failed to resolv
e 'chromium.googlesource.com' ([Errno -3] Temporary failure in name resolution)"))
2024-11-04 19:27:04.817199 kci-672903754fd588c32f3690b4-kbuild-gcc-12-arm64-chro-j6im2j6nf default Exception type: ConnectionError
2024-11-04 19:27:04.817222 kci-672903754fd588c32f3690b4-kbuild-gcc-12-arm64-chro-j6im2j6nf default Traceback (most recent call last):
2024-11-04 19:27:04.817244 kci-672903754fd588c32f3690b4-kbuild-gcc-12-arm64-chro-j6im2j6nf default   File "<string>", line 158, in <module>
2024-11-04 19:27:04.817268 kci-672903754fd588c32f3690b4-kbuild-gcc-12-arm64-chro-j6im2j6nf default   File "<string>", line 146, in main
2024-11-04 19:27:04.817305 kci-672903754fd588c32f3690b4-kbuild-gcc-12-arm64-chro-j6im2j6nf default   File "/usr/local/lib/python3.11/dist-packages/kernelci/kbuild.py", line 573, in write_script
2024-11-04 19:27:04.817355 kci-672903754fd588c32f3690b4-kbuild-gcc-12-arm64-chro-j6im2j6nf default   File "/usr/local/lib/python3.11/dist-packages/kernelci/kbuild.py", line 543, in _generate_script
2024-11-04 19:27:04.817407 kci-672903754fd588c32f3690b4-kbuild-gcc-12-arm64-chro-j6im2j6nf default   File "/usr/local/lib/python3.11/dist-packages/kernelci/kbuild.py", line 512, in _merge_frags
[...]

PR #2731, opened 2024-11-04
fix(kbuild): Implement file download retry

nuclearcat:
kbuild process prone to have networking issues due way k8s works. Implement proper retry.

Fixes: https://github.com/kernelci/kernelci-core/issues/2730


nuclearcat:
Tested on staging

PR #2735, opened 2024-11-06
fix(kbuild): Implement command retry to firmware clone

nuclearcat:
As often firmware git clone fails - implement retry, so build wont fail.


nuclearcat:
tested on staging

PR #2736, opened 2024-11-06
build-configs.yaml: Enable more configs for x86 AMD Chromebooks

laura-nao:
Enable additional configurations for x86 AMD Chromebooks in the x86-board fragment. These missing configurations were identified from preliminary results of the ACPI kselftest on AMD Chromebooks.


nuclearcat:
Thanks, i guess it is trivial enough, and staging seems works ok.

PR #2737, opened 2024-11-07
lava: Update FVP commands for latest shrinkwrap release

broonie:
The latest shrinkwrap release has updated the containers we're using for
the FVP to a new version which moves SVE support from a plugin into the
main model.  This has been causing all our FVP jobs to fail as the
arguments supplied are invalid, update with parameters suitable for the
new version of the model.


broonie:
There seems to be an awful lot of stuff still running on legacy, I'm seeing far more jobs going through from that than I am from Maestro. I'm certainly not seeing any FVP jobs, looks like that needs some updates in config/runtime/boot which didn't happen yet...

nuclearcat:
Merging, as we might use this info for migration

PR #2740, opened 2024-11-13
feat(cli/event): Implement promisc flag

nuclearcat:
We have API promisc flag for pub/sub events, and it is useful to have it available in cli too, to listen for all events on api instance.


nuclearcat:
Tested manually

PR #2745, opened 2024-11-25
kernelci.api.models: fix pydantic warnings for `url` fields

JenySadadia:
Fix the below pydantic v2 warnings:
```
Expected `url` but got `str` with value - serialized value may not be as expected
```
This is due to fields such as `artifacts` and `revision.url` expect pydantic `Url` type and afterwards convert the values to `str` to store it in mongo DB and for json response. Change field types to `str` and use `TypeAdapter` to validate `Url` values before validating them as strings.

Discussion thread for the issue:
https://github.com/pydantic/pydantic/discussions/6395


kernelci/api/models.py
```diff
@@ -117,7 +120,8 @@ class Revision(BaseModel):
     tree: str = Field(
         description="git tree of the revision"
     )
-    url: Annotated[AnyUrl | FileUrl, AfterValidator(str)] = Field(
+    url: Annotated[str, BeforeValidator(
+        lambda value: str(any_url_adapter.validate_python(value)))] = Field(
```
pawiecz, line 124: Support for [FileURLs without host set](https://docs.pydantic.dev/latest/api/networks/#pydantic.networks.FileUrl) is dropped here but I don't think any kind of currently supported Storage types allowed them anyway

pawiecz (approved):
LGTM

kernelci/api/models.py
JenySadadia: `AnyUrl` allows `FileUrl` as well. That's why dropped it from here.

PR #2748, opened 2024-11-29
kernelci.api.models: update `default` values for `Dict` fields

JenySadadia:
In principle, all the dictionary fields should have empty `dict` as their default value.
This also reduces chances of pipeline service crash due to issues such as iterating over a dict field with `None` as its value.


pawiecz (approved):
LGTM

PR #2751, opened 2024-12-04
config: dmesg.sh: Ignore missing handler errors in baseline tests

musamaanjum:
Ignore "No irq handler for vector" errors in baseline tests [1]. These errors are harmless. But because of these errors the baseline test gets marked as failed which makes results dirty. It would have been better to ignore these errors as known errors on the dashboard. But as we don't have the functionality at this time. We have decided to ignore these during the test until we have the functionality.

[1] https://kcidb.kernelci.org/d/test/test?orgId=1&var-datasource=default&var-build_architecture=x86_64&var-build_config_name=cros:%2F%2Fchromeos-6.6%2Fx86_64%2Fchromeos-amd-stoneyridge.flavour.config&var-id=maestro:67455b8c3be6da94b19fde34&from=now-100y&to=now&timezone=browser&var-origin=$__all&var-test_path=&var-issue_presence=$__all Close https://github.com/kernelci/kernelci-project/issues/496
Signed-off-by: Muhammad Usama Anjum <musamaanjum@collabora.com>

### DIFF
--- a/config/docker/cros-baseline.jinja2
+++ b/config/docker/cros-baseline.jinja2
@@ -20,7 +20,7 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
 {{ super() }}
 
 ADD https://raw.githubusercontent.com/kernelci/kernelci-core/\
-kernelci.org/\
+main/\
 config/rootfs/debos/overlays/baseline/opt/kernelci/dmesg.sh \
 /home/cros/dmesg.sh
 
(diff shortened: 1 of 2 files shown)


config/docker/cros-baseline.jinja2
```diff
@@ -20,7 +20,7 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
 {{ super() }}
 
 ADD https://raw.githubusercontent.com/kernelci/kernelci-core/\
-kernelci.org/\
+main/\
```
musamaanjum, line 23: @nuclearcat I'm not sure about the significance of kernelci.org branch. But dmesg.sh file is same on both main and kernelci.org.

nuclearcat: yes, at current moment better to use main

nuclearcat:
Lets try to merge and use it

PR #2760, opened 2024-12-16
config: docker: Update pahole to v1.28

laura-nao:
A recent regression[1] has been identified, and it does not occur with pahole v1.28. Update pahole to the appropriate version in the Docker containers used for kernel builds.

[1] https://lore.kernel.org/all/20241106160820.259829-1-laura.nao@collabora.com/

### DIFF
--- a/config/docker/base/host-tools.jinja2
+++ b/config/docker/base/host-tools.jinja2
@@ -62,8 +62,10 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
     bison \
     bsdmainutils \
     ccache \
+    cmake \
     cpio \
     dwarves \
+    elfutils \
     flex \
     g++ \
     gawk \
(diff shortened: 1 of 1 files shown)


laura-nao:
I tested the changes locally by building a Docker container for x86_64 with gcc-12 and kernelci+kselftest fragments. Checked that `pahole --version` reports the correct version and manually built linux-next using the docker container.

@nuclearcat is there any way to test the changes on staging before merging? i only checked one container manually, but these changes will affect all kernel build containers.

nuclearcat:
Yes, we are using staging- prefix for staging instance. So you can trigger by kci tool checkout of any tree, where this pahole matters.

laura-nao:
Thanks, I reviewed a sample of kernel builds on staging - pahole 1.28 is correctly reported in the config, the builds completed successfully, and the error from the regression report is no longer present. Should be good to be merged.

Not sure what the CI check failure is about, it doesn't seem to be related to the changes in the PR.

PR #2766, opened 2025-01-10
feat(rootfs-configs.yaml): prepare rootfs for netdev/kselftest

nuclearcat:
As part of work to prepare kselftest rootfs we need to add packages required to run netdev related tests properly.


nuclearcat:
Tested on staging

PR #2767, opened 2025-01-13
feat(helper.py): Better allow/deny list debug

nuclearcat:
In some situations we need better debug on allow/deny list


nuclearcat:
Tested on staging

PR #2768, opened 2025-01-14
fix(yaml/jinja): Add IGB and grub nfs support

nuclearcat:
Board Aaeon UP squared 6000 with X6425RE needs IGB driver to have functioning networking.


nuclearcat:
Tested on staging: https://staging.kernelci.org:9000/viewer?search=parent%3D6786a260029aeabb05bfc54c

PR #2769, opened 2025-01-14
fix(runtime/lava.py): Add timeout for POST

nuclearcat:
In some situations request might get stuck,
take precautions against such situations.


nuclearcat:
Tested on staging: https://staging.kernelci.org:9000/viewer?search=parent%3D6786a260029aeabb05bfc54c

PR #2771, opened 2025-01-15
fix(kbuild): Add CC/HOST_CC variables for kselftest

nuclearcat:
kselftest (for example net) require correct CC/HOST_CC variable to compile some of the components.


nuclearcat:
Testing on staging, fixed building kselftest components (bpf).

PR #2775, opened 2025-01-16
upd(build-configs.yaml): Add kselftest/net required options

nuclearcat:
kselftest/net require some non-default options, lets try to enable them.


nuclearcat:
Tested on staging

PR #2780, opened 2025-01-22
feat(runtime/tests/kselftest.jinja2): Add skipfile support

nuclearcat:
We need skipfile support to skip some of tests.


nuclearcat:
Tested on staging

PR #2782, opened 2025-01-23
config: runtime: boot: add support for barebox bootloader

hnez:
We are currently in the process of reviving the [Pengutronix LAVA lab](https://lava.pengutronix.de/) for use with the new KernelCI infrastsructure.

As a first step we need [Barebox](https://www.barebox.org/) bootloader support, since it is used by all boards in the Pengutronix lab.

This PR replaces #2600 by @pawiecz and includes the following suggestion by @a-wai https://github.com/kernelci/kernelci-core/pull/2600#issuecomment-2320699175:

> This doesn't take into account the case where `boot_commands == nfs`; I'd rather suggest to just copy the u-boot template and `s/u-boot/barebox/` there, as it seems they should be otherwise identical.

The barebox template is indeed a copy of the u-boot template with the method set to `barebox`.

### DIFF
--- a/config/runtime/boot/barebox.jinja2
+++ b/config/runtime/boot/barebox.jinja2
@@ -1,0 +1,48 @@
+{%- if boot_commands == 'nfs' and nfsroot is not defined %}
+{%-   set nfsroot = 'http://storage.kernelci.org/images/rootfs/debian/bookworm-kselftest/20250117.0/' + debarch %}
+{%- endif %}
+
+- deploy:
+    kernel:
+      type: {{ node.data.kernel_type }}
+      url: '{{ node.artifacts.kernel }}'
+    modules:
+      compression: xz
+      url: '{{ node.artifacts.modules }}'
+{%- if device_dtb %}
+    dtb:
+      url: '{{ node.artifacts.dtb }}'
+{%- endif %}
+{%- if boot_commands == 'nfs' %}
+    nfsrootfs:
+      url: '{{ nfsroot }}/full.rootfs.tar.xz'
+      compression: xz
+    ramdisk:
+      url: '{{ nfsroot }}/initrd.cpio.gz'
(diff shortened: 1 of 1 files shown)


config/runtime/boot/barebox.jinja2
```diff
@@ -0,0 +1,48 @@
+{%- if boot_commands == 'nfs' and nfsroot is not defined %}
+{%-   set nfsroot = 'http://storage.kernelci.org/images/rootfs/debian/bookworm-kselftest/20240221.0/' + debarch %}
```
pawiecz, line 2: Is there any specific reason to use `20240221.0` images? There were [later updates](https://github.com/kernelci/kernelci-core/blame/main/config/core/rootfs-configs.yaml#L200) to [this rootfs](https://storage.kernelci.org/images/rootfs/debian/bookworm-kselftest/)

pawiecz (commented):
LGTM

Older nfsroot might not be able to run all the currently enabled tests but it can be updated later

config/runtime/boot/barebox.jinja2
hnez: That's just what's currently used in `u-boot.jinja2`. I could add a commit that updates both to e.g. `20250117.0` if that makes sense.

pawiecz: @hnez right, that LAVA action template should have been updated, good catch!

Please bump nfsroot defaults in `u-boot.jinja2` as a separate PR and align them in this PR.

hnez: I've created https://github.com/kernelci/kernelci-core/pull/2787 and updated this PR to use `20250117.0` as well.

pawiecz (approved):
LGTM

Thanks @hnez for picking up this task and your contribution!

PR #2786, opened 2025-01-24
build-configs.yaml: Add MediaTek ethernet drivers to lab-setup fragment

nfraprado:
This allows boards like the Genio 700 EVK and Genio 350 EVK to run NFS jobs.


nfraprado:
@nuclearcat the CI failure is completely unrelated to the change.

PR #2787, opened 2025-01-27
config: runtime: boot: u-boot: update the default nfsroot to 20250117.0

hnez:
While adding barebox bootloader support in https://github.com/kernelci/kernelci-core/pull/2782#discussion_r1930452334 it was discovered that the default nfsroot image in the u-boot template has not been updated in some time. Update to the current most recent version 20250117.0.

Thanks to @pawiecz for noticing this in https://github.com/kernelci/kernelci-core/pull/2782.


pawiecz (approved):
LGTM

PR #2792, opened 2025-01-28
feat(helper.py): Add promisc to subscribe_filters

nuclearcat:
Add support of promisc flag to subscribe_filters
helper function. Will be used in kcidb-bridge, to
listen for all users events.


nuclearcat:
Tested on staging

PR #2806, opened 2025-02-17
feat(kbuild.py): Download linux-firmware from kci infra

nuclearcat:
git.kernel.org sometimes unreachable or Azure k8s issue, so lets keep at least firmware locally on Azure storage.


nuclearcat:
Tested on staging

PR #2819, opened 2025-03-02
(setup-networking.sh): Add cloudflare resolver

nuclearcat:
We are observing very strange resolving issues:
https://github.com/kernelci/kernelci-pipeline/issues/1064

This is attempt to improve DNS reliability.


nuclearcat:
Tested on staging

PR #2820, opened 2025-03-05
build-configs.yaml: Enable CONFIG_SND_SOC_INTEL_SOF_RT5682_MACH=m in …

laura-nao:
…x86-board fragment

The snd-soc-sof_nau8825.ko module fails to link without sof_realtek_common support. ChromeOS-specific platform fragments already address this by setting CONFIG_SND_SOC_INTEL_SOF_RT5682_MACH=m, which enables sof_realtek_common. Apply the same change to the x86-board fragment for consistency.

See: https://dashboard.kernelci.org/issue/maestro%3A4c243628961e1d9057e707cf55ea29e8b5a2761d


nuclearcat (approved):
looks good, on staging too

PR #2824, opened 2025-03-12
runtime: lava: Rename TestJob root Node to avoid collisions

pawiecz:


kernelci/runtime/lava.py
```diff
@@ -259,7 +259,7 @@ def get_hierarchy(self, results, job_node):
 
         return {
             'node': {
-                'name': job_node['name'],
+                'name': f"{job_node['name']}-lava-testjob",
```
padovan, line 262: `-main` or `-group` maybe instead of `-lava-testjob`? I assume this is the group that may contain multiple jobs insides.  So I guess we will have a <job>-group and a <job> entries in KCIDB?

Would this change affect eg kselftest where we don' t have such a duplication issue?

PR #2829, opened 2025-04-02
test-configs: remove kontron-pitx-imx8m

hthiery:
This board is no longer supported by the vendor and will be removed from the test farm.


hthiery:
Oh, I've only just noticed that no more tests have been running in our lab since November 12. That means we need to migrate our config to the new repo?

hthiery:
I migrated the configuration to the new pipeline: https://github.com/kernelci/kernelci-pipeline/pull/1101

PR #2830, opened 2025-04-02
lava: Install the kselftest tarball using an overlay

broonie:
Currently the kselftest LAVA jobs that KernelCI generates have the test
job download the kselftest tarball onto the device.  This has a couple
of disadvantages:

 - Any caching that is configured for the LAVA worker will be bypassed,
   increasing network traffic both for the lab and for the KernelCI
   storage server.
 - The entire kselftest tarball has to be downloaded onto the test
   system, and if (as is usual) it is a NFS filesystem then an
   uncompressed copy written over the network to the LAVA worker.

The test-definitions integration for kselftest also supports running a
copy that is already installed in the filesystem, and LAVA supports
adding files onto the root filesystem via it's overlay mechanism, so we
can avoid the overheads of downloading directly to the device by having
LAVA apply the kselftest tarball to the root filesystem as an overlay
instead.


broonie:
This really needs testing on staging but I don't seem to see any kselftest jobs being scheduled from staging (even before I did this, it's not that I broke it).

nuclearcat:
Maybe i can trigger some tree by kci-dev? mainline probably have kselftests

nuclearcat:
https://staging.kernelci.org:9000/viewer?search=parent%3D67ef07d290edc24662fb86a9
looks like many kselftests running ok?  so i guess it should be fine

broonie:
Looking at what I think are those jobs I don't think the change had actually been picked up... we still seem to be downloading at runtime.

nuclearcat:
Probably this is legacy files. I feel i need to do big cleanup, and i will try to take care about this problem on next week (kselftest).

PR #2834, opened 2025-04-09
(fragment/kernelci.jinja2) Fix, add --break-system-packages to pip

nuclearcat:
Suddenly pip stopped to work for kernelci docker image builds on staging. Most likely reason is --break-system-packages, probably one of base images changed distro and this flag now mandatory.


nuclearcat:
Merging, otherwise its a bit hard to verify updated workflows

PR #2840, opened 2025-04-10
lava: Install the kselftest tarball using an overlay

broonie:
Currently the kselftest LAVA jobs that KernelCI generates have the test
job download the kselftest tarball onto the device.  This has a couple
of disadvantages:

 - Any caching that is configured for the LAVA worker will be bypassed,
   increasing network traffic both for the lab and for the KernelCI
   storage server.
 - The entire kselftest tarball has to be downloaded onto the test
   system, and if (as is usual) it is a NFS filesystem then an
   uncompressed copy written over the network to the LAVA worker.

The test-definitions integration for kselftest also supports running a
copy that is already installed in the filesystem, and LAVA supports
adding files onto the root filesystem via it's overlay mechanism, so we
can avoid the overheads of downloading directly to the device by having
LAVA apply the kselftest tarball to the root filesystem as an overlay
instead.

A commit was previously merged doing this for the legacy LAVA templates,
this updates the Maestro ones.

### DIFF
--- a/config/runtime/boot/depthcharge.jinja2
+++ b/config/runtime/boot/depthcharge.jinja2
@@ -27,7 +27,16 @@
 {%- if boot_commands == 'nfs' and nfsroot %}
     nfsrootfs:
       compression: xz
+      format: tar
       url: '{{ nfsroot }}/full.rootfs.tar.xz'
+      {%- if node.artifacts.kselftest_tar_gz %}
+      overlays:
+        kselftest:
+          url: {{ node.artifacts.kselftest_tar_gz }}
+          compression: gz
+          format: tar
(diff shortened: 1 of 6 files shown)


broonie:
Not seeing this take effect in the staging run that's currently got jobs in my lab: https://lava.sirena.org.uk/scheduler/job/1283336/definition - not sure what's going on there?

nuclearcat:
Let's try to test it again on staging. I remember last time i tried and failed, but we should try again.
@broonie can you rebase it please?

padovan:
issues review meeting:
@broonie is still trying to figure how to test for the presence of the tarball (that has kselftest inside).

PR #2843, opened 2025-04-11
Docker: remove docker intermediate images after build

tollsimy:
Building Docker images using the kci tool results in dangling intermediate images, which can lead to unnecessary disk space usage over time and leave the system in a dirty state

This PR updates the build process to properly clean up intermediate images after the build is complete, preventing the accumulation of dangling images.


nuclearcat:
Nice find, indeed i notied there is a lot of dangling images, i will add it to staging tests now.

tollsimy:
Thanks @nuclearcat !

Issue #2842, opened 2025-04-11
Production crash on node with format symbols in commit info


nuclearcat:
```
Exception Error, node id: 67f8c7fac98e04810d2fd5ac, unmatched '{' in format spec
04/11/2025 07:45:43 AM UTC [ERROR] Traceback (most recent call last):
  File "/home/kernelci/pipeline/src/base.py", line 70, in run
    status = self._run(context)
             ^^^^^^^^^^^^^^^^^^
  File "/home/kernelci/pipeline/./src/scheduler.py", line 337, in _run
    self._run_job(job, runtime, platform, input_node)
  File "/home/kernelci/pipeline/./src/scheduler.py", line 81, in _run_job
    node = self._api_helper.create_job_node(job_config,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/kernelci/api/helper.py", line 467, in create_job_node
    raise error
  File "/usr/local/lib/python3.11/site-packages/kernelci/api/helper.py", line 464, in create_job_node
    job_node['data'] = platform.format_params(job_node['data'], extra_args)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/kernelci/config/base.py", line 167, in format_params
    return _format_dict_strings(param, args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/kernelci/config/base.py", line 79, in _format_dict_strings
    param[key] = _format_dict_strings(param[key], fmap)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/kernelci/config/base.py", line 79, in _format_dict_strings
    param[key] = _format_dict_strings(param[key], fmap)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.11/site-packages/kernelci/config/base.py", line 74, in _format_dict_strings
    param = param.format_map(fmap)
            ^^^^^^^^^^^^^^^^^^^^^^
ValueError: unmatched '{' in format spec
```
Node:
```
{
[...]

PR #2844, opened 2025-04-11
(api/helper.py) Escape all curly braces in f-strings

nuclearcat:
As we need to replace {kver} for ChromeOS config strings, we treat all node["data"] as f-string, but same time some entries are user supplied and might contain curly braces, which will cause confusion. Fix that by escaping curly braces.

Fixes: https://github.com/kernelci/kernelci-core/issues/2842

### DIFF
--- a/kernelci/api/helper.py
+++ b/kernelci/api/helper.py
@@ -388,6 +388,16 @@ class APIHelper:
 
         return True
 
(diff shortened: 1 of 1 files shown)


nuclearcat:
Tested at staging, as its affecting production - merge ASAP

PR #2847, opened 2025-04-14
(docker/lava-callback.jinja2) Update path to lava-callback requirements.txt

nuclearcat:
Path to requirements.txt is changed, reflect it in docker recipe


nuclearcat:
Tested on staging

PR #2849, opened 2025-04-14
build-configs.yaml: Add a fragment for fault injection

crazoes:
We are working on adding a test based on fault injection into KernelCI, hence, add a fragment to enable fault injection specific configs

### DIFF
--- a/config/core/build-configs.yaml
+++ b/config/core/build-configs.yaml
@@ -447,6 +447,31 @@ fragments:
   debug:
     path: "kernel/configs/debug.config"
 
+  fault-injection:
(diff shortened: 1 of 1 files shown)


crazoes:
> @crazoes you're missing the `=y`/`=m` assignments :)
> 

Oops, I missed that. Let me push the changes.

> These configs probably don't add much to the overall kernel size, right? We won't need all of them just to run an initial slab allocation test, but if the size impact is minimal, it might be worth adding them all from the start.

They don't really increase the kernel size that is why I added them all. They have very minimal dependency.

laura-nao (approved):

PR #2852, opened 2025-04-16
Docker: remove spawned containers after completion

tollsimy:
Similarly to #2843, remove spawned containers after completion.

Running containers using `kci` tool results in dangling containers, which can lead to unnecessary disk space usage over time and leave the system in a dirty state.

This PR updates the container run process to properly clean up containers after the execution is complete, preventing the accumulation of dangling containers.

This is evident when running some kbuilds, after which respective containers remains in exited state in the system.


tollsimy:
Not sure why the check `CI / Lint` failed, maybe not related to the PR.

nuclearcat:
LGTM

PR #2855, opened 2025-04-21
runtime:boot: Improve fastboot template for dynamic dtb blob extraction

quic-ylal:
Enhancing the fastboot template to dynamically extract the DTB blob. This approach aims to make post-processing more generic and adaptable for various fastboot boot targets.


nuclearcat (approved):
LGTM, thank you!

Issue #2859, opened 2025-04-25
Loading YAML configs works only for files found in top-level `config_path`


pawiecz:
Docstring for function [`iterate_yaml_files()`](https://github.com/kernelci/kernelci-core/blob/6ad79ac6d12026679e6b6f8a476888e817cdbdbd/kernelci/config/__init__.py#L17) states that it should traverse over given directory and "find multiple YAML files recursively".

Pull request https://github.com/kernelci/kernelci-pipeline/pull/1132 shows that this statement is false - PR introduces additional level of directories (`trees/`) which are ignored by said function:

```py
In [1]: glob.glob(os.path.join('config', '*.yaml'))
Out[1]: 
['config/trees.yaml',
 'config/scheduler.yaml',
 'config/pipeline.yaml',
 'config/platforms.yaml',
 'config/scheduler-chromeos.yaml',
 'config/jobs-chromeos.yaml',
 'config/result-summary.yaml',
 'config/jobs.yaml',
 'config/traces_config.yaml',
 'config/platforms-chromeos.yaml']
# 'config/trees/linux-pci.yaml' is missing
```

PR #2860, opened 2025-04-25
config: Find YAML files recursively

pawiecz:
Previously only top-level files were found in a given directory.

Fixes: #2859


kernelci/config/__init__.py
```diff
@@ -24,7 +24,7 @@ def iterate_yaml_files(config_path: str):
     if config_path.endswith('.yaml'):
         yaml_files = [config_path]
     else:
-        yaml_files = glob.glob(os.path.join(config_path, "*.yaml"))
+        yaml_files = glob.glob(os.path.join(config_path, "**", "*.yaml"), recursive=True)
```
kwilczynski, line 27:
```suggestion
        yaml_files = glob.iglob(os.path.join(config_path, "**", "*.yaml"), recursive=True)
```

kwilczynski: While we do not expect a large number of files there, as such memory usage is not a concern, we walk the list of the files we find almost immediately, so we might as well use the variant that returns an iterator.

kwilczynski (approved):
Looks good! A small nit, but it's non-blocking, for sure.

kwilczynski:
The following change should, once this branch is rebased, help resolve the CI linker failure:

- https://github.com/kernelci/kernelci-core/pull/2862

kernelci/config/__init__.py
pawiecz: Good catch! Applied

kwilczynski: Ah, so mypy is not happy now. Sorry, I didn't notice this earlier.

We can do a few things:

- `yaml_files = iter([config_path])` so that the type can be correctly inferred
- `yaml_files: list[str] | Iterator[str] = [config_path]` to add missing type hint

The latter requires an import from `typing`, which adds extra code, sadly.

There is also an option to not use `glob.iglob()` to keep the changes to a minimum.

nuclearcat:
@pawiecz can you rebase your PR, so probably latest workflows will be used for checks

kernelci/config/__init__.py
pawiecz: Thanks for suggestions! I went with the former one - aligned `yaml_files` types

PR #2864, opened 2025-04-29
runtime: boot: fastboot: add modules to initramfs

EmbeddedAndroid:
When reviewing the kerenlci devices in the Qualcomm lab I noticed we are not adding modules from the KCI artifacts into the initramfs. This patch adds the modules into the initramfs using the lava deploy schema.


config/runtime/boot/fastboot.jinja2
```diff
@@ -10,6 +10,11 @@
         format: cpio.newc
         overlays:
           lava: true
+          modules:
+            url: '{{ node.artifacts.modules }}'
+            compression: xz
+            format: tar
+            path: /
```
kwilczynski, line 17: These may not be necessary here, after reviewing other similar code that already exists.

That said, I am not 100% sure.

kwilczynski: @EmbeddedAndroid, thank you for double-checking! Appreciated. 👍 

I wasn't sure if the extra attributes were necessary after checking how other templates handle modules. However, none of the other ones include the `overlays` section under `ramdisk`, indeed.

kwilczynski (approved):
Nice catch! Thank you!

EmbeddedAndroid:
@kwilczynski thanks for your review. I took a quick at the other boot templates, and indeed we are not overlaying the modules into the initramfs there. In the v1 of KCI, we overlay'd the modules in the initramfs so they would probe during boot, I think we should do the same in v2. I'll take a go at adding the required changes and test them locally. If they work properly I'll propose those changes as well.

a-wai (approved):

PR #2865, opened 2025-04-30
Add fault injection template

crazoes:
Add a basic fault injection test template which will be currently used to add a suspend/resume test.

### DIFF
--- a/config/runtime/tests/fault-injection.jinja2
+++ b/config/runtime/tests/fault-injection.jinja2
@@ -1,0 +1,32 @@
+- test:
+    timeout:
+      minutes: {{ job_timeout|default(20) }}
+    definitions:
+    - from: inline
+      name: "fault-injection-{{ test_name }}"
+      path: inline/fault-injection.yaml
+      repository:
+        metadata:
+          format: Lava-Test Test Definition 1.0
+          name: fault-injection
+          description: |
+             Fault injection: {{ test_name }}
+             Type: {{ fail_type|default('failslab') }} | Iterations: {{ iterations|default(5) }}
+          os:
+          - debian
+          scope:
+          - functional
+        run:
+          steps:
+            - |
+              for i in $(seq 1 {{ iterations|default(5) }}); do
+                echo "Iteration $i: {{ test_name }}"
+                FAILCMD_TYPE={{ fail_type|default('failslab') }} \
+                  /usr/local/bin/failcmd \
+                  --probability={{ probability|default(100) }} \
+                  --times={{ times|default(100) }} \
+                  --interval={{ interval|default(100) }} \
+                  --verbose=2 \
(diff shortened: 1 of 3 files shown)


config/runtime/tests/fault-injection.jinja2
```diff
@@ -0,0 +1,32 @@
...
+          format: Lava-Test Test Definition 1.0
+          name: fault-injection
+          description: "Fault injection test: {{ test_name }}"
+          os:
+          - debian
+          scope:
+          - functional
+        run:
+          steps:
+            - wget {{ failcmd_url }} -O /tmp/failcmd.sh
```
laura-nao, line 19: We're going to fetch the script from https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/tools/testing/fault-injection/failcmd.sh right? if the URL is not going to change, I guess there's no need to make it a parameter.

Side note - the script could also be installed in the rootfs as an overlay. However, since it's specific for this type of tests and there's no other dependency required (except for wget, which is already installed on most rootfs images already available in KernelCI) I think getting the script with wget is fine in this case. @pawiecz @nuclearcat any thought on this?

config/runtime/tests/fault-injection.jinja2
```diff
@@ -0,0 +1,32 @@
...
+            - chmod +x /tmp/failcmd.sh
```
laura-nao, line 20: Maybe we can wrap the wget and chmod commands in a LAVA test set, so it's easier to spot a failure if it happens in this case. So something like:
```
- lava-test-set start setup
- <wget command>
- <chmod command>
- lava-test-set stop setup
```

config/runtime/tests/fault-injection.jinja2
laura-nao: Ack - then @crazoes we're going to need a new entry in the `rootfs-configs.yaml` file for a new rootfs for fault injection tests, that includes the failcmd script. I suggest copying the existing config for `bookworm` and adjust as needed. Either we create a script that gets run during the rootfs build and fetches the failcmd script (`script:` property), or we store the script as an overlay (`test_overlay:` property). I'd probably go with the former, as it would allow us to pick up any upstream script update simply by triggering a new rootfs build.

As we're adding a new rootfs config, it might be worth moving some of the test logic to a dedicated script too (via the `test_overlay:` property). This will simplify the jinja2 template and make it easier to add a result parser later if we need it.

laura-nao:
LGTM; the rootfs changes were also verified on staging: https://storage.staging.kernelci.org/rootfs/bookworm-fault-injection/20250507.0/amd64/

nuclearcat:
Excellent, thank you very much for verifying, merging

PR #2871, opened 2025-05-06
config: runtime: tests: ltp: enable support for code coverage

a-wai:
When enabling code coverage support using in-kernel GCOV, we must first reset coverage "recording" right before starting the test, then ensure we pack and upload the resulting artifacts.

The latter is implemented by POSTing the artifact to the storage server, providing the artifact name and URL as a LAVA test result using the following format:

  `artifact-upload:<artifact name>:<artifact URL>`

This allows the pipeline to add the uploaded artifact(s) to the corresponding node so the information can be easily re-used at a later stage.

Depends on https://github.com/kernelci/kernelci-pipeline/pull/1141

### DIFF
--- a/config/runtime/tests/ltp.jinja2
+++ b/config/runtime/tests/ltp.jinja2
@@ -3,12 +3,68 @@
     timeout:
       minutes: {{ job_timeout|default(15) }}
     definitions:
-       - repository: https://github.com/kernelci/test-definitions
-         from: git
-         revision: kernelci.org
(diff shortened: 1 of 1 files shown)


a-wai:
Marking as ready as it has been successfully tested on staging: [node](https://staging.kernelci.org:9000/viewer?node_id=68277a5f02ac9c758095df26) & [LAVA job](https://lava.collabora.dev/scheduler/job/18678899)

Issue #2873, opened 2025-05-08
kbuild not handling cip:// fragments


nuclearcat:
```
Generating shell script
Searching for fragment cip://6.1.y-cip/arm/qemu_arm_defconfig in /etc/kernelci/core/build-configs.yaml
Fragment cip://6.1.y-cip/arm/qemu_arm_defconfig not found
```
Reproduced over: `kci-dev checkout --giturl https://git.kernel.org/pub/scm/linux/kernel/git/cip/linux-cip.git --branch linux-6.1.y-cip --tipoftree`

PR #2874, opened 2025-05-08
(kbuild) Add cip:// fragments proper handling

nuclearcat:
cip:// was processed properly in legacy system, so we need just to port code from old system.
Fixes: https://github.com/kernelci/kernelci-core/issues/2873

### DIFF
--- a/kernelci/kbuild.py
+++ b/kernelci/kbuild.py
@@ -39,6 +39,8 @@ import kernelci.config.storage
 import kernelci.storage
 
 
+CIP_CONFIG_URL = \
+    "https://gitlab.com/cip-project/cip-kernel/cip-kernel-config/-/raw/master/{branch}/{config}"  # noqa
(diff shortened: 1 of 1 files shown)


kernelci/kbuild.py
```diff
@@ -38,7 +38,9 @@
 import kernelci.config.storage
 import kernelci.storage
 
-
+CIP_CONFIG_URL = \
+    "https://gitlab.com/cip-project/cip-kernel/cip-kernel-config/-\
+/raw/master/{branch}/{config}"
```
kwilczynski, line 43: A small nitpick: I think, it would be fine to keep this as a single longer line. The break looks a bit awkward.

nuclearcat: Fixed, but actually it was just copy-paste from legacy code and i guess it was like this due python linters.
But i agree, single line is cleaner, updated code.

PR #2878, opened 2025-05-21
config: core: rootfs-configs: add `ca-certificates` to bookworm-ltp

a-wai:
This package is needed for using `curl` on SSL/TLS connections, but it is only a `Recommends`, so it must be explicitly listed as a wanted package.


a-wai:
Successfully tested in https://lava.collabora.dev/scheduler/job/18721559 (using the staging-deployed rootfs)

PR #2879, opened 2025-05-21
kernelci: kbuild: fix "coverage source" artifact name

a-wai:
Replace dash with underscore for better consistency with other artifacts.


nuclearcat (approved):
LGTM

PR #2880, opened 2025-05-22
tests: Remove interval for fault injection testing

crazoes:
Interval is making the probablity of fault occurrences less, hence remove it from the template.


pawiecz (approved):
LGTM

PR #2882, opened 2025-05-23
kernelci: kbuild: transition to `available` state when complete

### LINKED ISSUE #2875: Add new node kind for post-processing test results
As part of code coverage enablement, we need to post-process the results from each `test` node under a given `kbuild` node.

This implies waiting for all test jobs to finish, then execute a script that would:
* retrieve the source code tarball from the parent `kbuild` node
* retrieve the coverage data tarballs from each `test` node
* run the `lcov` command to:
  1. transform raw coverage data into single files
  2. aggregate all those files into a single, global results file
  3. generate and upload the html report from the aggregated data

### Proposed solution

This could be done by creating a new node kind, e.g. `postprocess`, that would have the following properties:
* created upon successful completion of a `kbuild` node
* initially in a (new) `waiting` state
* change state to `running` once every `test` job on the same level (i.e children of the parent `kbuild` node) have completed
* template created and executed after the transition to `running` state

### Benefits

This would ensure that results post-processing can rely on a generic mechanism, rather than hardcoding coverage-related post-processing into the KernelCI services themselves.

We could also leverage the templating system to implement such post-processing scripts in a generic and/or portable way.

---

I obviously offer to implement this mechanism and enable it for coverage post-processing. However, I'm expecting there will be some side-effects I haven't anticipated, and would appreciate to read your thoughts and suggestions about this proposal.

a-wai:
Currently, `kbuild` nodes transition to state `done` as soon as they complete, even though they often have many child nodes running afterwards. This is confusing as `checkout` nodes first transition to `available` state, then `closing` (waiting for child nodes to complete) and only transition to `done` once all child nodes have completed.

This behaviour should be applied to `kbuild` nodes as well, both for consistency and to be able to easily schedule post-processing jobs (e.g. by triggering those on the `state: done` event).

Also add a `holdoff` value (10 minutes, identical to `checkout` nodes) so the corresponding service properly handles state transitions for those nodes.

Fixes  #2875

Depends on https://github.com/kernelci/kernelci-pipeline/pull/1164

### DIFF
--- a/kernelci/api/helper.py
+++ b/kernelci/api/helper.py
@@ -551,9 +551,12 @@ class APIHelper:
         root_node = merge(root_from_db, root)
         root_node = root.copy()
         root_node['result'] = results['node']['result']
+        root_node['state'] = results['node'].get('state', 'done')
         root_node['artifacts'].update(results['node']['artifacts'])
         root_node['data'].update(results['node'].get('data', {}))
         root_node['processed_by_kcidb_bridge'] = False
+        if 'holdoff' in results['node']:
+            root_node['holdoff'] = results['node']['holdoff']
         if root_node['result'] != 'incomplete':
             data = root_node.get('data', {})
             if data.get('error_code') == 'node_timeout':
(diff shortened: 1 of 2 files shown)


a-wai:
Note: https://github.com/kernelci/kernelci-pipeline/pull/1164 should be merged first, adding a note in the PR description.

Issue #2883, opened 2025-05-23
linux-firmware is too large


nuclearcat:
Each kbuild pod downloads linux-firmware.tar.gz, and recently it reached 1.3GB.
We have approx 1000 kbuild events per day, and since majority of clusters are other cloud, our egress bill can reach $100/day.
We need to deal with that ASAP.

PR #2884, opened 2025-05-23
(kubernetes.jinja2) Introduce volume for firmware file caching

nuclearcat:
Related: https://github.com/kernelci/kernelci-core/issues/2883 We create volume, where cron job daily update linux-firmware archive. This volume PWX, mounted read-only on each kbuild container, so they can use this file.


nuclearcat:
Tested on staging

PR #2886, opened 2025-05-26
config: rootfs: Clean up .git for Kirk

broonie:
Shrink the rootfs slightly by discarding the git repository data for
Kirk.


nuclearcat:
Nice catch, thank you Mark!

PR #2892, opened 2025-06-02
rootfs-configs.yaml: Add psmisc to the kselftest rootfs for mm

broonie:
The mm selftests use killall from one of the scripts, hanging if it's
not available.  Add psmisc so it's available.

### DIFF
--- a/config/core/rootfs-configs.yaml
+++ b/config/core/rootfs-configs.yaml
@@ -295,6 +295,7 @@ rootfs_configs:
       - perl
       - perl-modules-5.36
       - procps
+      - psmisc
       - publicsuffix
       - python3-minimal
       - python3-unittest2


pawiecz:
> The CI failure seems bogus, or at least unrelated to what I'm pushing (which doesn't touch the Python code at all).

Yup, [the reason is unrelated](https://github.com/broonie/kernelci-core/commit/bd5cd0e990db0d90c9cf3a2392032605b5079908#diff-badf13c91864f7d73300483dc3f75173bb6a9a965aa5dd578da80318e9f25261R430) and CI failure can be ignored (btw `main` [has been updated](https://github.com/kernelci/kernelci-core/commit/82d0e46d2a01adf6d0bd670524a83b3b762b62c6#diff-badf13c91864f7d73300483dc3f75173bb6a9a965aa5dd578da80318e9f25261R428) to prevent such failures in future)

PR #2897, opened 2025-06-05
config: runtime: util: gcov-upload: fix compression method

a-wai:
The LTP rootfs doesn't include `xz`, so we can't use that compression method. Revert to `gz` as those artifacts aren't that big anyway.


nuclearcat (approved):
Looks good, thanks!

PR #2905, opened 2025-06-14
roofs: ltp: Use a fixed release of LTP when building it's rootfs

broonie:
Currently we build LTP using whatever the current commit on the master
branch is. This means that any development on the rootfs could change the
set of LTP tests we're running and that we will be running just some random
snapshot of LTP, not an actual release, so could end up carrying bugs that
wouldn't make it into a release. Fix the version of LTP at the recently
released 20250530 version.


nuclearcat:
Very nice catch, thank you!

PR #2908, opened 2025-06-19
config: docker: base: host-tools: add gcovr

a-wai:
`gcovr` is needed to post-process coverage data. However, the version available in Debian Bookworm is old and lacks features we need, such as generating `lcov`-compatible tracefiles. Install it using `pip` as a temporary workaround.


pawiecz (approved):
LGTM

For the record it's:
- v5.2 in Bookworm
- v7.2 in Trixie
- v8.3 in PyPI (Released: Jan 19, 2025)

PR #2909, opened 2025-06-19
config: runtime: base: python: make more generic

a-wai:
The base `python` template was designed with post-checkout jobs (`kbuild`, `kunit`...) in mind and doesn't work too well with test or post-processing jobs due to the following issues:
* it assumes we need the kernel source from the parent `checkout` node
* it doesn't provide a way to retrieve arbitrary artifacts from a node or its parent
* the `_get_source()` function doesn't allow specifying an arbitrary target folder to extract tarballs to

Ensure we only download/extract/cd to the kernel source if the current node is a child to a `checkout` node, otherwise let the job definition prepare its workspace as needed.

Also add a generic `_get_artifact_url()` for downloading arbitrary artifacts and make the existing `_get_tarball_url()` use it to avoid duplicating code.

Finally, add an optional `path` parameter to `_get_source()` so the job can specify the destination folder if needed.

### DIFF
--- a/config/runtime/base/python.jinja2
+++ b/config/runtime/base/python.jinja2
@@ -86,22 +86,36 @@ class BaseJob:
         storage_cred = os.getenv('KCI_STORAGE_CREDENTIALS')
         return kernelci.storage.get_storage(storage_config, storage_cred)
 
-    def _get_tarball_url(self, node):
-        if node.get('artifacts') and 'tarball' in node['artifacts']:
-            return node['artifacts']['tarball']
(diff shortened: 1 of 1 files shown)


a-wai:
v2:
* drop the parent node download/storage as an object member as it's unneeded 99% of the time
* factor the kernel source download/unpacking code into a new function
* change condition for downloading kernel source to `kind == kbuild`; the only affected job would be `kunit`, which template is fixed in https://github.com/kernelci/kernelci-pipeline/pull/1223

PR #2919, opened 2025-07-02
config: docker: Bump Python version for base template

pawiecz:
This upgrade is inteded to make use of [recent changes](https://docs.python.org/3.12/library/tempfile.html#tempfile.NamedTemporaryFile) in handling temporary files (in "tempfile" module).

Depends on: https://github.com/kernelci/kernelci-pipeline/pull/1229

### DIFF
--- a/config/docker/base/python.jinja2
+++ b/config/docker/base/python.jinja2
@@ -4,7 +4,7 @@
  # Author: Guillaume Tucker <guillaume.tucker@collabora.com>
 -#}
 
-FROM mirror.gcr.io/python:3.11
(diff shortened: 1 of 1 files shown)


pawiecz:
This template is used [only in images for Maestro Services](https://github.com/search?q=repo%3Akernelci%2Fkernelci-core+path%3A%2F%5Econfig%5C%2Fdocker%5C%2F%2F+base%2Fpython&type=code) which [seem to be working fine](https://staging.kernelci.org:9088/container/c4e2e20a2568) since last Staging deployment

PR #2920, opened 2025-07-03
kci config: Implement pipeline job forecasting tool

nuclearcat:
Often we need to know, what builds and tests will be triggered on each checkout. We can add option to kci to "forecast" these builds, for example:
./kci config forecast -c ../kernelci-pipeline/config


nuclearcat:
Moved from https://github.com/kernelci/kernelci-pipeline/pull/1225

kernelci/cli/config.py
```diff
@@ -75,3 +77,127 @@ def dump(section, config, indent, recursive):
...
+        checkouts.append(data)
+
+    # sort checkouts by tree and branch
+    checkouts.sort(key=lambda x: (x.get("tree", ""), x.get("branch", "")))
+
+    # iterate over checkouts
+    for checkout in checkouts:
+        checkout["kbuilds"] = []
+        # iterate over events (jobs)
+        jobs = merged_data.get("scheduler", [])
```
JenySadadia, line 145: Does it include configs from different scheduler files such as `scheduler-cip.yaml` and `scheduler-chromeos.yaml`?

JenySadadia:
I haven't tested the PR yet. Other than some CI checks are failing, it looks good to me.

PR #2923, opened 2025-07-11
receive_event: Make it optionally semi-nonblocking

nuclearcat:
Right now receive_event is blocking, and even on receiving keep alive it stay inside its own busy loop, which makes impossible to implement proper scheduling, watchdogs, health monitoring.
We make optional parameter, where on keep alive message, (sent each 45 sec by kernelci-api) it will return None. We keep backward compatible defaults, as lot of code is not expecting None in return.

### DIFF
--- a/kernelci/api/helper.py
+++ b/kernelci/api/helper.py
@@ -64,9 +64,16 @@ class APIHelper:
             self._filters.pop(sub_id)
         self.api.unsubscribe(sub_id)
 
-    def receive_event_data(self, sub_id):
-        """Receive CloudEvent from Pub/Sub and return its data payload"""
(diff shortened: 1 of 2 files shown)


pawiecz (approved):
LGTM (functionally)

Argument semantics is non-obvious to me - shouldn't the event receiving loop be broken if the `keep_alive` value is `False` (this would require setting defaults to `True` and reversing loop condition)?

JenySadadia:
The below code snippet in `receive_event` makes sure that it keeps on listening to events after `BEEP` is received:
```
if event.data == 'BEEP':
    continue
```
Instead of this workaround, I'd suggest we should check why `receive_event` loop blocks.

PR #2931, opened 2025-07-23
config: update production instance URL

JenySadadia:
Update maestro production API URL to `https://api.kernelci.org`. The former URL is an alias of the new URL and it's still working fine but this looks like more readable and standard URL.
The new URL is also being used in other places such as KCIDB submissions and documentation.


a-wai (approved):
LGTM

PR #2937, opened 2025-08-04
kbuild: Decode CIP fragment properly, not over str()

nuclearcat:
CIP configurations was not working, as they was not decoded properly, looking like `b:'content\notherline\n' instead of properly formatted kernel config.
Decode() should fix that.


nuclearcat:
Tested on staging

PR #2942, opened 2025-08-13
Improve a log statement in `APIHelper.create_job_node`

JenySadadia:
Update a log print when the `create_job_node` method doesn't create a job node due to job filters.


JenySadadia:
Tested OK on staging:
```
Not creating node due to job filter. Job ltp-cap-bounds not found in jobfilter for node 689c354b17d8e167a68aea43
Not creating node due to job filter. Job ltp-containers not found in jobfilter for node 689c354b17d8e167a68aea43
```

PR #2945, opened 2025-08-15
Translate retry counter

JenySadadia:
Closes https://github.com/kernelci/kernelci-api/issues/626

In order to query nodes matching `retry_counter` field, translate query parameter value to integer since
all the query parameters are strings by default. For DB query, we need to change its data type to integer.


nuclearcat (approved):
Looks good

JenySadadia:
Tested OK on staging:
https://staging.kernelci.org:9000/nodes?retry_counter=3&limit=10

PR #2948, opened 2025-08-20
kernelci: api: models: add new "process" node type

a-wai:
Post-processing jobs are currently nodes of the `job` kind, which only works as long as they're related to `checkout` or `kbuild` nodes. However, due to performance issues of the `coverage-report` jobs, we need to post-process each coverage-enabled test job individually.

This leads to scheduling issues, as we could enter an infinite loop where new post-processing jobs would be triggered whenever one of those transitions to the `done` state.

We must therefore create a new node kind for post-processing jobs in order to avoid those scheduler limitations.

### DIFF
--- a/kernelci/api/models.py
+++ b/kernelci/api/models.py
@@ -627,6 +627,19 @@ class Job(Node):
     ]
 
 
+class Process(Node):
+    """API model for post-processing job nodes"""
+    class_kind: ClassVar[str] = 'process'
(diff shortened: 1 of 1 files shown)


a-wai:
Works fine on staging ([example node](https://staging.kernelci.org:9000/viewer?node_id=68a79c76657978facd2a2882))

PR #2956, opened 2025-08-22
config: docker: cros-tast: add gcov scripts to container

a-wai:
Those are needed for running gathering coverage data for Tast tests.

Depends on #2955

### DIFF
--- a/config/docker/cros-tast.jinja2
+++ b/config/docker/cros-tast.jinja2
@@ -23,6 +23,16 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
 {{ super() }}
 
 ADD https://raw.githubusercontent.com/kernelci/kernelci-core/\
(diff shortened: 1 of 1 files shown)


a-wai:
`staging-skip` for now as it requires #2955 to be merged **and deployed to the kernelci.org branch**

config/docker/cros-tast.jinja2
```diff
@@ -22,6 +22,16 @@ RUN apt-get update && apt-get install --no-install-recommends -y \
 
 {{ super() }}
 
+ADD https://raw.githubusercontent.com/kernelci/kernelci-core/\
+staging.kernelci.org/\
```
a-wai, line 26: This should be changed to
```suggestion
kernelci.org/\
```
once deployed to production

PR #2978, opened 2025-09-29
config/iPXE: increase timeouts

nuclearcat:
Based on observation, for example: https://lava.ciplatform.org/scheduler/job/1311593 3 minutes is not enough for bootloader.


aliceinwire:
ok, works for me

nuclearcat:
Tested, loading kernel ok now, but different problem appeared.
https://lava.ciplatform.org/scheduler/job/1311823

aliceinwire:
yes, I'm aware of that

PR #3001, opened 2025-10-17
Fragment for pipeline cron service

JenySadadia:
The pipeline cron service will rely on image `kernelci:pipeline-cron`. 
The PR includes the fragment file for the image.


nuclearcat:
We will need to make sure to include this fragment in all workflows (staging/production).

PR #3026, opened 2025-12-03
config: qemu: rename serial to console variable

bhcopeland:


bhcopeland:
Confirmed working: https://lava.ciplatform.org/scheduler/job/1342337

PR #3028, opened 2025-12-11
cli: Add --runtime-type parameter

nuclearcat:
Required by https://github.com/kernelci/kernelci-pipeline/pull/1396


nuclearcat:
Tested on staging

PR #3030, opened 2025-12-18
runtime/lava: Add get_data() method to LogParser

bhcopeland:
Add a method to return the parsed log data as a list of dicts in LAVA output.yaml format.


nuclearcat (approved):
LGTM

PR #3045, opened 2026-01-28
config: docker: Add riscv64 support

patersonc:
Needed to help the test jobs from https://github.com/kernelci/kernelci-pipeline/pull/1413/


patersonc:
It looks like qemu-system-riscv64 is already part of the image. Sorry for the noise.

PR #3062, opened 2026-03-02
config: Introduce rootfs_ref

nuclearcat:
At current moment we have rootfs scattered across configs, and even hardcoded inside kernelci-core.
We need to arrange that properly, similar to legacy system (even better), where we have single section with all rootfs kinds (templated), which can be referenced in other sections such as rootfs_ref parameter.


nuclearcat:
Tested on staging

PR #3084, opened 2026-04-20
kbuild: split kernel and kselftest builds

nuclearcat:
This is preparation step to track separately kernel build itself and kselftest builds.
For now we split them, make separate logs (this also addressing false logspec parsing errors of kselftest as kernel log) and avoid failing kernel build if kselftest only failed


nuclearcat:
It works fine, but needs follow up, as kselftest node missing logs in artifacts

PR #3088, opened 2026-04-23
kbuild: add extra_targets param for additional tuxmake targets

bhcopeland:
Add an extra_targets parameter to KBuild that allows job configs to request additional tuxmake build targets (e.g. binrpm-pkg) alongside the default kernel, modules and dtbs targets.

This is needed to produce RPM packages for AWS EC2 kernel testing, where kernels are installed via yum/dnf rather than from tarballs. The tuxmake artifact upload path already walks the full output directory, so RPMs are discovered and uploaded automatically.


nuclearcat:
Tested on staging, seems at least not breaking anything.

PR #3101, opened 2026-05-21
rootfs: strip: don't let apt cascade-purge the Python stack on trixie

bhcopeland:
In Debian trixie python3.13 and libpython3.13-stdlib Depends: tzdata, so the existing `apt-get remove --purge --yes tzdata` cascades and removes the whole Python stack (python3-tap, python3-yaml, asyncio stdlib...) from the crushed rootfs.cpio.gz.  trixie-kselftest is the visible casualty: parse-output.py fails with `No module named 'tap'`.

Use `dpkg --purge --force-depends` so only tzdata is removed.  Size saving is unchanged; dpkg-state warnings are cleared when crush.sh purges dpkg itself.

### DIFF
--- a/config/rootfs/debos/scripts/strip.sh
+++ b/config/rootfs/debos/scripts/strip.sh
@@ -16,11 +16,15 @@ export DEBIAN_FRONTEND=noninteractive
 exec 3>&-
 exec 4>&-
 
-# Removing unused packages
+# Removing unused packages.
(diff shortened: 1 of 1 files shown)


bhcopeland:
@yogeshlal is testing this locally it is linked to https://github.com/kernelci/kernelci-pipeline/issues/1484

yogeshlal:
Thanks @bhcopeland we are good with this PR, can you pls merge the change so that we consume in our pipeline.

PR #3104, opened 2026-05-23
rootfs: debos: fix dangling /init symlink

bhcopeland:
/init pointed at /usr/bin/systemd, which has never existed on Debian (systemd is at /usr/lib/systemd/systemd; /sbin/init is the working symlink).  LAVA's apply-overlay-ramdisk repacks the cpio so the bug is invisible there, but booting the published rootfs.cpio.gz as a raw initramfs (tuxrun, plain qemu -initrd) panics in mount_root_generic. Point /init at the real systemd binary.

### DIFF
--- a/config/rootfs/debos/rootfs.yaml
+++ b/config/rootfs/debos/rootfs.yaml
@@ -128,7 +128,7 @@ actions:
   - action: run
     description: Set symbolic link to init
     chroot: true
-    command: ln -s /usr/bin/systemd /init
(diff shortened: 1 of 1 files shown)


bhcopeland:
Tested locally with tuxrun, but not tested through LAVA, but since the previous symlink target was non-existent, LAVA's existing flow can't have been depending on it

nuclearcat:
Amazing find!

PR #3110, opened 2026-06-04
scheduler/lava: add priority-weighted queue depth limit

nuclearcat:
Add max_queue_depth_priority_factor config option and a get_queue_priority_fraction() helper so the queue ceiling can scale with job priority within configured priority bounds.


nuclearcat:
Staging looks ok

PR #3116, opened 2026-06-10
kbuild: log compiler version in tuxmake builds

bhcopeland:
When investigating a build failure it helps to know the exact
compiler version that produced it, but the build log never shows
one. The job parameters only carry the requested toolchain name
(e.g. clang-21), and with the null runtime tuxmake builds with
whatever compiler the build environment has installed, so the
requested name alone does not describe what actually ran.

Print the compiler's --version into build.log just before the
tuxmake invocation, so the version is visible even when the build
fails. The binary is resolved the same way tuxmake resolves it
(toolchain compiler plus tuxmake's arch-default CROSS_COMPILE)
rather than from the job's cross_compile parameter, because the two
can differ; arc jobs set arc-elf32- while tuxmake builds with
arc-linux-gnu-. Fall back to plain clang when tuxmake is not
importable.

### DIFF
--- a/kernelci/kbuild.py
+++ b/kernelci/kbuild.py
@@ -216,6 +216,7 @@ class KBuild:
             self._artifacts = []
             self._current_job = None
             self._config_full = ""
+            self._compiler_version = None
             self._srcdir = None
             self._firmware_dir = None
             self._af_dir = None
@@ -251,6 +252,7 @@ class KBuild:
             self._artifacts = jsonobj["artifacts"]
             self._current_job = jsonobj["current_job"]
             self._config_full = jsonobj["config_full"]
+            self._compiler_version = jsonobj["compiler_version"]
             self._srcdir = jsonobj["srcdir"]
             self._srctarball = jsonobj["srctarball"]
             self._firmware_dir = jsonobj["firmware_dir"]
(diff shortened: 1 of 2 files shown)


roxell:
tuxmake already collects this, it writes compiler.version_full into its metadata.json.
Can't we use that?

bhcopeland:
> tuxmake already collects this, it writes compiler.version_full into its metadata.json. Can't we use that?

It does, but I moved to this approach because it won't work for failed builds. We should make a similar change to metadata.json (for structured data).

roxell:
> > tuxmake already collects this, it writes compiler.version_full into its metadata.json. Can't we use that?
> 
> It does, but I moved to this approach because it won't work for failed builds. We should make a similar change to metadata.json (for structured data).

I think we should save and use what we want from tuxmake's metadata.json before we overwrite that with the "new" metadata.json file.

roxell:
LGTM, the only thing maybe add tests that goes along with the changes?

kernelci/kbuild.py
```diff
@@ -1243,7 +1256,24 @@ def _write_metadata(self):
...
+            if isinstance(tux_meta, dict) and "tuxmake" in tux_meta:
+                os.rename(
+                    metadata_path,
+                    os.path.join(self._af_dir, "tuxmake_metadata.json"),
+                )
+                version_full = tux_meta.get("compiler", {}).get("version_full")
+                if version_full:
+                    metadata["build"]["compiler_version"] = version_full
+                    self._compiler_version = version_full
+
```
roxell: Maybe create a function def _preserve_tuxmake_metadata() and call that just above the call to self._write_metadata() .

bhcopeland:
All addressed, thank you very much for your feedback @roxell!

roxell (approved):
LGTM

PR #3118, opened 2026-06-12
docker: install zstd in build images

nuclearcat:
Install zstd alongside the other host tools so kernels with CONFIG_MODULE_COMPRESS_ZSTD can complete modules_install. The shared host-tools base is used by gcc and clang images, covering every architecture that enables zstd module compression.


nuclearcat:
tested on staging

PR #3148, opened 2026-07-15
rootfs: trixie-ltp: put kirk on PATH

bhcopeland:
Since the recent test-definitions change, ltp default runner changed as a bare kirk through PATH. And looks only in the LTP install dir. So lets add a symlink just to catch these cases.


bhcopeland:
https://github.com/kernelci/kernelci-pipeline/pull/1551

PR #3166, opened 2026-07-26
runtime: raise TemplateRuntimeError from kci_raise

### LINKED ISSUE #1797: Replace generic `Exception` with a Jinja2 one in `kci_raise`
The `kci_raise` helper adds a command to raise an exception in the templates, for example when the variables aren't valid or some issue is detected when rendering the template.  It's better to validate the parameters before generating the output, but on the other hand it's better to raise a clear error when there's a problem then render a job definition that can't run and fails later in subtle ways.

We could pick one of the existing Jinja2 exceptions instead, see the documentation:

https://tedboy.github.io/jinja2/generated/jinja2.exceptions.html

giri256:
Fixes #1797.

The `kci_raise` Jinja helper currently raises a generic `Exception`, which loses the template-specific error type expected by Jinja callers.

This changes both the current runtime helper and the legacy LAVA helper to raise `jinja2.TemplateRuntimeError`. Focused regression tests cover both registrations.

Tests:

* `pytest tests/test_runtime.py -q` — 18 passed
* `pre-commit run --files kernelci/runtime/__init__.py kernelci/legacy/lava/__init__.py tests/test_runtime.py` — passed
* `pytest tests -q` — 139 passed, 7 unrelated Windows-only failures (`os.getuid`, POSIX path expectations, and slash handling)
* `git diff --check`

### DIFF
--- a/kernelci/legacy/lava/__init__.py
+++ b/kernelci/legacy/lava/__init__.py
@@ -12,6 +12,7 @@ import os
 import sys
 
 from jinja2 import Environment, FileSystemLoader
+from jinja2.exceptions import TemplateRuntimeError
 
 
 def add_kci_raise(jinja2_env):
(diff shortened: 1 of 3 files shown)


nuclearcat:
Fixes https://github.com/kernelci/kernelci-core/issues/1797

PR #3176, opened 2026-08-06
config/runtime: accept '~ #' shell prompt in LAVA boot templates

nuclearcat:
The new Buildroot-based baseline ramdisks ship BusyBox 1.38.0, whose lineedit get_homedir_or_NULL() now prefers $HOME over the passwd entry when expanding '\w' in the default PS1 ('\w \$ ').  The kernel hands init HOME=/, so with cwd '/' the prompt collapses to '~ #' instead of the '/ #' printed by older BusyBox, which only consulted getpwuid() (/root) and therefore never matched.

LAVA only knows about '/ #', so login-action times out and every baseline job on the new ramdisk fails, e.g.:

  login-action timed out after 117 seconds

Add '~ #' next to '/ #' in every boot template and in the watchdog-reset interactive test.  Prompts are regexes and '~' has no special meaning, so this stays compatible with the older ramdisks still in use.

### DIFF
--- a/config/runtime/boot/barebox.jinja2
+++ b/config/runtime/boot/barebox.jinja2
@@ -36,6 +36,7 @@
     failure_retry: 3
     prompts:
     - '/ #'
+    - '~ #'
     timeout:
       minutes: 20
     timeouts:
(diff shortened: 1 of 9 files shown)


nuclearcat:
Tested on staging: https://staging.kernelci.org:9000/viewer?search=parent%3D6a745614876eaefb4f4e548b

PR #3178, opened 2026-08-06
docker: bump tuxmake to 1.44.0

bhcopeland:
Update the pinned tuxmake version in the kernelci docker fragment from 1.38.0 to 1.44.0.

Notable changes for the builders:

- reproducer.sh is now written to the output directory, so it is picked up and uploaded along with the other build artifacts
- metadata.json is written early, so it survives a hard kill
- the compiler version is logged at the start of build.log
- arm was renamed to armhf and armv5 to armel; the old names are kept as aliases and armhf.ini is unchanged from the old arm.ini, so the arch=arm jobs are unaffected
- gcc-16, korg-gcc-16 and korg-clang-23 toolchains were added

### DIFF
--- a/config/docker/fragment/kernelci.jinja2
+++ b/config/docker/fragment/kernelci.jinja2
@@ -19,7 +19,7 @@ WORKDIR /root
 RUN rm -rf /tmp/kernelci-core
 
 # Install tuxmake kernel build tool
-RUN pip3 install --break-system-packages --no-cache-dir tuxmake==1.38.0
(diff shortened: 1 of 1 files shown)


roxell:
Since tuxmake now adds the compiler version into the build.log can we drop this commit 6ad0e70a86a2 ("kbuild: preserve tuxmake metadata as tuxmake_metadata.json") ?

roxell:
ba64058e794e ("kbuild: initialise rpmdb before tuxmake build") should probably be moved into the config/docker/fragment/kernelci.jinja2 and only generated once, and not every time we do a build.

PR #3179, opened 2026-08-06
Revert "kbuild: preserve tuxmake metadata as tuxmake_metadata.json"

bhcopeland:
This reverts commit 6ad0e70a86a2ac5bd3ba39a4ad8f90750a6867c1.

tuxmake now logs the compiler version at the start of build.log, so the exact compiler that produced a build is visible without kbuild carrying compiler.version_full over from tuxmake's metadata into build.compiler_version and the node data.

This also drops the rename of tuxmake's metadata.json to tuxmake_metadata.json, so _write_metadata overwrites it again.

### DIFF
--- a/kernelci/kbuild.py
+++ b/kernelci/kbuild.py
@@ -223,7 +223,6 @@ class KBuild:
             self._artifacts = []
             self._current_job = None
             self._config_full = ""
-            self._compiler_version = None
             self._srcdir = None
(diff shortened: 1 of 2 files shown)


roxell:
Yes that was what I meant with my comment in https://github.com/kernelci/kernelci-core/pull/3178

PR #3180, opened 2026-08-06
docker: initialise the rpm database in the container image

bhcopeland:
The rpm database that tuxmake's *rpm-pkg targets need was created by 'rpmdb --initdb' at the start of every build. It does not change between builds, so create it once in the kernelci docker fragment instead.

/var/lib/rpm is not writable by the build user, so the database lives under the kernelci home directory and ~/.rpmmacros points _dbpath at it via %{getenv:HOME}. _topdir stays in the per-build workspace, since rpmbuild output is collected from there into the artifacts directory.

The initdb is tolerant of rpm being missing, as images built from base/python.jinja2 do not install it.

### DIFF
--- a/config/docker/fragment/kernelci.jinja2
+++ b/config/docker/fragment/kernelci.jinja2
@@ -23,8 +23,14 @@ RUN pip3 install --break-system-packages --no-cache-dir tuxmake==1.38.0
 
 # Set up kernelci user
 RUN useradd kernelci -u 1000 -d /home/kernelci -s /bin/bash
-RUN mkdir -p /home/kernelci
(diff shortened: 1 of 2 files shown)


roxell:
Yes that was what I meant with my comment in https://github.com/kernelci/kernelci-core/pull/3178

bhcopeland:
Thanks @roxell (I split the pr's out on purpose)

nuclearcat:
https://staging.kernelci.org:9000/viewer?node_id=6a76069a37081d53fb49cfbd
seems no failures on rpm builds

PR #3181, opened 2026-08-07
kbuild: identify kselftest suite build results

nuclearcat:
Per-suite kselftest nodes report whether suite artifacts were built, not whether the tests were executed. Bare target names such as accel lose that context when dashboards display only the leaf node.

Prefix suite names with build.kselftest and continue replacing slashes with dots for nested targets. Add a focused test covering both ordinary and nested suite names.


nuclearcat:
Tested on staging: https://staging.kernelci.org:9000/viewer?search=parent%3D6a79970ddb05f633cbe790d8

PR #3193, opened 2026-08-24
docker: build pahole in Release mode

nuclearcat:
pahole's CMakeLists defaults CMAKE_BUILD_TYPE to Debug when it is not set, and CMAKE_C_FLAGS_DEBUG is "-Wall -Werror -ggdb -O0".  The build here never set a build type, so pahole has been compiled unoptimised with -Werror armed.

glibc 2.43 in forky makes strchr(), bsearch() and friends preserve the constness of their argument, so pahole's assignments of their results to non-const pointers now warn, and -Werror turns that into a build failure on the gcc-15 images:

```
  /pahole-1.29/btf_encoder.c:1856:16: error: assignment discards
  'const' qualifier from pointer target type
  [-Werror=discarded-qualifiers]
   1856 |         target = bsearch(&key, base, cnt, sizeof(key),
                            btf_func_cmp);
```

Even i tried, this is not fixed by moving to a newer pahole: v1.31 fails the same way. Still we will upgrade pahole later.

### DIFF
--- a/config/docker/base/host-tools.jinja2
+++ b/config/docker/base/host-tools.jinja2
@@ -100,12 +100,18 @@ RUN CFLAGS="-DPyString_FromString=PyUnicode_FromString \
             -DPyInt_AsLong=PyLong_AsLong" \
     pip3 install dtschema --break-system-packages
 
-# Download and build pahole v1.28
+# Download and build pahole v1.29
+#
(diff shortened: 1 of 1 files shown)


nuclearcat:
Tested on staging https://staging.kernelci.org/staging/1369

PR #3202, opened 2026-09-01
kbuild: make network readiness checks bounded (#3197)

### LINKED ISSUE #3197: Make Kubernetes job network readiness checks bounded and deployment-independent
## Problem

KernelCI contains two startup workarounds for Kubernetes nodes whose networking or DNS may not be immediately ready.

`KBuild._verify_network()` performs an HTTP request to Google before generating the build script:

```python
requests.get("https://google.com")
```

The current implementation has two failure modes:

- the request has no timeout and can hang indefinitely;
- a non-200 HTTP response does not decrement `retries`, causing an infinite loop.

The generic Python job template contains a separate DNS workaround that resolves `www.google.com` up to 30 times before starting a job.

Both checks depend on an unrelated third-party service. Reaching Google does not prove that the KernelCI API, artifact storage, source repositories, or other services required by the job are reachable. Conversely, a Google outage or network policy blocking Google can prevent otherwise healthy KernelCI jobs from starting.

The HTTP check was originally introduced because networking could be slow to initialize on newly started Kubernetes nodes.

## Relevant code

- `kernelci/kbuild.py`: `KBuild._verify_network()` and its call from `write_script()`
- `config/runtime/base/python.jinja2`: DNS readiness loop in `main()`

ChromeOS/Tast-specific Google dependencies are being removed separately in #3196 and are outside this issue.

## Desired outcome

Network initialization handling must:

- have a strict upper time limit;
- not depend on an unrelated public service;
- test connectivity relevant to the operation the job is about to perform;
- produce an actionable infrastructure error when readiness is not achieved;
- behave consistently between kbuild and other Python jobs.

## Maintainer decision required

Before implementation, a KernelCI maintainer or Kubernetes operator should choose one of these strategies:

1. **Retry the first required operation — recommended**

   Remove the generic preflight probes and apply bounded retry handling to the first real API, storage, source, or artifact request.

2. **Use a configurable readiness target**

   Keep a preflight check, but provide the hostname or URL through runtime configuration. The default should be a KernelCI-controlled service required by the job.

The selected maximum startup delay, request timeout, and retry interval must also be recorded in this issue.

## Implementation outline

After the strategy is confirmed:

1. Remove both hard-coded Google readiness checks.
2. Ensure every attempt has a connection and read timeout.
3. Use a fixed attempt count or monotonic deadline so every failure path terminates.
4. Handle expected network exceptions explicitly rather than catching every `Exception`.
5. Avoid calling `sys.exit()` from a low-level helper; return or raise an error that the job runner can classify.
6. Report exhausted network readiness as an infrastructure failure rather than a kernel build failure.
7. Include the attempted service and elapsed time in the final error without exposing credentials or tokens.
8. Add focused unit and template-rendering tests.

## Required tests

Depending on the selected strategy, cover:

- immediate success;
- initial failures followed by success;
- DNS failure;
- connection timeout;
- read timeout;
- repeated non-success HTTP responses;
- retry exhaustion;
- enforcement of the maximum elapsed time;
- correct infrastructure-error classification.

Tests must mock network requests and waiting so they do not contact external services or introduce real delays.

## Human participation required

This issue intentionally needs two concise operational checkpoints:

1. A Kubernetes operator confirms whether delayed networking still occurs and approves the readiness strategy and timing values.
2. After implementation, an operator runs a kbuild job on a newly scaled or cold Kubernetes node and links the result here.

Repository tests can prove that retry handling is bounded, but they cannot reproduce the real network initialization behavior of the production clusters.

## Acceptance criteria

- Generic KernelCI startup code no longer uses `google.com` as a connectivity probe.
- No readiness or retry path can loop or block indefinitely.
- The total readiness wait is bounded by the maintainer-approved duration.
- Failure identifies the relevant unavailable service and is reported as an infrastructure error.
- Unit tests cover success, recovery, timeout, non-success response, and exhaustion.
- Generated runtime templates contain the selected bounded behavior.
- A staging kbuild job succeeds on a newly started Kubernetes node.
- The staging validation result is linked in this issue.
- Existing test and lint checks pass.

## Out of scope

- Removing Google services that are required for an actual GKE or Google Cloud operation.
- Reworking unrelated download and upload retry policies.
- Changing Kubernetes cluster networking.
- Removing ChromeOS/Tast dependencies tracked by #3196.

Aniket1260:
This PR addresses #3197 by making Kubernetes network readiness checks bounded and independent of unrelated third-party services.

Problem

KernelCI had two generic startup checks that used Google connectivity as an indication that the Kubernetes node network was ready.

KBuild._verify_network() performed an HTTP request to google.com without a request timeout. Additionally, non-200 HTTP responses did not decrement the retry counter, which could cause the readiness check to loop indefinitely.

The generic Python runtime template also attempted to resolve www.google.com before starting a job. Besides introducing an unnecessary dependency on an external service, this does not verify connectivity to the KernelCI services that the job actually needs.

Solution

This PR updates the network readiness handling to use the KernelCI API as the readiness target.

For KBuild._verify_network():

Added a configurable target URL.
Use the configured KernelCI API URL when available, with https://api.kernelci.org as the fallback.
Added configurable retry count, retry delay, and request timeout.
Pass the timeout directly to requests.get().
Decrement the retry counter for every failed attempt, including non-200 HTTP responses.
Avoid sleeping after the final attempt.
Report the target service and elapsed time when readiness checks are exhausted.
Raise a RuntimeError instead of terminating the process directly with sys.exit().

For the generic Python runtime template:

Removed the hard-coded dependency on www.google.com.
Resolve the hostname of the configured KernelCI API instead.
Fall back to api.kernelci.org when the API URL cannot be determined.
Bound the DNS readiness check to a fixed number of attempts.
Report the actual hostname when DNS resolution fails.
Validation

The readiness logic was tested with mocked network conditions, including:

Immediate HTTP 200 success.
Repeated non-200 responses.
Temporary failures followed by a successful response.
Connection failures.
Request timeouts.
Retry exhaustion.

Closes #3197

### DIFF
--- a/config/runtime/base/python.jinja2
+++ b/config/runtime/base/python.jinja2
@@ -161,18 +161,24 @@ class Job(BaseJob):
 {% block python_main -%}
 def main(args):
     # Sometimes container DNS is failing if k8s node is still booting
-    # This is a workaround:
-    # test www.google.com until it works or retry 10 times
-    for _ in range(30):
+    # Workaround: resolve target KernelCI API host until ready (up to 10 retries)
+    try:
+        api_yaml = yaml.safe_load(API_CONFIG_YAML) if isinstance(API_CONFIG_YAML, str) else API_CONFIG_YAML
+        api_url = (api_yaml.get('url') if isinstance(api_yaml, dict) else None) or 'https://api.kernelci.org'
+        hostname = urllib.parse.urlparse(api_url).hostname or 'api.kernelci.org'
+    except Exception:
+        hostname = 'api.kernelci.org'
+
+    for _ in range(10):
         try:
-            socket.gethostbyname('www.google.com')
+            socket.gethostbyname(hostname)
             break
         except socket.gaierror:
             pass
         # sleep 1 second
         time.sleep(1)
     else:
-        print("DNS resolution failed, exiting", file=sys.stderr)
+        print(f"DNS resolution failed for {hostname}, exiting", file=sys.stderr)
         sys.exit(1)
 
     job = Job({% block python_job_constr %}api_config=API_CONFIG_YAML, node=NODE, workspace=WORKSPACE{% endblock %})
--- a/kernelci/kbuild.py
+++ b/kernelci/kbuild.py
@@ -574,24 +574,42 @@ trap - ERR
 
         return (buffer, fragment)
 
-    def _verify_network(self):
-        """Verify network connectivity"""
-        # TBD: Different URL? pool of urls?
-        retries = 10
+    def _verify_network(self, url=None, max_retries=5, retry_delay=5, timeout=10):
+        """Verify network connectivity to a KernelCI target service"""
+        target_url = url
+        if not target_url:
+            if hasattr(self, "_api_config") and self._api_config and getattr(self._api_config, "url", None):
+                target_url = self._api_config.url
+            elif hasattr(self, "_context") and self._context:
+                target_url = getattr(self._context, "get_api_url", lambda: None)()
+        if not target_url:
+            target_url = "https://api.kernelci.org"
+
+        retries = max_retries
+        start_time = time.time()
         while retries > 0:
             try:
-                r = requests.get("https://google.com")
+                r = requests.get(target_url, timeout=timeout)
+                if r.status_code == 200:
+                    return
+                print(
+                    f"[_verify_network] Non-200 response ({r.status_code}) "
+                    f"from {target_url}"
+                )
             except Exception as e:
-                print(f"[_verify_network] Error: {e}")
-                time.sleep(5)
-                retries -= 1
-                continue
-            if r.status_code == 200:
-                return
-            time.sleep(5)
+                print(f"[_verify_network] Error reaching {target_url}: {e}")
 
-        print("[_verify_network] Network is not available")
-        sys.exit(1)
+            retries -= 1
+            if retries > 0:
+                time.sleep(retry_delay)
+
+        elapsed = int(time.time() - start_time)
+        msg = (
+            f"[_verify_network] Network readiness check failed for {target_url} "
+            f"after {elapsed}s"
+        )
+        print(msg, file=sys.stderr)
+        raise RuntimeError(msg)
 
     def extract_config(self, frag):
         """Extract config fragments from legacy config file"""
@@ -1372,7 +1390,7 @@ trap 'case $stage in
                 for file in files:
                     file_rel = os.path.relpath(
                         os.path.join(root, file), self._af_dir
-                    )
+                    ).replace("\\", "/")
                     artifact_path = os.path.join(self._af_dir, file_rel)
                     upload_tasks.append((file_rel, artifact_path))
         else:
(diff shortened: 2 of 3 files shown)


config/runtime/base/python.jinja2
```diff
@@ -161,18 +161,24 @@ class Job(BaseJob):
...
+        api_yaml = yaml.safe_load(API_CONFIG_YAML) if isinstance(API_CONFIG_YAML, str) else API_CONFIG_YAML
+        api_url = (api_yaml.get('url') if isinstance(api_yaml, dict) else None) or 'https://api.kernelci.org'
+        hostname = urllib.parse.urlparse(api_url).hostname or 'api.kernelci.org'
+    except Exception:
+        hostname = 'api.kernelci.org'
+
+    for _ in range(10):
         try:
-            socket.gethostbyname('www.google.com')
+            socket.gethostbyname(hostname)
```
nuclearcat, line 174: socket.gethostbyname() has no deadline, so the ten-iteration loop cannot interrupt a stuck DNS lookup

kernelci/kbuild.py
```diff
@@ -574,24 +574,42 @@ def _getcrosfragment(self, fragment):
...
+                target_url = getattr(self._context, "get_api_url", lambda: None)()
+        if not target_url:
+            target_url = "https://api.kernelci.org"
+
+        retries = max_retries
+        start_time = time.time()
         while retries > 0:
             try:
-                r = requests.get("https://google.com")
+                r = requests.get(target_url, timeout=timeout)
```
nuclearcat, line 592: timeout=10 is an inactivity timeout, not a total wall-clock deadline

kernelci/kbuild.py
```diff
@@ -574,24 +574,42 @@ def _getcrosfragment(self, fragment):
 
         return (buffer, fragment)
 
-    def _verify_network(self):
-        """Verify network connectivity"""
-        # TBD: Different URL? pool of urls?
-        retries = 10
+    def _verify_network(self, url=None, max_retries=5, retry_delay=5, timeout=10):
```
nuclearcat, line 577: _verify_network() still catches every Exception; malformed URLs and programming/configuration errors become misleading network-readiness failures

kernelci/kbuild.py
```diff
@@ -1382,7 +1400,8 @@ def upload_artifacts(self):
                 upload_tasks.append((artifact, artifact_path))
 
         def is_dtb_artifact(artifact):
-            return artifact.startswith("dtbs/") and artifact.endswith(".dtb")
+            posix_art = artifact.replace("\\", "/")
+            return posix_art.startswith("dtbs/") and posix_art.endswith(".dtb")
```
nuclearcat, line 1404: The DTB path-normalization changes are unrelated to #3197 and are not explained in the PR description. Please remove them from this PR or submit them separately with their motivation and tests.

Their inclusion also raises concern that the complete patch may not have been carefully reviewed before submission. KernelCI’s contribution policy requires contributors to understand and take responsibility for all submitted changes, and to disclose meaningful AI assistance with an Assisted-by tag. Please review the full diff and disclose any applicable assistance.

nuclearcat (changes requested):

Aniket1260:
@nuclearcat before submitting my next commit, I'll observe the changes more carefully. Thanks for your valuable insights

nuclearcat:
No activity from submitter

PR #3215, opened 2026-09-16
Build armel rootfs images and let platforms override the Debian arch

nuclearcat:
ARMv5 platforms are handed an `armhf` rootfs they physically cannot execute.

`get_system_arch()` derives the Debian architecture from the kernel architecture, where `arm` maps to `armhf`. That is correct for ARMv7 but wrong for ARMv5: armhf requires hardware floating point, so on an ARM926EJ-S core the very first instruction of init traps. `at91sam9g20ek` boots the armhf `trixie-ltp` image and panics with:

```
Run /init as init process
Kernel panic - not syncing: Attempted to kill init! exitcode=0x00000004
CPU: 0 UID: 0 PID: 1 Comm: init Not tainted 6.13.12 #1
Hardware name: Atmel AT91SAM9
```

Exit code 4 is SIGILL. The board is `ARM926EJ-S [41069265] revision 5 (ARMv5TEJ)` with 64 MB of RAM. Its `ltp-smoke` job has never passed in any run retained by the API — I sampled 340 runs back to 2026-03-20 and every one is `incomplete`. That is expected, because `trixie-ltp` has never published an armel image in any version.

### config: allow platforms to override the Debian architecture

Adds an optional `debarch` attribute to the platform configuration, and stops the two derivation sites from clobbering an explicitly configured value:

- `kernelci/config/base.py` — `format_params()` only derives a system architecture name that isn't already set
- `kernelci/runtime/__init__.py` — same for `_get_params()`, which feeds the job templates

The guard covers all four of `brarch`/`crosarch`/`debarch`/`karch` rather than just `debarch`, so any future override behaves the same way. Platforms that don't set it are completely unaffected.

### rootfs: build trixie-ltp and trixie-kselftest for armel

Adds `armel` to both arch lists.

`trixie-kselftest` needs nothing further — its build script is architecture agnostic, as `trixie-vdso` already demonstrates by building armel today.

`trixie-ltp` cross-compiles LTP outside the target chroot against a triplet table that rejects anything unknown, so armel is mapped to the soft-float `arm-linux-gnueabi` triplet. Everything else in that script, including the `PKG_CONFIG_LIBDIR` multiarch path, is driven off that one variable. The debos image only carried the hard-float cross compiler, so `gcc-arm-linux-gnueabi` is installed alongside it.

### Testing

- 4 new unit tests in `tests/test_configs.py` covering the derived default, the override, the other architecture names staying derived, and loading from YAML
- `tests/test_configs.py`, `test_rootfs_config.py`, `test_rootfs_debos.py`, `test_rootfs_builder.py` all pass
- Verified against the kernelci-pipeline platform config that `at91sam9g20ek` resolves to `armel` while `imx6q-sabrelite`, `stm32mp157a-dhcor-avenger96`, `qemu-arm` and the arm64/x86_64 boards are unchanged

### Notes

The armel LTP cross-build has not been exercised yet; a soft-float build may surface something the way the armhf `trixie-rt` build did. Once these images are published, `trixie-ltp` and `trixie-kselftest` in kernelci-pipeline `config/rootfs.yaml` need bumping to the new version — it has to be a full rebuild of every architecture, since a single version directory must contain everything `{debarch}` can resolve to.

The companion platform change is kernelci/kernelci-pipeline#1604. It is inert until this lands, as unknown YAML keys are silently ignored.

Assisted-by: Claude


nuclearcat:
Related-to: https://github.com/kernelci/kernelci-pipeline/pull/1601

PR #3216, opened 2026-09-16
rootfs: ship netconfig and protocols in the NFS initramfs

nuclearcat:
NFS root mounts fail on the 20260915.0 images:

```
mount.nfs: Protocol not supported for 192.168.201.1:/var/lib/lava/dispatcher/tmp/22976572/extract-nfsrootfs-pn0we53n on /root
Begin: Retrying nfs mount ... mount.nfs: Protocol not supported for ... on /root
...
Target filesystem doesn't have requested /sbin/init.
No init found. Try passing init= bootarg.
```

The board then sits at the `(initramfs)` prompt until LAVA gives up with `login-action timed out`.

### Bisect

`ltp-dio` on `imx6q-sabrelite` is a clean before/after:

| Run | initrd | Result |
|---|---|---|
| 2026-09-15 08:49 | `trixie-ltp/20260219.0/armhf` | pass |
| 2026-09-15 16:52 | `trixie-ltp/20260915.0/armhf` | incomplete |
| 2026-09-16 00:23 | `trixie-ltp/20260915.0/armhf` | incomplete |

It had passed on every run going back to March. The first failure is the first run after 20260915.0 went live, which is the first image whose initramfs ships `mount.nfs` instead of klibc's `nfsmount`.

### Cause

`copy_exec` installs the binary and the shared libraries it links against, but `mount.nfs` also reads two databases at run time that nothing else pulls into the initramfs.

`nfs_get_netid()` in `support/nfs/getport.c` resolves the transport to an RPC netid two ways, and both need a file this initramfs does not have:

- `getprotobynumber()` — reads `/etc/protocols`
- `setnetconfig()` / `getnetconfig()` — reads `/etc/netconfig` via libtirpc

When either fails the function falls through to `rpc_createerr.cf_stat = RPC_UNKNOWNPROTO`, and `nfs_construct_new_options()` in `utils/mount/stropts.c` maps that to `errno = EPROTONOSUPPORT`, printed as "Protocol not supported". That matches the observed behaviour exactly: instantaneous failure, no RPC timeout, and no other diagnostic, since every `EPROTONOSUPPORT` site in `network.c` also prints a `Failed to find '%s' protocol` line that is absent from the log.

Confirmed by unpacking the published initrds. All four have `libtirpc.so.3` and `mount.nfs`, and none has either database:

```
trixie-kselftest armhf  nfsmount=wrapper  mount.nfs=yes  netconfig=NO  protocols=NO
trixie-kselftest arm64  nfsmount=wrapper  mount.nfs=yes  netconfig=NO  protocols=NO
trixie-ltp       armhf  nfsmount=wrapper  mount.nfs=yes  netconfig=NO  protocols=NO
trixie-ltp       arm64  nfsmount=wrapper  mount.nfs=yes  netconfig=NO  protocols=NO
```

klibc's `nfsmount` carried its own portmap and MOUNT client, so it needed neither database and mounted the same export from the same server without complaint.

### Ruled out

- **Server does not speak v3** — the last passing boot mounted from the same server with the same kernel command line.
- **Kernel lacks NFSv3** — the `multi_v7_defconfig` build has `CONFIG_NFS_FS=y`, `CONFIG_NFS_V3=y`, `CONFIG_ROOT_NFS=y`.
- **The wrapper mistranslates the bare `v3` option** — `v3` is in nfs-utils' `nfs_version_opttbl[]` and is present in the shipped binary, tail-merged into the `"tcp,v3"` string, so `mount.nfs` handles it natively and no translation is needed.
- **`scripts/nfs` changed** — it is byte identical between 20260219.0 and 20260915.0. The only difference is `nfsmount`.

### Fix

Copy both databases into the initramfs. glibc defaults the protocols database to `files` when `/etc/nsswitch.conf` is absent, so that file is not needed. Both sources are present in the built image, so `copy_file` resolves them.

### Scope

Affects every Debian image whose initramfs boots over NFS — `trixie-ltp` and `trixie-kselftest`, all architectures. Ramdisk boots are unaffected, as they never call `nfsmount`.

Needs a rebuild of both images to take effect, after which the `config/rootfs.yaml` pointers in kernelci-pipeline need bumping to the new version.

Assisted-by: Claude


nuclearcat:
Related-to: https://github.com/kernelci/kernelci-pipeline/pull/1601

